PR #24, opened 2026-01-12
fix!: introduce config_schema to define config shapes for capabilities

raginpirate:
# Description

Adds `config_schema` field to capability declarations, allowing capabilities to expose the JSON Schema URL that validates their `config` object. Previously, capabilities could include a `config` object but there was no mechanism to expose or discover the schema for that config—unlike payment handlers which already had `config_schema`.

This change:
- Adds `config_schema` field to the capability base schema
- Adds `dependentRequired` constraint: if `config` is present, `config_schema` is required
- Updates all documentation examples to include `config_schema` where `config` is used
- Clarifies the `schema` field description to "URL to JSON Schema for types this capability defines or extends"

**Motivation:** The original capability schema had no way to expose config schemas, making it impossible for platforms to validate or understand capability-specific configuration with just schemas; they would have to crawl the spec. This mirrors the pattern already established in payment handlers.

## Type of change

      or fields**)
- [ ] Documentation update

---

### DIFF
--- a/source/schemas/capability.json
+++ b/source/schemas/capability.json
@@ -25,16 +25,21 @@
         "schema": {
           "type": "string",
           "format": "uri",
-          "description": "URL to JSON Schema for this capability's payload."
+          "description": "URL to JSON Schema for types this capability defines or extends."
         },
         "extends": {
           "type": "string",
           "pattern": "^[a-z][a-z0-9]*(?:\\.[a-z][a-z0-9_]*)+$",
           "description": "Parent capability this extends. Present for extensions, absent for root capabilities."
         },
+        "config_schema": {
+          "type": "string",
+          "format": "uri",
+          "description": "URL to JSON Schema that validates the config object."
+        },
         "config": {
           "type": "object",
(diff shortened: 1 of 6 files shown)


yanheChen:
Hey daniel, is this PR intentionally putting payment under the checkout extensions bracket? 

if it is, @igrigorik I heard that the payment object has been designed dangling there for a reason. May I have you chime in for this one to give me some context of the original design?

source/schemas/capability.json
```diff
@@ -25,16 +25,21 @@
...
           "format": "uri",
-          "description": "URL to JSON Schema for this capability's payload."
+          "description": "URL to JSON Schema for types this capability defines or extends."
         },
         "extends": {
           "type": "string",
           "pattern": "^[a-z][a-z0-9]*(?:\\.[a-z][a-z0-9_]*)+$",
           "description": "Parent capability this extends. Present for extensions, absent for root capabilities."
         },
+        "config_schema": {
```
knightlin-shopify, line 35: Does this imply we should have `schema` property part of `config` object instead? This aligns with existing `capability.schema` pattern.

PR #28, opened 2026-01-12
refactor!: make "complete checkout" schemas consistent across actions/ transports

lemonmade:
# Description

This PR adds the generation of dedicated `complete` schemas, and aligns all transports to use the `checkout` object in those schemas. This lets us define a custom subset of the `checkout` object that can be provided for `complete_checkout` MCP methods and `POST /checkout-sessions/{id}/complete` REST actions. I have defined the inclusion criteria for `complete` based on a combination of the `update` configuration and my own understanding from our conversations on what we consider to be allowable on `complete`. I am open to changes by more knowledgeable domain experts, and I have annotated any choices I wasn’t confident in.

In this PR, I have not adjusted how `risk_signals` or `ap2` are defined; both need further adjustments to be transport-agnostic, as neither extends the `Checkout` object.

## Type of change

      or fields**)
- [ ] Documentation update

---

files or fields:

#### Justification

We want to fix inconsistency in the Checkout Complete input arguments across transports, and we want to make sure that all fields — whether they are core fields or added by extensions — can be defined as omitted, optional, or required for all actions in all transports.

---

## Checklist:


source/schemas/shopping/checkout.json
```diff
@@ -95,7 +111,12 @@
       "ucp_request": "omit"
     },
     "payment": {
-      "$ref": "payment.json"
+      "$ref": "payment.json",
+      "ucp_request": {
+        "create": "optional",
+        "update": "optional",
```
lemonmade, line 117: `payment` is actually required on `create` and `update` calls today. I feel this should at least be `optional`, but it wasn’t clear to me if this should actually be _disallowed entirely_ for create and update instead. I believe we should just make it optional, as it allows extensions to add fields to payment that might make sense pre-complete, but I wanted to confirm.

source/schemas/shopping/checkout.json
```diff
@@ -27,19 +27,30 @@
       "description": "Unique identifier of the checkout session.",
       "ucp_request": {
         "create": "omit",
-        "update": "required"
+        "update": "required",
+        "complete": "omit"
```
lemonmade, line 31: Based on discussions, my understanding is that all core fields other than `payment` should be omitted from `complete`, which is what I believe I have implemented.

source/services/shopping/openrpc.json
```diff
@@ -79,9 +79,9 @@
           "schema": {"type": "string"}
         },
         {
-          "name": "payment",
+          "name": "checkout",
```
lemonmade, line 82: This is the other breaking change in this PR: MCP `complete_checkout` now accepts `{checkout: {payment: instruments: []}}`, instead of `{payment: {instruments: []}}`. This allows it to accept other top-level, non-payment fields that extensions mark as optional or required in `ucp_request[complete]`.

source/services/shopping/openapi.json
```diff
@@ -246,7 +246,7 @@
             "application/json": {
               "schema": {
                 "allOf": [
-                  { "$ref": "#/components/schemas/payment_data" },
+                  { "$ref": "#/components/schemas/checkout" },
```
lemonmade, line 249: This is the first breaking change: REST `POST /checkout_sessions/:id/complete` now accepts `{payment: {instruments: []}}` in its body, instead of `{payment_data: {}}`.

raginpirate (approved):
Spot on; this sits well with me as the first of many breaking changes we need to push out this week 😄

source/schemas/shopping/checkout.json
raginpirate: Aligned; it should be totally fine to provide payment context ahead of submission, even for full credentials. Its the business's discretion to use the credentials beyond the scenario contracted by the signed checkout mandate from the payment handler.
But definitely not required 👍

yanheChen: based on my understand, we should also ucp_request: omit handlers field and selected_instrument_id field in the payment.json

source/services/shopping/openapi.json
yanheChen: my understanding that it will pass only the payment not the checkout

spec/schemas/shopping/discount.complete_req.json
```diff
@@ -0,0 +1,115 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
```
yanheChen, line 2: qq that if all the fields are omitted, why do we still have this fulfillment.complete_req.json  and discount.complete_req.json generated? should them been omitted?

source/services/shopping/openapi.json
lemonmade: @yanheChen this isn’t what I thought we had specified, and it would mean that all fields we want to add that are provided in complete (including `risk_signals` and `ap2`) would need to be nested under payment. I feel the added flexibility of allowing extensions to define non-payment fields that are accepted on complete is worth the requirement to nest `payment`, especially since that also matches the other REST actions.

spec/schemas/shopping/discount.complete_req.json
lemonmade: Yep, will update the generator to omit these files 👍

knightlin-shopify (approved):

igrigorik (changes requested):
LGTM overall, requesting we address one more nit before merging.

```json
{
    "name": "checkout",
    "required": false,  //👈🏻👈🏻👈🏻 
    "schema": { "$ref": "checkout.complete_req.json" }
  }
```

We mark `checkout` object as optional in MCP spec, which creates an inconsistency with the REST shape where it's required. Let's align these.. 

```
From fdcb000b4f3c03069bc3a72bb1c4f9cdf39a0cb3 Mon Sep 17 00:00:00 2001
From: Ilya Grigorik <ilya@grigorik.com>
Date: Mon, 12 Jan 2026 17:04:48 -0800
Subject: [PATCH] align complete_checkout with REST body requirements

- Change `checkout` param from optional to required in OpenRPC spec
- Update MCP docs to reflect `checkout` param (was incorrectly showing `payment`)

REST requires requestBody with payment inside; MCP now matches.
Free checkout still works via empty instruments array, not omitted param.
---
 docs/specification/checkout-mcp.md    | 3 +--
 source/services/shopping/openrpc.json | 2 +-
 2 files changed, 2 insertions(+), 3 deletions(-)

diff --git a/docs/specification/checkout-mcp.md b/docs/specification/checkout-mcp.md
index 3f61b17..e8373ce 100644
--- a/docs/specification/checkout-mcp.md
+++ b/docs/specification/checkout-mcp.md
@@ -539,8 +539,7 @@ Maps to the [Complete Checkout](checkout.md#complete-checkout) operation.
 #### Input Schema
 
 *   `id` (String): The ID of the checkout session.
-*   `payment` ([Payment](checkout.md#payment), Optional): Payment instrument instance submitted
-    by the buyer.
+*   `checkout` ([Checkout](checkout.md#complete-checkout)): **Required**. Contains payment credentials to execute the transaction.
 *   `idempotency_key` (String, UUID): **Required**. Unique key for retry
     safety.
 
diff --git a/source/services/shopping/openrpc.json b/source/services/shopping/openrpc.json
index 9ba6d49..a063448 100644
--- a/source/services/shopping/openrpc.json
+++ b/source/services/shopping/openrpc.json
@@ -80,7 +80,7 @@
         },
         {
           "name": "checkout",
-          "required": false,
+          "required": true,
           "schema": {"$ref": "https://ucp.dev/schemas/shopping/checkout.json"}
         },
         {
-- 
2.52.0
```

^ also fixes a docs bug referencing old schema. With that we have consistent behavior between MCP and REST. For no payment required flows, you can pass an empty instrument array.

Separately, A2A docs are still using `payment_data` shape. I'm OK with addressing that in separate fast-follow PR.

---

Stepping back, for future archeology, principles behind this update:

1. **Complete is a command, not an update** — credentials execute against server's authoritative state; no mutable checkout fields in complete payload.
2. **`payment.instruments[]` array** — future-proofs split payments without breaking change; consistent with fulfillment.methods[] pattern.
3. **Transport agnostic** — same logical shape across REST/MCP.

The new shapes are...

#### REST
```json
  POST /checkout-sessions/{id}/complete
  Content-Type: application/json
  Idempotency-Key: <uuid>

  {
    "payment": {
      "instruments": [{
        "handler_id": "com.google.pay",
        "credential": { "type": "PAYMENT_GATEWAY", "token": "..." }
      }]
    },
  }
```

#### MCP (JSON-RPC)
```json
  {
    "method": "tools/call",
    "params": {
      "name": "complete_checkout",
      "arguments": {
        "id": "cs_abc123",
        "checkout": {
          "payment": {
            "instruments": [{
              "handler_id": "com.google.pay",
              "credential": { "type": "PAYMENT_GATEWAY", "token": "..." }
            }]
          }
        },
        "idempotency_key": "<uuid>"
      }
    }
  }
```

pnpm-lock.yaml
lemonmade, line 1: This was accidentally committed, needs to be removed

source/schemas/shopping/checkout.json
lemonmade: I am going to leave `selected_payment_instrument` to a subsequent PR as well, if that’s OK, as it plays into a bigger conversation around split payment flexibility IIUC. [`payment.handlers` is already omitted from all requests](https://github.com/lemonmade/ucp/blob/c5badd74a596318ac423d91f80fb12908d4a46b8/source/schemas/shopping/payment.json#L17), so I think we are good on that front 👍

source/services/shopping/openrpc.json
```diff
@@ -79,9 +79,9 @@
           "schema": {"type": "string"}
         },
         {
-          "name": "payment",
-          "required": false,
-          "schema": {"$ref": "https://ucp.dev/schemas/shopping/payment.json"}
+          "name": "checkout",
+          "required": true,
```
lemonmade, line 83: To match REST, we are making the `complete_checkout` `checkout` field required in the MCP transport. Free checkouts would pass `{checkout: {payment: {instruments: []}}}` to satisfy the contract.

spec/schemas/shopping/types/fulfillment_available_method_req.json
lemonmade, line 1: All of these files were deleted by the new logic to remove unused per-operation schema types. None of these were referenced in any transport schema previously, as these fields were all omitted from complete and/or update requests.

PR #49, opened 2026-01-14
refactor!: organize payments, services, and capabilities under same hierarchy

raginpirate:
Stacked ontop of https://github.com/Universal-Commerce-Protocol/ucp/pull/54

Example of Shop Pay handler diff:
https://github.com/Universal-Commerce-Protocol/ucp/commit/ba7f800c2c19072050a02d969477510e42d10116#diff-515511ea32bd431b8b3094a1d51eb3e70b9607cc9c45c48eef2916c29eb52151

# Description

This PR reorganizes the schema structure to establish a unified hierarchy for payments, services, and capabilities. The refactoring consolidates schema files into a consistent location and structure, making the codebase more maintainable and intuitive.

Main wins are:
- Payments stops having a strange schema structure relative to capabilities.
- Services stop having an isolated solution for service <-> implementation pattern.
- Allows for multiple versions of a service to be hosted at once, and allows for transports to be surfaced at the same level. This fixes a two bugs: how do I explain to an agent that I support two versions of my service, and how do I version just one transport at a time?

**Key changes:**
- Moved `payment_handler.json` from `source/schemas/shopping/types/` to `source/schemas/`
- Moved `service_schema.json` from `source/services/` to `source/schemas/service.json`
- Unified all three types (capabilities, services, payment handlers) to use a common base pattern from `ucp.json#/$defs/entity_base`
- Introduced consistent variant definitions (`platform_config`, `business_config`, `response`) across all schema types
- Updated all documentation to reflect the new schema structure and organization
- Expose a unique structure for profile schemas, rather than mixing the typing between platform and business.

**Open Issues:**
- This PR intentionally breaks apart the ucp metadata between platform, business, and response, such that any service, capability, or extension has the right breakpoint to define in their schema (config shaped one way for discovery while shaped another way for checkout/order responses). However, this PR does not define the semantics of how to accomplish that: that is for another PR.

## Type of change

      or fields**)
- [ ] Documentation update

---

files or fields:

### DIFF
--- a/docs/documentation/schema-authoring.md
+++ b/docs/documentation/schema-authoring.md
@@ -30,7 +30,7 @@ UCP schemas use standard JSON Schema fields plus UCP-specific metadata:
 | `title` | JSON Schema | Human-readable display name | All schemas |
 | `description` | JSON Schema | Schema purpose and usage | All schemas |
 | `name` | UCP | Reverse-domain identifier; doubles as registry key | Capabilities, services, handlers |
-| `version` | UCP | Entity version (`YYYY-MM-DD` format) | Capabilities, services, handlers |
+| `version` | UCP | Entity version (`YYYY-MM-DD` format) | Capabilities, services, payment handlers |
 | `id` | UCP | Instance identifier for multiple configurations | Payment handlers only |
 
 ### Why Self-Describing?
@@ -89,7 +89,7 @@ UCP schemas fall into six categories based on their role in the protocol.
 
 Define negotiated capabilities that appear in `ucp.capabilities{}` registries.
 
-- **File-level fields**: `$schema`, `$id`, `title`, `description`, `name`, `version`
+- **Top-level fields**: `$schema`, `$id`, `title`, `description`, `name`, `version`
 - **Variants**: `platform_schema`, `business_schema`, `response_schema`
 
 Examples: `checkout.json`, `fulfillment.json`, `discount.json`, `order.json`
@@ -99,7 +99,7 @@ Examples: `checkout.json`, `fulfillment.json`, `discount.json`, `order.json`
 Define transport bindings that appear in `ucp.services{}` registries. Each transport
 (REST, MCP, A2A, Embedded) is a separate entry.
 
-- **File-level fields**: `$schema`, `$id`, `title`, `description`, `name`, `version`
+- **Top-level fields**: `$schema`, `$id`, `title`, `description`, `name`, `version`
 - **Variants**: `platform_schema`, `business_schema`
 - **Transport requirements**:
     - REST/MCP: `endpoint`, `schema` (OpenAPI/OpenRPC URL)
@@ -110,7 +110,7 @@ Define transport bindings that appear in `ucp.services{}` registries. Each trans
 
 Define payment handler configurations in `ucp.payment_handlers{}` registries.
 
-- **File-level fields**: `$schema`, `$id`, `title`, `description`, `name`, `version`
+- **Top-level fields**: `$schema`, `$id`, `title`, `description`, `name`, `version`
 - **Variants**: `platform_schema`, `business_schema`, `response_schema`
 - **Instance `id`**: Required to distinguish multiple configurations of the same handler
 
@@ -125,7 +125,7 @@ specification template.
 Data structures embedded within capabilities but not independently negotiated.
 Do **not** appear in registries.
 
-- **File-level fields**: `$schema`, `$id`, `title`, `description`
+- **Top-level fields**: `$schema`, `$id`, `title`, `description`
 - **Omit**: `name`, `version` (not independently versioned)
 
 Examples:
@@ -136,7 +136,7 @@ Examples:
 
 Reusable definitions referenced by other schemas. Do **not** appear in registries.
 
-- **File-level fields**: `$schema`, `$id`, `title`, `description`
+- **Top-level fields**: `$schema`, `$id`, `title`, `description`
 - **Omit**: `name`, `version`
 
 Examples: `types/buyer.json`, `types/line_item.json`, `types/postal_address.json`
@@ -145,7 +145,7 @@ Examples: `types/buyer.json`, `types/line_item.json`, `types/postal_address.json
 
 Define protocol structure rather than entity payloads.
 
-- **File-level fields**: `$schema`, `$id`, `title`, `description`
+- **Top-level fields**: `$schema`, `$id`, `title`, `description`
 - **Omit**: `name`, `version`
 
 Examples: `ucp.json` (entity base), `capability.json`, `service.json`, `payment_handler.json`
@@ -318,6 +318,6 @@ A capability schema defines both payload structure and declaration variants:
 
 Key points:
 
-- **File-level `name` and `version`** make the schema self-describing
+- **Top-level `name` and `version`** make the schema self-describing
 - **`$defs` variants** enable validation in different contexts
 - **Payload properties** define the actual checkout response structure
--- a/docs/playground.md
+++ b/docs/playground.md
@@ -425,7 +425,7 @@ const UcpData = {
     "com.shopify.shop_pay": [
       {
         id: "shop_pay",
-        version: "2025-12-08",
+        version: "2026-01-11",
         spec: "https://shopify.dev/docs/agents/checkout/shop-pay-handler",
         config_schema: "https://shopify.dev/ucp/shop-pay-handler/2026-01-11/config.json",
         instrument_schemas: ["https://shopify.dev/ucp/shop-pay-handler/2026-01-11/instrument.json"],


docs/documentation/schema-authoring.md
```diff
@@ -75,14 +75,14 @@ capability negotiation.
...
-    {"name": "dev.ucp.shopping.checkout", "version": "2026-01-11"},
-    {
-      "name": "dev.ucp.shopping.fulfillment",
-      "version": "2026-01-11",
-      "extends": "dev.ucp.shopping.checkout"
-    }
-  ]
+  "capabilities": {
+    "dev.ucp.shopping.checkout": [
+      {"version": "2026-01-11"}
```
knightlin-shopify: Since we remove the `name` and put it as key, the document above should be updated as well to reflect it.

docs/documentation/schema-authoring.md
```diff
@@ -75,14 +75,14 @@ capability negotiation.
...
+    ],
+    "dev.ucp.shopping.fulfillment": [
+      {"version": "2026-01-11", "extends": "dev.ucp.shopping.checkout"}
+    ]
+  }
 }
```
knightlin-shopify: nit: should have `payment_handlers` example payload here.

docs/playground.md
```diff
@@ -572,13 +572,19 @@ class UcpBackend {
   }
 
   getDiscoveryProfile() {
+    const paymentHandlers = {};
```
knightlin-shopify: Should update 'UcpData' mock data shape so we don't need to convert here?

docs/playground.md
```diff
@@ -572,13 +572,19 @@ class UcpBackend {
...
+    });
+    
     return {
       ucp: {
         version: UcpData.version,
-        capabilities: [...UcpData.capabilities.core, ...UcpData.capabilities.extensions]
-      },
-      payment: {
-        handlers: UcpData.handlers
+        capabilities: [...UcpData.capabilities.core, ...UcpData.capabilities.extensions],
```
knightlin-shopify: 'UcpData' need to be updated, otherwise, it won't reflect the `name` to key change.

docs/playground.md
```diff
@@ -730,7 +735,7 @@ class UcpBackend {
       fulfillment: fulfillmentObj,
       adjustments: [],
       totals: this.session.totals,
-      payment: { selected_instrument_id: paymentInstrument.id }
+      payment: { instruments: [{ ...paymentInstrument, selected: true }] }
```
knightlin-shopify: This is order object, there is no payment property here. Should be remvoed?

docs/specification/checkout-rest.md
```diff
@@ -285,41 +277,15 @@ so clients must include all previously set fields they wish to retain.
...
-                }
-              ]
-            }
-          }
-        ],
-        "selected_instrument_id": "pi_gpay_5678",
         "instruments": [
           {
             "id": "pi_gpay_5678",
             "handler_id": "com.google.pay",
```
knightlin-shopify: The value should be `gpay` instead? It refers to the handler.id, right? Not the key. 'shop_pay' example above does it right.

docs/specification/checkout-rest.md
```diff
@@ -706,41 +664,15 @@ Follow-up calls after initial `fulfillment` data to update selection.
...
```
knightlin-shopify: Same here, it should be `gpay` isntead.

docs/specification/checkout-rest.md
```diff
@@ -1070,41 +992,15 @@ place to set these expectations via `messages`.
...
```
knightlin-shopify: `gpay` instead.

docs/specification/checkout.md
```diff
@@ -34,11 +34,15 @@ PCI DSS compliant to support this Capability.
 
 **Payments**
 
-On checkout creation, businesses are required to define a payment configuration.
-The Checkout object includes `payment.handlers` which define the processing
-specifications for collecting payment instruments (e.g., Google Pay, direct
-tokenization). When the buyer submits payment, the platform populates the
-`payment.instruments` array with the collected instrument data.
+Payment handlers are discovered from the business's UCP profile at
+`/.well-known/ucp`. The handlers define the processing specifications for
```
knightlin-shopify: And through the checkout? `ucp.payment_handlers`

docs/specification/examples/business-tokenizer-payment-handler.md
```diff
@@ -61,7 +61,7 @@ happens internally within the Processor's secure environment.
 │ (Collector)│                         │      (Business or PSP)            │
 └─────┬──────┘                         └─────────────────┬─────────────────┘
       │                                                  │
-      │  1. GET payment.handlers                         │
+      │  1. GET payment_handlers                         │
```
knightlin-shopify: nit: `ucp.payment_handlers`

docs/specification/overview.md
```diff
@@ -591,7 +623,7 @@ participants together.
 **Important distinction:**
 
 - **Payment Credential Provider** = The participant (entity like Google Pay, Stripe)
```
knightlin-shopify: nit: Update Stripe to Shop Pay to match the example handler name provided below.

docs/specification/overview.md
```diff
@@ -635,55 +667,66 @@ an encrypted payment token.
...
         }
-      }
-    ]
+      ],
+      "dev.shopify.shop_pay": [
+        {
+          "id": "shop_pay_handler",
+          "version": "2026-01-11",
+          "spec": "https://shopify.dev/ucp/shop-pay-handler",
+          "schema": "https://shopify.dev/ucp/schemas/shop-pay-config.json",
```
knightlin-shopify: nit: add version in the schema and spec URL.

docs/specification/overview.md
```diff
@@ -635,55 +667,66 @@ an encrypted payment token.
...
+    }
   }
 }
 ```
 
 **2. Token Execution (Platform Side)**
-The platform recognizes `com.google.pay`. It passes the `config` into the
-Google Pay API. Google Pay returns the encrypted token data.
+The platform recognizes `com.google.pay` or `dev.shopify.shop_pay`. It passes the `config` into the
+respective wallet API. The wallet returns the encrypted token data.
```
knightlin-shopify: nit: Would be more consistent to use `Payment Credential Provider` instead of `wallet`.

source/schemas/capability.json
```diff
@@ -6,54 +6,47 @@
...
       ]
     },
 
-    "response": {
-      "title": "Capability (Response)",
+    "business_schema": {
+      "title": "Capability (Business Schema)",
+      "description": "Capability configuration for business/merchant level. May include business-specific config overrides.",
+      "allOf": [
+        {"$ref": "#/$defs/base"}
```
knightlin-shopify: Do we need `{"required": ["spec", "schema"]}` here as well?

source/schemas/shopping/payment.json
```diff
@@ -4,22 +4,7 @@
...
-        "$ref": "types/payment_handler.json"
-      },
-      "description": "Processing configurations that define how payment instruments can be collected. Each handler specifies a tokenization or payment collection strategy.",
-      "ucp_request": "omit"
-    },
-    "selected_instrument_id": {
-      "type": "string",
-      "description": "The id of the currently selected payment instrument from the instruments array. Set by the agent when submitting payment, and echoed back by the merchant in finalized state."
-    },
     "instruments": {
```
knightlin-shopify, line 8: Should the instruments be required now?

docs/documentation/schema-authoring.md
0x8000-0000: Should we rev the version? Will clients that are built with the previous version of the schema work with servers built with the updated schema?

raginpirate:
Thanks for the quick review @igrigorik!

> I'd be in favor of the above: expose id and config as optional in base type. It doesn't change the existing shape but it gives us flexibility and symmetry between these primitives.

👍 😄 while I was working on this, I totally wanted to add config and schema to services since it makes sense as a way to expose interesting configuration inputs for the service, regardless of transport. Will do!

Thanks for scrubbing through the docs here @knightlin-shopify! I'll admit that was mostly AI driven changes, and I was going to wait until the approach was cemented before I actually made sure every example was perfect 😂

docs/documentation/schema-authoring.md
raginpirate: Hey there! Versions will be revised as a part of cutting a release I think 😄

source/schemas/capability.json
raginpirate: From the business side, I was thinking they do not need that context.
They have already implemented against the spec and schema and a platform should NEVER serve back a wider collection. It should only prune as the business and platform communicate.

source/schemas/shopping/payment.json
raginpirate: I just made all of payments optional; I don't see a reason to require pushing or exposing payments if theres nothing in it right now.

docs/documentation/schema-authoring.md
```diff
@@ -16,172 +16,237 @@
...
 |-------|----------|---------|--------------|
-| `$schema` | JSON Schema | Declares JSON Schema draft version (**SHOULD** use `draft/2020-12`) | All schemas |
+| `$schema` | JSON Schema | Declares JSON Schema draft version (`draft/2020-12`) | All schemas |
 | `$id` | JSON Schema | Schema's canonical URI for `$ref` resolution | All schemas |
 | `title` | JSON Schema | Human-readable display name | All schemas |
 | `description` | JSON Schema | Schema purpose and usage | All schemas |
-| `name` | UCP | Reverse-domain capability identifier | Capability schemas only |
-| `version` | UCP | Capability version (YYYY-MM-DD format) | Capability schemas only |
+| `name` | UCP | Reverse-domain identifier; doubles as registry key | Capabilities, services, handlers |
+| `version` | UCP | Entity version (`YYYY-MM-DD` format) | Capabilities, services, handlers |
```
knightlin-shopify: nit: payment handlers

docs/documentation/schema-authoring.md
```diff
@@ -16,172 +16,237 @@
...
-schema, it should be able to determine exactly what capability and version it
-represents without cross-referencing other documents.
+UCP schemas fall into six categories based on their role in the protocol.
 
-### Why self-describing?
+### Capability Schemas
 
-1. **Independent versioning**: Capabilities **MAY** version independently. The
-   schema must declare its version explicitly.
+Define negotiated capabilities that appear in `ucp.capabilities{}` registries.
```
knightlin-shopify: nit: `registry`  as sigular?

docs/documentation/schema-authoring.md
```diff
@@ -16,172 +16,237 @@
...
 
-2. **Validation**: Validators can cross-check that a capability declaration's
-   `schema` URL points to a schema whose embedded `name`/`version` match the
-   declaration. Mismatches are authoring errors caught at build time.
+- **File-level fields**: `$schema`, `$id`, `title`, `description`, `name`, `version`
```
knightlin-shopify: `File-level fields` is not JSON schema term. Those are the `schema keywords` declared. Or, we are talking about `schema metadata keywords`?

source/discovery/profile_schema.json
```diff
@@ -1,74 +1,93 @@
...
+					"description": "Key usage. Should be 'sig' for signing keys."
+				},
+				"alg": {
+					"type": "string",
+					"description": "Algorithm (e.g., 'ES256', 'RS256')."
+				}
+			}
+		},
+
+		"platform_profile": {
```
igrigorik: Hmm, can we pull this out into separate convo and PR? 

As implemented, we lost properties and there isn't a default schema to validate against: we only have defs. We need to think through more carefully on how we want to do this, if we want this separation.

docs/playground.md
```diff
@@ -419,99 +419,113 @@ const UcpData = {
...
-              allowed_auth_methods: ["PAN_ONLY", "CRYPTOGRAM_3DS"],
-              allowed_card_networks: ["VISA", "MASTERCARD"]
-            },
-            tokenization_specification: {
-              type: "PAYMENT_GATEWAY",
+  payment_handlers: {
+    "com.shopify.shop_pay": [
+      {
+        id: "shop_pay",
+        version: "2025-12-08",
```
knightlin-shopify: nit: `2026-01-11` to align with the version in the URLs

source/schemas/shopping/payment.json
igrigorik: To confirm, we're saying: handlers moved up and out into ucp{} object, so what we have left is optional instruments. If so, this then conform to the pattern we apply everywhere else: if all fields are optional, the parent object is also optional.

docs/documentation/schema-authoring.md
```diff
@@ -16,172 +16,237 @@
...
 | Field | Standard | Purpose | Required For |
 |-------|----------|---------|--------------|
-| `$schema` | JSON Schema | Declares JSON Schema draft version (**SHOULD** use `draft/2020-12`) | All schemas |
+| `$schema` | JSON Schema | Declares JSON Schema draft version (`draft/2020-12`) | All schemas |
 | `$id` | JSON Schema | Schema's canonical URI for `$ref` resolution | All schemas |
 | `title` | JSON Schema | Human-readable display name | All schemas |
 | `description` | JSON Schema | Schema purpose and usage | All schemas |
-| `name` | UCP | Reverse-domain capability identifier | Capability schemas only |
-| `version` | UCP | Capability version (YYYY-MM-DD format) | Capability schemas only |
+| `name` | UCP | Reverse-domain identifier; doubles as registry key | Capabilities, services, handlers |
```
knightlin-shopify: So we decided to have both `name` as property and registry key?

source/discovery/profile_schema.json
raginpirate: Its really tough to rip this out of the PR: core UCP already has this improperly typed expectation about split profiles advertised through this object, so it impacts basically every part of this PR in terms of shape.

I took a stab at just applying your simple base class proposal this added commit, I think it addresses your concerns pretty simply: https://github.com/Universal-Commerce-Protocol/ucp/pull/49/changes/b8d7ec1f305a3024f047a529b9cfa56eb2dd9ec0

However if you think there's room to debate here, maybe you're intention was for me to isolate this change ONLY at the profile schema level, which I can pop out to another PR. LMK 👍

docs/documentation/schema-authoring.md
```diff
@@ -16,172 +16,237 @@
```
drewolson-google, line 16: It seems a lot of the documentation changes are unrelated to the goal of this PR. Might be better to split out these changes to another PR as this is already quite large.

docs/specification/checkout-mcp.md
```diff
@@ -30,30 +30,44 @@ Businesses advertise MCP transport availability through their UCP profile at
...
-      {
-        "name": "dev.ucp.shopping.fulfillment",
-        "version": "2026-01-11",
-        "spec": "https://ucp.dev/specification/fulfillment",
-        "schema": "https://ucp.dev/schemas/shopping/fulfillment.json",
-        "extends": "dev.ucp.shopping.checkout"
-      }
-    ]
+    "capabilities": {
+      "dev.ucp.shopping.checkout": [
```
igrigorik, line 45: I think we lost validation on keys in this refactor? Can you update to enforce?

``` 
"$defs": {
    "reverse_domain_name": {
      "type": "string",
      "pattern": "^[a-z][a-z0-9]*(?:\\.[a-z][a-z0-9_]*)+$",
      "description": "Reverse-domain identifier (e.g., com.google.pay, dev.ucp.shopping.checkout)"
    },

    "platform_schema": {
      "properties": {
        "capabilities": {
          "propertyNames": { "$ref": "#/$defs/reverse_domain_name" },
          // ...
        }
      }
    }
  }
```
And all the rest...

docs/documentation/schema-authoring.md
```diff
@@ -16,172 +16,237 @@
...
+    ]
+  },
+  "payment_handlers": {
+    "com.google.pay": [{"id": "gpay_1", "version": "2026-01-11"}]
+  }
+}
 ```
 
-This indicates which specification version the schema implements, enabling:
+**Why registries?**
```
igrigorik: FWIW, I don't think is necessary in this doc.

docs/documentation/schema-authoring.md
```diff
@@ -16,172 +16,237 @@
...
+| `schema` | URI | JSON Schema URL |
+| `id` | string | Instance identifier (handlers only) |
+| `config` | object | Entity-specific configuration |
 
-### Component Schemas
+### Schema Variants
 
-Schemas that define data structures embedded within capabilities but are not
-independently negotiated.
+Each entity type defines **three variants** for different contexts:
```
igrigorik, line 200: Missing the third..

docs/documentation/schema-authoring.md
```diff
@@ -16,172 +16,237 @@
...
-| `name` | UCP | Reverse-domain capability identifier | Capability schemas only |
-| `version` | UCP | Capability version (YYYY-MM-DD format) | Capability schemas only |
+| `name` | UCP | Reverse-domain identifier; doubles as registry key | Capabilities, services, handlers |
+| `version` | UCP | Entity version (`YYYY-MM-DD` format) | Capabilities, services, handlers |
+| `id` | UCP | Instance identifier for multiple configurations | Payment handlers only |
 
-## Self-Describing Schemas
+## Schema Categories
 
-**Capability schemas MUST be self-describing.** When a platform fetches a
```
igrigorik, line 38: Updated version lost the context of "why" behind the design: why self describing, why name+id are required, etc.

docs/documentation/schema-authoring.md
igrigorik: Agree, it's a big one. That said, splitting means the doc will be factually wrong / show wrong schemas. I'd prefer to land correct & self-contained updates, without incurring forward debt. 

@raginpirate ptal at the comments in my review.. let's be more surgical.

drewolson-google: +1 on including the relevant doc changes, but there seems to be a lot of unrelated wordsmithing as well. As long as the docs are relevant I feel good about it.

igrigorik:
For those following along, a tl;dr of what this PR achieves...

**Before:**
```
  checkout
  ├── ucp
  │   ├── version
  │   └── services{}
  │   └── capabilities[]
  ├── payment
  │   ├── handlers[]        ← 
  │   └── instruments[]
  └── ...
```
Earlier versions of the spec had ~instruments modelled under handlers, hence it made sense to expose it on checkout object. But after we split and moved instruments as sibling, handlers are now back to ~handshake.

**After this PR lands:**
```
  checkout
  ├── ucp
  │   ├── version
  │   ├── services{}
  │   ├── capabilities{}
  │   └── payment_handlers{}  ← promoted to top-level
  ├── payment
  │   └── instruments[]     ← handlers removed
  └── ...
```

Three key improvements from this refactor:
1. Services, capabilities, and payment handlers are now peers under `ucp.*` with identical structure: reverse-domain keys mapping to arrays of versioned entities.
2. Handlers are correctly reflected as protocol-level metadata alongside capabilities and services.
3. Discovery profiles now have explicit split between`platform_schema`, `business_schema`, `response_schema`, allowing proper validation.

drewolson-google (approved):

yanheChen:
[non-blocking qq]: I saw that we have `response_schema` and also we split the payment_handlers out from the payment. 

during the checkout phase when short-listing the handlers, did we plan to put the payment_handlers within payment again? or we plan to put a payment_handlers object beside the payment object. It's not very clear which path it is and I want to get clarification here.

igrigorik (commented):
@raginpirate doing Nth pass.. did we lose `instrument_schemas` and `config_schema` on payment_handler? I don't see it defined anywhere in the schema.

source/schemas/ucp.json
```diff
@@ -8,59 +8,182 @@
...
+                  "$ref": "payment_handler.json#/$defs/business_schema"
+                }
+              }
+            }
+          }
+        }
+      ]
+    },
+
+    "response_checkout_schema": {
```
lemonmade, line 147: I find the `ucp` object is quite unlike other objects in the schema, where it has two distinct "variants" that other files reference, in addition to a bunch of other helper schema shapes. To me, the separation between checkout vs order is part of what made the old structure (nested under `checkout.payment`) more natural to me — handlers only make sense in the context of some resources, but defining them on `ucp` means we must jump through some hoops to omit them where they feel awkward.

docs/specification/payment-handler-guide.md
```diff
@@ -161,95 +161,356 @@ HANDLER_DECLARATION(prerequisites_output) → handler_declaration
...
+#### Handler Declaration Variants
+
+The `PaymentHandler` schema defines three variants for different contexts. While only
+`id` and `version` are technically required, each variant serves a distinct purpose
+and typically includes different configuration:
+
+| Variant | Context | Purpose |
+|:--------|:--------|:--------|
+| **business_schema** | Business discovery (`/.well-known/ucp`) | Declares the business identity and how they're configured for this handler. Contains merchant-specific settings. |
+| **platform_schema** | Platform discovery (`/.well-known/ucp`) | Declares the platform identity and how it supports this handler. Includes `spec` and `schema` URLs for implementers. |
```
lemonmade: Is there an expectation that this will be a well-known URL? The limited examples I saw in the docs did not use `.well-known`, and I had assumed this might be a dynamically-chosen profile based on the business.

source/schemas/ucp.json
yanheChen: qq: how did we envision this matching back our schema authoring part (capabilites / components/ type / meta)? are we saying that we allow schema to be in two categories at the same time?

raginpirate: @yanheChen hmm not sure I understand but the schema authoring guide was heavily updated in this PR!

@lemonmade if we think, after going through this exercise of attempting unification, that the shape of payment handlers being identical but placement shifting just beneath ucp is a perfect outcome to me! Happy to sync on this one 😄

docs/specification/payment-handler-guide.md
raginpirate: Hmm missed this when auditing; totally right the description was just mirrored from the business schema everywhere 😢

raginpirate:
👋 PR is refreshed. Also sharing an example of the diff to payment handler schemas after this, Shop Pay as an example.
https://github.com/Universal-Commerce-Protocol/ucp/commit/ba7f800c2c19072050a02d969477510e42d10116#diff-515511ea32bd431b8b3094a1d51eb3e70b9607cc9c45c48eef2916c29eb52151

igrigorik (approved):
Phew, this is a big one -- nice work @raginpirate and ty for your patience on it! 

LGTM

knightlin-shopify (approved):

PR #51, opened 2026-01-14
feat: add AP2 mandate as an extension to `Checkout` object

lemonmade:
# Description

This PR updates the AP2 extension schema to add its `ap2` fields as an extension of `Checkout`, with UCP request annotations. This brings it in line with the other extensions like fulfillments and discounts, which mark their additional fields on checkout in this same way. Previously, `ap2` was just added to special, isolated types, and so was not referenced from any of the transport schemas. By adding it to `checkout`, it will be exposed in all transports and, once https://github.com/Universal-Commerce-Protocol/ucp/pull/28 merges, in all UCP operations.

## Type of change

      or fields**)
- [ ] Documentation update

## Checklist:

### DIFF
--- a/docs/specification/ap2-mandates.md
+++ b/docs/specification/ap2-mandates.md
@@ -235,7 +235,7 @@ a completion request without mandates **MUST** result in a session failure.
 The platform initiates the session. The business returns the `Checkout` object
 with `ap2.merchant_authorization` embedded in the response body.
 
-{{ extension_schema_fields('ap2_mandate.json#/$defs/checkout_response_with_ap2', 'ap2-mandates') }}
+{{ extension_schema_fields('ap2_mandate.json#/$defs/checkout', 'ap2-mandates') }}
 
 **Example Response:**
 ```json
@@ -321,7 +321,7 @@ Binding (+kb) signature.
 Once the mandates are generated, the platform submits them in the completion
 request:
 
-{{ extension_schema_fields('ap2_mandate.json#/$defs/complete_request_with_ap2', 'ap2-mandates') }}
+{{ extension_schema_fields('ap2_mandate.json#/$defs/ap2_with_checkout_mandate', 'ap2-mandates') }}
 
 ```json
 {
@@ -347,7 +347,7 @@ request:
     ]
   },
   "ap2": {
-     "checkout_mandate": "eyJhbGciOiJFUzI1NiIsInR5cCI6InZjK3NkLWp3dCJ9..." // The User-Signed SD-JWT+kb / platform provider signed SD-JWT / delegated SD-JWT-KB
+    "checkout_mandate": "eyJhbGciOiJFUzI1NiIsInR5cCI6InZjK3NkLWp3dCJ9..." // The User-Signed SD-JWT+kb / platform provider signed SD-JWT / delegated SD-JWT-KB
   }
 }
 ```
@@ -409,19 +409,19 @@ checkout.
 
 ### AP2 Checkout Response
 
-The `ap2` object included in CREATE / UPDATE checkout responses.
+The `ap2` object included in checkout responses.
 
-{{ extension_schema_fields('ap2_mandate.json#/$defs/ap2_checkout_response', 'ap2-mandates') }}
+{{ extension_schema_fields('ap2_mandate.json#/$defs/ap2_with_merchant_authorization', 'ap2-mandates') }}
+
+### Checkout Mandate
+
+{{ extension_schema_fields('ap2_mandate.json#/$defs/checkout_mandate', 'ap2-mandates') }}
 
 ### AP2 Complete Request
 
 The `ap2` object included in COMPLETE checkout requests.
 
-{{ extension_schema_fields('ap2_mandate.json#/$defs/ap2_complete_request', 'ap2-mandates') }}
-
-### Checkout Mandate
-
-{{ extension_schema_fields('ap2_mandate.json#/$defs/checkout_mandate', 'ap2-mandates') }}
+{{ extension_schema_fields('ap2_mandate.json#/$defs/ap2_with_checkout_mandate', 'ap2-mandates') }}
 
 ### Error Codes
 
--- a/spec/schemas/shopping/ap2_mandate.create_req.json
+++ b/spec/schemas/shopping/ap2_mandate.create_req.json
@@ -1,80 +1,48 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://ucp.dev/schemas/shopping/ap2_mandate.json",
+  "$id": "https://ucp.dev/schemas/shopping/ap2_mandate.create_req.json",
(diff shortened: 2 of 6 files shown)


docs/specification/ap2-mandates.md
```diff
@@ -342,14 +342,16 @@ request:
         "token": "examplePaymentMethodToken"
       }
   },
-  "ap2": {
-     "checkout_mandate": "eyJhbGciOiJFUzI1NiIsInR5cCI6InZjK3NkLWp3dCJ9..." // The User-Signed SD-JWT+kb / platform provider signed SD-JWT / delegated SD-JWT-KB
+  "checkout": {
```
lemonmade: This will need additional updates once https://github.com/Universal-Commerce-Protocol/ucp/pull/28 lands, since `payment_data` will also be nested under this object.

docs/specification/ap2-mandates.md
```diff
@@ -405,19 +407,19 @@ checkout.
...
 ### AP2 Checkout Response
 
-The `ap2` object included in CREATE / UPDATE checkout responses.
+The `ap2` object included in checkout responses.
 
-{{ extension_schema_fields('ap2_mandate.json#/$defs/ap2_checkout_response', 'ap2-mandates') }}
+{{ extension_schema_fields('ap2_mandate.json#/$defs/ap2_with_merchant_authorization', 'ap2-mandates') }}
 
-### AP2 Complete Request
+### Checkout Mandate
```
lemonmade, line 416: Reordered this for symmetry with the above merchant authorization (show the token type, then the object with the token).

source/schemas/shopping/ap2_mandate.json
```diff
@@ -21,32 +21,56 @@
...
-      "required": ["checkout_mandate"],
+      "description": "AP2 extension data including checkout mandate.",
       "properties": {
         "checkout_mandate": {
           "$ref": "#/$defs/checkout_mandate",
-          "description": "SD-JWT+kb proving user authorized this checkout."
+          "description": "SD-JWT+kb proving user authorized this checkout.",
+          "ucp_request": {
+            "create": "omit",
+            "update": "omit"
```
lemonmade: Once https://github.com/Universal-Commerce-Protocol/ucp/pull/28 lands and adds `complete` as an allowed UCP operation name, we will add `"complete": "required"` here, as my understanding is that the desire is to make the checkout mandate required on complete requests.

docs/specification/ap2-mandates.md
```diff
@@ -342,14 +342,16 @@ request:
         "token": "examplePaymentMethodToken"
       }
   },
-  "ap2": {
-     "checkout_mandate": "eyJhbGciOiJFUzI1NiIsInR5cCI6InZjK3NkLWp3dCJ9..." // The User-Signed SD-JWT+kb / platform provider signed SD-JWT / delegated SD-JWT-KB
+  "checkout": {
+    "ap2": {
+      "checkout_mandate": "eyJhbGciOiJFUzI1NiIsInR5cCI6InZjK3NkLWp3dCJ9..." // The User-Signed SD-JWT+kb / platform provider signed SD-JWT / delegated SD-JWT-KB
```
drewolson-google: Do we want the `checkout_` prefix repeated here in the nested key? Perhaps `checkout.ap2.mandate` is more appropriate now, given the new namespace.

source/schemas/shopping/ap2_mandate.json
```diff
@@ -21,32 +21,56 @@
...
+          "type": "object",
+          "properties": {
+            "ap2": {
+              "allOf": [
+                { "$ref": "#/$defs/ap2_with_merchant_authorization" },
+                { "$ref": "#/$defs/ap2_with_checkout_mandate" }
+              ],
+              "ucp_request": {
+                "create": "omit",
+                "update": "omit"
```
lemonmade: We will also add `"complete": "required"` here, to mark the whole `ap2` object as allowed and required on complete. `ap2.merchant_authorization` is still marked as omitted in all operations, so it can never be provided as input.

docs/specification/ap2-mandates.md
lemonmade: I definitely like that better, happy to make the change if you folks prefer it. I wasn’t sure if you’d still want a second word for symmetry with `merchant_authorization`, though, or if you wanted to rename `merchant_` to `business_` in line with other spec language changes.

yanheChen: That is an interesting point. In our AP2, there are two technical terms that has been defined during the flow and they are checkout_mandate and payment_mandate. so if we want to have the name match, I would recommend us to leave it as it.

docs/specification/ap2-mandates.md
yanheChen: If I understand correctly, we would not have payment_data anymore right?

it will be looks like
```
{
    "payment": {
        instruments: [ ... ]
    },
    "ap2.checkout_mandate": xxxxxx, 
}
```

docs/specification/ap2-mandates.md
drewolson-google: Got it, in that case let's just leave as is, thanks @yanheChen!

source/schemas/shopping/ap2_mandate.json
yanheChen: yes. if we ap2 has turned on in the flow, this field will be required for complete endpoint

docs/specification/ap2-mandates.md
yanheChen: small nits: I felt this one might need to be based on the huge one and not the origin/main

spec/schemas/shopping/ap2_mandate.create_req.json
```diff
@@ -0,0 +1,59 @@
...
+      "properties": {}
+    },
+    "checkout": {
+      "title": "Checkout with AP2 Mandate Create Request",
+      "description": "Checkout extended with AP2 mandate support.",
+      "allOf": [
+        {
+          "$ref": "checkout.create_req.json"
+        },
+        {
```
igrigorik, line 38: This is a schema generator bug we shoud fix. If all properties are omitted, we should not be emitting the object into the schema.

docs/specification/ap2-mandates.md
lemonmade: Yes, `ap2` and `payment_data=>payment` will be siblings. I will do the rebase after whichever of these merges first to achieve that structure 👍

spec/schemas/shopping/ap2_mandate.create_req.json
lemonmade: Yes, I had this in the original version, but it is an existing issue and fixing it was not as simple as I had hoped. @yanheChen I believe you mentioned you would address this in a follow-up?

yanheChen: yes i will address this as a follow-up. thanks for the eagle eyes here!

PR #54, opened 2026-01-15
refactor!: restructure payment instruments

raginpirate:
# Description

This PR refactors the payment instrument schema structure to improve consistency and extensibility:

**Schema Consolidation:**
I've gone ahead and removed these "base" base objects and just reclaimed the namespace for the base. Taking the same approach as handlers, capabilities, and services in https://github.com/Universal-Commerce-Protocol/ucp/pull/49 to simply associate to a base class type, and expect people to extend it, rather than creating artificial allOf patterns in the base repo for objects that implementers may not use.
- Removed `payment_instrument_base.json` and consolidated its functionality into `payment_instrument.json`
- Updated credential schemas (`card_credential.json`, `token_credential.json`) to properly extend `payment_credential.json` using `allOf` pattern

**Selection Model Change:**
We've had some discussion around this one; this was historically done to align with how fulfillment landed for their destinations. Their concern: multiselect was NOT valid, required wrapping their objects, etc. For instruments, we have the best of both worlds: multiselect is a real thing as we introduce split payments, and we are not seeing an issue with bringing selection onto this base object (unless address which is shared across fulfillment and billing). Thus, we're moving away from selected_instrument_id which makes no sense alongside multi-selectable instruments.
- Replaced `payment.selected_instrument_id` string field with `instrument.selected` boolean flag
- Instruments now carry their own selection state rather than being referenced by ID

**Display Data Restructuring:**
We had open discussion about making display context nilable on payment instruments. The original proposal was just making the brand and last4 optional as well on card, but this doesn't easily extend itself obviously to payment handler developers. Thus, @yanheChen and I propose grouping this under a display_data field on the base object, helping communicate that display should always be optional in UCP to support flows like re-submission of an existing instrument. Since we're force to make a breaking change, we might as well make the primitive better while we're at it 👍 
- Moved display-specific fields (brand, last_digits, expiry, etc.) into a nested `display_data` object
- This separates presentation concerns from core instrument data and provides cleaner extensibility for handler-specific display fields

**Documentation:**
I found a bunch of broken handler links still... I really don't know where these came from.
- Updated all examples across specification docs to reflect new structure
- Updated payment handler guide to document the new schema patterns

## Type of change

      or fields**)
- [ ] Documentation update

---

files or fields:

```

(Please provide a detailed technical and strategic rationale here for why this
breaking change or removal is necessary.)
```

These breaking changes are vital to the long-term state of UCP. See the original description for why.

---

## Checklist:


docs/specification/checkout-rest.md
```diff
@@ -141,39 +141,28 @@ All REST endpoints **MUST** be served over HTTPS with minimum TLS version
       "payment": {
         "handlers": [
           {
-            "id": "com.google.pay",
-            "name": "gpay",
+            "id": "shop_pay",
+            "name": "com.shopify.shop_pay",
             "version": "2024-12-03",
```
knightlin-shopify: nit: `2026-01-1` to match the version on the schemas' URL.

docs/specification/checkout-rest.md
```diff
@@ -287,39 +276,28 @@ so clients must include all previously set fields they wish to retain.
...
```
knightlin-shopify: nit: `2026-01-1` to match the version on the schemas' URL.

docs/specification/checkout-rest.md
```diff
@@ -498,13 +476,13 @@ type & addresses.
       "payment": {
         "handlers": [
           {
-            "id": "com.google.pay",
-            "name": "gpay",
+            "id": "gpay",
+            "name": "com.google.pay",
             "version": "2024-12-03",
```
knightlin-shopify: nit: `2026-01-1` to match the version on the schemas' URL.

knightlin-shopify: Same for all the examples.

docs/specification/checkout-rest.md
```diff
@@ -141,39 +141,28 @@ All REST endpoints **MUST** be served over HTTPS with minimum TLS version
...
-                  "parameters": {
-                    "allowed_card_networks": [
-                      "VISA",
-                      "MASTERCARD",
-                      "AMEX"
-                    ]
-                  }
-                }
-              ]
+              "merchant_id": "shop_merchant_123"
```
knightlin-shopify: shop_id

docs/specification/checkout-rest.md
```diff
@@ -287,39 +276,28 @@ so clients must include all previously set fields they wish to retain.
...
```
knightlin-shopify: shop_id

docs/specification/checkout-rest.md
```diff
@@ -708,39 +688,28 @@ Follow-up calls after initial `fulfillment` data to update selection.
...
```
knightlin-shopify: shop_id

docs/specification/checkout-rest.md
```diff
@@ -1072,39 +1045,28 @@ place to set these expectations via `messages`.
...
```
knightlin-shopify: shop_id

docs/specification/examples/business-tokenizer-payment-handler.md
```diff
@@ -187,12 +187,15 @@ Content-Type: application/json
 
 {
   "payment_data": {
```
knightlin-shopify: missing `selected` property

docs/specification/examples/encrypted-credential-handler.md
```diff
@@ -212,10 +212,12 @@ Content-Type: application/json
     "id": "instr_1",
     "handler_id": "platform_encrypted",
     "type": "card",
```
knightlin-shopify: missing `selected` property.

docs/specification/examples/platform-tokenizer-payment-handler.md
```diff
@@ -284,8 +284,10 @@ Content-Type: application/json
     "id": "instr_1",
     "handler_id": "platform_wallet",
     "type": "card",
```
knightlin-shopify: missing `selected` property.

docs/specification/overview.md
```diff
@@ -693,8 +693,10 @@ POST /checkout-sessions/{id}/complete
     "id": "pm_1234567890abc",
     "handler_id": "8c9202bd-63cc-4241-8d24-d57ce69ea31c",
     "type": "card",
```
knightlin-shopify: missing `selected` property.

generated/schema-types.ts
```diff
@@ -28,28 +28,28 @@ export type CheckoutMandate = string;
...
  */
 export type AP2ErrorCode =
-  | 'mandate_required'
-  | 'agent_missing_key'
-  | 'mandate_invalid_signature'
-  | 'mandate_expired'
-  | 'mandate_scope_mismatch'
-  | 'merchant_authorization_invalid'
-  | 'merchant_authorization_missing';
+	| "mandate_required"
```
knightlin-shopify: Is the indentation update expected in this file?

source/schemas/shopping/types/payment_instrument.json
```diff
@@ -2,10 +2,41 @@
...
+    "type": {
+      "type": "string",
+      "description": "The broad category of the instrument (e.g., 'card', 'tokenized_card'). Specific schemas will constrain this to a constant value."
+    },
+    "billing_address": {
+      "$ref": "postal_address.json",
+      "description": "The billing address associated with this payment method."
+    },
+    "credential": {
+      "$ref": "payment_credential.json"
```
knightlin-shopify, line 30: Should be `token_credential.json` instead given we are making `payment_credential.json` as the base for both `token_credential.json`  and `card_credential.json`.

knightlin-shopify: Or that should be override in `card_payment_instrument.json`?

docs/specification/checkout-rest.md
```diff
@@ -763,12 +732,14 @@ place to set these expectations via `messages`.
...
         "type": "card",
-        "brand": "mastercard",
-        "last_digits": "5678",
-        "rich_card_art": "https://cart-art-1.html",
-        "rich_text_description": "Google Pay •••• 5678",
+        "display_data": {
+          "brand": "mastercard",
+          "last_digits": "5678",
+          "rich_card_art": "https://cart-art-1.html",
+          "rich_text_description": "Google Pay •••• 5678"
```
igrigorik: What is "rich" about these? Is it markdown? In other places we do description: { {type: plain|markdown, value}. Is it worth adopting here?

docs/specification/embedded-checkout.md
```diff
@@ -903,9 +904,8 @@ submission.
 
 The host **MUST** respond with either an error, or the credential for the
 selected payment instrument. In successful responses, the host **MUST** supply a
-partial update to the `checkout` object, updating only the instrument indicated
-by `payment.selected_instrument_id` with the new `credentials` field. The
-Embedded Checkout **MUST** treat this update as a PUT-style change by entirely
+partial update to the `checkout` object, updating the instrument with `selected: true` 
+with the new `credentials` field. The Embedded Checkout **MUST** treat this update as a PUT-style change by entirely
```
igrigorik: nit: broken length / formatting.

docs/playground.md
```diff
@@ -538,8 +538,10 @@ const UcpData = {
     gpay: {
       "handler_id": "gpay",
       "type": "card",
-      "brand": "visa",
-      "last_digits": "4242",
+      "display_data": {
```
igrigorik: nit: is `_data` adding any value here? Implicit.

source/schemas/shopping/types/payment_instrument.json
```diff
@@ -2,10 +2,41 @@
...
+  "type": "object",
+  "required": [
+    "id",
+    "handler_id",
+    "type"
+  ],
+  "properties": {
+    "id": {
+      "type": "string",
+      "description": "A unique identifier for this instrument instance, assigned by the Agent."
```
igrigorik:
```suggestion
      "description": "A unique identifier for this instrument instance, assigned by the platform."
```

docs/specification/checkout-rest.md
igrigorik: If it's not carrying html or markdown, then I'd simplify to: description and card_art. BTW, the link for card art points to "cart-art" and should that be HTML or an image resource?

yanheChen: agree. I have also discussed that with mark, if we want to change that for the schema, this will be a nice one.

source/schemas/shopping/types/payment_instrument.json
raginpirate: I was trying to follow the same approach for how 1p capabilities are not setting themselves as a oneOf for the capability list for example; just keeping all the base objects separated and extended via the schemas defined. So in this case, those credential and token types that we've defined in the repo are there and ready for people to extend as needed, but not forced into the existing type path.

docs/specification/checkout-rest.md
```diff
@@ -141,39 +141,28 @@ All REST endpoints **MUST** be served over HTTPS with minimum TLS version
       "payment": {
         "handlers": [
           {
-            "id": "com.google.pay",
-            "name": "gpay",
-            "version": "2024-12-03",
-            "spec": "https://developers.google.com/merchant/ucp/guides/gpay-payment-handler",
-            "config_schema": "https://pay.google.com/gp/p/ucp/2026-01-11/schemas/gpay_config.json",
+            "id": "shop_pay",
```
yanheChen, line 146: may we follow the id agreement else where? e.g. this should be shop_pay_123 and not the name space.

source/schemas/shopping/types/payment_instrument.json
yanheChen: In this structure, the `payment_instrument` doesn't know it's supposed to use the `token_credential.json` definition; it only knows about the base `payment_credential.json` and then if we have a payment_instrument with credential field as token_credential, the credential validation will fail (but it might not break because we have "additionalProperties" below)

To make the credential field accept a token_credential (and potentially other types like card_credential later) while still validating the specific fields of that type, we need to use polymorphism.

During the implementations, the only dynamically validation works for me is when I use 
```
 "payment_credential": {
  "oneOf": [
    { "$ref": "token_credential.json" },
    { "$ref": "card_credential.json" },
    // also other handler credentials 
  ]
}
```

However, this approach could not scale well.

and then I found $dynamicAnchor and $dynamicRef

I think this might be the proper way to solve the problem here.

source/schemas/shopping/types/payment_instrument.json
lemonmade: Do we need this anymore, now that the ID is not used to indicate selection?

source/schemas/shopping/types/payment_instrument.json
```diff
@@ -2,10 +2,41 @@
...
+      "$ref": "postal_address.json",
+      "description": "The billing address associated with this payment method."
+    },
+    "credential": {
+      "$ref": "payment_credential.json"
+    },
+    "selected": {
+      "type": "boolean",
+      "description": "Whether this instrument is selected by the user."
+    },
```
lemonmade, line 35: Personally, I find it super weird to include this state field here. If we were to have payment instruments appear in other places in the schema, I doubt they would have the same field. What do you think about doing this via an `allOf` declaration on `payment.instruments` instead?

spec/schemas/shopping/types/card_credential.json
lemonmade, line 1: I couldn’t figure out how this gets into the REST/ MCP schemas, and notably it is not referenced by `card_payment_instrument`. is this just an example shape we would urge people to use if they were accepting a raw card credential?

raginpirate: thats exactly it; its referred to by the tokenizer guide.

source/schemas/shopping/types/payment_instrument.json
raginpirate: happy to isolate 😄

source/schemas/shopping/types/payment_instrument.json
raginpirate: I think that ID is going to be a solid ref for re-submitting checkouts with the same instrument or using vaulted credentials which will come after identity is added. Its also natural to UCP, so it doesn't hurt existing here 😄

source/schemas/shopping/types/payment_instrument.json
raginpirate: Hey!!! Yes you're seeing the problem I started outlining in our chat.
Just to set a foundation: even if we defined the anyOf here, when someone defines a **_new_** payment handler, if you walk the type chain down from checkout you will never find a payment handler's new types and never validate them; it will just work because of additional properties. So the solution we had before patched the current handler types in, but it didn't solve any real problem for the base of UCP.

Maybe we can align on that as a starting point: this PR did not actually harm our typing for payment handlers.

Now, lets talk about your dynamic ref and beyond instruments. I agree instruments in specific can be solved using dynamic ref because we _**have an anchor point that we intend to target**_, i.e. the payment_instruments field in checkout.

However, I'm thinking we just shouldn't solve instruments this way, _**because we aren't solving extensions/capabilities this way**_. With those, they DEMAND the ability to overwrite an existing type in it's context, without redefining the base schema or a dynamic entrypoint. So sadly, dynamic ref won't fix their issues 😢 

My opinion: lets keep this problem open and talk about it. Its possible we land on a unified ucp-extends custom rule which both handlers and capabilities use, or just not resolve this issue at all (patched via documentation explaining how we expect people to signal extending, i.e. redef checkout and describe it in your docs).

docs/playground.md
```diff
@@ -424,9 +424,9 @@ const UcpData = {
       id: "shop_pay",
       name: "com.shopify.shop_pay",
       version: "2025-12-08",
```
knightlin-shopify, line 428: nit: 2026-01-11 to match the version in the URL.

docs/specification/examples/business-tokenizer-payment-handler.md
```diff
@@ -187,12 +187,16 @@ Content-Type: application/json
...
+    "id": "instrument_1",
```
lemonmade: What is the purpose of `id` now that it’s not used for selection?

raginpirate: hmm repeating my prior answer: I think that ID is going to be a solid ref for re-submitting checkouts with the same instrument or using vaulted credentials which will come after identity is added. Its also natural to UCP, so it doesn't hurt existing here 😄

source/schemas/shopping/types/payment_instrument.json
yanheChen: I agree. then how about let's keep this open and not trying to solve it in this PR (also could keep this PR clean). when we figure out a real solution, we solve at that point?

docs/specification/overview.md
```diff
@@ -373,10 +373,10 @@ example:
         "id": "gpay",
         "name": "com.google.pay",
         "version": "2024-12-03",
```
knightlin-shopify, line 375: nit: Match the version below.

knightlin-shopify (approved):

PR #55, opened 2026-01-15
feat(catalog): Catalog Search+Lookup capabilities for product discovery

igrigorik:
## Summary

* `dev.ucp.shopping.catalog.search`: search for products using query text and filters.
* `dev.ucp.shopping.catalog.lookup`: retrieve products or variants by identifier.

These two new capabilities enable platforms to search merchant product catalogs and retrieve product/variant details. This fills the "discovery gap" before checkout—while checkout assumes you already have a variant ID, catalog enables scenarios like "find me blue running shoes under $150" that lead to cart building and purchase.

Discovery can be done via free-text search with filters and pagination, and direct product/variant lookup by ID. Both return a consistent `Product` structure containing variants, media, options, and pricing that seamlessly connects to checkout's `line_items[].item.id`.

## Motivation

UCP's checkout capability assumes the platform already knows what to buy. But agents and platforms need to discover products first—browsing catalogs, comparing options, filtering by price or category. Even when a product ID is already known (e.g., from a product feed integration), platforms need live lookups to fetch current facts—price, availability, and other details that may have changed. Without a standardized catalog interface, every integration requires custom product APIs or out-of-band discovery mechanisms—N custom integrations instead of one standardized protocol.

### Goals

* Enable product discovery via free-text search with category and price filters
* Support semantic search via optional `intent` context field
* Provide real-time product/variant lookup for cart validation and deep links
* Return variant IDs directly usable in checkout `line_items[].item.id`
* Establish consistent Product/Variant schema for cross-merchant compatibility

### Non-Goals

* Inventory management or stock updates (read-only catalog access)
* Product creation, modification, or deletion (merchant-side operations)
* Recommendations or personalization algorithms (merchant implementation detail)
* Full-text search ranking or relevance tuning (merchant implementation detail)

## Data Structures

**Product** (catalog entry):

```
├─ id, title, description, url, category
├─ price: PriceRange (min/max across variants)
├─ media[]: images, videos, 3D models (first = featured)
├─ options[]: dimensions like Size, Color
├─ variants[]: purchasable SKUs (first = featured)
├─ rating: aggregate reviews
└─ metadata: merchant-defined extensions
```

**Variant** (purchasable SKU):

```
├─ id: used as item.id in checkout
├─ sku, barcode: inventory identifiers
├─ title, price, availability
├─ selected_options[]: option values for this variant
├─ media[], rating, tags, metadata
└─ seller: optional marketplace context
```

## Key Behaviors

**Ordering Convention**: `media[]` and `variants[]` are ordered arrays. Merchants SHOULD return featured items first; platforms SHOULD treat first element as featured.
**ID Resolution**: `get_catalog_item` accepts product OR variant IDs:
* Product ID → returns product with representative variant set
* Variant ID → returns parent product with only requested variant

**Error Handling**: NOT_FOUND returns HTTP 200 with `messages[].type: "error"` (not 404). All catalog errors use `severity: "recoverable"` for programmatic handling.

## Risks and Mitigations

* **Schema Drift**: Product schemas vary wildly across merchants. *Mitigation*: Core fields are minimal and universal; `metadata` and `additionalProperties` enable merchant extensions without protocol changes.
* **Search Quality**: Free-text search results depend entirely on merchant implementation. *Mitigation*: Protocol defines interface only; merchants own ranking/relevance. `intent` field enables semantic hints without prescribing implementation.
* **Backward Compatibility**: N/A—new capability, no breaking changes to existing checkout flow.
* **Performance**: Large catalogs could return excessive data. *Mitigation*: Pagination with default 10, max 25 results; representative variant sets for products with many variants.

-----
## Type of change

## Checklist:

### DIFF
--- a/source/schemas/ucp.json
+++ b/source/schemas/ucp.json
@@ -17,6 +17,41 @@
       "description": "Reverse-domain identifier (e.g., com.google.pay, dev.ucp.shopping.checkout)"
     },
 
+    "version_constraint": {
+      "type": "object",
+      "description": "Version range requirement with minimum and optional maximum.",
+      "properties": {
+        "min": {
+          "$ref": "#/$defs/version",
+          "description": "Minimum required version (inclusive)."
+        },
+        "max": {
+          "$ref": "#/$defs/version",
+          "description": "Maximum compatible version (inclusive). When absent, no upper bound."
+        }
+      },
+      "required": ["min"],
+      "additionalProperties": true
+    },
+
+    "requires": {
+      "type": "object",
+      "description": "Version requirements for extension schemas. Declares minimum (and optionally maximum) protocol and capability versions needed for correct operation.",
+      "properties": {
+        "protocol": {
+          "$ref": "#/$defs/version_constraint",
+          "description": "Required protocol version."
+        },
+        "capabilities": {
+          "type": "object",
+          "description": "Required capability versions, keyed by capability name. Keys must be a subset of the extension's $defs keys.",
+          "propertyNames": { "$ref": "#/$defs/reverse_domain_name" },
+          "additionalProperties": { "$ref": "#/$defs/version_constraint" }
+        }
+      },
+      "additionalProperties": true
+    },
+
     "entity": {
       "type": "object",
       "description": "Shared foundation for all UCP entities.",
@@ -128,6 +163,15 @@
         {
           "required": ["services", "payment_handlers"],
           "properties": {
+            "supported_versions": {
+              "type": "object",
+              "description": "Previous protocol versions this business supports, mapped to profile URIs. Businesses that support older protocol versions SHOULD advertise each version and link to its profile. Each URI points to a complete, self-contained profile for that version. When omitted, only `version` is supported.",
+              "propertyNames": { "$ref": "#/$defs/version" },
+              "additionalProperties": {
+                "type": "string",
+                "format": "uri"
+              }
+            },
             "services": {
               "additionalProperties": {
                 "items": { "$ref": "service.json#/$defs/business_schema" }
--- a/docs/specification/overview.md
+++ b/docs/specification/overview.md
@@ -33,11 +33,21 @@ Schema notes:
 
 ## Discovery, Governance, and Negotiation
 
-UCP employs a server-selects architecture where the business (server) chooses
-the protocol version and capabilities from the intersection of both parties'
-capabilities. Both business and platform profiles can be cached by both parties,
-allowing efficient capability negotiation within the normal request/response
-flow between platform and business.
+UCP separates protocol version compatibility from capability negotiation.
+The business's profile at `/.well-known/ucp` describes capabilities for
+the protocol version it declares. Businesses that support older protocol
+versions **SHOULD** publish version-specific profiles and advertise them
+via the `supported_versions` field — a map from protocol version to
+profile URI, enabling platforms to discover the exact capabilities for a
+specific protocol version. Version lifecycle, including when to deprecate
+or remove older versions from `supported_versions`, is a business policy
+decision. The protocol does not prescribe a deprecation schedule.
+Capability negotiation follows a server-selects architecture where the
+business (server) determines the active capabilities from the
+intersection of both parties' declared capabilities. Both business and
+platform profiles can be cached by both parties, allowing efficient
+capability negotiation within the normal request/response flow between
+platform and business.
 
 ### Namespace Governance
 
@@ -264,6 +274,56 @@ This convention ensures:
 - **Verifiable**: Build-time checks can confirm each `extends` entry has a
     matching `$defs` key
 
+##### Version Requirements
+
+Extension schemas **SHOULD** declare a `requires` object (alongside
+`name`, `title`, `description`) to indicate the protocol and
+capability versions required for correct operation:
+
+```json
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://acme.com/ucp/schemas/loyalty.json",
+  "name": "com.acme.shopping.loyalty",
+  "title": "Acme Loyalty Points",
+  "requires": {
+    "protocol": { "min": "2026-01-23" },
+    "capabilities": {
+      "dev.ucp.shopping.checkout": { "min": "2026-06-01" }
+    }
+  },
+  "$defs": {
+    "dev.ucp.shopping.checkout": { ... }
+  }
+}
+```
+
+The schema author — not the profile publisher — declares version
+requirements. The profile publisher selects and advertises compatible
+versions in their profile.
+
+Each constraint is an object with a required `min` (inclusive) and
+optional `max` (inclusive) version. When `max` is absent, there is
+no upper bound:
+
+```json
+"requires": {
+  "protocol": { "min": "2026-01-23", "max": "2026-09-01" },
+  "capabilities": {
+    "dev.ucp.shopping.checkout": { "min": "2026-06-01" }
+  }
+}
+```
+
+Keys in `requires.capabilities` **MUST** be a subset of the
+extension's `$defs` keys. If `requires` is present, platforms and
+businesses **MUST** verify the negotiated protocol version and
+capability versions satisfy the declared constraints during schema
+resolution. Incompatible extensions are excluded from the active
+capability set (see [Resolution Flow](#resolution-flow)). If
+`requires` is absent, the extension is assumed to be compatible
+with the versions declared by the profile.
+
 #### Schema Resolution Convention
 
 To validate payloads, implementations resolve extension schemas as follows:
@@ -286,8 +346,13 @@ Platforms **MUST** resolve schemas following this sequence:
 2. **Negotiation**: Compute capability intersection (see
     [Intersection Algorithm](#intersection-algorithm))
 3. **Schema Fetch**: Fetch base schema and all active extension schemas
-4. **Compose**: Merge schemas via `allOf` chains based on active extensions
-5. **Validate**: Validate requests and responses against the composed schema
+4. **Version Compatibility**: For each fetched extension schema,
+    if `requires` is present, verify the negotiated protocol version
+    and capability versions satisfy the declared constraints. Exclude
+    incompatible extensions and re-prune orphaned extensions
+    (steps 3-4 of the [Intersection Algorithm](#intersection-algorithm))
+5. **Compose**: Merge schemas via `allOf` chains based on active extensions
+6. **Validate**: Validate requests and responses against the composed schema
 
 ### Profile Structure
 
@@ -401,6 +466,11 @@ and payment handlers. The `signing_keys` array contains public keys (JWK format)
 used to verify signatures on webhooks and other authenticated messages from the
 business. See [Key Discovery](#key-discovery) for key lookup and resolution,
 and [Message Signatures](signatures.md) for signing mechanics.
+
+Businesses that support older protocol versions **SHOULD** include a
+`supported_versions` object mapping each older version to a
+version-specific profile URI. See [Protocol Version](#protocol-version)
+for details.
 
 #### Platform Profile
 
@@ -574,17 +644,23 @@ for a session:
 1. **Compute intersection**: For each business capability, include it in the
     result if a platform capability with the same `name` exists.
 
-2. **Prune orphaned extensions**: Remove any capability where `extends` is
+2. **Select version**: For each capability in the intersection, compute the
+    set of version strings present in **both** the business and platform
+    arrays. If the set is non-empty, select the **highest** version
+    (latest date). If the set is empty (no mutual version), **exclude** the
+    capability from the intersection.
+
+3. **Prune orphaned extensions**: Remove any capability where `extends` is
     set but **none** of its parent capabilities are in the intersection.
     - For single-parent extensions (`extends: "string"`): parent must be present
     - For multi-parent extensions (`extends: ["a", "b"]`): at least one parent
         must be present
 
-3. **Repeat pruning**: Continue step 2 until no more capabilities are removed
+4. **Repeat pruning**: Continue step 3 until no more capabilities are removed
     (handles transitive extension chains).
 
-The result is the set of capabilities both parties support, with extension
-dependencies satisfied.
+The result is the set of capabilities both parties support at mutually
+compatible versions, with extension dependencies satisfied.
 
 #### Error Handling
 
@@ -601,8 +677,8 @@ not be retrieved or were malformed. Negotiation failures are business
 outcomes — the handler executed on the provided inputs and reported
 the result in the UCP response:
 
-- **Discovery failure** → transport error with optional `continue_url`
-- **Negotiation failure** → UCP response with optional `continue_url`
+- **Discovery or version failure** → transport error with optional `continue_url`
+- **Capability negotiation failure** → UCP response with optional `continue_url`
 
 ##### Error Codes
 
@@ -613,8 +689,8 @@ the result in the UCP response:
 | `invalid_profile_url`       | Profile URL is malformed, missing, or unresolvable   | 400  | -32001 |
 | `profile_unreachable`       | Resolved URL but fetch failed (timeout, non-2xx)     | 424  | -32001 |
 | `profile_malformed`         | Fetched content is not valid JSON or violates schema | 422  | -32001 |
+| `version_unsupported`       | Platform's protocol version not supported            | 422  | -32001 |
 | `capabilities_incompatible` | No compatible capabilities in intersection           | 200  | result |
-| `version_unsupported`       | Platform's UCP version is not supported              | 200  | result |
 
 **Signature Errors:**
 
@@ -673,7 +749,20 @@ task through the standard web interface.
     }
     ```
 
-    **Negotiation Failure (200):**
+    **Version Unsupported (422):**
+
+    ```http
+    HTTP/1.1 422 Unprocessable Content
+    Content-Type: application/json
+
+    {
+      "code": "version_unsupported",
+      "content": "Protocol version 2026-01-12 is not supported. This business supports versions 2026-01-11 and 2026-01-23.",
+      "continue_url": "https://merchant.com/cart"
+    }
+    ```
+
+    **Capabilities Incompatible (200):**
 
     ```http
     HTTP/1.1 200 OK
@@ -730,7 +819,25 @@ task through the standard web interface.
     }
     ```
 
-    **Negotiation Failure (JSON-RPC result):**
+    **Version Unsupported (JSON-RPC error):**
+
+    ```json
+    {
+      "jsonrpc": "2.0",
+      "id": 1,
+      "error": {
+        "code": -32001,
+        "message": "Protocol version not supported",
+        "data": {
+          "code": "version_unsupported",
+          "content": "Protocol version 2026-01-12 is not supported. This business supports versions 2026-01-11 and 2026-01-23.",
+          "continue_url": "https://merchant.com/cart"
+        }
+      }
+    }
+    ```
+
+    **Capabilities Incompatible (JSON-RPC result):**
 
     ```json
     {
@@ -1639,16 +1746,78 @@ Both businesses and platforms declare a single version in their profiles:
 
 ![High-level resolution flow sequence diagram](site:specification/images/ucp-discovery-negotiation.png)
 
-Businesses **MUST** validate the platform's version and determine compatibility:
+Version compatibility operates at two levels: the **protocol version**
+and **capability versions**. The protocol version (`ucp.version`)
+governs core protocol mechanisms — discovery, negotiation flow,
+transport bindings, and signature requirements. Capability versions
+govern the semantics of each feature independently, as defined in
+[Independent Component Versioning](#independent-component-versioning).
 
-1. Platform declares version via profile referenced in request
+#### Protocol Version
+
+The `version` field declares the business's current protocol version.
+The profile at `/.well-known/ucp` describes the capabilities, services,
+and payment handlers available at that version.
+
+Businesses that support older protocol versions **SHOULD** declare a
+`supported_versions` object mapping each older version to a profile
+URI. Each URI points to a complete, self-contained profile for that
+version — including its own capabilities, services, payment handlers,
+and signing keys. When `supported_versions` is omitted, only
+`version` is supported.
+
+```json
+{
+  "ucp": {
+    "version": "2026-01-23",
+    "supported_versions": {
+      "2026-01-11": "https://business.example.com/.well-known/ucp/2026-01-11"
+    }
+  }
+}
+```
+
+##### Initial Service and Capability Discovery
+
+Platforms discover a business's capabilities through the following flow:
+
+1. Platform fetches `/.well-known/ucp` — this is the current version
+    profile.
+2. If the platform's protocol version matches `version`: use this
+    profile directly. Proceed to capability negotiation.
+3. If the platform's protocol version is a key in
+    `supported_versions`: fetch the profile at the mapped URI. This
+    profile describes the capabilities available at that protocol
+    version. Proceed to capability negotiation.
+4. Otherwise: the business does not support the platform's protocol
+    version. Platforms **SHOULD NOT** send requests with an incompatible
+    version; businesses **MUST** respond with a `version_unsupported`
+    error.
+
+Version-specific profiles are leaf documents — they describe exactly
+one protocol version and **MUST NOT** contain a `supported_versions`
+field.
+
+##### Request-Time Validation
+
+Businesses **MUST** validate the platform's protocol version on
+every request:
+
+1. Platform declares the protocol version it uses via the
+    `version` field in the profile referenced in the request.
 2. Business validates:
-    - If platform version ≤ business version: Business **MUST**
-        process the request
-    - If platform version > business version: Business **MUST** return
-        `version_unsupported` error
-3. Businesses **MUST** include the version used for processing in every
-    response.
+    - If the platform's `version` matches the business's `version`
+        or is a key in `supported_versions`: the request **MAY**
+        proceed to capability negotiation using the matching
+        version of the business profile.
+    - Otherwise: Business **MUST** return a `version_unsupported`
+        error.
+3. If capability negotiation yields no mutually supported version
+    for a capability required by the requested operation, the
+    business **MUST** return a `capabilities_incompatible` error
+    (see [Error Handling](#error-handling)).
+4. Businesses **MUST** include the negotiated protocol version in
+    every response.
 
 Response with version confirmation:
 
@@ -1679,6 +1848,20 @@ Version unsupported error — no resource is created:
   "continue_url": "https://merchant.com/"
 }
 ```
+
+#### Capability Versions
+
+Capability versions are negotiated independently of the protocol
+version. Each capability in the profile is an array. Multiple entries
+for the same capability, each with a different `version`, advertise
+support for multiple versions of that capability. The capability
+intersection algorithm considers only capability versions supported
+by both parties.
+
+Businesses **MUST** include only capabilities compatible with the
+negotiated protocol version in their response. A capability that
+depends on features introduced in a newer protocol version **MUST
+NOT** be included when processing at an older protocol version.
 
 ### Backwards Compatibility
 


docs/specification/catalog.md
```diff
@@ -0,0 +1,275 @@
...
+
+{{ extension_schema_fields('catalog.json#/$defs/context', 'catalog') }}
+
+### Product
+
+A catalog entry representing a sellable item with one or more purchasable variants.
+
+`media` and `variants` are ordered arrays. Businesses SHOULD return the featured
+image and default variant as the first element. Platforms SHOULD treat the first
+element as the featured item for display.
```
knightlin-shopify: Clarify that `media`/`variants` ordering should be query-aware in search results (e.g., "yellow shoes" → yellow variant first), but default-based for direct lookups?

docs/specification/catalog.md
```diff
@@ -0,0 +1,275 @@
...
+## Messages and Error Handling
+
+All catalog responses include an optional `messages` array that allows businesses
+to provide context about errors, warnings, or informational notices.
+
+### Message Types
+
+Messages communicate business outcomes and provide context:
+
+| Type | When to Use | Example Codes |
```
knightlin-shopify: The code defined here uses `UPPER_SNAKE_CASE` for error codes, but the schema examples use `snake_case`.

docs/specification/catalog.md
```diff
@@ -0,0 +1,275 @@
...
+
+#### Message (Info)
+
+{{ schema_fields('types/message_info', 'catalog') }}
+
+### Common Scenarios
+
+**Empty Search**
+
+When search finds no matches, return an empty array without messages.
```
knightlin-shopify: Should we return a more explicit `info` message with the proper `code` (`not_found`)? This would make it clearer for the agent/platform to handle and avoid retrying the request.

docs/specification/catalog-rest.md
```diff
@@ -0,0 +1,309 @@
...
+| Status | Meaning |
+| :--- | :--- |
+| 400 | Bad Request - Malformed JSON or missing required parameters |
+| 401 | Unauthorized - Missing or invalid authentication |
+| 429 | Too Many Requests - Rate limited |
+| 500 | Internal Server Error |
+
+### Business Outcomes
+
+All application-level outcomes return HTTP 200 with the UCP envelope and optional
```
knightlin-shopify: GET by ID is resource retrieval - if the resource doesn't exist, the URI is invalid → should return HTTP 404.

- `POST /catalog/search` with no matches -> 200 OK (query succeeded, empty results)
- `GET /catalog/item/{invalid_id}` -> 404 Not Found (resource doesn't exist)
- `GET /catalog/item/{valid_id}` that's out of stock -> 200 + message (resource exists, business constraint)

docs/specification/catalog-mcp.md
```diff
@@ -0,0 +1,373 @@
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "search_catalog",
+  "params": {
+    "_meta": {
+      "ucp": {
+        "profile": "https://platform.example/profiles/v2026-01/shopping-agent.json"
```
knightlin-shopify: Are we expecting the version info should be advertise in the URI path, after `profiles`? And the version format should be `YYYY-MM-DD` format instead?

docs/specification/catalog.md
```diff
@@ -0,0 +1,275 @@
...
+This is not an error—the query was valid but returned no results.
+
+**Backorder Warning**
+
+When a product is available but has delayed fulfillment, return the product with a warning message. Use the `path` field to target specific variants.
+
+```json
+{
+  "ucp": {...},
+  "product": {
```
knightlin-shopify: nit: Add required fields `description` and `price`.

docs/specification/catalog-mcp.md
```diff
@@ -0,0 +1,373 @@
...
+        "query": "blue running shoes",
+        "context": {
+          "country": "US",
+          "region": "CA",
+          "intent": "looking for comfortable everyday shoes"
+        },
+        "filters": {
+          "category": "Footwear",
+          "price": {
+            "max": 15000
```
knightlin-shopify: What currency this is for? Should it be explicit in the price_filter schema?

source/schemas/shopping/types/price_filter.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/price_filter.json",
+  "title": "Price Filter",
+  "description": "Price range filter in minor currency units. Currency determined by context.",
```
knightlin-shopify: It is not clear where the currency info comes from for catalog search w/ this price filter.

knightlin-shopify: Looks like price_range.json should do the job, we don't need this schema.

source/schemas/shopping/types/product.json
```diff
@@ -0,0 +1,101 @@
...
+    "description",
+    "price",
+    "variants"
+  ],
+  "properties": {
+    "id": {
+      "type": "string",
+      "description": "Global ID (GID) uniquely identifying this product."
+    },
+    "handle": {
```
knightlin-shopify, line 19: We should document its intended usage and how it relates to `id`.

source/schemas/shopping/types/rating.json
```diff
@@ -0,0 +1,28 @@
...
+    "value",
+    "scale_max"
+  ],
+  "properties": {
+    "value": {
+      "type": "number",
+      "minimum": 0,
+      "description": "Average rating value."
+    },
+    "scale_max": {
```
knightlin-shopify, line 23: Do we need `scale_min`? For example, the 1-5 stars review system.

source/schemas/shopping/types/search_filters.json
```diff
@@ -0,0 +1,17 @@
...
+  "properties": {
+    "category": {
+      "type": "string",
+      "description": "Filter by product category. Categories are merchant-defined strings representing product taxonomy. Valid values can be discovered from the category field in search results, merchant documentation, or standard taxonomies that businesses may align with."
+    },
+    "price": {
+      "$ref": "price_filter.json"
+    }
+  },
+  "additionalProperties": true
```
knightlin-shopify, line 17: I see not all newly add schema has this set, `additionalProperties` default is true, just to double check the expect value across all schema file added.

source/schemas/shopping/types/variant.json
```diff
@@ -0,0 +1,127 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/variant.json",
+  "title": "Variant",
+  "description": "A purchasable variant of a product with specific option selections.",
+  "type": "object",
+  "required": [
+    "id",
+    "title",
+    "description",
```
knightlin-shopify, line 10: Should `description` optional, which should inherit from the product it belongs to?

source/schemas/shopping/types/variant.json
```diff
@@ -0,0 +1,127 @@
...
+        }
+      },
+      "minProperties": 1
+    },
+    "url": {
+      "type": "string",
+      "format": "uri",
+      "description": "Canonical variant page URL."
+    },
+    "category": {
```
knightlin-shopify: This should be product only property?

source/schemas/shopping/types/variant.json
```diff
@@ -0,0 +1,127 @@
...
+      "description": "Option selections that define this variant."
+    },
+    "media": {
+      "type": "array",
+      "items": {
+        "$ref": "media.json"
+      },
+      "description": "Variant media (images, videos, 3D models). First item is the featured media for listings."
+    },
+    "rating": {
```
knightlin-shopify, line 134: This should be product only property?

source/schemas/shopping/types/variant.json
```diff
@@ -0,0 +1,127 @@
...
+      "$ref": "rating.json",
+      "description": "Variant rating."
+    },
+    "tags": {
```
knightlin-shopify, line 138: This should be product only property?

knightlin-shopify (commented):
What is the language expectation on the text content reposed from business? We have product and variant description and message content. Should have explicit content language expectation provided by platform/agent (through `context`), which should reflect user's display language choice (can not be determined through `context.country`). And should expect agent to display as it instead of doing the auto translation given some of field (product name, variant name, legal /safety/liability disclaimer ...) should not be translated but mixed inside the text blob.

source/schemas/ucp.json
```diff
@@ -33,9 +33,9 @@
       }
     },
 
-    "response_checkout": {
-      "title": "UCP Checkout Response",
-      "description": "UCP metadata for checkout responses.",
+    "response": {
```
raginpirate: Just to clarify how this would work once you target the working branch: are we thinking that capabilities exposed through discovery would be the same as cart and checkout, or would it be a unique subset? I'm imagining this should be isolated and not explain how I do fulfillment from product discovery.

Either way, if we choose to include all capabilities, should it really expose their configs like it does at checkout, or just negotiate versioning?

source/schemas/shopping/types/price.json
```diff
@@ -0,0 +1,23 @@
...
+    "amount": {
+      "type": "integer",
+      "description": "Amount in minor currency units (e.g., 1000 = $10.00 USD). Use 0 for free items.",
+      "minimum": 0
+    },
+    "currency": {
+      "type": "string",
+      "description": "ISO 4217 currency code (e.g., 'USD', 'EUR', 'GBP').",
+      "pattern": "^[A-Z]{3}$"
+    }
```
raginpirate, line 20: Checkout doesn't push currency on every amount; it uses bare amounts and has a top-level contextual currency. Should we be thinking the same way for catalog?

docs/specification/catalog.md
```diff
@@ -0,0 +1,275 @@
...
+| **Get Catalog Item** | Retrieve a specific product or variant by ID. |
+
+### Search Catalog
+
+Performs a search against the business's product catalog. Supports free-text
+queries, filtering by category and price, and pagination.
+
+**Use Cases:**
+
+* User searches for "blue running shoes"
```
maximenajim: The use cases include “fetch featured or trending products,” but the conformance section and schema require query. Is this supported via an empty string query?  (that’s not stated).

Suggestion: explicitly document one of:

- query MAY be "" to request default/trending/featured results, or
- introduce an explicit mode / sort / collection field, or
- allow omission of query and define behavior.

docs/specification/catalog.md
```diff
@@ -0,0 +1,275 @@
...
+buyer context or other criteria. This ensures agents always have product context
+for display while getting exactly what they requested.
+
+{{ method_fields('get_catalog_item', 'rest.openapi.json', 'catalog') }}
+
+## Entities
+
+### Context
+
+Location and market context for catalog operations. All fields are optional. Platforms MAY geo-detect context from request IP/headers. When context fields are provided, they MUST override any auto-detected values.
```
maximenajim: Platforms MAY geo-detect… provided fields MUST override any auto-detected values”. If interpreted literally, it encourages merchants to trust platform-supplied country/region/postal_code (and now intent) even when it conflicts with authenticated user/account signals, IP geolocation, shipping destination, export controls, or fraud rules.

Concrete fix: reword to something like:
- Provided context fields are hints, not authorization; merchants MAY ignore or down-rank them if inconsistent with stronger signals.

- If merchants do accept them, they MUST still enforce eligibility at checkout/order time.

Also consider adding a normative note that context must never be used as the sole input for policy enforcement.

source/schemas/shopping/types/product.json
```diff
@@ -0,0 +1,101 @@
...
+    },
+    "description": {
+      "type": "object",
+      "description": "Product description in one or more formats. At least one format must be provided.",
+      "properties": {
+        "plain": {
+          "type": "string",
+          "description": "Plain text description."
+        },
+        "html": {
```
maximenajim: That’s useful, but it creates XSS/content-injection risk if platforms render it directly.
Suggestion: add explicit security guidance: HTML must be sanitized (no scripts/event handlers), and platforms must treat all rich text as untrusted input.

docs/specification/catalog-rest.md
```diff
@@ -0,0 +1,309 @@
...
+      ],
+      "pagination": {
+        "cursor": "eyJwYWdlIjoxfQ==",
+        "has_next_page": true,
+        "total_count": 47
+      }
+    }
+    ```
+
+### `GET /catalog/item/{id}`
```
maximenajim: The capability design depends on context (pricing/currency, availability, “representative variants” selection), but the REST binding for GET /catalog/item/{id} provides no way to supply context fields. 

Concrete fix options:
- Add query parameters country, region, postal_code, and intent (if you allow intent on GET), OR
- Change this operation to POST /catalog/item with JSON body {id, context} (less cache-friendly but consistent), OR
- Add a dedicated X-UCP-Context header carrying a structured object (needs canonicalization rules).

source/schemas/shopping/types/context.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/context.json",
+  "title": "Context",
+  "description": "Location and market context. All fields optional. Platforms MAY geo-detect from request; provided fields MUST override.",
+  "type": "object",
+  "properties": {
```
thompson-tomo, line 8: Should we have an optional `preferred currency` property. Scenario I am thinking of is places like Switzerland where Swiss franc is the official currency but euro is widely accepted hence a customer might want euro prices.

source/schemas/shopping/types/search_filters.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/search_filters.json",
+  "title": "Search Filters",
+  "description": "Filter criteria to narrow search results. All specified filters combine with AND logic.",
+  "type": "object",
+  "properties": {
```
thompson-tomo, line 7: Would be useful to have a date range search filter

Use cases would be:
- I want a flight on the morning of 30th Jan, in which case the `min` would be set to 30th 00:00 & `Max` 29th 11:00. This results in only flights between those dates to be returned
- I need the product for 30th Jan in which case the `Max` would be set to 30th. This results in only products which can be delivered before that date being returned.
- I need accommodation from 30th Jan to 5th Feb hence `min` & `Max` would be set.

source/schemas/shopping/types/product.json
```diff
@@ -0,0 +1,101 @@
...
+  "description": "A product in the catalog with variants and options.",
+  "type": "object",
+  "required": [
+    "id",
+    "title",
+    "description",
+    "price",
+    "variants"
+  ],
+  "properties": {
```
thompson-tomo, line 14: Should we add an `add-on product` property to provide a list of products which can be offered to the user? Scenario is if I buy sunglasses, an add-on would be a case for them.

This also applies to variant.

spec/schemas/shopping/catalog.json
```diff
@@ -0,0 +1,100 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/catalog.json",
+  "name": "dev.ucp.shopping.catalog",
+  "version": "2026-01-11",
+  "title": "Catalog",
+  "description": "Product catalog search and item retrieval capability.",
```
amithanda: I am thinking if we should have item retrieval capability as its own separate capability rather than bundling it together with Product catalog search in a single Catalog capability. Some reasons:
1) As an Agent platform I may want to advertise clearly that I support item retrieval (or look up) for getting latest info about a product that is already submitted offline (e.g via feeds) but not catalog search. Current structure makes it difficult to do that.
2) Search and Item retrieval operations can evolve separately as capabilities, there is no real reason for these operations to be together in a single capability unlike checkout or cart e.g. which operate on a single object.

docs/specification/catalog.md
igrigorik: Hmm, `not_found` implies a lookup to me.. i think an empty result set is fairly direct and self-explanatory.

docs/specification/catalog-rest.md
igrigorik: With REST hat on, agree, but it's inconsistent with how we defined same operations for cart/checkout, where failed lookup is a modelled and exposed at UCP layer, not transport. I agree this is ~meh, but I'll stick with consistency and we can debate whether that's a broad pattern we want to revisit.

source/schemas/shopping/types/price_filter.json
igrigorik: They look very similar but have different requirements for optional vs required. It is duplication, but we need it.

source/schemas/shopping/types/search_filters.json
igrigorik: Yep, it's implicit but made explicit here because I wanted to underscore that this is an extensible set.

source/schemas/shopping/types/variant.json
igrigorik: Other way around. A variant is the purchasable unit - it should be self-describing.

source/schemas/shopping/types/variant.json
igrigorik: True in general but has gotchas -- e.g. variant materials may affect category, a print can have different categories based on what it's printed on, etc.

source/schemas/shopping/types/variant.json
igrigorik: Same as above, different variants can have different ratings -- e.g. the leather shoe is great, the suede one is meh.

source/schemas/shopping/types/price.json
igrigorik: Ack on inconsistency. The reason it's modelled here this way is to support marketplaces which _may_ have mixed currency items. As defined, checkout doesn't have same flexibility, but you could create discrete carts/checkouts and finalize as separate baskets.

docs/specification/catalog.md
igrigorik: Good call, I'm going to remove this -- trending, recommendations, etc, is a different thing entirely.

docs/specification/catalog.md
igrigorik: 👍  updated

docs/specification/catalog-rest.md
igrigorik: Fixed 👍

source/schemas/shopping/types/search_filters.json
igrigorik: I don't think that's releant for product search as part of base schema.. but extensions are welcome to introduce this.

source/schemas/shopping/types/product.json
igrigorik: Yes, but via extension! We don't need to model it in base schema. Related products / recommendations / addons is perfect use case for extensions.

docs/specification/catalog/lookup.md
```diff
@@ -0,0 +1,53 @@
...
+
+{{ extension_schema_fields('catalog_lookup.json#/$defs/lookup_request', 'catalog') }}
+
+### Response
+
+{{ extension_schema_fields('catalog_lookup.json#/$defs/lookup_response', 'catalog') }}
+
+## Transport Bindings
+
+* [REST Binding](rest.md#get-catalogitemid): `GET /catalog/item/{id}`
```
TioBorracho: rest resources should be plural. I think catalog is ok as a feauter in the path, but here it shouldbe `items`

TioBorracho:
I left a comment on the file,but I think the rest path for get should be `items`, and not `item`.

docs/specification/catalog/lookup.md
igrigorik: Why plural when you're looking up a single ID?

spec/schemas/shopping/types/variant.json
```diff
@@ -0,0 +1,127 @@
...
+      },
+      "minProperties": 1
+    },
+    "url": {
+      "type": "string",
+      "format": "uri",
+      "description": "Canonical variant page URL."
+    },
+    "category": {
+      "type": "string",
```
alex-jansen: Would it make sense to make this extensible and potentially repeating (i.e, an array of objects)? There are multiple public taxonomies in the industry, and we might want to support the return of both a standard taxonomy value and a merchant-defined value.
@amithanda

spec/schemas/shopping/types/variant.json
```diff
@@ -0,0 +1,127 @@
...
+    "list_price": {
+      "$ref": "price.json",
+      "description": "List price before discounts (for strikethrough display)."
+    },
+    "availability": {
+      "type": "object",
+      "description": "Variant availability for purchase.",
+      "properties": {
+        "available": {
+          "type": "boolean",
```
alex-jansen: eventually we probably want to support item availabilities including preorder, backorder and built-to-order (in addition to the basic in-stock and out-of-stock). Is the idea to model this using a new property under this object or should this be an extensible enumeration of values instead of a boolean?
@amithanda

igrigorik: @alex-jansen absolutely, this is exactly why we're modelling `availability` as an open object. We can extend and add additional fields via composition, which will live as siblings to `available` boolean. As an example, see: https://github.com/Universal-Commerce-Protocol/ucp/pull/55#issuecomment-3824223394

spec/schemas/shopping/types/variant.json
igrigorik: @alex-jansen good call, added in https://github.com/Universal-Commerce-Protocol/ucp/pull/55/changes/8bbe14890c3017563f7a458176f65ccc0fec1279

docs/specification/catalog/rest.md
```diff
@@ -0,0 +1,386 @@
...
+        {
+          "id": "prod_abc123",
+          "handle": "blue-runner-pro",
+          "title": "Blue Runner Pro",
+          "description": {
+            "plain": "Lightweight running shoes with responsive cushioning."
+          },
+          "url": "https://business.example.com/products/blue-runner-pro",
+          "category": "Footwear > Running",
+          "price": {
```
koenbouwmans: should this be `price_range` to align with [schemas/types/product.json](https://github.com/Universal-Commerce-Protocol/ucp/pull/55/changes#diff-530483c79720636206684610215c65d73fb244c81d3f8993761724022d2b88f6R58)

source/schemas/shopping/types/category.json
```diff
@@ -0,0 +1,18 @@
...
+  "description": "A product category with optional taxonomy identifier.",
+  "type": "object",
+  "required": ["value"],
+  "properties": {
+    "value": {
+      "type": "string",
+      "description": "Category value or path (e.g., 'Apparel > Shirts', '1604')."
+    },
+    "taxonomy": {
+      "type": "string",
```
alex-jansen, line 14: we could consider making this an enum instead of a free-form string

docs/specification/catalog/rest.md
```diff
@@ -0,0 +1,386 @@
...
+      "products": [
+        {
+          "id": "prod_abc123",
+          "handle": "blue-runner-pro",
+          "title": "Blue Runner Pro",
+          "description": {
+            "plain": "Lightweight running shoes with responsive cushioning."
+          },
+          "url": "https://business.example.com/products/blue-runner-pro",
+          "category": "Footwear > Running",
```
alex-jansen: needs to be updated to match source/schemas/shopping/types/category.json

docs/specification/catalog/lookup.md
igrigorik: Resolving, no longer applies in light of https://github.com/Universal-Commerce-Protocol/ucp/pull/55/commits/46d1e4874ce1a234f9bec29b269bd43c3a4ef78f

source/schemas/shopping/types/category.json
igrigorik: The problem with enum in JSONSchema is that they're closed and can't be extended via our composition pattern, and we want to explicitly allow that ability here.

docs/specification/catalog/rest.md
igrigorik: Good catch, fixed

source/schemas/shopping/types/variant.json
```diff
@@ -0,0 +1,115 @@
...
+    "price": {
+      "$ref": "price.json",
+      "description": "Current selling price."
+    },
+    "list_price": {
+      "$ref": "price.json",
+      "description": "List price before discounts (for strikethrough display)."
+    },
+    "availability": {
+      "type": "object",
```
koenbouwmans, line 107: should this be pulled out as a type

amithanda:
Hi Ilya, 
Do you want to update the PR description, since we have split this into two separate capabilities?  
I am ready to approve, once you do that and address any other open comments or resolve any conficts.

source/schemas/shopping/types/category.json
gsmith85: Another, and perhaps more pernicious pitfall to consider with JSONSchema enums is that they are not forward compatible.  They are not the equivalent of Protobuf enums that are designed to be open by default.  You cannot add additional enum values without a major version rev.

source/schemas/shopping/types/context.json
```diff
@@ -2,7 +2,7 @@
...
+  "description": "Provisional buyer signals for relevance and localization—not authorization. Businesses SHOULD use these values when authoritative data (e.g., shipping address) is absent, and MAY ignore or down-rank them if inconsistent with stronger signals (authenticated account, fraud rules, export controls). Eligibility and policy enforcement MUST occur at checkout/order time with authoritative data. Context SHOULD be non-identifying and can be disclosed progressively—coarse signals early, finer resolution as the session progresses. Higher-resolution data (shipping address, billing address) supersedes context.",
```
gsmith85: A minor copy suggestion for teasing apart provisional and progressively authoritative signals:

> Provisional buyer signals for relevance and localization—not authoritative data. Businesses SHOULD use these values when verified inputs (e.g., shipping address) are absent, and MAY ignore or down-rank them if inconsistent with higher-confidence signals (authenticated account, risk detection) or regulatory constraints (export controls). Eligibility and policy enforcement MUST occur at checkout time using binding transaction data. ...

source/schemas/shopping/types/description.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/description.json",
+  "title": "Description",
+  "description": "Description content in one or more formats. At least one format must be provided.",
+  "type": "object",
+  "properties": {
+    "plain": {
```
gsmith85, line 8: **Suggestion**: Use JSONSchema's `contentMediaType`:

https://json-schema.org/understanding-json-schema/reference/non_json_data

This enforces each property's value at the schema level rather than requiring adherence to the description.

source/schemas/shopping/types/input_correlation.json
```diff
@@ -0,0 +1,19 @@
...
+  "type": "object",
+  "required": ["id"],
+  "properties": {
+    "id": {
+      "type": "string",
+      "description": "The identifier from the lookup request that resolved to this variant."
+    },
+    "match": {
+      "type": "string",
+      "enum": ["exact", "featured"],
```
gsmith85: Based on the earlier thread on enum usage, and the inability to add new enum values without breaking older clients, I'm wondering if we actually want to move away from using them if there is _any_ possibility of future supplementation?

In the future, it is easy to imagine the server logic evolving to include additional resolution strategies like:
- `fuzzy` (matched on name)
- `semantic` (matched on intent)
- `personal` (matched on user history)

Using an `enum` forces a major version revision for any of these additions because v1 clients would hard-fail upon seeing an unrecognized string. I suggest using an "Open Set" pattern instead:

```
"anyOf": [
    { "const": "exact", "description": "Input directly identifies this variant." },
    { "const": "featured", "description": "Server selected this variant as representative." },
    { "type": "string" }
  ],
```

Depending on the serialization library and language features used (e.g., Kotlin sealed classes or Jackson custom creators), this adds minimal code-overhead while ensuring that legacy clients degrade gracefully; ignoring the new signal rather than crashing the entire request.

gsmith85 (commented):
Some initial comments, primarily focused on schema.  More to come!

source/schemas/shopping/types/description.json
igrigorik: My understanding is that it's an annotation, it does not enforce any form of validation: and how would it for arbitrary content type? We need to support multi-format, we'd end up with something like...

```
  {
    "properties": {
      "plain": {
        "type": "string",
        "contentMediaType": "text/plain"
      },
      "html": {
        "type": "string",
        "contentMediaType": "text/html"
      },
      "markdown": {
        "type": "string",
        "contentMediaType": "text/markdown"
      }
    }
  }
```

The property names are already mapped to `text/{name}`. Adding `contentMediaType` is harmless, but I also don't think it wins us anything.

source/schemas/shopping/types/input_correlation.json
igrigorik: Good flag, fixed in https://github.com/Universal-Commerce-Protocol/ucp/pull/55/changes/040684cbed60031fbeb4b6d89c926398db82612f. 

Note: anyOf with const branches is are a no-op because`{ "type": "string" }` already matches everything; the const branches don't do anything for validation. They only serve as documentation, which description + examples handle more simply without the structural overhead. This is consistent how we've handled this in other places in the spec.

source/schemas/shopping/types/media.json
```diff
@@ -0,0 +1,37 @@
...
+  "description": "Product media item (image, video, etc.).",
+  "type": "object",
+  "required": [
+    "type",
+    "url"
+  ],
+  "properties": {
+    "type": {
+      "type": "string",
+      "enum": ["image", "video", "model_3d"],
```
gsmith85: Thought: If we agree that enums are problematic, perhaps we codify "avoid enums" in `docs/documentation/schema-authoring.md` ?

gsmith85: Suggestion: Consider polymorphic structure. Split `image`, `video` and `model_3d` into distinct definitions so that required fields are strictly enforced on a per type basis.

Something like:

```
{
  "$id": "https://ucp.dev/schemas/shopping/types/media.json",
  "title": "Media",
  "description": "Product media item (image, video, or 3D model).",
  "oneOf": [
    { "$ref": "#/$defs/image" },
    { "$ref": "#/$defs/video" },
    { "$ref": "#/$defs/model_3d" }
  ],
  "$defs": {
    "base_media": {
      "type": "object",
      "required": ["type", "url"],
      "properties": {
        "url": { "type": "string", "format": "uri" },
        "alt_text": { "type": "string" }
      }
    },
    "image": {
      "allOf": [{ "$ref": "#/$defs/base_media" }],
      "properties": {
        "type": { "const": "image" },
        "width": { "type": "integer", "minimum": 1 },
        "height": { "type": "integer", "minimum": 1 }
      },
      "required": ["width", "height"]
    },
    "video": {
      "allOf": [{ "$ref": "#/$defs/base_media" }],
      "properties": {
        "type": { "const": "video" },
        "width": { "type": "integer", "minimum": 1 },
        "height": { "type": "integer", "minimum": 1 },
        "duration": { "type": "number", "description": "Length in seconds" }
      },
      "required": ["width", "height"]
    },
    ...
```

source/schemas/shopping/types/pagination.json
```diff
@@ -0,0 +1,44 @@
...
+          "minimum": 1,
+          "default": 10,
+          "description": "Requested page size. Implementations MAY clamp to a lower maximum."
+        }
+      }
+    },
+    "response": {
+      "type": "object",
+      "description": "Pagination information in responses.",
+      "properties": {
```
gsmith85, line 27: Suggestion: make `has_next_page` and `total_count` required.

source/schemas/shopping/types/pagination.json
```diff
@@ -0,0 +1,44 @@
...
+        "cursor": {
+          "type": "string",
+          "description": "Cursor to fetch the next page of results."
```
gsmith85: Missing Constraints: Note that cursor is required if `has_next_page` is `true`?

source/schemas/shopping/types/price.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/price.json",
+  "title": "Price",
+  "description": "Price with explicit currency. Amount is always in minor units (e.g., cents for USD).",
```
gsmith85: Suggestion: Describe proper handling for currencies without minor units (e.g. the Yen)?

I'm wary of implementation bugs where it's assumed either (i) all currencies have minor units and (ii) all currencies have 100 minor units per major unit.

source/schemas/shopping/types/price_filter.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/price_filter.json",
+  "title": "Price Filter",
+  "description": "Price range filter in minor currency units. Merchant determines currency from context (e.g., country=US → USD). Response prices include explicit currency.",
```
gsmith85: Question: Are there means to convey to the client the currency the merchant has concluded is the applicable to the buyer?

That is, is there a way to aid in ensuring the first request in not miscommunicated?  Example: I've chosen my amounts based on USDs but I'm on vacation and the merchant thinks MXN, so the response includes no results.

source/schemas/shopping/types/variant.json
```diff
@@ -0,0 +1,122 @@
...
+      "description": "Global ID (GID) uniquely identifying this variant. Used as item.id in checkout."
+    },
+    "sku": {
+      "type": "string",
+      "description": "Business-assigned identifier for inventory and fulfillment."
+    },
+    "barcode": {
+      "type": "string",
+      "description": "Industry-standard identifier (UPC, EAN, ISBN) for the variant."
+    },
```
gsmith85, line 39: Suggestion: Extract an object for `barcode` and use an `array` to support multiple barcodes:

```
"barcodes": [
    { "type": "UPC", "value": "123456789012" },
    { "type": "EAN", "value": "01234567890123" },
]
```

In agentic contexts having barcodes actually allows for correlating relevant product knowledge that may have been harvested across geographies where the code varies.

source/schemas/shopping/types/product_option.json
```diff
@@ -0,0 +1,25 @@
...
+  "$id": "https://ucp.dev/schemas/shopping/types/product_option.json",
+  "title": "Product Option",
+  "description": "A product option such as size, color, or material.",
+  "type": "object",
+  "required": [
+    "name",
+    "values"
+  ],
+  "properties": {
+    "name": {
```
gsmith85, line 12: Should we distinguish between "display name" and "machine-readable keys?"

For example:

```
{ "id": "color", "display_name": "Couleur" }
```

source/schemas/shopping/types/variant.json
```diff
@@ -0,0 +1,122 @@
...
+    "list_price": {
+      "$ref": "price.json",
+      "description": "List price before discounts (for strikethrough display)."
+    },
+    "availability": {
+      "type": "object",
+      "description": "Variant availability for purchase.",
+      "properties": {
+        "available": {
+          "type": "boolean",
```
gsmith85, line 111: Thought: Should we limit ourselves to a boolean here, I think there's a more diverse set of states to communicate:
- In/out of stock.
- Low stock (and X Remaining)
- Preorder

Is the plan to layer these in later?

source/schemas/shopping/types/input_correlation.json
gsmith85: I agree that `{"type": "string"}` is the 'truth' for validation.  The nudge toward the `anyOf` structure was really about affordance for code generators and IDEs.  This is a two-way door, I'm totally comfortable with using `examples` in the interim.

docs/specification/catalog/mcp.md
```diff
@@ -0,0 +1,493 @@
...
+Business outcomes use the JSON-RPC `result` field with messages in the response
+payload. See the [Partial Success](#partial-success) section for handling mixed
+results.
+
+## Conformance
+
+A conforming MCP transport implementation **MUST**:
+
+1. Implement JSON-RPC 2.0 protocol correctly.
+2. Provide both `search_catalog` and `lookup_catalog` tools.
```
amithanda: I thought these were independent and merchants/platforms can choose to use or implement only lookup_catalog capability, if they want. Any particular reason why MCP transport should provide both tools?

source/schemas/shopping/catalog_lookup.json
```diff
@@ -0,0 +1,53 @@
...
+    "lookup_request": {
+      "type": "object",
+      "description": "Request body for catalog lookup.",
+      "required": ["ids"],
+      "properties": {
+        "ids": {
+          "type": "array",
+          "items": { "type": "string" },
+          "minItems": 1,
+          "description": "Identifiers to lookup. Implementations MUST support product ID and variant ID; MAY support secondary identifiers (SKU, handle, etc.)."
```
amithanda, line 36: I am curious if you anticipate that this will create any implementation challenges e.g how will a business know if the identifier that is shared is a product Id, variant Id or a SKU? If they are maintaining separate tables for these, will they be required to look up these identifiers in all their systems and what if there are conflicts (although very rare) as these identifiers are not meant to be unique among each other?

source/schemas/shopping/types/variant.json
igrigorik: I deferred this, but I think you're right, we need a qualifier. Updated: https://github.com/Universal-Commerce-Protocol/ucp/pull/55/changes/7d327241ee39a884c78599be78b985318c751f70

docs/specification/catalog/rest.md
```diff
@@ -0,0 +1,396 @@
...
+      "dev.ucp.shopping": {
+        "version": "2026-01-11",
+        "spec": "https://ucp.dev/specification/overview",
+        "rest": {
+          "schema": "https://ucp.dev/services/shopping/rest.openapi.json",
+          "endpoint": "https://business.example.com/ucp"
+        }
+      }
+    },
+    "capabilities": [
```
maximenajim: The REST response example shows ucp.capabilities as an array of {name, version} entries, while MCP examples (and the protocol spec) show capabilities as a map keyed by capability name with an array of versions/configs. We need one canonical envelope shape across all transports and examples.
Fix: Update REST examples to match the object-of-arrays form (e.g., "capabilities": { "dev.ucp.shopping.catalog.search": [ { "version": "…" } ] }).

docs/specification/catalog/rest.md
```diff
@@ -0,0 +1,396 @@
...
+
+#### Example
+
+=== "Request"
+
+    ```json
+    {
+      "query": "blue running shoes",
+      "context": {
+        "country": "US",
```
maximenajim: The request example uses context.country / context.region, but the shared Context schema defines address_country / address_region. Because Context.additionalProperties=true, this won’t fail validation, but it will create two de facto field names in the ecosystem.

docs/specification/catalog/rest.md
```diff
@@ -0,0 +1,396 @@
...
+    {
+      "type": "info",
+      "code": "not_found",
+      "content": "prod_invalid2"
+    }
+  ]
+}
+```
+
+Business outcomes use the standard HTTP 200 status with messages in the response body.
```
maximenajim, line 379: Error model states “business outcomes return HTTP 200 with messages” and reserves non-200 for transport errors. That’s fine, but it needs to be reflected in the OpenAPI responses for the new endpoints (400/401/429/etc.) so client generators and implementers don’t miss required failure modes.
Fix: Ensure /catalog/search and /catalog/lookup operations define the same transport error responses as the other shopping endpoints, and document the response body shape for transport errors (even if it’s a minimal {ucp, messages} envelope or a separate error schema).

docs/specification/catalog/index.md
```diff
@@ -0,0 +1,238 @@
...
+
+# Catalog Capability
+
+## Overview
+
+The Catalog capability allows platforms to search and browse business product catalogs.
+This enables product discovery before checkout, supporting use cases like:
+
+* Free-text product search
+* Category and filter-based browsing
```
maximenajim, line 25: The overview lists “Category and filter-based browsing” as a core use case, but Catalog Search conformance appears to require a query parameter. If query is truly required, “filter-only browse” isn’t possible unless clients send an empty/placeholder query, which is brittle.

docs/specification/catalog/lookup.md
```diff
@@ -0,0 +1,82 @@
...
+
+The response does not guarantee order. Each variant carries an `input`
+array identifying which request identifiers resolved to it, and how.
+
+{{ schema_fields('types/input_correlation', 'catalog') }}
+
+Multiple request identifiers may resolve to the same variant (e.g., a
+product ID and one of its variant IDs). When this occurs, the variant's
+`input` array contains one entry per resolved identifier, each with its
+own match type. Variants without an `input` entry MUST NOT appear in
```
maximenajim: “Variants without an input entry MUST NOT appear in lookup responses” is a strong and useful guarantee for correlation—but it’s not clear the response schema actually enforces this (since Variant.input appears optional in the shared variant type). If this is a protocol guarantee, it should be enforced either via:

- a lookup-specific variant schema where input is required, or
- a oneOf / allOf constraint in the lookup response that requires input on variants.

Otherwise we’ll get partial implementations that violate correlation and break agents.

source/services/shopping/rest.openapi.json
```diff
@@ -578,6 +578,94 @@
...
+          "content": {
+            "application/json": {
+              "schema": { "$ref": "#/components/schemas/catalog_search_request" }
+            }
+          }
+        },
+        "responses": {
+          "200": {
+            "description": "Search results",
+            "headers": {
```
maximenajim, line 609: The marks Signature, Signature-Input, and Content-Digest as required: false. This conflicts with the normative “Message Signatures” spec, which treats RFC 9421 signing headers as required for HTTP signing, and Content-Digest as required when a body is present. Leaving them optional in the canonical OpenAPI spec will lead to unsigned/unauthenticated implementations “passing conformance”
Fix: Either:

- Mark them required on operations where signing is mandatory (and define the auth negotiation rules if exceptions exist), or
- Update docs/specification/signatures.md to explicitly make signing optional and specify when it is REQUIRED vs OPTIONAL.

Right now, spec + OpenAPI disagree

source/schemas/shopping/types/variant.json
```diff
@@ -0,0 +1,161 @@
...
+    "category": {
+      "type": "array",
+      "items": {
+        "$ref": "category.json"
+      },
+      "description": "Variant categories with optional taxonomy identifiers."
+    },
+    "price": {
+      "$ref": "price.json",
+      "description": "Current selling price."
```
alex-jansen, line 66: consider adding a "price_valid_until" date field to indicate until when a (sale) price will be valid and after which the price will  increase.

source/schemas/shopping/types/pagination.json
igrigorik: 👍  

Made has_next_page required. Keeping total_count optional — many backends can't produce a total count cheaply, and forcing it would lead to dummy values or expensive queries.

source/schemas/shopping/types/pagination.json
igrigorik: 👍

source/schemas/shopping/types/price.json
igrigorik: Great catch! This requires broader surgery.. updated.

source/schemas/shopping/types/variant.json
igrigorik: 👍  good call, updated.

source/schemas/shopping/types/product_option.json
igrigorik: I _think_ this one we can get away with. 

The join works within a single response: product.options[].name → variant.selected_options[].name tells the agent which value was selected on which dimension. This join only operates within one response — the agent renders options, the buyer picks, the selected variant carries the selections. Unlike product/variant IDs which flow as durable references into cart and checkout, option names don't persist across requests. Each request carries context.language and the response is self-{contained, consistent} for that locale.

source/schemas/shopping/types/media.json
igrigorik: Practice what you preach! Great catch and updated 😄 👍

igrigorik: I really like the strict validation, but `oneOf` would close the extension point. You can't extend a oneOf from outside. If an extension wants to add type: "audio" with custom fields, they'd need to modify the oneOf array in the base schema. There's no allOf composition path into a oneOf discriminator — it's structurally closed.

docs/specification/catalog/rest.md
igrigorik: Good catch, fixed!

docs/specification/catalog/rest.md
igrigorik: Related: https://github.com/Universal-Commerce-Protocol/ucp/pull/216. I believe we're consistent and have all the things covered once that PR lands. 

>Fix: Ensure /catalog/search and /catalog/lookup operations define the same transport error responses as the other shopping endpoints, and document the response body shape for transport errors (even if it’s a minimal {ucp, messages} envelope or a separate error schema).

Catalog endpoints are stateless reads — empty results are valid responses, not error conditions. The error_response pattern() applies to stateful resource operations where a resource can't be created or no longer exists (checkout, cart). Catalog search/lookup always return the normal response shape with products: [] for no results and optional messages for diagnostics.

docs/specification/catalog/index.md
igrigorik: Good catch, we relaxed "must" for query earlier but the text did not capture this. Updated.

docs/specification/catalog/lookup.md
igrigorik: 👍  refactored.

docs/specification/catalog/rest.md
igrigorik: good catch, fixed!

source/services/shopping/rest.openapi.json
igrigorik: The signing spec (0426800) explicitly makes signatures SHOULD for requests and OPTIONAL for catalog queries (see "When Signatures Apply" section).

source/schemas/shopping/catalog_lookup.json
igrigorik: Some of the IDs have well-known shapes, but yes, the expectation is that business can support lookup against primary and some set of secondary identifiers. Conflicts are possible (but should be rare, as you noted) and business can return both matches. Let's keep this and revisit if necessary.

docs/specification/catalog/mcp.md
igrigorik: You're right, missed this! Fixed.

source/schemas/shopping/types/variant.json
igrigorik: Hmm, that's a good and interesting one. 

I do see this modeled on some feed specs and that makes sense, since that's cached data, and you need to explicitly capture start/stop dates. What we're modeling here though is real-time: each response is a point in time snapshot of current price and offer. 

Let's pull this out into followup and separate conversation, because it requires a deeper thought process: sensitivity, timezones, etc.

source/schemas/shopping/types/price_filter.json
igrigorik: That's a "fun" one. 

Response prices carry explicit {amount, currency} on every price object, so the agent sees the merchant's resolved currency on the first response. Also, clarified the contract: https://github.com/Universal-Commerce-Protocol/ucp/pull/55/commits/fa0b5e3b281c645c944405fcd68e8f1e890c71d9

amithanda (approved):
I've reviewed the PR and it looks great to me. Here's a summary of my findings:

1. **Schemas (\`catalog_lookup.json\`, \`catalog_search.json\`, \`types/...\`)**: 
   - The schemas properly separate concerns between finding items (Search) and retrieving items (Lookup).
   - The metadata, variants, and other options (like \`barcodes\` array, \`inputs\` correlation) are robust and provide the exact context needed.
2. **Docs**: 
   - Comprehensive examples are included.
   - Using open vocabularies rather than fixed enums is documented cleanly in \`schema-authoring.md\`!

The CI failure on \`mkdocs build\` in \`docs.yml\` is an upstream MkDocs-Material version compatibility issue (\`material.extensions.emoji\` module changes) affecting the \`main\` branch as well, so this PR is not the origin of the failure and doesn't need to be blocked on it.

Approving! Great work on catalog capabilities!

PR #58, opened 2026-01-15
refactor!: Remove `full_name` fields across UCP APIs

lemonmade:
# Description

This PR removes the optional `full_name` field exposed for a few different objects in the schema. While implementing, it felt quite awkward that every merchant and partner would need to handle all the different cases of `first_/last_/full_name` being provided, and the rules for how those different parties should reflect this back was not clear.

We think separate, optional `first_name` and `last_name` fields provide the highest fidelity information, as it is possible for a business or agent to go from `first_name` + `last_name` to full name (provided they know the locale), but it is not always possible to do the reverse with just a `full_name`. UCP consumers that only know a single name have to do extra work to split out a `first_name` and `last_name` in their input, but they are the ones with the context to separate first and last name (including, optionally, omitting one of those fields, such as for mononyms).

## Type of change

      or fields**)
- [ ] Documentation update

---

files or fields:

```

(Please provide a detailed technical and strategic rationale here for why this
breaking change or removal is necessary.)
```
---

## Checklist:

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -230,7 +230,8 @@ image: assets/banner.png
           "destinations": [
             {
               "id": "dest_1",
-              "full_name": "Elisa Beckett",
+              "first_name": "Elisa",
+              "last_name": "Beckett",
               "street_address": "1600 Amphitheatre Pkwy",
               "address_locality": "Mountain View",
               "address_region": "CA",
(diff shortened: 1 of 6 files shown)


igrigorik (approved):
LGTM

raginpirate (approved):
Thanks for pushing this in Chris!

To port over some of our discussion: I think the clean interface is worth the sometimes-awkward notion of having to convert a mononym or inverted name structure into first and last; payments may be the most common edge where this happens (payment APIs using just one name field). But even then, we see flows within payments that force us to deal with this split anyways, so it shouldn't be shocking to any implementor. 🚀

PR #59, opened 2026-01-15
chore: correctly set up cspell to use config

wry-ry:
# Description

Correctly set up cspell to use config, clean up trivy issue

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -10,7 +10,8 @@
     ".vscode/**",
     ".cspell.json",
     "**/*.svg",
-    ".gitignore"
+    ".gitignore",
+    "package.json"
   ],
   "dictionaryDefinitions": [
     {


docs/assets/partner/codeveloped/Etsy.svg
0x8000-0000, line 1: What has Etsy got to do with spelling?

.cspell.json
```diff
@@ -10,7 +10,9 @@
     ".vscode/**",
     ".cspell.json",
     "**/*.svg",
-    ".gitignore"
+    ".gitignore",
+    "**/*.yml",
```
0x8000-0000: Should this be .github/** instead? We presumably want to spellcheck yaml files in the regular repo, if we ever have them.

Otherwise, why single out YAML files? We can have the same typos everywhere else.

0x8000-0000 (approved):

.cspell.json
wry-ry: I was thinking all yaml files in general because they are filled with config things that are easily flagged with minimal stuff that actually needs to be spelled correctly. But I'm ok with keeping the exception narrower for now.

docs/assets/partner/codeveloped/Etsy.svg
wry-ry: didn't sync in advance

carol-w-tech (approved):

PR #62, opened 2026-01-16
feat: add `context`; make currency output-only

igrigorik:
`currency` as client input is flawed: merchants control accepted currencies, not buyers. Allowing "show me price in $X" implies forex operations that break at payment time when the merchant's actual accepted currency differs. Practically, merchants would treat it as a hint and ignore it for anything they don't explicitly allow. 

Correct model: buyer provides context signals (country, region, postal code, affiliations), merchant determines currency and other market-relevant attributes such as product availability, shipping and delivery times, price, etc.

Changes:
- Add extensible `context` type with country, region, postal_code for progressive disclosure
- Add `context` to checkout (optional on create/update, omit on complete)
- Make currency output-only (omit on all operations)

This primitive will be relevant & used by catalog, cart, and checkout.

---

## Checklist:

### DIFF
--- a/source/schemas/shopping/types/context.json
+++ b/source/schemas/shopping/types/context.json
@@ -4,19 +4,19 @@
   "title": "Context",
   "description": "Provisional buyer signals for relevance and localization: product availability, pricing, currency, tax, shipping, payment methods, and eligibility (e.g., student or affiliation discounts). Businesses SHOULD use these values when authoritative data (e.g., address) is absent, and MAY ignore unsupported values without returning errors. Context can be disclosed progressively—coarse signals early, finer resolution as the session progresses. Higher-resolution data (shipping address, billing address) supersedes context. Platforms SHOULD progressively enhance context throughout the buyer journey.",
   "type": "object",
+  "additionalProperties": true,
   "properties": {
     "address_country": {
       "type": "string",
-      "pattern": "^[A-Z]{2}$",
-      "description": "ISO 3166-1 alpha-2 country code (e.g., 'US', 'CA'). Informs product availability, pricing, and currency. Detected from request if omitted."
+      "description": "The country. Recommended to be in 2-letter ISO 3166-1 alpha-2 format, for example \"US\". For backward compatibility, a 3-letter ISO 3166-1 alpha-3 country code such as \"SGP\" or a full country name such as \"Singapore\" can also be used. Optional hint for market context (currency, availability, pricing)—higher-resolution data (e.g., shipping address) supersedes this value."
     },
     "address_region": {
       "type": "string",
-      "description": "State, province, emirate, or district (e.g., 'CA', 'ON', 'Dubai'). Format varies by country."
+      "description": "The region in which the locality is, and which is in the country. For example, California or another appropriate first-level Administrative division. Optional hint for progressive localization—higher-resolution data (e.g., shipping address) supersedes this value."
     },
     "postal_code": {
       "type": "string",
-      "description": "Postal or ZIP code for regional refinement. Not applicable in all countries."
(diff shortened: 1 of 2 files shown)


amithanda (commented):
I am wondering if there will be use cases where there is an explicit buyer preference to see amounts in a specified currency, although I'll admit that I have not seen this to be super common e.g shoppers from different country choosing to pay in local v/s foreign currency.

source/schemas/shopping/types/context.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/context.json",
+  "title": "Context",
+  "description": "Buyer signals that help the business tailor responses for relevance and localization: product availability, pricing, currency, tax, shipping, payment methods, and eligibility (e.g., student or affiliation discounts). Context can be disclosed progressively—coarse signals early in the journey, finer resolution as the session progresses to improve relevance and accuracy. Higher-resolution data provided elsewhere (shipping address, billing address) supersedes provided context. Platforms SHOULD provide and progressively enhance context throughout the buyer journey to optimize relevance and localization.",
+  "type": "object",
+  "properties": {
+    "country": {
```
lemonmade: Should we use `address_country` or `address_region`, the key names we use for these in addresses?

source/schemas/shopping/checkout.json
```diff
@@ -41,6 +41,14 @@
       "$ref": "types/buyer.json",
       "description": "Representation of the buyer."
     },
+    "context": {
+      "$ref": "types/context.json",
+      "ucp_request": {
```
lemonmade, line 56: Discussed in person: it seems like we want to omit this from the response, in addition to omitting it from the `complete` requests. I am not sure if we actually have a mechanism to achieve this today.

richmolj: There's a `ucp_response: "omit"` see: https://github.com/Universal-Commerce-Protocol/ucp/blob/16f071e106f2bb2fa814c24e30686309d31ead4d/source/schemas/shopping/types/token_credential.json#L16

source/schemas/shopping/types/context.json
```diff
@@ -0,0 +1,22 @@
+{
```
knightlin-shopify, line 1: `"additionalProperties": true` is missing?

source/schemas/shopping/types/context.json
```diff
@@ -0,0 +1,22 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/context.json",
+  "title": "Context",
+  "description": "Provisional buyer signals for relevance and localization: product availability, pricing, currency, tax, shipping, payment methods, and eligibility (e.g., student or affiliation discounts). Businesses SHOULD use these values when authoritative data (e.g., address) is absent, and MAY ignore unsupported values without returning errors. Context can be disclosed progressively—coarse signals early, finer resolution as the session progresses. Higher-resolution data (shipping address, billing address) supersedes context. Platforms SHOULD progressively enhance context throughout the buyer journey.",
+  "type": "object",
+  "properties": {
+    "address_country": {
+      "type": "string",
+      "pattern": "^[A-Z]{2}$",
```
knightlin-shopify: nit: This matches the “ISO 3166-1 alpha-2 country code” schema, but it’s inconsistent with `address_country` in `postal_address.json`. As hint, should we accept lowercase values as well (e.g., `^[A-Za-z]{2}$`)?

igrigorik: Good flag, updated.

richmolj (approved):

drewolson-google (approved):

PR #68, opened 2026-01-16
chore: Trying to enable the CI checks

yanheChen:
# Description

This PR initializes the Python project configuration in `pyproject.toml` and updates the `ci_check_models.sh` script to leverage `uv` for dependency management and execution.

Key changes:
- Populated `pyproject.toml` with project metadata and dependencies (specifically `datamodel-code-generator[ruff]`).
- Updated `scripts/ci_check_models.sh` to use `uv run` with the `--with` flag for ephemeral dependency installation.
- Added `ruff` formatters to the data model generation process in the CI script.
- Added `uv.lock` file to pin dependencies.

## Type of change

- [x] Chore (Changes to the build process or auxiliary tools and libraries)

---

files or fields:

### DIFF
--- a/docs/specification/ap2-mandates.md
+++ b/docs/specification/ap2-mandates.md
@@ -235,7 +235,7 @@ a completion request without mandates **MUST** result in a session failure.
 The platform initiates the session. The business returns the `Checkout` object
 with `ap2.merchant_authorization` embedded in the response body.
 
-{{ extension_schema_fields('ap2_mandate.json#/$defs/checkout_response_with_ap2', 'ap2-mandates') }}
+{{ extension_schema_fields('ap2_mandate.json#/$defs/checkout', 'ap2-mandates') }}
 
 **Example Response:**
 ```json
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -1,3 +1,19 @@
+[project]
+name = "ucp"
+version = "0.1.0"
```
0x8000-0000, line 3: should the version be 20260111?

scripts/ci_check_models.sh
```diff
@@ -42,8 +42,9 @@ echo "Generating Pydantic models from $SCHEMA_DIR to temporary dir $OUTPUT_DIR..
 # Run generation using uv
 uv run \
     --link-mode=copy \
-    --extra-index-url https://pypi.org/simple python \
-    -m datamodel_code_generator \
+    --with "datamodel-code-generator[ruff]" \
```
0x8000-0000, line 45: This should be kept in sync with https://github.com/Universal-Commerce-Protocol/python-sdk/blob/main/generate_models.sh; perhaps even add a comment to that effect.

0x8000-0000: How is this script run in CI?

0x8000-0000 (changes requested):

scripts/ci_check_models.sh
yanheChen: seems has been run here: https://github.com/Universal-Commerce-Protocol/ucp/blob/main/.github/workflows/docs.yml#L28

PR #72, opened 2026-01-16
chore: configuration for some linter issues

wry-ry:
# Description

Configure prettier to ignore files in generated/.
Turn off gitleaks linter.


.prettierignore
```diff
@@ -0,0 +1,2 @@
+# Ignore generated files
+generated/
```
0x8000-0000: End of file new line missing?

drewolson-google (approved):

PR #73, opened 2026-01-17
feat(cart): Cart capability for basket building

igrigorik:
## Summary

Cart capability (`dev.ucp.shopping.cart`) enables pre-purchase exploration without checkout complexity. Checkout models buying — payment handlers, status lifecycle, order finalization. Cart models shopping — browsing, comparing, saving items for later, all without payment commitment.

## Motivation

Platforms need a lightweight way to build baskets before purchase intent is established — session persistence, sharing, and exploratory workflows without the checkout contract. Checkout operations are expensive for businesses (payment handler negotiations, 3P calls, tax calculations with full address resolution, etc), and often subject to strict rate limits. Carts address this via lower-fidelity policies that accommodate efficient pre-purchase basket building and ability to service much higher volumes.

### Goals

* Enable efficient pre-purchase basket building with accurate estimates
* Provide cart-to-checkout conversion via `cart_id` parameter
* Maintain schema consistency with checkout (reuse line_item, buyer, context, totals)

### Non-Goals

* Payment configuration (checkout concern)
* Status lifecycle management (cart is binary: exists/404)
* Fulfillment, discounting, etc (addressed via capability extensions)
* Subscriptions, loyalty, B2B pricing (ditto, capability extensions)

## Detailed Design

**Key differences from Checkout:**

| Aspect | Cart | Checkout |
|--------|------|----------|
| Purpose | Pre-purchase exploration | Purchase finalization |
| Payment | None | Required |
| Status | Binary (exists/404) | Lifecycle (incomplete → completed) |
| Totals | Estimates | Final pricing |
| Handoff URL | `checkout_url` | `continue_url` |

**API Operations:**

| Operation | REST | MCP |
|-----------|------|-----|
| Create | `POST /carts` | `create_cart` |
| Get | `GET /carts/{id}` | `get_cart` |
| Update | `PUT /carts/{id}` | `update_cart` |
| Cancel | `POST /carts/{id}/cancel` | `cancel_cart` |

**Cart-to-Checkout Conversion:**

When checkout is initialized via `cart_id`, sessions SHOULD be linked for checkout duration. Business SHOULD maintain cart for back-to-storefront flows. After checkout completion, business MAY clear cart.

**Schema:**

* Cart reuses checkout entities: line_item, buyer, context, totals, messages, links
* New `cart_id` field on checkout (via schema extension) for conversion
* `response_cart` added to ucp.json for cart response metadata

**Update Semantics:** Full replacement (mirrors checkout)
**Cancel Behavior:** Business MUST return cart state before deletion. Subsequent operations SHOULD return 404.
**Tax Guidance:** Taxes MAY be included where calculable. Platforms SHOULD assume cart totals are estimates when provided.

## Risks and Mitigations

* **Complexity:** Cart adds another capability to implement. *Mitigation:* Cart reuses checkout schemas — no new entity types. Simpler than checkout (no payment, no status lifecycle).
* **Backward Compatibility:** None — new capability, opt-in via profile advertisement.
* **Security:** Cart stores buyer signals (context, optional buyer info). *Mitigation:* Same security model as checkout. TTL via `expires_at`. Explicit `cancel_cart` for privacy.
* **Schema Drift:** Cart and checkout share entities. *Mitigation:* Cart explicitly references checkout schemas (`$ref`), ensuring consistency.

----

- [x] Documentation update

## Checklist:

### DIFF
--- a/source/services/shopping/openapi.json
+++ b/source/services/shopping/openapi.json
@@ -675,7 +675,7 @@
         "$ref": "../../schemas/shopping/checkout.json"
       },
       "cart": {
-        "$ref": "https://ucp.dev/schemas/shopping/cart.json"
+        "$ref": "../../schemas/shopping/cart.json"
       },
       "order": {
         "$ref": "../../schemas/shopping/order.json"
--- a/source/services/shopping/openrpc.json
+++ b/source/services/shopping/openrpc.json
@@ -182,12 +182,12 @@
         {
           "name": "cart",
           "required": true,
-          "schema": {"$ref": "https://ucp.dev/schemas/shopping/cart.json"}
+          "schema": {"$ref": "../../schemas/shopping/cart.json"}
         }
       ],
       "result": {
         "name": "cart",
-        "schema": {"$ref": "https://ucp.dev/schemas/shopping/cart.json"}
+        "schema": {"$ref": "../../schemas/shopping/cart.json"}
       }
     },
     {
@@ -202,7 +202,7 @@
       ],
       "result": {
         "name": "cart",
-        "schema": {"$ref": "https://ucp.dev/schemas/shopping/cart.json"}
+        "schema": {"$ref": "../../schemas/shopping/cart.json"}
       }
     },
     {
@@ -217,12 +217,12 @@
         {
           "name": "cart",
           "required": true,
-          "schema": {"$ref": "https://ucp.dev/schemas/shopping/cart.json"}
+          "schema": {"$ref": "../../schemas/shopping/cart.json"}
         }
       ],
       "result": {
         "name": "cart",
-        "schema": {"$ref": "https://ucp.dev/schemas/shopping/cart.json"}
+        "schema": {"$ref": "../../schemas/shopping/cart.json"}
       }
     },
     {
@@ -242,7 +242,7 @@
       ],
       "result": {
         "name": "cart",
-        "schema": {"$ref": "https://ucp.dev/schemas/shopping/cart.json"}
+        "schema": {"$ref": "../../schemas/shopping/cart.json"}
       }
     }
   ]


docs/specification/cart.md
```diff
@@ -0,0 +1,206 @@
...
+* **Version:** `DRAFT`
+
+## Overview
+
+The Cart capability enables basket building without the complexity of checkout.
+While [Checkout](checkout.md) manages payment handlers, status lifecycle, and
+order finalization, cart provides a lightweight CRUD interface for item
+collection before purchase intent is established.
+
+**When to use Cart vs Checkout:**
```
knightlin-shopify, line 29: I’m having trouble understanding why we need Cart capability here. It seems more like a Checkout in `draft` status and make payment (it is actually already supporting empty object) optional during the Checkout creation?

docs/specification/cart.md
```diff
@@ -0,0 +1,206 @@
...
+*   **MAY** use carts for pre-purchase exploration and session persistence.
+*   **SHOULD** convert cart to checkout when user expresses purchase intent.
+*   **MAY** display `checkout_url` for handoff to business UI.
+*   **SHOULD** handle 404 gracefully when cart expires or is canceled.
+
+**Business**
+
+*   **MUST** provide `checkout_url` in all cart responses.
+*   **SHOULD** provide estimated totals when calculable.
+*   **MAY** omit shipping totals until checkout when address is unknown.
```
raginpirate:
```suggestion
*   **MAY** omit fulfillment totals until checkout when address is unknown.
```

docs/specification/cart.md
```diff
@@ -0,0 +1,206 @@
...
+*   **SHOULD** return informational messages for validation warnings.
+*   **MAY** set cart expiry via `expires_at`.
+*   **SHOULD** follow [cart lifecycle requirements](#cart-to-checkout-conversion)
+    when checkout is initialized via `cart_id`.
+
+## Capability Schema Definition
```
raginpirate:
```suggestion
## Cart Schema Definition
```

docs/specification/cart.md
raginpirate: This has opened a series of questions for me:
- Looks like we need changes to the fulfillment capability to extend cart 🤔 
- or... would we express this as an extension on the base fulfillment capability and cart capability simultaneously?
- And if so... how do we express the dependency graph? At the moment we allow one parent <-> child. Do we care to express this?

docs/specification/cart.md
```diff
@@ -0,0 +1,206 @@
...
+- **Incremental building**: Add/remove items across sessions
+- **Localized estimates**: Context-aware pricing without full checkout overhead
+- **Sharing**: `checkout_url` enables cart sharing and recovery
+
+## Cart vs Checkout
+
+| Aspect | Cart | Checkout |
+|--------|------|----------|
+| **Purpose** | Pre-purchase exploration | Purchase finalization |
+| **Payment** | None | Required (handlers, instruments) |
```
raginpirate, line 49: Challenge: if we're ok with fulfillment extending cart to build fulfillment estimates, shouldn't we be ok with payment handlers being negotiated and pruned down based on cart shape, to assist with accelerated checkouts from cart?

I'm fine without instruments though, we don't need store credit / vaulted cards surfacing through cart.

docs/specification/cart.md
raginpirate: With this title I was thinking I was going to be reading the config schema, so just adding some clarity to wording.

docs/specification/cart.md
```diff
@@ -0,0 +1,206 @@
...
+Cancels a cart session. Business MUST return the cart state before deletion.
+Subsequent operations for this cart ID SHOULD return 404.
+
+*   [REST Binding](cart-rest.md#cancel-cart)
+*   [MCP Binding](cart-mcp.md#cancel_cart)
+
+## Entities
+
+Cart reuses the same entity schemas as [Checkout](checkout.md). This ensures
+consistent data structures when converting a cart to a checkout session.
```
raginpirate, line 163: This description is good, but adding a table in this page for every existing checkout entity feels like a lot of repetition and makes me feel like maybe there is going to be a discrepancy in the descriptions or something.

My proposal: turn this into a simple list:
* line items
* buyer
* context
* total
* message

With links to the checkout page for each.

docs/specification/cart-rest.md
```diff
@@ -0,0 +1,487 @@
...
+   limitations under the License.
+-->
+
+# Cart Capability - REST Binding
+
+This document specifies the REST binding for the [Cart Capability](cart.md).
+
+## Protocol Fundamentals
+
+### Base URL
```
lemonmade, line 60: I put up a PR this morning to add a Discovery section on the Checkout REST docs, matching the equivalent section in the MCP docs: https://github.com/Universal-Commerce-Protocol/ucp/pull/82/changes. I think the same change would be useful here, as the same problem (the ucp.dev schema URL for REST is not provided anywhere on the page) applies to this new Cart REST page.

source/schemas/shopping/cart.json
```diff
@@ -0,0 +1,115 @@
...
+    },
+    "links": {
+      "type": "array",
+      "items": {
+        "$ref": "types/link.json"
+      },
+      "description": "Optional merchant links (policies, FAQs).",
+      "ucp_request": "omit"
+    },
+    "checkout_url": {
```
lemonmade: Did you consider calling this `continue_url` for symmetry with Checkout? I personally find this name a little confusing, since the API endpoint is really the "checkout URL" for many UCP implementers.

source/schemas/shopping/cart.json
```diff
@@ -0,0 +1,115 @@
...
+                "update": "omit"
+              }
+            }
+          }
+        }
+      ]
+    }
+  },
+  "type": "object",
+  "required": ["ucp", "id", "line_items", "currency", "totals", "checkout_url"],
```
lemonmade: `continue_url` is optional on Checkout. Is there a reason `checkout_url` needs to be mandatory for Carts? I assumed it covered similar "human in the loop" flows, and would therefore have the same optionality in the schema as `continue_url`.

docs/specification/cart.md
```diff
@@ -0,0 +1,206 @@
...
+
+**Cart lifecycle after conversion:**
+
+When checkout is initialized via `cart_id`, the cart and checkout sessions
+SHOULD be linked for the duration of the checkout.
+
+*   **During active checkout** — Business SHOULD maintain the cart and reflect
+    relevant checkout modifications (quantity changes, item removals) back to
+    the cart. This supports back-to-storefront flows when buyers transition
+    between checkout and storefront.
```
lemonmade, line 85: What should a business do if multiple checkouts are created from the same cart?

source/services/shopping/openapi.json
```diff
@@ -338,6 +338,156 @@
...
+            "content": {
+              "application/json": {
+                "schema": { "$ref": "#/components/schemas/cart" }
+              }
+            }
+          }
+        }
+      }
+    },
+    "/carts/{id}/cancel": {
```
lemonmade, line 457: Did you consider making this a DELETE instead, since there is no Checkout-style `status=cancelled`, and since we seem to be trying to express "it exists or it doesn’t" semantics for Carts? There are a few places where we have to say something like, "doesn’t exist, has expired, or was cancelled" to describe the 404 case, which I assume could also be simplified a bit if we treated this as a DELETE.

lemonmade (approved):

source/schemas/shopping/cart.json
sinhanurag: @igrigorik we will need an optional continue_url construct in cart as well. We will have to also define embedded transport for cart operation as needed for some upcoming integrations.

source/schemas/shopping/cart.json
sinhanurag: +1 refer to my comment above too. I would also call it `continue_url` with the expectation that the business populating it is aware of what is the intended use of this url when invoked by the agent. Such as:

- convert to checkout
- continue with cart building in embedded transport
- resolve human in the loop interaction for cart building.

source/services/shopping/openapi.json
sinhanurag: I agree as opposed to checkout where its an actual cancellation (not just a resource deletion) this is pretty much a CRUD entity with no functional operation as we envision it. DELETE makes sense for REST.

docs/specification/cart.md
raginpirate: Why allow this to be 1-N instead of just creating a 1-1 relationship between carts and checkouts within UCP?
I think businesses could just fork the UUID if a cart ends up being re-used to build a second checkout, as once a checkout completes the associated cart should be frozen / destroyed, making it impossible to modify the state of a second checkout.

source/schemas/shopping/cart.json
```diff
@@ -0,0 +1,115 @@
...
+      "title": "Checkout with Cart",
+      "description": "Checkout extended with cart capability. Adds cart_id to create_checkout for cart-to-checkout conversion.",
+      "allOf": [
+        { "$ref": "checkout.json" },
+        {
+          "type": "object",
+          "properties": {
+            "cart_id": {
+              "type": "string",
+              "description": "Cart ID to convert. Cart contents (line_items, context, buyer) initialize checkout and take precedence over checkout payload for overlapping fields.",
```
maximenajim: “Cart contents … take precedence over checkout payload for overlapping fields.” 

This is a high-risk ambiguity for both security and correctness:
- It enables “confused deputy” behavior if implementations merge payloads loosely.
- It’s unclear whether conflicting fields should be rejected vs silently overridden.

Fix: tighten the contract in the spec + schema:
Define conflict handling: either MUST ignore overlapping checkout fields when cart_id is present (and optionally emit a warning message), or MUST return 4xx on conflicts—pick one and make it normative.

source/schemas/shopping/cart.json
maximenajim: `required` includes "totals" and "currency" but the capability text emphasizes estimates and partial availability. If the intention is “always present, possibly partial,” explicitly state that:

- `totals` MAY be empty and/or MAY omit specific total types until calculable.
- `currency` is always determinable by the merchant (or define fallback behavior).

source/schemas/shopping/cart.json
```diff
@@ -0,0 +1,115 @@
...
+      "items": {
+        "$ref": "types/line_item.json"
+      },
+      "description": "Cart line items. Same structure as checkout. Full replacement on update.",
+      "ucp_request": {
+        "create": "required",
+        "update": "required"
+      }
+    },
+    "context": {
```
maximenajim, line 57: The context description references country, region, postal_code, but later patches rename Context fields to address_country / address_region.

docs/specification/cart-rest.md
```diff
@@ -0,0 +1,487 @@
...
+    UCP-Agent: profile="https://platform.example/profile"
+    Content-Type: application/json
+
+    {
+      "line_items": [
+        {
+          "item": {
+            "id": "item_123",
+            "title": "Red T-Shirt",
+            "price": 2500
```
maximenajim: The Create Cart request example includes response-controlled fields (line_items[*].id, item.title, item.price). Client-supplied price is particularly dangerous in an agentic commerce protocol: naive implementations might accidentally trust it, creating a direct integrity vulnerability.

docs/specification/cart-mcp.md
```diff
@@ -0,0 +1,566 @@
...
+{{ schema_fields('cart_resp', 'cart') }}
+
+#### Example
+
+=== "Request"
+
+    ```json
+    {
+      "jsonrpc": "2.0",
+      "method": "update_cart",
```
maximenajim, line 327: `update_cart` does not include `idempotency_key` in its documented input schema or examples, while `create_cart` and `cancel_cart` do.

docs/specification/cart.md
igrigorik: I would prefer to have authoritative and full info. These are generated from same schemas, I think the risk here is ~low.

docs/specification/cart-rest.md
igrigorik: Good catch, mirrored

source/schemas/shopping/cart.json
igrigorik: 👍  updated

docs/specification/cart.md
igrigorik: Agree on 1:1, made this explicit.

source/schemas/shopping/cart.json
igrigorik: Updated

source/schemas/shopping/cart.json
igrigorik: Good catch! Docs bug, fixed.

source/services/shopping/openapi.json
igrigorik: I agree DELETE is more "restful" but this would create inconsistency with Checkout where we used POST for cancel operations. I think symmetry here is more valuable; if we want to change it here we should do the same in both places. Let's split this out into followup discussion.

source/schemas/shopping/cart.json
igrigorik: Good catch, ty!

docs/specification/cart-mcp.md
igrigorik: Good catch on the asymmetry. REST requires Idempotency-Key for all mutations. MCP docs only explicitly mention it for cancel/complete. I'll move this into async followup.. we should fix and be symmetric.

docs/specification/cart.md
igrigorik: Good prompt. I'll pull this out for async followup.

PR #96, opened 2026-01-20
feat(extends): multi-parent support + deterministic schema resolution

igrigorik:
Problems: 
 1. Schema resolution currently relies on implicit conventions. Extension `$defs` use short names ("checkout") without a direct mapping or clear convention that implementers can rely on.
 2. Multi-parent extensions were possible via multiple array entries, but this is ~clunky / verbose pattern.

**Solution:**
  - `extends` accepts `string | array` (cleaner multi-parent declaration)
  - Extension `$defs` keys **MUST** use full parent names (`dev.ucp.shopping.checkout`)
  - Added schema resolution convention to the spec: $defs[{root_capability}]
  - Added response capability selection algorithm

## Example: Multi-parent extension (Discount extends Checkout + Cart)

**Discovery profile:**
```json
  "dev.ucp.shopping.discount": [{
    "version": "2026-01-11",
    "extends": ["dev.ucp.shopping.checkout", "dev.ucp.shopping.cart"],
    "schema": "https://ucp.dev/schemas/shopping/discount.json"
  }]
```

**Extension schema ($defs use full parent names):**
```json
  {
    "$defs": {
      "discounts_object": { ... },
        "allOf": [
          {"$ref": "checkout.json"},
          {"properties": {"discounts": {"$ref": "#/$defs/discounts_object"}}}
        ]
      },
      "dev.ucp.shopping.cart": {
        "allOf": [
          {"$ref": "cart.json"},
          {"properties": {"discounts": {"$ref": "#/$defs/discounts_object"}}}
        ]
      }
    }
  }
```

The extension mechanism is now explicit and self-documenting rather than relying on implicit naming conventions.

## Checklist:

### DIFF
--- a/docs/specification/overview.md
+++ b/docs/specification/overview.md
@@ -171,9 +171,9 @@ Extensions use the `extends` field to declare their parent(s):
 {
   "dev.ucp.shopping.fulfillment": [
     {
-      "version": "2026-01-11",
-      "spec": "https://ucp.dev/specification/fulfillment",
-      "schema": "https://ucp.dev/schemas/shopping/fulfillment.json",
+      "version": "2026-01-23",
+      "spec": "https://ucp.dev/2026-01-23/specification/fulfillment",
+      "schema": "https://ucp.dev/2026-01-23/schemas/shopping/fulfillment.json",
       "extends": "dev.ucp.shopping.checkout"
     }
   ]
@@ -188,9 +188,9 @@ Extensions **MAY** extend multiple parent capabilities by using an array:
 {
   "dev.ucp.shopping.discount": [
     {
-      "version": "2026-01-11",
-      "spec": "https://ucp.dev/specification/discount",
-      "schema": "https://ucp.dev/schemas/shopping/discount.json",
+      "version": "2026-01-23",
+      "spec": "https://ucp.dev/2026-01-23/specification/discount",
+      "schema": "https://ucp.dev/2026-01-23/schemas/shopping/discount.json",
       "extends": ["dev.ucp.shopping.checkout", "dev.ucp.shopping.cart"]
     }
   ]
@@ -261,7 +261,7 @@ This convention ensures:
 
 - **Self-documenting**: The schema declares exactly which parents it extends
 - **Deterministic resolution**: The `extends` value maps directly to the `$defs` key
-- **Validatable**: Build-time validation can verify each `extends` entry has a
+- **Verifiable**: Build-time checks can confirm each `extends` entry has a
     matching `$defs` key
 
 #### Schema Resolution Convention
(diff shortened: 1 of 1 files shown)


raginpirate (approved):
Awesome, this addressed my question posed here exactly 😄 https://github.com/Universal-Commerce-Protocol/ucp/pull/73#discussion_r2704846704

docs/specification/overview.md
```diff
@@ -167,7 +167,7 @@ functionality is supported and where to find documentation and schemas.
 #### Extensions
 
 An **extension** is an optional module that augments another capability.
-Extensions use the `extends` field to declare their parent:
+Extensions use the `extends` field to declare their parent(s):
```
lemonmade, line 168: Did you choose to keep the string version just to avoid the breaking change? It feels a little awkward to have the union type to me (every tool dealing with profiles will need to do an array check/ normalize the shape), but I can see the value in not breaking this very important type.

docs/specification/overview.md
```diff
@@ -167,7 +167,7 @@ functionality is supported and where to find documentation and schemas.
 #### Extensions
```
lemonmade, line 165: Can you remind me why we wanted to explicitly list `extends` on the capability response objects? Now that we have a more explicit indication within the schema of how the extension augments various other types, I am not sure I see the need for this field. Would an implementer be expected to not augment a type from a base capability if an extension does not list that capability in `extends`, even if they have an definition with that exact name in their schema?

docs/specification/overview.md
```diff
@@ -558,6 +612,35 @@ The `capabilities` registry in responses indicates active capabilities:
 }
 ```
 
+#### Response Capability Selection
```
lemonmade, line 616: This section would not apply to `/.well-known/ucp`, right?

lemonmade (commented):
This definitely helps a lot to make schema resolution more clear, thank you for the update 👍

docs/specification/overview.md
knightlin-shopify: I had the same question. What convinced me is that it enables capability intersection/pruning without loading and parsing the extension schema during the capability pruning in run time. The downside is the "parent" info need to duplicated in two places and keep in sync.

docs/specification/overview.md
igrigorik: Correct - this is about responses, not discovery profiles.

docs/specification/overview.md
igrigorik: Yes. Open to changing it to array-only, it didn't feel onerous and I suspect majority use case is single extends.

docs/specification/overview.md
igrigorik: @knightlin-shopify yes exactly, you can do pruning without fetching the schema, but even more importantly.. It allows you to be selective about what you enable -- e.g. if discount supports extensions for both cart and checkout, without explicit declaration we'd have to assume that it applies everywhere, and by explicitly enumerating via `extends` we allow targeted opt-in.

docs/specification/overview.md
maximenajim: In capability negotiation small inconsistencies like duplicate parents (["a","a"]) or non-deterministic ordering can lead to subtle divergence (“we negotiated the same thing” but serialized differently).

We should add spec language such as:

- extends array MUST NOT contain duplicates
- extends array SHOULD be sorted lexicographically (or another canonical order)
- Implementations MUST treat a string extends as equivalent to a singleton array for processing.

igrigorik: @maximenajim we could, but I think our [intersection guidance](https://ucp.dev/latest/specification/overview/#intersection-algorithm) already covers these edge cases?

PR #103, opened 2026-01-22
docs: Cleanup payment responses from checkout schema

raginpirate:
# Description

A member of TC flagged that we have some disjointed references in checkout for display data and payment responses.
To resolve this, I'm clarifying that:
- display is the right obj on instruments for all examples.
- Payment has a single shape now (not the fragmented create/req objects).
- Payment handlers are not in base checkout anymore (already visible under the UCP schema).
- Card and Token instruments are not directly associated with checkout; these are entirely dependent on the merchants defined handlers, so they are removed from the base doc of checkout.

## Type of change
- [x] Documentation update

---

## Checklist:

### DIFF
--- a/docs/specification/checkout.md
+++ b/docs/specification/checkout.md
@@ -444,41 +444,15 @@ field or omitting them.
 
 ### Payment
 
-#### Payment Create Request
-
-{{ schema_fields('payment.create_req', 'checkout') }}
-
-#### Payment Update Request
-
-{{ schema_fields('payment.update_req', 'checkout') }}
-
-#### Payment Response
-
-{{ schema_fields('payment_resp', 'checkout') }}
-
-### Payment Handler Response
-
-{{ schema_fields('types/payment_handler_resp', 'checkout') }}
+{{ schema_fields('payment', 'checkout') }}
 
 ### Payment Instrument
 
 {{ schema_fields('payment_instrument', 'checkout') }}
 
-### Card Payment Instrument
-
-{{ schema_fields('card_payment_instrument', 'checkout') }}
-
 ### Payment Credential
 
 {{ schema_fields('payment_credential', 'checkout') }}
-
-### Token Credential Response
-
-{{ schema_fields('token_credential.create_req', 'checkout') }}
-
-### Card Credential
-
-{{ schema_fields('card_credential', 'checkout') }}
 
 ### Postal Address
 
--- a/spec/schemas/shopping/types/payment_instrument.json
+++ b/spec/schemas/shopping/types/payment_instrument.json
@@ -31,7 +31,7 @@
     },
     "display": {
       "type": "object",
-      "description": "Display information for this payment instrument."
+      "description": "Display information for this payment instrument. Each payment instrument schema defines its specific display properties, as outlined by the payment handler."
     }
   },
   "additionalProperties": true,
--- a/docs/specification/checkout-rest.md
+++ b/docs/specification/checkout-rest.md
@@ -157,7 +157,7 @@ All REST endpoints **MUST** be served over HTTPS with minimum TLS version
             "handler_id": "shop_pay_1234",
             "type": "shop_pay",
             "selected": true,
-            "display_data": {
+            "display": {
               "email": "buyer@example.com"
             }
           }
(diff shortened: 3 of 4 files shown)


drewolson-google (approved):

docs/specification/checkout.md
```diff
@@ -444,42 +444,16 @@ field or omitting them.
 
 ### Payment
 
-#### Payment Create Request
-
-{{ schema_fields('payment.create_req', 'checkout') }}
```
lemonmade, line 449: We don’t need to show the per-operation types because they are all basically identical to the "GET" shape for payment, right?

lemonmade (approved):

docs/specification/checkout.md
```diff
@@ -444,42 +444,16 @@ field or omitting them.
 
 ### Payment
 
-#### Payment Create Request
```
maximenajim, line 447: This change removes the documentation references. Rather than removing these documentation sections, should we:
1. Create the actual schema files
2. Keep the documentation sections
3. Ensure they render properly in the generated docs

spec/schemas/shopping/types/payment_instrument.json
```diff
@@ -31,7 +31,7 @@
     },
     "display": {
       "type": "object",
-      "description": "Display information for this payment instrument."
+      "description": "Display information for this payment instrument. Defined by each payment instrument schema."
```
maximenajim: Should we create a concrete schema instead of a generic description?

While "Defined by each payment instrument schema" is flexible, I think having a display object definition with fields like might be better:
- brand
- last_digits
- card_art
- description (or rich_text_description)

I'd recommend creating `types/payment_instrument_display.json` with a proper schema:
```json
{
  "type": "object",
  "properties": {
    "brand": { "type": "string" },
    "last_digits": { "type": "string" },
    "card_art": { "type": "string", "format": "uri" },
    "description": { "type": "string" },
    "email": { "type": "string", "format": "email" }
  },
  "additionalProperties": true
}
```
Then reference it:
```
display: {
  description: Display information for this payment instrument.,
  allOf: [{ $ref: payment_instrument_display.json }]
}
```
This provides:
✅ Schema validation
✅ Consistent field names across handlers
✅ Extensibility via additionalProperties: true
✅ Clear API contract for implementers

raginpirate: @maximenajim display is up to handlers to define, right?
Payment instruments are not just credit cards. These can range from standard methods like card/bank to opaque wallets like PayPal/ShopPay to LPMs like iDEAL etc.
I don't think we have any standards to bring into UCP at this time for this.

The above struct you are looking for currently exists on the card payment instrument, because that is a solid definition of a card's display properties 😄 It could have been split out into a card_payment_instrument_display.json and have a base type like credential or instrument, but theres nothing to really share between any of them 😅

docs/specification/checkout.md
raginpirate: Those objs are actually dead, there are no unique create vs update vs resp shapes for the base payment schema.
This still links to the schema for payment which is shared across all of them.

spec/schemas/shopping/types/payment_instrument.json
maximenajim: @raginpirate  thank you for the clarification!

This highlights a gap:
1. Each payment instrument type (card, wallet, etc.) defines its own display structure
2. The base `payment_instrument.json` intentionally keeps it flexible
3. `card_payment_instrument.json` already has the concrete definition for cards

Should we add clarifying documentation? Something like:
```
"display": {
  "type": "object",
  "description": "Display information for this payment instrument. Each payment instrument type (card_payment_instrument, etc.) defines its specific display properties. Payment handlers MAY add additional handler-specific fields."
}
```

**In the docs (`checkout.md`):**
Add a note explaining:
> The `display` object structure varies by payment instrument type. See specific payment instrument schemas (e.g., `card_payment_instrument`) for their display field definitions.

Thoughts on this approach?

raginpirate: Sounds good, I already was updating the json, happy to make it a bit more verbose and throw a bit extra into the checkout.md 😄

raginpirate: Hmm, actually I'm really struggling to find a place in checkout.md to add this. I think this context is already very well described under the payment handler guide ("Instrument Shapes" section).

igrigorik (approved):

PR #109, opened 2026-01-23
feat(cart): Embedded protocol transport binding for cart capability

jingyli:
## Summary

Cart capability (`dev.ucp.shopping.cart`) is being introduced in https://github.com/Universal-Commerce-Protocol/ucp/pull/73. This adds an additional transport binding (EP) on the capability (building directly upon the referenced PR), on top of existing REST & MCP bindings.

## Motivation

Businesses need a way to embed their cart building UI into eligible platforms, especially when complex experiences (i.e. item recommendations & upsells) are involved during cart building. Embedded Cart Protocol (ECaP) addresses this need and also offers a seamless connection to existing Embedded Checkout Protocol to continue the purchase experience.

### Goals

* Provide seamless transition between embedded protocols to enable cart-to-checkout conversions.
* Maintain general schema & design consistency with existing ECP (i.e. reuse naming conventions, message types, etc.)
    * There remains some fundamental design differences in the design (i.e. host-initiated notification to share required authorization data, change requests that require host responses, etc.)

### Non-Goals

## Detailed Design

| Category          | Communication Direction  | Purpose                                                               | Pattern      | Core Messages                                |
| :---------------- | :----------------------- | :---------------------------------------------------------------------| :----------- | :------------------------------------------- |
| **Handshake**     | Embedded Cart -> Host   | Establish connection between host and Embedded Cart.                  | Request      | `ect.ready`                                  |
| **Authentication**| Host -> Embedded  Cart    | Inform any additional auth data required by Embedded Cart from host.  | Notification | `ect.auth`                                   |
| **Lifecycle**     | Embedded Cart -> Host   | Inform of cart state transitions.                                     | Notification | `ect.start`, `ect.transition.checkout`|
| **State Change**  | Embedded Cart -> Host   | Inform of cart field changes.                                         | Notification | `ect.line_items.change`, `ect.buyer.change`, `ect.context.change`, `ect.messages.change` |
| **State Change**  | Embedded Cart -> Host   | Inform of cart field changes but also require host input.             | Request      | `ect.line_items.change_request`              |

**Protocol transition**
* Introducing a new message type: `[protocol namespace].transition.[capability transitioning to]` that can be generally extended if conversions between multiple capabilities are possible.
  * First use case we are introducing is `ect.transition.checkout` that would denote a cart transitioning to a checkout session on the business's iframe.

## Risks and Mitigations

* **Security:** Introduced new host-initiated notification mechanism `ect.auth` to exchange required authorization data instead of adding them as query parameters in `continue_url` to avoid hijacking attacks.

## Graduation Criteria

**Candidate → Stable:**

### Implementation History

TBD

### DIFF
--- a/docs/specification/embedded-cart.md
+++ b/docs/specification/embedded-cart.md
@@ -204,7 +204,7 @@ all implementations.
 | Category          | Communication Direction | Purpose                                                               | Pattern      | Core Messages                                |
 | :---------------- | :---------------------- | :---------------------------------------------------------------------| :----------- | :------------------------------------------- |
 | **Handshake**     | Embedded Cart -> Host   | Establish connection between host and Embedded Cart.                  | Request      | `ect.ready`                                  |
-| **Authentication**| Host -> Embedded        | Inform any additional auth data required by Embedded Cart from host.  | Notification | `ect.auth`                                   |
+| **Authentication**| Host -> Embedded Cart   | Inform any additional auth data required by Embedded Cart from host.  | Notification | `ect.auth`                                   |
 | **Lifecycle**     | Embedded Cart -> Host   | Inform of cart state transitions.                                     | Notification | `ect.start`, `ect.transition.checkout`|
 | **State Change**  | Embedded Cart -> Host   | Inform of cart field changes.                                         | Notification | `ect.line_items.change`, `ect.buyer.change`, `ect.context.change`, `ect.messages.change` |
 | **State Change**  | Embedded Cart -> Host   | Inform of cart field changes but also require host input.             | Request      | `ect.line_items.change_request`              |


docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,726 @@
...
+}
+```
+
+When the host responds with an `upgrade` object, the Embedded Cart **MUST**
+discard any other information in the message, send a new `ect.ready` message
+over the upgraded communication channel, and wait for a new response. All
+subsequent messages **MUST** be sent only over the upgraded communication
+channel.
+
+The host **MAY** also respond with a `cart` object, if no additional auth
```
lemonmade, line 292: IMO, we should not have added this to the checkout protocol, so I am worried about an even more powerful version of it being recommended here (in checkout, we restrict to only allowing payment info). Making the whole cart potentially editable in `ec.ready` means a business can’t render any content on the server, and they must follow the worst-performing way to build an embedded webview: HTML loads => JS loads => `ec.ready` is received => network request to get full cart data => paint the page.

The assumption is that a buyer only enters the Embedded Cart flow after a UCP request, so I would much prefer we encourage the platform who made that UCP request make an `update_cart` with all this data, *then* open the embedded flow.

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,726 @@
...
+
+#### `ect.auth`
+
+Informs Embedded Cart of any required auth data from host per business
+requirements (i.e. when identity linking is a pre-requisite).
+
+- **Direction:** Host → Embedded Cart
+- **Type:** Notification
+- **Payload:**
+    - `authorization` (string, **REQUIRED**): The required authorization data by
```
lemonmade, line 329: Why aren’t we following the same `ec_auth` approach we have in the embedded checkout protocol?

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,726 @@
...
+    "jsonrpc": "2.0",
+    "method": "ect.auth",
+    "params": {
+        "authorization": "fake_token_from_identity_linking"
+    }
+}
+```
+
+The host **MAY** also specify a `cart` object to Embedded Cart, especially if
+it was not communicated over UCP cart actions or `ect.ready`, which will be
```
lemonmade, line 348: Why would it not be communicated over cart UCP actions instead? This flow seems backwards — you get into ECaP *after* creating a UCP cart with data, not before.

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,726 @@
...
+occurred partially and depending on host response, the final changes
+occur in Embedded Cart.
+
+#### `ec.line_items.change_request`
+
+Embedded Cart would like to request line items modifications (quantity
+changed, items added/removed) in the cart UI. However, it's waiting
+for host input (i.e. validating the update to see if it's compliant
+with host's requirement) before applying the changes. Embedded Cart
+**MUST NOT** fully apply the changes without receiving a response
```
lemonmade, line 589: Do you have examples of UI for this? How would an embedded view present the loading state while waiting for the response, and how should they communicate the changes that were/ weren’t accepted by the host, in the case of partial acceptance?

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,726 @@
...
+| :---------------- | :---------------------- | :---------------------------------------------------------------------| :----------- | :------------------------------------------- |
+| **Handshake**     | Embedded Cart -> Host   | Establish connection between host and Embedded Cart.                  | Request      | `ect.ready`                                  |
+| **Authentication**| Host -> Embedded Cart   | Inform any additional auth data required by Embedded Cart from host.  | Notification | `ect.auth`                                   |
+| **Lifecycle**     | Embedded Cart -> Host   | Inform of cart state transitions.                                     | Notification | `ect.start`, `ect.transition.checkout`|
+| **State Change**  | Embedded Cart -> Host   | Inform of cart field changes.                                         | Notification | `ect.line_items.change`, `ect.buyer.change`, `ect.context.change`, `ect.messages.change` |
+| **State Change**  | Embedded Cart -> Host   | Inform of cart field changes but also require host input.             | Request      | `ect.line_items.change_request`              |
```
lemonmade, line 210: Should this not have an accompanying "delegation", like in ECP? There is no mention of delegation in this doc, unless I missed it.

lemonmade: Also, this category name is duplicated from the one above it. And it is not really "informing" the host, as the host has full control over the final update.

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,726 @@
...
+            "totals": [/* ... */],
+            "line_items": [/* ... */],
+            "buyer": {/* ... */},
+            "context": {/* ... */}
+        }
+    }
+}
+```
+
+#### `ect.transition.checkout`
```
lemonmade, line 409: Do all carts need to be able to support this message? This would be a fairly big lift for Shopify, and I assume any other platform that actually has a need to distinguish cart vs checkout.

I would personally prefer to more clearly mark the transition between cart and checkout, like we do in UCP actions. This would be something like an `ect.checkout` action, which just informs the host that the embedded cart has shown intent to checkout. A host could then use this hook to run the `create_checkout(cart_id)` operation, and either submit natively (I assume this would be preferred?), or escalate separately into Checkout by establishing an ECP connection from scratch. This would make the common "native submission" flow more natural, and would avoid some tricky implementation questions (e.g., what if you can’t host cart and checkout on the same origin? Do we skip `ec.ready` altogether if you already got `ect.ready`?)

docs/specification/embedded-cart.md
jingyli: Resolved in https://github.com/Universal-Commerce-Protocol/ucp/pull/116/commits/4252786597206d1d75b149b9f6728f970fce4101

docs/specification/embedded-cart.md
jingyli: Resolved in https://github.com/Universal-Commerce-Protocol/ucp/pull/116/commits/4252786597206d1d75b149b9f6728f970fce4101

PR #115, opened 2026-01-23
refactor!: updates for 2026-01-23 release

wry-ry:


.github/workflows/linter.yaml
```diff
@@ -16,7 +16,7 @@ name: Lint Code Base
 
 on:
   pull_request:
-    branches: [main]
+    branches: [main, working-draft]
```
drewolson-google: We can remove `working-draft` after this merge.

PR #116, opened 2026-01-24
feat(cart): Embedded protocol transport binding for cart capability

jingyli:
Reopening the PR to avoid recreating the `working-draft` branch. This is a mirror copy of https://github.com/Universal-Commerce-Protocol/ucp/pull/109 and will also cross-reference/carry-over comments from the old PR.

## Summary

Cart capability (`dev.ucp.shopping.cart`) is being introduced in https://github.com/Universal-Commerce-Protocol/ucp/pull/73. This adds an additional transport binding (EP) on the capability (building directly upon the referenced PR), on top of existing REST & MCP bindings.

## Motivation

Businesses need a way to embed their cart building UI into eligible platforms, especially when complex experiences (i.e. item recommendations & upsells) are involved during cart building. Embedded Cart Protocol (ECaP) addresses this need and also offers a seamless connection to existing Embedded Checkout Protocol to continue the purchase experience.

### Goals

* Provide seamless transition between embedded protocols to enable cart-to-checkout conversions.
* Maintain general schema & design consistency with existing ECP (i.e. reuse naming conventions, message types, etc.)
    * There remains some fundamental design differences in the design (i.e. host-initiated notification to share required authorization data, change requests that require host responses, etc.)

### Non-Goals

## Detailed Design

| Category          | Communication Direction | Purpose                                                                   | Pattern                | Core Messages                                                                             |
| :---------------- | :---------------------- | :------------------------------------------------------------------------ | :--------------------- | :---------------------------------------------------------------------------------------- |
| **Handshake**     | Embedded Cart -> Host   | Establish connection between host and Embedded Cart.                      | Request                | `ect.ready`                                                                               |
| **Authentication**| Host <-> Embedded Cart  | Communicate auth data exchanges between Embedded Cart and host.           | Notification & Request | `ect.auth` (Request), `ect.auth.change` (Notification)                                    |
| **Lifecycle**     | Embedded Cart -> Host   | Inform of cart state in Embedded Cart.                                    | Notification           | `ect.start`                                                                               |
| **Transition**    | Embedded Cart -> Host   | Establish transition from cart to other capabilities.                     | Request                | `ect.transition.checkout`                                                                 |
| **State Change**  | Embedded Cart -> Host   | Inform of cart field changes.                                             | Notification           | `ect.line_items.change`, `ect.buyer.change`, `ect.context.change`, `ect.messages.change`  |

**Protocol transition**
* Introducing a new message type: `[protocol namespace].transition.[capability transitioning to]` that can be generally extended if conversions between multiple capabilities are possible.
  * First use case we are introducing is `ect.transition.checkout` that would denote a cart transitioning to a checkout session on the business's iframe.

## Risks and Mitigations

* **Security:** Introduced new host-initiated notification mechanism `ect.auth` to exchange required authorization data instead of adding them as query parameters in `continue_url` to avoid hijacking attacks.

## Graduation Criteria

**Candidate → Stable:**

### Implementation History

TBD

### DIFF
--- a/source/schemas/shopping/order.json
+++ b/source/schemas/shopping/order.json
@@ -82,6 +82,10 @@
       },
       "description": "Append-only event log of money movements (refunds, returns, credits, disputes, cancellations, etc.) that exist independently of fulfillment."
     },
+    "currency": {
+      "type": "string",
+      "description": "ISO 4217 currency code. MUST match the currency from the originating checkout session."
+    },
     "totals": {
       "type": "array",
       "items": {


.cspell/custom-words.txt
```diff
@@ -80,3 +80,8 @@ upsells
 vulnz
 yaml
 yml
+ZEGTBCTMJ
```
lemonmade: Intentional?

.cspell/custom-words.txt
```diff
@@ -80,3 +80,8 @@ upsells
...
+llms
+paymentmethodchange
```
lemonmade: I don’t see this elsewhere in the PR

.cspell/custom-words.txt
```diff
@@ -80,3 +80,8 @@ upsells
...
+reauth
+reprepare
```
lemonmade: This is not a word, I’d suggest rewriting the sentence it’s in instead.

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,770 @@
...
+            "version": "2026-01-23",
+            "rest": {
+                "schema": "https://ucp.dev/services/shopping/rest.openapi.json",
+                "endpoint": "https://merchant.example.com/ucp/v1"
+            },
+            "mcp": {
+                "schema": "https://ucp.dev/services/shopping/mcp.openrpc.json",
+                "endpoint": "https://merchant.example.com/ucp/mcp"
+            },
+            "embedded": {
```
lemonmade: Can you remind me if a business implementing UCP must implement every part of the transport for a given service (e.g., every message in embedded protocol, endpoint/ action in REST/ MCP)? Up until now it had not occurred to me that this might be an issue, since the business I represent would happily implement all the MCP cart actions alongside checkout, but this may not be the case with embedded, where Shopify may prefer to/ temporarily have only the checkout embed messages hooked up.

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,770 @@
...
+
+When the host responds with an `upgrade` object, the Embedded Cart **MUST**
+discard any other information in the message, send a new `ect.ready` message
+over the upgraded communication channel, and wait for a new response. All
+subsequent messages **MUST** be sent only over the upgraded communication
+channel.
+
+### Authentication
+
+#### `ect.auth`
```
lemonmade: I don’t think I understand how this message works from the Cart’s perspective. If the Cart sends `ec.ready` with `requires_auth`, what do they do while they wait for this auth message to come in? Why wouldn’t the auth be part of the response to that ready message, if it is actually required?

More generally, it’s not clear to me why this would exist for Cart and not Checkout. If a business wanted to enact this same identity linking requirement in Checkout, I assume they would do it with a `messages` error indicating that identity linking is required, and they would expect a UCP consumer to attach the necessary information before escalating to an embedded cart. Why not follow that same approach here, rather than creating an additional request-response cycle between host and business?

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,770 @@
...
+
+To initiate the session, the host **MUST** augment the `continue_url` with ECaP
+query parameters using the `ect_` prefix.
+
+All ECaP parameters are passed via URL query string, not HTTP headers, to ensure
+maximum compatibility across different embedding environments. Parameters use
+the `ect_` prefix to avoid namespace pollution and clearly distinguish ECaP
+parameters from business-specific query parameters:
+
+- `ect_version` (string, **REQUIRED**): The UCP version for this session
```
lemonmade: Can we add a `ect_auth` param (as a parallel to `ec_auth`)? In ECP, that param serves to identify the embedder and to unlock customizations out-of-band from UCP (e.g., the branding overrides of an embedded checkout or cart); I feel this is needed in the initial request for cart in the same way it is for checkout (so the server render of the page can incorporate knowledge of the embedder).

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,770 @@
...
+            "buyer": {/* ... */},
+            "context": {/* ... */}
+        }
+    }
+}
+```
+
+### Transition Messages
+
+#### `ect.transition.checkout`
```
lemonmade: As I noted in my previous review, I feel this message is quite odd. As described, it implies that the buyer would always see a web checkout, and that the same HTML page will morph from cart to checkout, which I do not believe will be a common implementation pattern. It also means that the cart creates the checkout and navigates to it, which requires a separate parallel way to do everything you might do in that transition (you added `ec_delegate` param as a response, but omitted `ec_auth`; similarly, there is no way to attach data to `create_checkout` that is not attached to the cart).

I would have expected a much simpler notification here instead, indicating that the user has attempted to checkout (e.g., something like `ect.checkout_request`). This puts the embedder in control to use whatever approach to checkout is most appropriate, exactly as they would if they were building their own, non-embedded cart; they can call `create_checkout` with anything they like, and can construct an ECP URL/ initiate an ECP escalation in the standard way.

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,770 @@
...
+occurred partially and depending on host response, the final changes
+occur in Embedded Cart.
+
+#### `ec.line_items.change_request`
+
+Embedded Cart would like to request line items modifications (quantity
+changed, items added/removed) in the cart UI. However, it's waiting
+for host input (i.e. validating the update to see if it's compliant
+with host's requirement) before applying the changes. Embedded Cart
+**MUST NOT** fully apply the changes without receiving a response
```
lemonmade: What does "fully apply the changes" mean? I think we need to be more specific here — the embedded checkout would obviously need to indicate something changed, otherwise the buyer would think their edit had been ignored. I also think there should be some sort of guidance on timeouts here, since the action being blocked is one that every implementation would expect to be synronous or near-instant (if I change line item quantity by 1, I do not want to have to wait for 2 seconds to see that change reflected).

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,770 @@
...
+```
+
+### State Change Messages Requesting for Host Input
+
+ECaP support specific state change messages that also require host
+input. We consider these as change message requests where the UI update
+occurred partially and depending on host response, the final changes
+occur in Embedded Cart.
+
+#### `ec.line_items.change_request`
```
lemonmade: Why does this exist in cart, but not checkout? I feel we are creating a strange patchwork, where one side has request messages for some fields (`fulfillment.address_change_request`, `payment.credential_request`) and the other side has messages for other fields (`line_items.change_request`). Is there a rule that guides what requests are available where, or is it just, "whatever the first implementers wanted in their particular UX"?

Additionally, should this not be a formal delegation, instead of something that is expected of all embedded carts? We did this for the checkout "let host handle some parts of updates" flows, and I think it applies equally well to this action.

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,770 @@
...
+
+**Example Message:**
+
+```json
+{
+    "jsonrpc": "2.0",
+    "id": "items_change_request_1",
+    "result": {
+        "cart": {
+            "line_items": [
```
lemonmade: Is the host allowed to edit every detail of these lines, or only some aspects?

PR #125, opened 2026-01-27
fix(schema): publish UCP annotated schemas

igrigorik:
Our contract is that UCP schemas are self-describing. We define and advertise one URL in capability metadata, which agents and servers can fetch and validate against. At the moment, this contract is broken. 

```json
  "dev.ucp.shopping.fulfillment": [
    {
      "version": "2026-01-11",
      "spec": "https://ucp.dev/specification/fulfillment",
      "schema": "https://ucp.dev/schemas/shopping/fulfillment.json",  // <~ single schema
      "extends": "dev.ucp.shopping.checkout"
    }
  ]
```

A schema yields different shapes based on operation (e.g., create vs update vs complete) and direction (request vs response). We encode these rules via `ucp_request`/`ucp_response` annotations in our `source/*` schemas, but then run a preprocessor that splits them into N permutations based on operation × direction (e.g., `checkout.create_req.json` maps to create + request). The schema URL we advertise in UCP metadata doesn't contain all the necessary information, and we've defined a convention that off-the-shelf schema validators are not aware of and would need to be taught to understand. As is, this makes UCP validation very hard to implement.

### Proposed desired state

> **Deliver self-describing schemas. Keep servers+agents simple: one URL. Handle complexity in shared tooling.**

By count and burden: servers > agents > validators. If we require that servers dynamically resolve and return operation-specific schema refs, that's complexity multiplied across every UCP business / implementer. Same for agents, who would need to vary their profile per operation — clunky. We can eliminate and shift all of this complexity into validators: restore the single file contract, a ref to which can be passed in both directions by servers+agents, and lean on UCP-aware tooling. 

| Actor | Per-op files (today) | Single schema (proposed) |
|-------|---------------------------|----------------------|
| **Spec authors** | N files per capability (via preprocessor) | 1 file |
| **Server implementers** | Hardcode operation→URL mapping | Return one URL |
| **Validator** | Simple fetch + validate | Resolve annotations |
| **Discovery** | Multiple URLs, custom convention | One URL serves all |
| **Readability** | Rules scattered across files | Self-documenting |

### Recommendation

1. Publish source schemas with `ucp_request`/`ucp_response` annotations intact
2. Remove the preprocessor and `/spec/schemas/` directory
3. Adopt portable validator for linting, resolution, validation. 

Prototype validator: [ucp-schema](https://github.com/Shopify/ucp-schema)

```bash
# Lint schemas in CI
ucp-schema lint schemas/

# Resolve to vanilla JSON Schema (for tools that need it)
ucp-schema resolve checkout.json --request --op create

# Validate directly (self-describing from ucp.capabilities)
ucp-schema validate response.json --op read
```

The validator is written in Rust for portability & re-use. We can use it for our build process, and we can provide it as official binary for validating schemas for capability developers, and for agent+server developers to validate payloads on the wire — see docs on above repo for linting, resolving, validation, etc.

### This PR

Big diff, but mostly deletions and simplification of the build pipeline in favor of using the shared ucp-schema validator.

1. Integrates `ucp-schema`, deprecates Python preprocessor
2. Publishes `source` schemas with annotations intact (non-breaking: standard validators ignore unknown keywords)
3. Updates CI to use `ucp-schema lint` + `ucp-schema validate`

#### Big picture changes

**1. Runtime schema resolution via `ucp-schema` CLI** (`main.py`)
- Calls `ucp-schema resolve` during mkdocs build
- Caches resolved schemas per-schema to avoid redundant work
- Detects polymorphic types for correct anchor generation

**2. Versioned schema publishing** (`hooks.py`)
- Publishes source schemas to `/schemas/{version}/` during build
- Rewrites `$id` and `$ref` URLs to include version path
- Sets `version` field in capability metadata
- Handles date versions (`2026-01-26`) vs. named versions (`draft`)

**3. CI integration** (`.github/workflows/docs.yml`)
- Installs `ucp-schema` via `cargo install`
- Runs `ucp-schema lint source/` before build
- Schema errors fail the build

### Before/After

```
Before:                              After:
source/schemas/checkout.json         source/schemas/checkout.json
       ↓                                    ↓
generate_schemas.py                  (direct publish)
       ↓                                    ↓
spec/schemas/checkout.json           https://ucp.dev/{version}/schemas/shopping/checkout.json
spec/schemas/checkout.create_req.json       (with annotations intact)
spec/schemas/checkout.update_req.json
spec/schemas/checkout_resp.json
```

Single schema URL, versioned. All site-generation markdown logic remains functional & intact.

-----

## Type of change

- [x] Bug fix

## Checklist

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,4 +1,9 @@
 {
+  "vcs": {
+    "enabled": true,
+    "clientKind": "git",
+    "useIgnoreFile": true
+  },
   "linter": {
     "enabled": true,
     "rules": {


.github/workflows/docs.yml
```diff
@@ -81,19 +81,26 @@ jobs:
...
         uses: tj-actions/changed-files@v46
         with:
-          files: spec/**
+          files: source/**
 
       - name: Check Python SDK Models Consistency
-        if: steps.spec_files_changed.outputs.any_changed == 'true'
+        if: steps.source_files_changed.outputs.any_changed == 'true'
         run: |
           chmod +x scripts/ci_check_models.sh
```
drewolson-google: We should be able to check the executable bit into `git` so we don't have to run this every time.

hooks.py
```diff
@@ -77,15 +89,73 @@ def _process_refs(data, current_file_dir):
...
+      else:
+        _rewrite_version_urls(value, url_version)
+  elif isinstance(data, list):
+    for item in data:
+      _rewrite_version_urls(item, url_version)
+
+
+def _set_schema_version(data, version):
+  """Set version field in capability schemas (top-level 'version' property)."""
+  if "version" in data:
```
wry-ry: Would it make sense to remove version from the source file completely, and add it here for all files?

wry-ry (commented):
LGTM

drewolson-google (approved):

hooks.py
igrigorik: 👍 updated

PR #132, opened 2026-01-28
chore: add script to run super linter locally

drewolson-google:
# Description

This PR introduces `scripts/super_linter_local.py` that runs our super linter Github action locally. The script reads the `yaml` config for the action and converts it into a command that runs a local container for super linter.

The command supports two container runtimes via the `--runtime` flag: `docker` and `podman`. A default is chosen based on the contents of your `PATH`.

The command supports providing a branch to compare against via the `--branch` flag. The default is `main`.

Here's an example invocation:

```
./scripts/super_linter_local.py --runtime podman --branch main
```

For usage information, run:

```
./scripts/super_linter_local.py -h
```

## Type of change

- [x] Chore

## Checklist

### DIFF
--- a/scripts/super_linter_local.py
+++ b/scripts/super_linter_local.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env -S uv run
+
+"""Run super-linter locally using configuration from GitHub Actions."""
+
+import argparse
+import re
+import shutil
+import subprocess
+import sys
+from pathlib import Path
+
+import yaml
+
+
+def map_action_to_image(action_str):
+  """Map GitHub Action string to ghcr.io image."""
+  # super-linter/super-linter/slim@v8 ->
+  # ghcr.io/super-linter/super-linter:slim-v8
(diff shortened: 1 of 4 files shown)


scripts/super_linter_local.py
```diff
@@ -0,0 +1,149 @@
...
+import argparse
+import re
+import subprocess
+import sys
+from pathlib import Path
+
+import yaml
+
+
+def get_current_commit(vcs):
```
0x8000-0000: Can it walk from ${pwd} up to root until it finds either .jj or .git, so this function (and tool) does not need the vcs argument?

drewolson-google: Good point. Mind if I address this in a follow up PR?

drewolson-google: Ok, this is fixed. I misunderstood the `DEFAULT_BRANCH` parameter -- it is the branch to compare against. Given that, it defaults to `main` and we no longer need to detect the current commit at all.

wry-ry (approved):

PR #133, opened 2026-01-28
refactor: openapi dedupe errors

oericdacosta:
refactor: deduplicate parameters and standardize error responses

# Description

This PR refactors the Shopping Service OpenAPI specification source/services/shopping/openapi.json to improve maintainability and strictly follow the DRY principle.

**Key Changes:**
1.  **Deduplicated Header Parameters:** Moved 9 common header parameters (e.g., `Authorization`, `X-API-Key`, `Request-Id`) from individual operations to the path level. This significantly reduces redundancy across endpoints like `/checkout-sessions` and `/carts`.
2.  **Standardized Error Responses:** Introduced a reusable `error_response` schema and added standard HTTP error responses (400, 401, 404, 500) to all operations, referencing shared definitions in `components/responses`.
3.  **Synced with Upstream:** Aligned with the latest project structure by removing legacy spec/ artifacts, ensuring a clean merge with main

These changes make the OpenAPI spec more concise, easier to read, and ensure consistent error handling documentation across the API.

## Type of change

- [x] Refactor (non-breaking change which improves code quality)

---

files or fields:

```text
N/A - This is a purely structural refactor of the source OpenAPI file.
```

---

## Checklist


jingyli:
Thank you for the contribution! Some comments after going through the PR:
1) It's a great idea to refactor out the common header parameters definition to path level to reduce duplication and make it less error prone when we add new operations.
2) Error responses actually have been standardized since this PR is open: https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/common/types/error_response.json - so I don't think we need to do it again here.
3) I'm not 100% convinced if we want to outline all the non-200 responses in our `openapi` definition. Given https://ucp.dev/latest/specification/overview/#error-handling, if we do want to specify everything, then we should make sure we include all that have been spec'ed out.

Given the file has evolved quite a bit, if you want to still keep this PR open, it'd be great to resolve the conflicts so we can take a look at the updated deltas again! Otherwise, we can close this PR and reopen a fresh one for the first point listed above.

oericdacosta:
@jingyli Thanks for the detailed review and feedback!

I have just rebased the branch to resolve all conflicts and updated the PR.

Following your suggestion, I have completely removed the error response standardization from this PR, as it's already being handled elsewhere. The PR is now strictly focused on the first point: deduplicating the common header parameters by moving them to the path level.

Let me know if the update look good to you!

source/schemas/common/cart.complete_req.json
damaz91, line 1: @oericdacosta this file seems a bit out-of-scope for the PR intent of deduplication. Please remove it from the PR

oericdacosta:
@damaz91 Apologies for that! The cart.complete_req.json file slipped in accidentally during my local commits. I have just removed it from the branch and updated the PR. Thanks for catching that!

gsmith85:
Thanks for sticking with this one, @oericdacosta. You turned it around promptly every time we asked, including dropping the error-response work when @jingyli pointed at `error_response.json`, and pulling the stray `cart.complete_req.json`. Sorry this is landing as a close rather than a merge.

Two things have moved upstream since the last round that between them settle it.

**The parameters this PR hoists are being deleted.** #716 (open, same file) removes `#/components/parameters/authorization` and `x_api_key` entirely, replacing them with `securitySchemes` and a top-level `security` block. That is the fix for #668: declaring authentication as header parameters is invalid under OpenAPI 3.1 and ignored by most tooling. Two of the nine headers this PR relocates are on their way out, and hoisting them entrenches the pattern we are moving away from.

**This file is becoming generated output.** RFC #800 proposes compiling the REST binding from the JSON Schema sources via `ucp-schema export-openapi`, with the result committed to `dist/` and guarded by `--check`. FR-6 of that work makes protocol header binding a compiler responsibility emitted from `operations.rs`. It is still in Technical Council review and the companion PR is a draft, so it is not imminent, but it is the direction of travel.

A smaller supporting point on the refactor itself. The shipped spec has 11 paths and 13 operations, and only two paths carry more than one operation (`/checkout-sessions/{id}` and `/carts/{id}`). Path-item parameters only dedupe across operations sharing a path, so hoisting the 11 universal headers takes us from 143 parameter references to 121, roughly 15%. It also does not hoist cleanly: `content_digest` sits on 10 of the 13 operations, so it would end up advertised on three that deliberately omit it.

Worth flagging for completeness: the June rebase looks like it resolved conflicts by taking the branch's copy of the file wholesale. The current head is missing four paths (`/catalog/search`, `/catalog/lookup`, `/catalog/product`, `/orders/{id}`), four operations and six schema bindings, and carries seven `$ref`s that do not resolve (`request_signature` is referenced from every path item but defined nowhere). `webhooks.orderEvent.post` has also picked up `operationId: lookup_catalog`. That this stayed green is on us, not on you: we have `VALIDATE_OPENAPI: false` in `linter.yaml`, so nothing checks reference integrity. I am fixing that separately.

Closing this one. The discussion here is what prompted me to go measure the duplication and find the disabled linter, so it was not wasted.

If you would like to keep going, there is a better-shaped version of this same idea. Once `export-openapi` lands, header emission lives in `operations.rs` in [`ucp-schema`](https://github.com/Universal-Commerce-Protocol/ucp-schema), and deciding there whether common headers are emitted per-operation or per-path-item would apply uniformly across every generated profile rather than to one hand-maintained file. Happy to point you at the relevant code if that is of interest.

Thanks again for the patience across the review cycle.

PR #141, opened 2026-01-29
docs(mcp): fix MCP examples & errors

igrigorik:
MCP has own envelope that runs over JSONRPC. Existing MCP docs show JSONRPC without the MCP envelope, which this PR addresses and fixes outdated `meta` and missing `checkout` \ `cart` object wrappers while we're at it.

JSON-RPC request:
```json
{
  "jsonrpc": "2.0",
  "method": "create_checkout",
  "params": { /* UCP payload */ },
  "id": 1
}
```

Response:
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": { /* UCP payload */ }
}
```

MCP envelope (tools/call binding over JSON-RPC):

Request:
```json
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "create_checkout",
    "arguments": { /* UCP payload */ }
  },
  "id": 1
}
```

Updated examples now show correct MCP envelope. 

---
For error+warning examples
- standardize all error/warning codes to snake_case per schema
- fix adjusted_quantity example (warning, not error)

## Checklist
- [x] Documentation update

### DIFF
--- a/docs/specification/checkout-mcp.md
+++ b/docs/specification/checkout-mcp.md
@@ -594,8 +594,15 @@ Maps to the [Cancel Checkout](checkout.md#cancel-checkout) operation.
 
 ## Error Handling
 
-See the [Core Specification](overview.md#error-handling) for negotiation error
-handling (discovery failures, negotiation failures).
+UCP distinguishes between protocol errors and business outcomes. See the
+[Core Specification](overview.md#error-handling) for the complete error code
+registry and transport binding examples.
+
+* **Protocol errors**: Transport-level failures (authentication, rate limiting,
+    unavailability) that prevent request processing. Returned as JSON-RPC
+    `error` with code `-32000` (or `-32001` for discovery errors).
+* **Business outcomes**: Application-level results from successful request
+    processing, returned as JSON-RPC `result` with UCP envelope and `messages`.
 
 ### Business Outcomes
 
@@ -649,8 +656,7 @@ A conforming MCP transport implementation **MUST**:
 
 1. Implement JSON-RPC 2.0 protocol correctly.
 2. Provide all core checkout tools defined in this specification.
-3. Return negotiation failures per the
-    [Core Specification](overview.md#error-handling).
+3. Return errors per the [Core Specification](overview.md#error-handling).
 4. Return business outcomes as JSON-RPC `result` with UCP envelope and
     `messages` array.
 5. Validate tool inputs against UCP schemas.
--- a/docs/specification/overview.md
+++ b/docs/specification/overview.md
@@ -588,13 +588,31 @@ These failure types require different handling:
 
 ##### Error Codes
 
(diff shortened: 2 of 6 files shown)


richmolj (approved):

docs/specification/overview.md
```diff
@@ -637,15 +657,32 @@ task through the standard web interface.
...
           "type": "error",
-          "code": "VERSION_UNSUPPORTED",
-          "content": "Platform UCP version 2024-01-01 is not supported",
+          "code": "version_unsupported",
+          "content": "UCP version 2024-01-01 is not supported",
           "severity": "requires_buyer_input"
         }
       ],
-      "continue_url": "https://merchant.com/cart"
+      "continue_url": "https://merchant.com"
```
lemonmade, line 665: Should this field actually be here? I thought it would only be `ucp`+`messages`

docs/specification/overview.md
```diff
@@ -637,15 +657,32 @@ task through the standard web interface.
       "messages": [
         {
           "type": "error",
-          "code": "VERSION_UNSUPPORTED",
-          "content": "Platform UCP version 2024-01-01 is not supported",
+          "code": "version_unsupported",
+          "content": "UCP version 2024-01-01 is not supported",
           "severity": "requires_buyer_input"
```
lemonmade, line 662: To me, `requires_buyer_input` does not feel right here (or with any "business errors"). Would `recoverable` be more appropriate, since this is something the platform can correct with a follow-up API call.

docs/specification/overview.md
```diff
@@ -673,23 +710,62 @@ task through the standard web interface.
...
+          },
+          "messages": [
+            {
+              "type": "error",
+              "code": "version_unsupported",
+              "content": "UCP version 2024-01-01 is not supported",
+              "severity": "requires_buyer_input"
+            }
+          ],
+          "continue_url": "https://merchant.com"
```
lemonmade, line 726: Same question as above about the presence of this field

docs/specification/checkout-mcp.md
```diff
@@ -568,44 +594,58 @@ Maps to the [Cancel Checkout](checkout.md#cancel-checkout) operation.
...
 {
   "jsonrpc": "2.0",
   "id": 1,
   "result": {
-    "ucp": {
-      "version": "2026-01-11",
-      "capabilities": {
-        "dev.ucp.shopping.checkout": [{"version": "2026-01-11"}]
+    "structuredContent": {
+      "checkout": {
```
lemonmade, line 619: In quite a few of these examples, `structuredContent` and `content` include a wrapping `checkout`/ `cart` object. While I would have preferred this approach, the spec indicates that the return object *is* a cart/ checkout, so the wrappers should be removed.

lemonmade (approved):

PR #144, opened 2026-01-30
chore: update ts schema generator to use files in source/

wry-ry:
# Description

Update schema to generate_ts_schema_types.js to use source/ now that we have deleted the pre-processing step.

## Type of change


igrigorik:
@yanheChen is this repo the right place for these though? Why can't or shouldn't these live in the SDK?

PR #147, opened 2026-01-30
feat: Add first-class errors concept

richmolj:
It's helpful for Platforms to understand certain error codes and handle these scenarios with first-class treatment. Consider:

* ✅ `last_name_required`: No problem, maps to input field so can be handled with generic form validation
* ✅ `payment_required`: No problem, buyer hasn't reached this step yet
* ✅ `insufficient_stock`: No problem, business can allocate what's possible and give warning message on remaining quantity
* ❌ `OUT_OF_STOCK`: Problem! Platform may want to handle this immediately with a modal (e.g. `create_checkout`), only show in their Cart page UX (item goes out of stock some time after adding to cart), or elsewise provide specific/appropriate UX for this critical scenario.

Without this, Platforms need to understand each merchant's bespoke error code implementation. A small set of codes with first-class treatment is in the best interest of both merchant and buyer.

Given we also want merchant-defined codes, this cannot be an enum and is only defined in documentation.

This is not an exhaustive list, just a basic setup. PR reviewers: please note any other low-hanging fruit.

> Alt: A more generic pattern was considered, combining error messages with additional metadata. This approach seemed error-prone and higher-effort, and is not mutually exclusive with first-class errors if we want to add it down the line.

## Type of change

      or fields**)
- [ ] Documentation update

---

## Checklist

### DIFF
--- a/docs/specification/checkout.md
+++ b/docs/specification/checkout.md
@@ -196,6 +196,26 @@ ELSE IF requires_buyer_review is not empty
   handoff_context = "ready for final review by the buyer"
 ```
 
+#### Standard Errors
+
+Standard errors are standardized error codes that platforms are expected to
+handle with specific, appropriate UX rather than generic error treatment.
+
+| Code                     | Description                                                                |
+| :----------------------- | :------------------------------------------------------------------------- |
+| `out_of_stock`           | Specific item or variant is unavailable                                    |
+| `item_unavailable`       | Item cannot be purchased (e.g. delisted)                                   |
+| `address_undeliverable`  | Cannot deliver to the provided address                                     |
+| `payment_failed`         | Payment processing failed                                                  |
+
+Businesses **SHOULD** mark standard errors with `severity: recoverable` to
+signal that platforms should provide appropriate UX (out-of-stock messaging,
+address validation prompts, payment method changes) rather than generic error
+messages or deferring to checkout completion.
+
+Example: `out_of_stock` requires specific upfront UX, whereas
+`payment_required` can be handled generically at submission.
+
 ## Continue URL
 
 The `continue_url` field enables checkout handoff from platform to business UI,
--- a/source/schemas/shopping/types/error_code.json
+++ b/source/schemas/shopping/types/error_code.json
@@ -1,0 +1,13 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/error_code.json",
+  "title": "Error Code",
+  "description": "Error code identifying the type of error. Standard errors are defined in specification (see examples), and have standardized semantics; freeform codes are permitted.",
+  "type": "string",
+  "examples": [
+    "out_of_stock",
+    "item_unavailable",
+    "address_undeliverable",
+    "payment_failed"
+  ]
+}
--- a/.cspell/custom-words.txt
+++ b/.cspell/custom-words.txt
@@ -49,6 +49,7 @@ checkout
 credentialless
 credentialization
 datamodel
+delisted
 dpan
 ewallet
 fontawesome
--- a/source/schemas/shopping/types/message_error.json
+++ b/source/schemas/shopping/types/message_error.json
@@ -16,8 +16,7 @@
       "description": "Message type discriminator."
     },
     "code": {
-      "type": "string",
-      "description": "Error code. Possible values include: missing, invalid, out_of_stock, payment_declined, requires_sign_in, requires_3ds, requires_identity_linking. Freeform codes also allowed."
+      "$ref": "error_code.json"
     },
     "path": {
       "type": "string",


richmolj:
@raginpirate good call, updated!

yanheChen:
naive qq here because we are talking about the error codes in EP transport as well. How do we envision the name "first party"? 
1) does first party indicate first party code existing in all flows in UCP cross all transport or first party means that this code needs proper UX handling? 
2) for the name of those code, how did you think ENUM comparing with plain text constants?

docs/specification/checkout.md
```diff
@@ -196,6 +196,26 @@ ELSE IF requires_buyer_review is not empty
   handoff_context = "ready for final review by the buyer"
 ```
 
+#### First-Class Errors
```
igrigorik: Why not just call it "Standard errors" instead of defining first-class as standardized errors?

source/schemas/shopping/types/error_code.json
```diff
@@ -0,0 +1,28 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/error_code.json",
+  "title": "Error Code",
+  "description": "Error code identifying the type of error. First-class errors have standardized semantics; freeform codes are also permitted.",
+  "oneOf": [
```
igrigorik: This doesn't work. 

`oneOf` requires exactly one match. "out_of_stock" matches both its `const` AND `type: string` and validation will fail for any const we define here. In theory, `anyOf` walks around that, but it renders all the branches redundant because `type: string` will match everything regardless.

Suggestion: 
```
   "description": "Error code identifying the type of error. First-class errors have standardized semantics; freeform codes are also permitted.",
    "type": "string",
    "examples": [
      "out_of_stock",
      "item_unavailable",
      "address_undeliverable",
      "payment_failed"
    ]
```

`examples` is a no-op at validation time but it's a built-in annotation for tooling that matches content in the spec.

docs/specification/checkout.md
```diff
@@ -196,6 +196,26 @@ ELSE IF requires_buyer_review is not empty
...
+handle with specific, appropriate UX rather than generic error treatment.
+
+| Code                     | Description                                                                |
+| :----------------------- | :------------------------------------------------------------------------- |
+| `out_of_stock`           | Specific item or variant is unavailable                                    |
+| `item_unavailable`       | Item cannot be purchased (e.g. delisted)                                   |
+| `address_undeliverable`  | Cannot deliver to the provided address                                     |
+| `payment_failed`         | Payment processing failed                                                  |
+
+Businesses **SHOULD** mark first-class errors with `severity: recoverable` to
```
igrigorik: Can we generalize this? I suspect we'll end up with a growing list of standard codes here and not all of them will naturally map to default of `recoverable`. Something like...

> Businesses MUST assign an appropriate severity to signal whether the agent should attempt to resolve the issue programmatically or escalate to the buyer.

igrigorik (changes requested):

source/schemas/shopping/types/error_code.json
raginpirate: Ah good catch, yes the intention was to write this as an anyOf!
Sounds like you're just disagreeing with the premise here: https://github.com/Universal-Commerce-Protocol/ucp/pull/147#issuecomment-3825743072. The intention is not to actually improve static typing; a const is obviously a string, its to co-locate context about different error codes in the schema. Examples doesn't do this by default: its unclear if those examples are just random strings which obviously I can use as an error code, or if they have specific meaning for implementation.

We can solve this via comments in this file, or we can solve it via anyOf consts + fallback string, or we can choose to just let external docs deal with informing implementers they must do something special for these codes. IMO I like this solution because it nudges implementers looking at schemas in the right direction, rather than docs they may never read, but comments over examples can too.

docs/specification/checkout.md
richmolj: I was looking for a name that conveyed optimized experiences. But WFM, updated.

source/schemas/shopping/types/error_code.json
richmolj: Agree with Daniel but don't have a strong opinion. Updated to Ilya's approach since this is not a one-way door.

docs/specification/checkout.md
richmolj: I think this defeats the purpose. The reason we have this concept is for Platforms to give it special treatment, but they can't do so if escalating and ceding control to the Business UX. The Platform `SHOULD` be the one handling these.

source/schemas/shopping/types/error_code.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/error_code.json",
+  "title": "Error Code",
+  "description": "Error code identifying the type of error. Standard errors have standardized semantics; freeform codes are also permitted.",
```
igrigorik:
```suggestion
  "description": "Error code identifying the type of error. Standard errors are defined in specification (see examples), and have standardized semantics; freeform codes are permitted.",
```

PR #155, opened 2026-02-04
docs: use versioned schema URLs in docs

lemonmade:
I noticed that many of the docs were still using `2026-01-11` API versions, and almost none of the docs had been updated to use a versioned schema URL. This PR fixes both issues, with the `2026-01-23` API version used consistently. I assume we will want a better, more automated way of doing this as we continue to publish versions, but in the interim, I’d love to have the docs speak a little more consistently, and point folks to permanent URLs (`ucp.dev/*` and `ucp.dev/latest/*` will always work, but their content will be variable over time, and their versions will not match with the hardcoded version strings in examples that reference those URLs).

### DIFF
--- a/docs/documentation/schema-authoring.md
+++ b/docs/documentation/schema-authoring.md
@@ -158,17 +158,17 @@ by `name` rather than arrays of objects with `name` fields.
 ```json
 {
   "capabilities": {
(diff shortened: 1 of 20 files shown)


wry-ry (approved):
👍

Issue #135, opened 2026-01-28
[Spec] Clarify Request-Signature for Checkout REST binding (define signing/verification; RFC 9421 discussion)


ACSchil:
## Description
I’m working on the HTTP bindings for UCP and hit an ambiguity around `Request-Signature` in the Checkout REST spec.

In https://ucp.dev/specification/checkout-rest/ the `Request-Signature` header is **required**, but the documentation doesn’t say how the signature is computed or verified (what bytes are signed, what format/alg, how `kid`/key lookup works, expected failure behavior, etc.). The Shopping REST OpenAPI has the same short description for `Request-Signature` (no normative details): https://ucp.dev/services/shopping/rest.openapi.json

By contrast, the Order capability has a concrete definition using a detached JWT/JWS (RFC 7797) for webhook signing/verification and also uses `Request-Signature`: https://ucp.dev/2026-01-23/specification/order/

It feels like Checkout REST probably intended to reuse that same signing scheme, but it’s not stated.

## Why this matters
Implementers can’t reliably interop without the “how” here. Right now different clients could sign:
- body only vs body + method/path/headers
- different canonicalization / serialization
- different key selection / `kid` conventions
…while still “matching” the current description.

## Proposed doc update (minimum)
If the intent is to match the Order capability approach, I think Checkout REST should explicitly say:
- `Request-Signature` is a detached JWS/JWT per RFC 7797 (same as Order webhook guidance)
- what exactly is signed (at minimum: request body)
- required JOSE header params (`kid`, `alg`, etc.)
- key discovery/verification: verifier matches `kid` to a key in the sender’s published `signing_keys`
- expected error behavior when missing/invalid

## Discussion: should HTTP bindings use RFC 9421 instead?
Separately, I’d like to open discussion on using RFC 9421 HTTP Message Signatures for HTTP bindings.

Reason: detached JWS over only the body doesn’t naturally bind the request to method + target URI + important headers. RFC 9421 is designed to sign those HTTP components directly, which avoids replay/endpoint swap issues unless every implementation reinvents “what else to include” on top of a body signature.

If UCP wanted to go this direction, it likely means adopting:
- `Signature-Input` + `Signature` headers (RFC 9421)
- a UCP profile: required covered components (e.g. `@method`, `@target-uri` or `@path`+query, `content-digest`, and any UCP-specific headers)
- mapping `keyid`/`kid` to published `signing_keys`
[...]

PR #156, opened 2026-02-04
feat: signing for UCP requests & responses

igrigorik:
Context / Closes https://github.com/Universal-Commerce-Protocol/ucp/issues/135. Defines a layered signature architecture:
```
  SHARED FOUNDATION
  ├── Canonicalization: JCS (RFC 8785)
  ├── Algorithms: ES256 (required), ES384, ES512
  ├── Key Format: JWK (RFC 7517)
  ├── Key Discovery: signing_keys[] in /.well-known/ucp
  └── Replay Protection: idempotency-key (business layer)
          │
          ├── REST BINDING (RFC 9421)
          │   Headers: Signature, Signature-Input, UCP-Content-Digest-JCS
          │
          └── MCP BINDING (RFC 7515 Appendix F)
              Fields: meta.signature, meta.idempotency-key, meta.ucp-agent
```

**JCS canonicalization (RFC 8785):**
  - Deterministic JSON serialization before signing
  - Avoids whitespace/key-ordering verification failures
  - Custom header `UCP-Content-Digest-JCS` (not RFC 9530 Content-Digest)
    because we hash canonicalized JSON, not raw bytes

**Transport-specific formats:**
  - REST uses RFC 9421 HTTP Message Signatures (modern standard)
  - MCP uses detached JWS (RFC 7515 Appendix F) in meta.signature
  - Both sign the full message; MCP excludes only meta.signature field

**Changes to OpenAPI:**
  - `Request-Signature` header → `Signature` + `Signature-Input` headers
  - `X-Detached-JWT` response header → `Signature` + `Signature-Input`
  - `UCP-Content-Digest-JCS` header for body digest

**Other updates:**
  - checkout-rest.md, checkout-mcp.md: Added Message Signing sections
  - order.md: Rewrote webhook signing to use RFC 9421
  - ap2-mandates.md: Now references signatures.md for crypto primitives
  - openrpc.json: Added signature field to meta, added meta to cart methods
  - mkdocs.yml: Added signatures.md to nav and llmstxt sections

---

## Checklist
- [x] **Breaking change**

### DIFF
--- a/main.py
+++ b/main.py
@@ -132,7 +132,7 @@ def _resolve_schema(
       data = json.loads(result.stdout)
       _resolved_schema_cache[cache_key] = data
       return data
-  except (subprocess.SubprocessError, json.JSONDecodeError, FileNotFoundError):
+  except subprocess.SubprocessError:
     pass
   return None
 


docs/specification/ap2-mandates.md
```diff
@@ -123,15 +123,16 @@ If a public key cannot be resolved, or if the signature is invalid, the business
 
 ## Cryptographic Requirements
 
-### Signature Algorithm
+This extension uses the cryptographic primitives defined in the
+[Message Signatures](signatures.md) specification:
 
-All signatures **MUST** use one of the following algorithms:
+* **Algorithms:** ES256 (required), ES384, ES512
```
drewolson-google, line 129: What does the `(required)` parenthetical mean here? Should this read `recommended`?

docs/specification/signatures.md
```diff
@@ -0,0 +1,977 @@
...
+{"checkout":{"buyer":{"email":"alice@example.com"},"line_items":[{"id":"prod_123","price":1000,"quantity":2}]}}
+```
+
+### Signature Algorithms
+
+All signatures **MUST** use one of the following algorithms:
+
+| Algorithm | Curve   | Hash    | Required | Description                    |
+| :-------- | :------ | :------ | :------- | :----------------------------- |
+| `ES256`   | P-256   | SHA-256 | Yes      | ECDSA (**RECOMMENDED**)        |
```
drewolson-google: Again, we say `recommended` here but `MUST` below. Which is it?

docs/specification/signatures.md
```diff
@@ -0,0 +1,977 @@
...
+1. **Add new key** — Publish new key in `signing_keys[]` alongside existing keys
+2. **Start signing** — Begin signing with the new key
+3. **Grace period** — Continue accepting signatures from old keys (minimum 7 days)
+4. **Remove old key** — Remove the old key from `signing_keys[]`
+
+**Recommendations:**
+
+* Rotate keys every 90 days
+* Support multiple active keys during transitions
+* Signers: use newest key
```
drewolson-google: What does "newest" mean in this context? First in the list? There's no notion of age in the key specification.

drewolson-google (approved):
Generally looks very good to me. I've included a few small comments.

docs/specification/signatures.md
igrigorik: You're right, and I don't think it's necessary -- will remove.

docs/specification/ap2-mandates.md
igrigorik: Yep, updated 👍

docs/specification/ap2-mandates.md
```diff
@@ -213,17 +214,14 @@ selective disclosure, key binding) is defined by the
 
 ### Canonicalization
 
-For signature computation over JSON payloads, implementations **MUST** use
-**JSON Canonicalization Scheme (JCS)** as defined in
-[RFC 8785](https://datatracker.ietf.org/doc/html/rfc8785).
+All JSON payloads **MUST** be canonicalized using **JSON Canonicalization
```
GarethCOliver, line 218: AIUI canonicalized JSON is generally considered much more problematic than signing the body bytes of the serialized JSON. I would highly recommend we avoid doing so.

igrigorik:
A couple of key takeaways from today's working group call...

### 1. Do we really need request signatures?
Yes and no.

 - No when alternative auth mechanisms exist: API keys, OAuth client credentials, mTLS. If you've already established trust out-of-band, signatures are redundant for authentication.
 - Yes for the distributed web use case. A self-hosted site can't negotiate API keys with every agent. Signatures with advertised public keys solve the one-to-many scaling problem — merchants can declaratively say "I trust agent A & B" without any out-of-band coordination.

Recommendation: "Businesses SHOULD authenticate agents". The spec should allow multiple mechanisms (API keys, signatures, MTLS) without mandating a single approach. Signatures become valuable when you want permissionless agent onboarding, and we should spec this path.

### 2. If agent auth is the primary goal, can we simplify what we're signing?
Likely, yes. Current proposal signs the entire request body with JCS canonicalization, which is complex and might be unnecessary — see 3. We could, potentially, skip signing body or consider `Content-Digest`: sha-256 of the payload bytes, which is a lot simpler. For identity we could lean on RFC 9421 request binding (method, path, idempotency-key, ucp-agent, content-digest?) which would cover request payload & agent attestation.

As a next step, we'll tap security folks to review and help converge on a recommendation.

### 3. Bonus: can we unify on RFC 9421?
Likely, yes. UCP's MCP transport runs over HTTP, which means we can rely on 9421: single signature mechanism, CDN/edge can validate, no JCS needed. This would be a significant simplification.

docs/specification/signatures.md
```diff
@@ -0,0 +1,678 @@
...
+
+# Message Signatures
+
+This specification defines how UCP messages are cryptographically signed to
+ensure authenticity and integrity.
+
+## Overview
+
+Businesses **SHOULD** authenticate agents to prevent impersonation and ensure
+message integrity. UCP supports multiple authentication mechanisms:
```
maximenajim: The doc introduces “multiple authentication mechanisms” (API keys, OAuth, mTLS, HTTP Message Signatures) but doesn’t specify how identity is bound to UCP-Agent for non-signature mechanisms, or what to do when multiple mechanisms are present. An implementation could authenticate via API key but accept a conflicting UCP-Agent (or vice-versa), leading to mis-attribution and authorization bugs.

Possibly add the following:

- If HTTP Message Signatures are present, verifiers MUST verify them (no silent ignore).
- If API key / OAuth / mTLS is used, the server MUST derive the caller identity from that mechanism and either ignore UCP-Agent or require it to match the authenticated principal.
- Define explicit precedence rules and reject “mix-and-match” identities.

docs/specification/signatures.md
```diff
@@ -0,0 +1,678 @@
...
+
+### Signature Algorithms
+
+UCP supports ECDSA signatures with the following algorithms:
+
+| Algorithm | Curve   | Hash    |
+| :-------- | :------ | :------ |
+| `ES256`   | P-256   | SHA-256 |
+| `ES384`   | P-384   | SHA-384 |
+| `ES512`   | P-521   | SHA-512 |
```
maximenajim: Nit: The algorithm guidance is written in JWA/JWS terms (ES256, ES384, ES512) while this document is now centered on RFC 9421. In RFC 9421, the alg signature parameter uses the HTTP Signature Algorithms registry (e.g., ecdsa-p256-sha256, ecdsa-p384-sha384) and does not include P‑521/ES512 as a standard algorithm identifier.

docs/specification/signatures.md
```diff
@@ -0,0 +1,678 @@
...
+**Intermediary Warning:** Proxies, API gateways, and other intermediaries
+**MUST NOT** re-serialize JSON bodies, as this would invalidate the signature.
+The `Content-Digest` is computed over raw bytes; any modification breaks
+verification.
+
+### REST Request Signing
+
+**Signed Components:**
+
+| Component         | Required | Description                             |
```
maximenajim: The required covered components omit @authority. A valid signed request can potentially be replayed to a different host that also trusts the same agent key, because the signature doesn’t bind to the intended origin.

docs/specification/signatures.md
```diff
@@ -0,0 +1,678 @@
...
+Host: merchant.example.com
+Signature-Input: sig1=("@method" "@path");keyid="platform-2026"
+Signature: sig1=:MEQCIBx7kL9nM2oP5qR8sT1uV4wX6yZaB3cD...:
+```
+
+### REST Response Signing
+
+Response signatures use `@status` instead of `@method`:
+
+**Signed Components:**
```
maximenajim, line 284: The spec needs to be consistent about whether UCP-Agent is covered. The checkout MCP example uses "ucp-agent" as a covered component, but the “Signed Components” table in this doc doesn’t list it.
We should Add ucp-agent to the covered component set for requests (and update verification pseudo-code accordingly).

docs/specification/signatures.md
```diff
@@ -0,0 +1,678 @@
...
+            "content-digest": digest_header,
+            "content-type": "application/json"
+        },
+        keyid=kid
+    )
+
+    // 4. Sign
+    signature = ecdsa_sign(signature_base, private_key)
+
+    // 5. Return headers
```
maximenajim, line 240: The spec doesn’t clearly state the signature byte encoding for ECDSA. Many crypto libraries output ASN.1/DER ECDSA signatures by default; RFC 9421’s ECDSA algorithms use a fixed-width raw r||s encoding (64 bytes for P‑256, 96 bytes for P‑384).

docs/specification/signatures.md
```diff
@@ -0,0 +1,678 @@
...
+
+```text
+verify_rest_request(request):
+    // 1. Parse Signature-Input
+    sig_input = parse_signature_input(request.headers["Signature-Input"])
+    keyid = sig_input.keyid
+    components = sig_input.components
+
+    // 2. Fetch signer's public key
+    profile_url = get_profile_url_from_ucp_agent(request.headers["UCP-Agent"])
```
maximenajim, line 381: Profile fetching based on UCP-Agent is a major SSRF + DoS boundary, especially given the “permissionless onboarding” positioning.  Maybe add explicit verifier requirements:

- Verifiers MUST NOT fetch arbitrary profiles unless the profile is in a configured trust policy (allowlist, registry, onboarding record, etc.).
- If “permissionless” is a goal, define additional SSRF mitigations: block private/loopback/link-local IPs, limit redirects strictly, enforce DNS rebinding defenses, set aggressive timeouts, and cap response size.
- Add rate limiting/backoff on “unknown kid / profile” failures to prevent request-amplification attacks.

docs/specification/signatures.md
```diff
@@ -0,0 +1,678 @@
...
+
+### When Signatures Are Recommended
+
+**Requests:** Platforms **SHOULD** sign all requests when using HTTP Message
+Signatures. Alternative authentication mechanisms (API keys, OAuth, mTLS) may
+be used instead.
+
+**Responses:** Signatures are **RECOMMENDED** for:
+
+* Order webhook notifications
```
maximenajim: This section says signatures are “RECOMMENDED” for order webhooks / payment auth / checkout completion responses, but other docs in this PR still use MUST language for webhook signing. We need one consistent normative stance because this materially affects security posture.

Either:
- Make webhooks MUST be signed (and keep “recommended” for other responses), or
- Downgrade webhook doc language and explicitly state the risk/tradeoff.

For merchant safety, I strongly recommend webhooks remain MUST signed.

docs/specification/checkout-rest.md
```diff
@@ -1282,6 +1282,67 @@ with HTTP 200 and the UCP envelope containing `messages`:
 }
 ```
 
+## Message Signing
+
+Platforms **SHOULD** authenticate agents when using REST transport. When using
```
maximenajim: The text says “Platforms using HTTP Message Signatures SHOULD sign requests…”. Once a mechanism is chosen, signing needs to be MUST, otherwise you end up with partial/accidental unauthenticated traffic that “kind of works” in dev and fails open in prod.

Rephrase to:
“Businesses MAY choose among auth mechanisms. If HTTP Message Signatures are used/required for an operation, the request MUST include valid Signature-Input + Signature (and Content-Digest when a body is present).”

source/services/shopping/rest.openapi.json
```diff
@@ -604,14 +679,32 @@
...
         "schema": {
           "type": "string"
         },
-        "description": "Ensure the authenticity and integrity of an HTTP message."
+        "description": "RFC 9421 HTTP Message Signature. Contains the signature value in the format `sig1=:<base64-signature>:`."
+      },
+      "signature_input": {
+        "name": "Signature-Input",
+        "in": "header",
+        "required": true,
```
maximenajim: Signature and Signature-Input are still modeled as "required": true parameters, but the docs now treat HTTP Message Signatures as one option among several auth mechanisms. This is a schema-level contradiction that will force clients to send signature headers even when using OAuth/mTLS/API keys.

docs/specification/signatures.md
igrigorik: Good flag. 

UCP derives the algorithm entirely from the key's `crv` field in the JWK and does not include `alg` in `Signature-Input` parameters. This is permitted by RFC 9421 (algorithm determined from key material). Since crv uniquely determines the ECDSA algorithm for each curve, there's no ambiguity and implementors only deal with one naming system (JWA, in the JWK).

I've updated the table, which now leads with the curve and shows the corresponding JWK alg as a reference column — making it clear these are JWK identifiers, not RFC 9421 Signature-Input values.  Also dropped ES512/P-521 — not in the IANA HTTP Signature Algorithms registry.

docs/specification/signatures.md
igrigorik: Ah, great catch! Agree this is a MUST and contradicts other places where we already state that. Updated.

source/services/shopping/rest.openapi.json
igrigorik: Good catch, again! Updated to required: false with instructions in the description.

docs/specification/signatures.md
igrigorik: Fair. Added normative section for this.

docs/specification/signatures.md
igrigorik: You're pointing to a much larger gap that's been on my TODO list. This should do it: https://github.com/Universal-Commerce-Protocol/ucp/pull/156/changes/84f00eb0d84cd8214713bb2b2ebf34d72184bc2c

igrigorik:
@maximenajim ty sir, sharp eyes! I believe all the flags should be addressed, please take another pass. 🙇🏻

maximenajim (approved):

docs/specification/overview.md
```diff
@@ -835,6 +836,144 @@ root capability.
...
+
+#### Hosting
+
+Both profiles must be reliably hosted. An unreliable or misconfigured
+profile endpoint may prevent the other party from processing requests.
+
+1. Profiles **MUST** be served over HTTPS.
+2. Profile endpoints **MUST NOT** use redirects (3xx).
+3. Profile responses **MUST** include a `Cache-Control` header with
+   `public` and `max-age` of at least 60 seconds. Profiles **MUST NOT**
```
manumilou, line 918: How should we handle the profile discovery when we get `no-store` or `no-cache` directives ? Reject, default to min TTL ?

Issue #170, opened 2026-02-11
[Payment] Allow merchant to specify preferred ordering of payment


prasad-stripe:
## Summary

Today the merchant advertises `payment.handlers` (in the business profile and/or in the checkout response). The ordering of payment methods on the platform or agent side is unspecified: agents may treat array index as implicit ordering or use ordering in an unspecified way. This proposal adds a way for the merchant to specify a preferred ordering in the spec. The ordering is **suggestive** (not deterministic): the agent/platform may still override it, but the spec should allow the merchant to express preference at least.

## Motivation

- Merchants have a preferred order for payment methods (e.g. “show our wallet first”, “prioritize BNPL in certain contexts”). They are best placed to suggest this ordering because they know what works for their **fraud vs. conversion trade-off**—which methods they trust, which convert best for their audience, and how they want to balance risk and revenue.
- Without a specified mechanism, every platform/agent invents its own interpretation (e.g. array order, alphabetical, or arbitrary), leading to inconsistent UX and no way for the merchant to communicate intent.
- Specifying a simple, suggestive ordering gives merchants a standard way to express preference while preserving platform/agent flexibility.

### Goals

- Define in the spec how a merchant can indicate preferred order of payment handlers (e.g. in the `handlers` array or via an explicit field).
- Keep the semantics **suggestive**: platforms/agents MAY reorder or override; the spec does not require strict adherence.
- Avoid breaking existing implementations: ordering can be optional (e.g. “array order is the suggested order” by default or an optional `display_order` / `priority` field).

### Non-Goals

- Mandating that platforms/agents always follow the merchant’s order.
- Defining UI/UX rules (e.g. how to display); only the data contract for expressing order.

## Detailed Design

**Option A — Explicit ordering field**  
- Add an optional field to each handler entry (e.g. `display_order: number` or `priority: number`) so that ordering is explicit even when serialization order differs. Again, semantics are suggestive; platform/agent may override.

**Option B — Single ordered list**  
- Add an optional top-level field (e.g. `handler_order: ["handler_id_1", "handler_id_2", ...]`) that lists handler IDs in preferred order. Handlers not listed can be shown in any order after the listed ones.

Recommendation: Option A is the smallest change (clarify existing array order). Option B can be considered if the working group prefers explicit ordering specified outside of the core payment handlers.

**Relevant spec surface:**  
- Checkout response: `payment.handlers` in the create-checkout (or equivalent) response 

## Risks and Mitigations

[...]

PR #176, opened 2026-02-13
feat: add optional display_order to payment handlers

prasad-stripe:
# Payment handler display order (merchant-suggested ordering)

## Description

Please include a summary of the changes and the related issue. Please also
include relevant motivation and context.

Adds an optional **`display_order`** field to each payment handler in
`payment_handlers`. Businesses (merchants) can set an integer per handler
(lower = higher preference) to suggest how payment methods should be ordered
when presented to the user. The ordering is **suggestive only**: platforms and
agents MAY reorder (e.g. for user preference, localization, or A/B tests). Addresses #170 

This gives merchants a standard way to express their preferred order—e.g. for
fraud vs. conversion trade-offs—without requiring agents to follow it strictly.
Aligns with the same capability in the Agentic Commerce Protocol (ACP).

## Type of change

      or fields**)
- [ ] Documentation update

---

N/A — optional additive field only.

---

## Motivation and context

Today the ordering of payment methods on the platform or agent side is
unspecified. Merchants know what works best for their fraud/conversion
trade-off and should be able to suggest an order at the spec level.

Fixes/Addresses: #[issue number] _(fill in the UCP issue number you created)_

---

## Example

**Handler with `display_order`:**

```json
{
  "id": "processor_tokenizer_1234",
  "version": "2026-01-11",
  "spec": "https://example.com/ucp/handler",
  "schema": "https://example.com/ucp/handler/schema.json",
  "config": { "environment": "production", "business_id": "business_xyz_789" },
  "display_order": 0
}
```

Lower `display_order` = higher preference (e.g. show first). Omission means no
suggestion.

---

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code

---

### Pull Request Title

Use Conventional Commits, e.g.:

**`feat: add optional display_order to payment handlers`**

---

## Files changed

- **Schema:** `source/schemas/payment_handler.json` — added optional
  `display_order` (integer) to `base` definition.
- **Docs:** `docs/specification/payment-handler-guide.md` — added
  `display_order` to Business Schema example and a short description of the
  field.

---

## Additional notes

- **Backward compatible:** Optional field; existing implementations can ignore
  unknown properties. No breaking changes.
- If this project regenerates TypeScript types from the JSON schemas (e.g. via
  a script that reads `source/schemas`), re-run the generator so
  `display_order` appears in the generated types.
- Same capability is implemented in the Agentic Commerce Protocol (ACP); this
  keeps UCP and ACP aligned for merchant-suggested payment handler ordering.

### DIFF
--- a/docs/specification/payment-handler-guide.md
+++ b/docs/specification/payment-handler-guide.md
@@ -314,6 +314,69 @@ authoritative value returned in the `response_schema`.
 In this example, the business's PSP is not configured for Discover, so Discover
 is excluded from the response even though the platform supports it.
 
+#### Display Ordering
+
+Businesses **MAY** include a `display_order` integer on each entry in
+`available_instruments` to suggest how payment options should be presented to
+the buyer. Lower values indicate higher preference (e.g., `0` = show first).
+
+Ordering is relative across **all** `available_instruments` entries across
+**all** handlers in the response. When the same instrument type appears in
+multiple handlers (e.g., `"card"` via both a wallet and a direct tokenizer),
+each entry carries its own `display_order` — this allows the business to rank
+distinct payment paths independently (e.g., "card via Google Pay" before "card
+via direct entry").
+
+**Semantics:**
+
+- `display_order` is **suggestive only** — platforms **MAY** override based on
+    user preference, localization, A/B tests, or other factors.
+- Entries without `display_order` express no preference and **SHOULD** be
+    placed after entries that declare one.
+- When multiple entries share the same `display_order` value, the platform
+    determines their relative order.
+- `display_order` participates in the `available_instruments` resolution flow:
+    the business includes it in the resolved `response_schema`, and platforms
+    **MUST** treat the response values as authoritative.
+
+**Example:**
+
+A business advertising three payment paths with suggested ordering:
+
+```json
+{
+  "ucp": {
+    "payment_handlers": {
+      "com.google.pay": [
+        {
+          "id": "gpay_handler",
+          "version": "{{ ucp_version }}",
+          "available_instruments": [
+            {"type": "card", "display_order": 0}
+          ],
+          "config": { "..." : "..." }
+        }
+      ],
+      "com.example.tokenizer": [
+        {
+          "id": "tokenizer_handler",
+          "version": "{{ ucp_version }}",
+          "available_instruments": [
+            {"type": "card", "constraints": {"brands": ["visa", "mastercard"]}, "display_order": 1},
+            {"type": "bank", "display_order": 2}
+          ],
+          "config": { "..." : "..." }
+        }
+      ]
+    }
+  }
+}
+```
+
+In this example the business suggests: Google Pay card first, then direct card
+entry, then bank transfer. The platform may follow this suggestion or adjust
+based on buyer context.
+
 ---
 
 #### Defining the Schema
--- a/source/schemas/shopping/types/available_payment_instrument.json
+++ b/source/schemas/shopping/types/available_payment_instrument.json
@@ -15,6 +15,11 @@
       "additionalProperties": true,
       "description": "Constraints on this instrument type. Structure depends on instrument type and active capabilities.",
       "minProperties": 1
+    },
+    "display_order": {
+      "type": "integer",
+      "description": "Business-suggested display priority for this instrument across all handlers. Lower values indicate higher preference. Ordering is suggestive; platforms MAY override based on user preference, localization, or other factors.",
+      "minimum": 0
     }
   }
 }


docs/specification/payment-handler-guide.md
```diff
@@ -314,6 +314,69 @@ authoritative value returned in the `response_schema`.
...
+distinct payment paths independently (e.g., "card via Google Pay" before "card
+via direct entry").
+
+**Semantics:**
+
+- `display_order` is **suggestive only** — platforms **MAY** override based on
+    user preference, localization, A/B tests, or other factors.
+- Entries without `display_order` express no preference and **SHOULD** be
+    placed after entries that declare one.
+- When multiple entries share the same `display_order` value, the platform
```
kmcduffie, line 336: An alternative might be that the spec does not support duplicate values in a given payload. 

I'd like to better understand the use case you're thinking of that's driving this modeling where there could be collisions?

docs/specification/payment-handler-guide.md
```diff
@@ -314,6 +314,69 @@ authoritative value returned in the `response_schema`.
...
+    determines their relative order.
+- `display_order` participates in the `available_instruments` resolution flow:
```
kmcduffie, line 338: This field shares the same name as the first field in this comment. Did you mean for this to be different than the `display_order` above?

source/schemas/shopping/types/available_payment_instrument.json
```diff
@@ -15,6 +15,11 @@
       "additionalProperties": true,
       "description": "Constraints on this instrument type. Structure depends on instrument type and active capabilities.",
       "minProperties": 1
+    },
+    "display_order": {
+      "type": "integer",
+      "description": "Business-suggested display priority for this instrument across all handlers. Lower values indicate higher preference. Ordering is suggestive; platforms MAY override based on user preference, localization, or other factors.",
```
kmcduffie, line 21: I don't believe a handler will know of any other handlers. Can the `display_order` be across all handlers, or is more signaling the priority within the instruments returned by this handler?

kmcduffie (commented):
@prasad-stripe Thanks for the PR!  A few questions in line.

raginpirate:
fwiw this PR is great @prasad-stripe, I got to catch @igrigorik async for a moment on this and I think there might be a higher level abstraction here for how we generally offer ordering hints across UCP for arrays... we're thinking about this and will circle back, if you don't also come up with something 😄

TLDR; seems like this isn't just a payments concern!

archrao:
Adding some thoughts for discussion to provide ordering hints for arrays within UCP objects (like available_instruments, line_items, totals, etc.)

Option 1: Sibling _hints Object
This approach adds a new JSON object alongside the array, named by convention as [array_field_name]_hints.
 "available_instruments": [
    { "id": "instr_card", "type": "card", "description": "Major credit/debit cards" },
    { "id": "instr_paypal", "type": "ewallet", "description": "Pay with PayPal" },
    { "id": "instr_bnpl_klarna", "type": "bnpl", "description": "Pay over time with Klarna" }
  ],
  "available_instruments_hints": {
    "item_ids": ["instr_card", "instr_paypal", "instr_bnpl_klarna"],
    "order": [
      "instr_card",
      "instr_paypal",
      "instr_bnpl_klarna"
    ]}
This option is clean but could increase payload.

Option 2: In-Item _display Hints
This approach adds a reserved _display object within each item of the array that needs hints. 
 "available_instruments": [
    {
      "id": "instr_card",
      "type": "card",
      "description": "Major credit/debit cards",
      "_display": {
             "order": 1
      }
    }]
This option is more payload efficient and hints are co-located with the item.

source/schemas/shopping/types/available_payment_instrument.json
gsmith85, line 21: The resolution flow bullet says platforms MUST treat response values as authoritative, but the semantics bullet says platforms MAY override. If the intent is that platforms should follow the suggested ordering by default unless they have a specific reason to deviate (user preference, localization, etc.), something like "SHOULD follow unless overriding for contextual reasons" would help with clarity.

source/schemas/shopping/types/available_payment_instrument.json
```diff
@@ -15,6 +15,11 @@
       "additionalProperties": true,
       "description": "Constraints on this instrument type. Structure depends on instrument type and active capabilities.",
       "minProperties": 1
+    },
+    "display_order": {
```
gsmith85, line 19: `available_payment_instrument` is used in [payment_handler.base](https://github.com/Universal-Commerce-Protocol/ucp/blob/aa84ce15b5cc575e5c123651d2d62513cd1d2c9a/source/schemas/payment_handler.json#L8), which is inherited by all three handler variants: p`latform_schema`, `business_schema`, and r`esponse_schema`. That means `display_order `is now valid in a business' static `/.well-known/ucp profile`, not just in a checkout response.

In a profile context, the semantics are undefined. There's no buyer, no session, and no other handlers to rank against. If a merchant `publishes display_order: 0` on their GPay entry in `/.well-known/ucp`, should a platform treat it as a default preference or ignore it?

I suspect the goal is to only affect only the `response_schema`.  In that case, in [payment_handler.json](https://github.com/Universal-Commerce-Protocol/ucp/blob/aa84ce15b5cc575e5c123651d2d62513cd1d2c9a/source/schemas/payment_handler.json#L44), override `available_instruments` with an inline `allOf` that extends `available_payment_instrument` with `display_order`, leaving base clean. The field can then only appear where it has defined meaning.

igrigorik:
Closing, continued in https://github.com/Universal-Commerce-Protocol/ucp/issues/525.

PR #187, opened 2026-02-17
feat: add available_instruments to payment handler configurations

raginpirate:
# Description

Introduce available_instruments as an optional field on payment handler configurations. This allows handlers to declare which instrument types they support along with type-specific constraints (e.g., card brand restrictions like ["visa", "mastercard"]). When absent, all instruments are considered available.

This functionality is possible today through the config payload blob; this PR is making available instruments and constraints a first-class schema and removing the divergence that will emerge as every payment handler redefines the same concepts.

As well, this is a vital introduction to UCP to support future payments extensions like payment terms or purchase options. In order to only offer instalments on credit cards, UCP must have a base structure to expose that context deterministically, instead of in json blobs as it does today, such that we can target the available instrument for a given payment term id.

This change adds:
- A new available_payment_instrument.json base schema and utilizes the new pattern in card_payment_instrument.json
- The available_instruments array to the payment_handler.json schema
- Constraint pattern documentation in the schema authoring guide and payment handler guide
- Updated examples across all specification docs to demonstrate the new field

## Type of change
- [x] Documentation update

---

No, this is an optional new field.

---

## Checklist

---

### DIFF
--- a/docs/specification/payment-handler-guide.md
+++ b/docs/specification/payment-handler-guide.md
@@ -191,10 +191,11 @@ and desired configuration.
 }
 ```
 
-**`available_instruments`** is optional. When absent, it means **everything is
-available without constraints** (the handler supports all instrument types and
-constraints that it is capable of). When present, it restricts which instrument
-types and constraints the handler advertises for this declaration.
+**`available_instruments`** is optional. When absent, the handler places no
+restrictions on instrument types or constraints — it supports the full set of
+instrument types defined by its handler schema. When present, it narrows the
+advertised types and/or applies additional constraints (e.g., limiting card
+brands to `["visa", "mastercard"]`).
 
 ---
 
@@ -208,7 +209,7 @@ and typically includes different configuration:
 | :------ | :------ | :------ |
 | **business_schema** | Business discovery (`/.well-known/ucp`) | Declares the business identity and how they're configured for this handler. Contains merchant-specific settings. |
 | **platform_schema** | Platform profile (advertised URI) | Declares the platform identity and how it supports this handler. Includes `spec` and `schema` URLs for implementers. |
-| **response_schema** | Checkout/Order API responses | **Runtime configuration** with merged context: merchant identity, available payment methods, tokenization specs, and other state needed to process the transaction. Often the richest of the three. |
+| **response_schema** | Checkout/Order API responses | **Runtime configuration** with resolved context: merchant identity, resolved `available_instruments` for this checkout, tokenization specs, and other state needed to process the transaction. Platforms **MUST** treat this as authoritative. |
 
 **Business Schema Example** (business declares handler configuration):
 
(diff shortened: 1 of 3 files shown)


source/schemas/shopping/types/card_payment_instrument.json
```diff
@@ -3,6 +3,33 @@
...
+            "type": {
+              "const": "card"
+            },
+            "constraints": {
+              "type": "object",
+              "properties": {
+                "brands": {
+                  "type": "array",
+                  "items": { "type": "string" },
+                  "description": "Limit to specific card brands (e.g., ['visa', 'mastercard', 'amex'])."
```
gsmith85, line 26: Consider `"pattern": "^[a-z0-9-]+$",` to provide canonicalization and security by enforcing a lowercase, URL-safe format that prevents injection, avoids case-sensitivity collisions, and ensures predictable data for downstream processing.

source/schemas/shopping/types/card_payment_instrument.json
```diff
@@ -3,6 +3,33 @@
...
+        {
+          "type": "object",
+          "properties": {
+            "type": {
+              "const": "card"
+            },
+            "constraints": {
+              "type": "object",
+              "properties": {
+                "brands": {
```
gsmith85, line 21: Consider:
- `"uniqueItems": true`: Ensures brands behaves as a logical set.
- `"minItems": 1`: Prevents ambiguous semantics; if constraints exist, at least one requirement must be specified to avoid "empty filter" confusion.

source/schemas/shopping/types/card_payment_instrument.json
gsmith85: If you're amenable to this change, what that likely implies is the need to consolidate [brand](https://github.com/Universal-Commerce-Protocol/ucp/blob/92d601bddfffd52838c408573a60e8aa15e06185/source/schemas/shopping/types/card_payment_instrument.json#L23) into a global type so its constraints are defined once.

source/schemas/shopping/types/available_payment_instrument.json
```diff
@@ -0,0 +1,19 @@
...
+  "title": "Available Payment Instrument",
+  "description": "An instrument type available from a payment handler with optional constraints.",
+  "type": "object",
+  "required": ["type"],
+  "properties": {
+    "type": {
+      "type": "string",
+      "description": "The instrument type identifier (e.g., 'card', 'gift_card'). References an instrument schema's type constant."
+    },
+    "constraints": {
```
gsmith85, line 13: Does an empty constraints object mean "no restrictions" or "incomplete data"?  If the `constraints` is specified, it _probably_ should contain at least one piece of data, you could consider adding `"minProperties": 1` to be more explicit.

source/schemas/payment_handler.json
```diff
@@ -11,6 +11,16 @@
...
           "type": "object",
           "required": ["id"]
+        },
+        {
+          "type": "object",
+          "properties": {
+            "available_instruments": {
+              "type": "array",
+              "items": { "$ref": "shopping/types/available_payment_instrument.json" },
+              "description": "Instrument types this handler supports, with optional constraints. When absent, every instrument should be considered available."
```
gsmith85: Nit: When absent or empty, every instrument should be considered available.

source/schemas/shopping/types/card_payment_instrument.json
raginpirate: Good questions, I asked myself the same thing while writing this.

To start with the larger point, I'm concerned trying to centralize a brand list in UCP. This might sign us up for countless PRs appearing over time extending this list as it continues to grow, pushing small brands to have to directly merge something into UCP in order to get adoption, which feels backwards compared to how we've looked at payments and payment handlers in UCP.

For now, lets go with a non-controversial string, and if we feel this is very painful in the future we can change course and come up with a centralized list.

Now, extending from this decision point, I also have reservations about using that exact regex pattern:
- We end up with divergence between representations like "diners" and "diners-club" and "dinersclub" anyways. Limiting to kebab-case strings helps reduce the amount of variance here, but it doesn't really solve the problem; only a centralized brand list does.
- kebab-case is an opinion that starts to cause visible convention friction with any other payment handler written constraint or config variables currently accepted. Example: the Google Pay handler would have it's card brand constraint look... disconnected ("type": "PAYMENT_GATEWAY" ... "brand": "diners-club").
- security is not a concern here; injection MUST be handled out-of-band by all adopters of UCP, we have many open, non-patterned strings in the protocol.

With all of that said, I would love for there to be a standard, but let me first gather thoughts from the rest of the TC. If anything, I will propose we move forward with snake_case since that matches other existing examples in UCP and is our default for the schema shape. But I still have reservations that everyone in the room would agree.

source/schemas/payment_handler.json
raginpirate: I set a minimumItems on this as well, so we don't have to deal with empty here 😄

raginpirate:
Thank you very much for the schema feedback! Addressed all of those, and only held-out on the pattern match on brand for now. Will cycle back here tomorrow after validating with others!

source/schemas/shopping/types/card_payment_instrument.json
drewolson-google: While I agree with the general idea of canonicalizing, I worry about the complexity of even a "simple" canonicalization. For example, we'd likely need to support all of UTF8, taking into account graphemes, etc.

drewolson-google (approved):

source/schemas/shopping/types/card_payment_instrument.json
raginpirate: 👍 To Drew's statement: we discussed this in the TC today and aligned that although there are good ideas here, any form of pattern-matching done across UCP is likely going to be fraught with errors over time. 
A card brand with a $ in it was a good example.

Since we see this pattern matching does not directly resolve the deduplication problem, we will move this PR forward without this change, and we can always reconsider in isolation of this PR.

docs/specification/payment-handler-guide.md
```diff
@@ -188,6 +191,11 @@ and desired configuration.
 }
 ```
 
+**`available_instruments`** is optional. When absent, it means **everything is
```
igrigorik: What does "everything" mean here? I think this is trying to say...

> available_instruments is optional. When absent, the handler places no restrictions on instrument types or constraints — it supports the full set of instrument types defined by its handler schema. When present, it narrows the advertised types and/or applies additional constraints (e.g., limiting card brands to ["visa", "mastercard"]).

~Ish?

PR #195, opened 2026-02-20
feat: get product operation for catalog.lookup

igrigorik:
`search_catalog` and `lookup_catalog` are discovery operations — they return multiple products with featured variant(s). But once a user picks a product, the agent needs a different interaction: full product detail with all options, real-time availability as the user selects options (Color, Size), and the exact variant to purchase. This is the product detail page (PDP) flow, which is best modelled as a distinct operation — this pattern conforms to common API shapes in the wild.

`get_product` is a single-resource operation (part of `dev.ucp.shopping.catalog.lookup`) for servicing purchase flow decisions. It returns one product with a relevant subset of variants, option-level availability signals, and support for interactive variant narrowing.

**REST:** `POST /catalog/product`
**MCP:** `get_product` tool

### Example request
```json
{
  "id": "prod_abc123",
  "selected": [
    { "name": "Color", "label": "Blue" }
  ],
  "preferences": ["Color", "Size"],
  "context": { "country": "US" }
}
```

Only `id` is required. `selected` and `preferences` are for interactive narrowing.

### Example (redacted) response

```json
{
  "product": {
    "id": "prod_abc123",
    "title": "Runner Pro",
    "price_range": {
      "min": { "amount": 12000, "currency": "USD" },
      "max": { "amount": 15000, "currency": "USD" }
    },
    "options": [
      {
        "name": "Color",
        "values": [
          { "label": "Blue",  "available": true,  "exists": true },
          { "label": "Green", "available": false, "exists": true }
        ]
      },
      {
        "name": "Size",
        "values": [
          { "label": "10", "available": true,  "exists": true },
          { "label": "11", "available": false, "exists": false }
        ]
      }
    ],
    "selected": [{ "name": "Color", "label": "Blue" }],
    "variants": [
      {
        "id": "var_abc123_blue_10",
        "sku": "RP-BLU-10",
        "title": "Blue / Size 10",
        "price": { "amount": 12000, "currency": "USD" },
        "availability": { "available": true },
        "options": [
          { "name": "Color", "label": "Blue" },
          { "name": "Size", "label": "10" }
        ],
        "media": [{ "type": "image", "url": "https://cdn.example.com/runner-pro-blue.jpg" }]
      }
    ]
  }
}
```

## Iterative Flow

1. **Agent calls `get_product(id: "prod_abc123")`** — no selections. Server returns the product with featured variant and option map.
1. **User picks Color=Blue.** Agent calls `get_product(id: "prod_abc123", selected: [{name: "Color", label: "Blue"}])`. Response narrows: `product.selected` confirms Blue, variants are all Blue, availability on Size values updates to reflect Blue inventory.
1. **User picks Size=10.** Agent adds to selections. Response returns the exact Blue/10 variant — price, SKU, availability — ready for checkout.
1. **User picks an impossible combination.** Agent sends `selected: [{Color: Red}, {Size: 15}]` with `preferences: ["Color", "Size"]`. No Red/15 exists. Server relaxes from the end of `preferences` — drops Size, keeps Color. Response `product.selected` is `[{Color: Red}]`. Agent diffs request vs response `selected`, sees Size was dropped, and can surface that to the user.

Each round-trip is stateless. The agent sends the full selection state, the server returns the full product state.

## Key Design Decisions

- **`product.selected` is the response anchor.** It determines the featured variant, the variant subset, and all availability signals. One concept, not three.
- **All returned variants match `product.selected`.** No "adjacent" or "contextually relevant" variants outside the selection. Availability context for non-matching options is carried by `options[].values[].available/exists`, not the variant array.
- **`selected_options` → `options` on variants.** Separates variant identity (what a variant *is*) from user selection state (what the user *chose*). These were previously conflated under the same name.
- **`input` correlation moved to `lookup_variant`.** Correlation is a batch-lookup concern, not intrinsic to variants. Base variant type stays clean; operation-specific extensions via `allOf`.
- **Singular response (`product`) not array.** Single-resource semantics — not found is an error (404 / `-32602`), not an empty result.

----

## Checklist

### DIFF
--- a/docs/specification/catalog/rest.md
+++ b/docs/specification/catalog/rest.md
@@ -556,6 +556,14 @@ Business outcomes use the standard HTTP 200 status with messages in the response
 
 {{ extension_schema_fields('catalog_lookup.json#/$defs/detail_product', 'catalog/rest') }}
 
+### Get Product Response {: #catalog-lookup-get-product-response }
+
+{{ extension_schema_fields('catalog_lookup.json#/$defs/get_product_response', 'catalog/rest') }}
+
+### Error Response {: #error-response }
+
+{{ schema_fields('types/error_response', 'catalog/rest') }}
+
 ## Conformance
 
 A conforming REST transport implementation **MUST**:
--- a/docs/specification/catalog/mcp.md
+++ b/docs/specification/catalog/mcp.md
@@ -680,6 +680,14 @@ results.
 
 {{ extension_schema_fields('catalog_lookup.json#/$defs/detail_product', 'catalog/mcp') }}
 
+### Get Product Response {: #catalog-lookup-get-product-response }
+
+{{ extension_schema_fields('catalog_lookup.json#/$defs/get_product_response', 'catalog/mcp') }}
+
+### Error Response {: #error-response }
+
+{{ schema_fields('types/error_response', 'catalog/mcp') }}
+
 ## Conformance
 
 A conforming MCP transport implementation **MUST**:


docs/specification/catalog/rest.md
```diff
@@ -329,6 +330,174 @@ messages indicating which identifiers were not found.
...
+    }
+    ```
+
+Green is out of stock (`available: false, exists: true`). Size 11 doesn't
+exist in Blue (`exists: false`). Variants returned match the Color=Blue
+selection.
+
+#### Product Not Found
+
+When the identifier does not resolve to a product, return HTTP 404:
```
maximenajim: REST docs specify POST /catalog/product returns HTTP 404 for not found, but rest.openapi.json only declares a 200 response for this operation (no 404, no 400 for invalid selections/preferences, no 401/403).
Suggested fix: update OpenAPI responses for /catalog/product to include at least 400, 401, 403, 404, and reference the standard error schema.

docs/specification/catalog/lookup.md
```diff
@@ -19,13 +19,32 @@
...
 
-| Operation | Description |
-| :--- | :--- |
-| **Lookup Catalog** | Retrieve products or variants by identifier. |
+| Operation | Tool / Endpoint | Description |
+| :--- | :--- | :--- |
+| **Batch Lookup** | `lookup_catalog` / `POST /catalog/lookup` | Retrieve multiple products by identifier. |
+| **Get Product** | `get_product` / `POST /catalog/product` | Retrieve full detail for a single product. |
+
+Both operations accept product and variant identifiers. They differ in cardinality
```
maximenajim: The “Both operations accept product and variant identifiers” table implies get_product supports the same identifier surface as lookup_catalog (SKU, URL, handle, etc.), but the later Get Product → Supported Identifiers section says “The id parameter accepts a single product ID or variant ID.” These are materially different guarantees.

Suggested fix: explicitly state:
“MUST support product ID + variant ID”
“MAY support SKU/URL/handle/etc.” (list optional identifiers)
and add a normative note: identifiers are opaque strings and MUST NOT be dereferenced or fetched as URLs.

source/schemas/shopping/catalog_lookup.json
```diff
@@ -48,6 +75,69 @@
...
+          "$ref": "types/context.json"
+        }
+      }
+    },
+    "detail_product": {
+      "description": "A product in a get_product response, extended with effective selections.",
+      "type": "object",
+      "allOf": [{ "$ref": "types/product.json" }],
+      "properties": {
+        "selected": {
```
maximenajim, line 124: detail_product.selected is optional and the description says “Present when the request includes selected options,” but the docs in lookup.md say the response MUST include product.selected and that it determines featured variant + availability anchor.

source/schemas/shopping/catalog_lookup.json
```diff
@@ -48,6 +75,69 @@
           "description": "Errors, warnings, or informational messages about the requested items."
         }
       }
+    },
+    "get_product_request": {
+      "type": "object",
+      "description": "Request body for single-product retrieval with optional option selection.",
```
gsmith85: Nit: "variant option selection"?  Or maybe just drop the terminating fragment altogether, I think the API is self explanatory.

source/schemas/shopping/catalog_lookup.json
```diff
@@ -48,6 +75,69 @@
...
+          "description": "Product or variant identifier. Implementations MUST support product ID and variant ID."
+        },
+        "selected": {
+          "type": "array",
+          "items": {
+            "$ref": "types/selected_option.json"
+          },
+          "description": "Partial or full option selections for interactive variant narrowing. When provided, response option values include availability signals (available, exists) relative to these selections."
+        },
+        "preferences": {
```
gsmith85, line 100: Question for the TC: As we think about the long-term scalability of get_product, how do we envision this relaxation logic behaving in multi-seller or marketplace scenarios?

Currently, since `seller` is metadata on a `variant` and not a first-class option, it cannot participate in the `preferences` stack. This forces the server to guess when inventory conflicts arise:
- Does it relax the attribute (show the wrong color to keep a preferred seller) or
- Relax the merchant (keep the color but switch to a non-Prime, higher-priced seller)?

To support heterogeneous inventory environments (where multiple sellers offer the same SKU with varying fulfillment terms), we should consider if 'Seller' or 'Fulfillment' needs to be a dimension the Agent can actually prioritize. Otherwise, the relaxation logic remains 'merchant-blind,' potentially dropping a user's primary intent (e.g., Prime eligibility, a specific vendor in B2B procurement) to satisfy a secondary one (e.g., color).

Should 'Seller' somehow be elevated to participate in the `preferences` stack, or should we handle marketplace-ranking logic through a more robust context object?

source/schemas/shopping/catalog_lookup.json
```diff
@@ -48,6 +75,69 @@
...
       }
+    },
+    "get_product_request": {
+      "type": "object",
+      "description": "Request body for single-product retrieval with optional option selection.",
+      "required": ["id"],
+      "properties": {
+        "id": {
+          "type": "string",
+          "description": "Product or variant identifier. Implementations MUST support product ID and variant ID."
```
gsmith85, line 91: Suggestion: Split out separate IDs for product and variant.  Using a single id for both may be brittle. As proposed, if an agent provides a variant `id` that is stale or purged, the server loses the anchor to the parent product and is forced to return a 404, even if the product itself is still active.

With both `product_id` and `variant_id` available you have an option for "soft recovery": If the server doesn't recognize the `variant_id` it can anchor on the product and still respect `preferences` to find the next best match.

Lower priority, but decoupling provides a clear signal to the server on the ID type being received, reducing lookups to determine if an identifier is a product or a variant.

source/schemas/shopping/catalog_lookup.json
```diff
@@ -48,6 +75,69 @@
...
+        },
+        "selected": {
```
gsmith85, line 93: Suggestion: Document how a conflict between a specific variant and `selected` is handled if the `options` aren't aligned.  I.e. if variant has option `Color:Red` but selected is `Color:Blue`.

source/schemas/shopping/types/selected_option.json
ihoosain, line 11: wondering if we should have an id as part of the object here, it would make calculating the result easier than matching on name/label i think.

source/schemas/shopping/catalog_lookup.json
igrigorik: I think we already address this. ID is defined as a GID, which is server assigend and opaque to the agent. The server can/should encode the semantics of product vs variant into the GID itself (e.g. `gid://{business}/Product/123` vs `gid://{business}/ProductVariant/456`), which gives the server everything it needs for type disambiguation and soft recovery.

source/schemas/shopping/catalog_lookup.json
igrigorik: In the aggregator model, each variant is an offer tied to a specific seller, and options are scoped to that seller's inventory. A product cluster aggregates offers across sellers...

```
Product: "Shoe Max 90"
├── Variant: SellerA / Blue / 10 / $120 / Same-day
├── Variant: SellerA / Red / 10 / $125 / Same-day
├── Variant: SellerB / Blue / 10 / $115 / Standard
├── Variant: SellerB / Blue / 12 / $118 / Standard
└── Variant: SellerC / Red / 10 / $110 / 2-day
```

When the agent selects Color=Blue/Size=10, the server returns matching variants from all sellers:

```
selected: [{Color: Blue}, {Size: 10}]

→ Variant: SellerA / Blue / 10 / $120 / Same-day  (available: false)
→ Variant: SellerB / Blue / 10 / $115 / Standard  (available: true)
```

SellerA is out of stock but SellerB's offer is there. No relaxation needed — the product selection is satisfied, just from a different seller. Offer ranking (preferred seller, fulfillment terms) is the business/aggregator's responsibility when deciding which variants to surface and in what order.

Relaxation only triggers when no variant from *any* seller matches — at that point it's purely product-attribute relaxation (drop Size before Color per `preferences`). The server never needs to choose "relax attribute vs relax seller" because those are different layers.

That said, I agree that choice of seller you purchase from can be valid criteria, and I think this is better modelled as an addition to filters -- i.e. you scope search and offers to a particular seller, or set of sellers. If there is demand, we can explore that as separate and followup PR.

docs/specification/catalog/rest.md
```diff
@@ -333,6 +334,164 @@ messages indicating which identifiers were not found.
...
+  "ucp": {
+    "version": "{{ ucp_version }}",
+    "status": "error",
+    "capabilities": {
+      "dev.ucp.shopping.catalog.lookup": [
+        {"version": "{{ ucp_version }}"}
+      ]
+    }
+  },
+  "messages": [
```
amithanda, line 479: The `get_product_response` schema declares required: `["ucp", "product"]`, but the error response example here produces a body with only `ucp` + `messages` — no `product` field. This makes every error response schema-invalid by the protocol's own definition. Is this expected?

source/schemas/shopping/catalog_lookup.json
```diff
@@ -34,6 +34,10 @@
           "minItems": 1,
           "description": "Identifiers to lookup. Implementations MUST support product ID and variant ID; MAY support secondary identifiers (SKU, handle, etc.)."
         },
+        "filters": {
```
amithanda, line 37: `search_filters` was added to `lookup_request `(the batch operation), but the Lookup capability is explicitly identifier-first: "use this when you already have identifiers." Applying `search_filters` here blurs the hard boundary with catalog_search.

Additionally, could not find documentation in the updated lookup.md, so no description of semantics, interaction with ids etc.

Can you elaborate more on why this is needed for catalog_lookup?

igrigorik: Good catch on docs, added.

Filters still have merit and use on lookup. The caller may have a set of IDs but still wants to provide a price or ships-to constraint: tell me which of these under $50 and ship to Canada. Similarly on get product: "show me available variants under $50".

docs/specification/catalog/rest.md
igrigorik: fixed

source/schemas/shopping/catalog_lookup.json
amithanda: I agree with Greg that we'll need to take a closer look at modeling sellers for marketplaces etc e.g filtering by sellers. Ilya do you think this would be a non-breaking future change to add that in a follow up PR, if we are introducing it at different level e.g context or filters or introducing new field in the main product schema etc and still keep this seller info more as metadata or do you think it will be a breaking change?

amithanda: Discussed in the TC that we can introduce later in a non-breaking way.

PR #199, opened 2026-02-20
chore: follow-up to update CODEOWNERS with devops-maintainers for additional tooling files 

nearlyforget:
Update CODEOWNERS to assign ownership of root-level tooling and generated files (hooks.py, main.py, generate_ts_schema_types.js, and /generated/). This ensures that infrastructure and automation changes are reviewed by the correct domain experts.


.github/CODEOWNERS
wry-ry, line 1: Sorry for /docs/, it should be devops-maintainers, and then /docs/specification/ should be maintainers I think?

PR #200, opened 2026-02-20
fix: clarify version negotiation

igrigorik:
Currently, spec requires businesses to process any request from a platform at an older protocol version (via "MUST process if platform ≤ business"). This imposed an unbounded backwards-compatibility obligation — every business must support every historical version indefinitely. Further, the negotiation mechanism is underspecified and missing a primitive for business to advertise which older versions are supported.

Update: businesses can list accepted versions via a `supported_versions` array in their profile. Platforms can discover the list by fetching the profile from well-known and MUST support the business's declared version to initiate a request -- the request carries a profile link with the version platform wants to use. The business validates the platform's version against its supported_versions list and rejects with 422 if there is no match.

## Checklist

- [x] Documentation update

### DIFF
--- a/docs/specification/overview.md
+++ b/docs/specification/overview.md
@@ -1836,6 +1836,7 @@ Response with version confirmation:
 
 Version unsupported error — no resource is created:
 
+```json
 {
   "ucp": { "version": "2026-01-11", "status": "error" },
   "messages": [{


docs/specification/overview.md
```diff
@@ -1629,20 +1695,34 @@ Response with version confirmation:
...
+#### Capability Versions
+
+Capability versions are negotiated independently of the protocol
+version. Each capability in the profile is an array. Multiple entries
+for the same capability, each with a different `version`, advertise
+support for multiple versions of that capability. The capability
+intersection algorithm considers only capability versions supported
+by both parties.
+
+Businesses **MUST** include only capabilities compatible with the
```
deinck6, line 1861: How are we determining compatible capabilities? Is this something each new version should declare in the documentation? Or is this something that will be dependent on the specific implementations used by businesses and platforms?

docs/specification/overview.md
```diff
@@ -1629,20 +1695,34 @@ Response with version confirmation:
...
-  }]
+  "code": "version_unsupported",
+  "content": "Protocol version 2026-01-12 is not supported. This business supports versions 2026-01-11 and 2026-01-23.",
+  "continue_url": "https://merchant.com/cart"
 }
 ```
 
+#### Capability Versions
+
+Capability versions are negotiated independently of the protocol
```
deinck6, line 1854: If multiple versions are supported for a capability, both platforms and businesses need to come to independent conclusions on which capability version to use. To avoid unnecessary ambiguity, should the platform also directly send the capability version that was used when building the request?

Also, the intersection algorithm identifies the capability overlap but does not include any reference to how a version is determined if multiple are present. Maybe it can be assumed to always choose the greatest mutually supported version but could be useful to call out explicitly.

docs/specification/overview.md
```diff
@@ -1603,16 +1639,46 @@ Both businesses and platforms declare a single version in their profiles:
...
+listing the protocol versions they accept. When present,
+`supported_versions` **MUST** include the `version` value. When
+`supported_versions` is omitted, it implicitly signals that only
+the declared `version` is supported.
+
+```json
+{
+  "ucp": {
+    "version": "2026-01-11",
+    "supported_versions": ["2026-01-11", "2026-01-23"]
```
deinck6: Other supported versions will have a separate `/.well-known/ucp` files, correct? Should the URIs for those well-known files also be included alongside the version?

docs/specification/overview.md
igrigorik: @deinck6 yes, the platform advertises both the protocol version and enumerates the list of capabilities (with version on each) in their profile: https://ucp.dev/latest/specification/overview/#platform-profile

Good catch on version selection, added.

docs/specification/overview.md
igrigorik: Great question. 

As is, it's self-declared: if a party lists both `ucp.version: 2026-01-11` and `checkout: 2026-09-01` in their profile, they're declaring that it's a compatible combination. How did they determine that this combination is compatible is left as DIY exercise — not great.

Stepping back, our contract today is:
 - dev.ucp.* capabilities + extensions version with protocol releases by default 
 - Individual extensions MAY version independently
 - Vendor extensions are fully independent

What we're missing is a declarative binding between "extension X supports protocol version Y and above". One way we could address this is by asking extensions to explicitly declare the "min version" they depend on. 

```json
{
    "$id": "https://acme.com/ucp/schemas/loyalty.json",
    "name": "com.acme.shopping.loyalty",
    "title": "Acme Loyalty Points",
    "min_protocol_version": "2026-01-23", // <---
    "$defs": { ... }
  }
```

The contract then could be...

> If min_protocol_version is present, platforms MUST verify the negotiated protocol version is ≥ that value during schema resolution. If absent, the capability is assumed compatible with the protocol version declared in the advertising party's profile.

docs/specification/overview.md
igrigorik: Yeah, I think you're right. Followup commit..

docs/specification/overview.md
igrigorik: @deinck6 ptal at latest draft, implemented ☝🏻

docs/specification/overview.md
```diff
@@ -33,11 +33,18 @@ Schema notes:
...
 ## Discovery, Governance, and Negotiation
 
-UCP employs a server-selects architecture where the business (server) chooses
-the protocol version and capabilities from the intersection of both parties'
-capabilities. Both business and platform profiles can be cached by both parties,
-allowing efficient capability negotiation within the normal request/response
-flow between platform and business.
+UCP separates protocol version compatibility from capability negotiation.
+The business's profile at `/.well-known/ucp` describes capabilities for
+its current protocol version. Businesses that support older protocol
```
lemonmade: Is "current" synonymous with "latest"/ "most recent" here?

docs/specification/overview.md
```diff
@@ -264,6 +271,35 @@ This convention ensures:
...
+  "$defs": { ... }
+}
+```
+
+The schema author — not the profile publisher — declares the minimum
+protocol version requirement. The profile publisher selects and
+advertises compatible versions in their profile.
+
+If `min_protocol_version` is present, platforms and businesses
+**SHOULD** verify the negotiated protocol version is >=
```
lemonmade: Why should, and not must here?

docs/specification/overview.md
```diff
@@ -33,11 +33,18 @@ Schema notes:
...
-UCP employs a server-selects architecture where the business (server) chooses
-the protocol version and capabilities from the intersection of both parties'
-capabilities. Both business and platform profiles can be cached by both parties,
-allowing efficient capability negotiation within the normal request/response
-flow between platform and business.
+UCP separates protocol version compatibility from capability negotiation.
+The business's profile at `/.well-known/ucp` describes capabilities for
+its current protocol version. Businesses that support older protocol
+versions **SHOULD** publish version-specific profiles and advertise them
+via the `supported_versions` field — a map from protocol version to
```
richmolj, line 40: I think we'd also need some concept of deprecation cycle baked in. I assume businesses will not want to support every version forever, but need some way to signal to platforms "this will no longer be in `supported_versions` on this date"

lemonmade (approved):

docs/specification/overview.md
```diff
@@ -402,6 +443,11 @@ used to verify signatures on webhooks and other authenticated messages from the
 business. See [Key Discovery](#key-discovery) for key lookup and resolution,
 and [Message Signatures](signatures.md) for signing mechanics.
 
+Businesses that support older protocol versions **SHOULD** include a
+`supported_versions` object mapping each older version to a
+version-specific profile URI. See [Protocol Version](#protocol-version)
+for details.
```
manumilou, line 473: This flow requires an additional HTTP fetch when the platform's version doesn't match version.

docs/specification/overview.md
```diff
@@ -1642,16 +1725,74 @@ Both businesses and platforms declare a single version in their profiles:
...
+    error.
+
+Version-specific profiles are leaf documents — they describe exactly
+one protocol version and **MUST NOT** contain a `supported_versions`
+field.
 
-1. Platform declares version via profile referenced in request
+##### Request-Time Validation
+
+Businesses **MUST** validate the platform's protocol version on
```
deinck6, line 1803: Do we want to call out incompatible Capabilities here too? For example, after the intersection is performed and there are is not mutually supported capability versions for this type of request.

docs/specification/overview.md
igrigorik: Correct. Version mismatch requires fetching the version-specific profile. This is intentional: each profile is a self-contained document for exactly one protocol version. Platforms cache profiles for their active integrations, so this is a cold-start cost, not a per-request cost.

docs/specification/overview.md
igrigorik: I'll clarify.

docs/specification/overview.md
igrigorik: Added a note: business policy / decision.

docs/specification/overview.md
deinck6: Thanks for this response!

Do you mind explaining this a bit more? 

- dev.ucp.* capabilities + extensions version with protocol releases by default
- Individual extensions MAY version independently
- Vendor extensions are fully independent

The [Independent Component Versioning] section mentions:

- UCP protocol versions independently from capabilities
- Each capability versions independently from other capabilities

Why do "capabilities + extensions version with protocol releases by default" if they are expected to version independently?

For example, if we make a change to the negotiation and create a new protocol version, 03-03-2026, does that create a new Checkout version 03-03-2026 even though the capability did not change?

For "extension X supports protocol version Y and above",

- what about a min_capability_version as well? for the capability it is extending.

docs/specification/overview.md
```diff
@@ -1642,16 +1728,78 @@ Both businesses and platforms declare a single version in their profiles:
...
```
deinck6, line 1803: I'm curious what is the rationale is for not including UCP metadata (profile intersection) in the platform’s request to the business? I'm thinking about the Checkout capability specifically but maybe this applies more generally in this Request-time validation phase. 

Possible risks of not passing UCP metadata with request:

Cache Invalidation: Its mentioned above that the Profiles can be cached by both Businesses and Platforms.  While the Platform can refresh its cache after detecting a UCP metadata mismatch in a response, the Business has no way of knowing a transaction was abandoned for that reason. Without a specific signal from the Platform, the Business will continue to hold an out-of-sync Platform Profile. 

Silent Corruption: If the Platform fails to detect versioning mismatches via the business’s UCP response, transactions could be processed under a false pretense of how the other party is handling the request/response.

Reactive vs. Proactive Detection: Currently, the business completes processing before a version mismatch could be identified by the platform initiating the request. This might be risky for more sensitive or stateful operations.

igrigorik: Good questions. 

The UCP-Agent header on every request gives the business the platform's profile URI. The business fetches (and caches) that profile, which contains the platform's version and full capability set. Version validation and capability intersection happen on the business side before processing the request — so mismatches are caught proactively by the business, not reactively by the platform. I don't think we need to (or should) reinvent new layers of profile validation -- use the HTTP mechanisms at our disposal.

Cache staleness: standard HTTP caching semantics apply to profile fetches — ETag, Cache-Control, conditional requests. The lifetime of the profile is controlled by the platform, and platform _can_ advertise a versioned URL to provide immediate invalidation. 

Inline profile: we discussed this in early drafts (having the platform send its UCP metadata inline with the request) but deferred it as it was not critical. Supporting an inline profile _would_ offer some benefits, but at a cost of meaningful overhead on initial request -- open to (re)opening that conversation but let's tackle that as a separate conversation.

deinck6: Ah that's what I was missing, we can signal a change to the current well-known file using versioned URL (e.g. adding v={hash} to the URL). Makes sense, thank you!

docs/specification/overview.md
richmolj: @igrigorik business decision for sure. But there is no signal businesses can provide that tells Platforms "things will start breaking on date X". It is in both sides' interests for such a signal to be present.

docs/specification/overview.md
```diff
@@ -274,34 +274,55 @@ This convention ensures:
...
 
 ```json
 {
   "$id": "https://acme.com/ucp/schemas/loyalty.json",
   "name": "com.acme.shopping.loyalty",
   "title": "Acme Loyalty Points",
-  "min_protocol_version": "2026-01-23",
-  "$defs": { ... }
+  "requires": {
```
deinck6, line 289: Thanks for this change, I think this makes sense!

PR #203, opened 2026-02-23
feat!: signals for authorization & abuse 

### LINKED ISSUE #153: EP: risk & abuse signals
Merchants need buyer environment signals (IP, user-agent, device fingerprints) for fraud prevention, rate limiting, bot detection, etc. UCP has no standard location for these signals and no mechanism for merchants to communicate what signals affect their authorization confidence.

We briefly explored this topic in earlier draft of the spec via `risk_signals` and deferred as a TODO. After a number of attempts and explorations, couple key insights:

1. It's not just about "risk", many signals also apply for rate-limiting and similar operational concerns.
2. Signals are dynamic and can change in response to every operation — e.g. buyer info, cart item, totals, etc.
3. UCP has first-class concept of step-up via escalation & `continueUrl`, and a state machine that performs validation of the checkout state, and what we need is a ~confidence layer & and mechanism to dynamically request signals from the platform.
 
To elaborate on last point, two distinct concerns govern checkout progression:

| Concern | Question | Mechanism | Blocks Status? |
|---------|----------|-----------|----------------|
| **Completeness** | Is checkout structurally valid? | `error` message | Yes |
| **Confidence** | Can I authorize without friction? | `hint` message | No |

A checkout can be complete but low-confidence:
- `ready_for_complete` = "You have everything needed to ATTEMPT completion"
- `hint` messages = "My confidence is low; expect friction unless you provide X"

Validation failures are deterministic: e.g, no email = can't complete. Confidence is probabilistic: e.g., no device_id = higher probability of step-up response, not certainty.

```
Request → [error?] → incomplete (hard enforcement)
                          ↓ resolve
       → ready_for_complete + hints (confidence signals)
                          ↓ complete()
       → POLICY EVALUATION: f(signals, context, policy)
                          ↓
       → completed | requires_escalation (step-up) | declined
```

---

### 1. Introduce new `signals` object
Transport-agnostic and extensible container for environment signals on checkout, cart, and catalog:

```json
{
  "checkout": {
    "id": "chk_123",
    "signals": {
      "buyer_ip": "203.0.113.42",
      "user_agent": "Mozilla/5.0...",
      "device_id": "fp_abc123",
      "ja4": "t13d1516h2_8daaf6152771_b186095e22b6"
    }
  }
}
```

The spec would define a small set of well-known signals: buyer_ip, device_id, user_agent, ja4. New signals can be added via extensions and negotiated between platform & business as part of handshake; core signals are assumed to be ~universal and do not need negotiation.

### 2. Add new Message type: `hint`
Agent-targeted message expressing what signals affect merchant confidence:

```json
{
  "type": "hint",
  "path": "$.signals.device_id",
  "code": "risk",
  "content": "Device fingerprint may reduce verification friction"
}
```

Hint codes:
- `risk` — fraud prevention, authorization decisions
- `abuse` — rate limiting, bot detection

---

### Hard enforcement
Platform MUST provide email. Deterministic:
```json
{
  "checkout": {
    "status": "incomplete",
    "messages": [{
      "type": "error",
      "path": "$.buyer.email",
      "code": "REQUIRED_FIELD",
      "content": "Email address is required"
    }]
  }
}
```

### Non-blocking hint signal 
```json
{
  "checkout": {
     ...
    "messages": [{
      "type": "hint",
      "path": "$.signals.device_id",
      "code": "risk",
      "content": "Device fingerprint may reduce verification friction"
    }]
  }
}
```
Platform CAN attempt to complete without signal. Probabilistic — step-up more likely without it.

```json
  {
    "catalog": {
      "products": [...],
      "messages": [{
        "type": "hint",
        "path": "$.signals.buyer_ip",
        "code": "abuse",
        "content": "Without IP address, stricter rate limits may apply"
      }]
    }
  }
```
As message says, missing buyer_ip may result in stricter rate limit.

## Requirements
Both parent `signals` object and all the signals would be marked as `optional` by default. This allows us to introduce this as a non-breaking change.

igrigorik:
Platforms mediate buyer interaction, making them the sole party able to observe the transaction environment (IP, user agent, etc). Businesses need this data for authorization decisions, rate limiting, and abuse prevention — but have no direct access to it. This PR defines `signals` contract that allows businesses to dynamically notify platforms of requested signals (via info messages on the response) and for platforms to provide data on the request(s). 

- Signals serve authorization, rate limiting, and abuse prevention across cart and checkout
- Signals are platform attestations: values MUST reflect direct observations, not relayed buyer claims
- Signal requests from business are delivered via info messages with `risk`, `abuse` codes
- Extensions can define proprietary signals that can be negotiated and delivered via same mechanism

1. platform creates checkout session and provides buyer IP for rate limiting and auth:
 ```json
  POST /checkout-sessions
  {
    "line_items": [{ "product_id": "sku_123", "quantity": 1 }],
    "signals": {
      "buyer_ip": "203.0.113.42"
    }
  }
```

2. Business responds — checkout is valid, requests additional data to improve authorization:
```json
  {
    "id": "cs_abc",
    "line_items": [...],
    "messages": [
      {
        "type": "info",
        "code": "risk",
        "path": "$.signals.user_agent",
        "content": "User agent improves authorization confidence for high-value orders"
      }
    ]
  }
```

3. Platform provides the requested signal on next request:
```json
POST /checkout-sessions/cs_abc/complete
{
  "payment": {
    "instruments": [...]
  },
  "signals": {
    "buyer_ip": "203.0.113.42",
    "user_agent": "Mozilla/5.0 ..."
  }
}
```

Closes #153. 

---

## Checklist

- [x] Documentation update

### DIFF
--- a/source/schemas/shopping/checkout.json
+++ b/source/schemas/shopping/checkout.json
@@ -66,6 +66,22 @@
         "complete": "optional"
       }
     },
+    "risk_signals": {
+      "type": "object",
+      "description": "Deprecated. Use signals instead. Will be removed in the next version.",
+      "deprecated": true,
+      "additionalProperties": true,
+      "ucp_request": {
+        "complete": {
+          "transition": {
+            "from": "optional",
+            "to": "omit",
+            "description": "Replaced by signals. Will be removed in the next version."
+          }
+        }
+      },
+      "ucp_response": "omit"
+    },
     "status": {
       "type": "string",
       "enum": [


docs/specification/overview.md
```diff
@@ -1541,6 +1521,62 @@ Sensitive data (such as Payment Credentials or PII) **MUST** be handled
...
+context to support authorization and abuse prevention. Info messages do not
+affect resource status or block progression. Platforms **SHOULD** surface
+actionable information to the buyer.
+
+Well-known info codes for signal feedback:
+
+| Code    | Description                                            |
+| :------ | :----------------------------------------------------- |
+| `risk`  | Fraud assessment and authorization confidence advisory |
+| `abuse` | Rate limiting and bot detection advisory               |
```
lemonmade: I like this as an alternative to the extra message type 👍

source/schemas/shopping/checkout.json
```diff
@@ -59,6 +59,16 @@
...
     },
+    "signals": {
+      "$ref": "types/signals.json",
+      "description": "Transaction environment data attested by the platform to support authorization decisions and abuse prevention.",
+      "ucp_request": {
+        "create": "optional",
+        "update": "optional",
+        "complete": "optional"
+      },
+      "ucp_response": "omit"
```
lemonmade: Can you comment on why we don’t reflect these back in the response? I am not going to die on this hill, but as I mentioned in our initial discussion, I find it odd that this is one of the few platform-provided fields they would need to keep the state of themselves, since they can’t read the current values from a `get_checkout` call.

igrigorik: You found the honeypot. 

I don't have a strong philosophical objection to removing `omit` annotation on response, exactly for the reasons you mentioned. However, if you peek two lines above, you'll see that we have same pattern on `context` object and hence why I replicated the pattern here. I think we can make the exact same argument for `context`, and perhaps we should. Let's discuss on the TC call tomorrow.

source/schemas/shopping/types/signals.json
```diff
@@ -0,0 +1,18 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/signals.json",
+  "title": "Signals",
+  "description": "Transaction environment data attested by the platform to support authorization decisions and abuse prevention. Values MUST reflect direct platform observations, not buyer-asserted data. Well-known signals are defined as typed properties; proprietary signals MUST use reverse-domain naming (e.g., com.example.score).",
```
raginpirate: I disagree with this proposal for `MUST use reverse-domain naming`.

If our intention for this to be done the UCP-way, its via an extension which is properly namespaced and extends the signals field, and could let the business communicate with the platform what it necessitates.

If our intention is for this to be done out-of-band, then the platform and agent have formed a relationship and can land on unique field names with validation around their known identities with one another; again, no need to enforce namespacing on the field.

docs/specification/overview.md
```diff
@@ -1541,6 +1521,62 @@ Sensitive data (such as Payment Credentials or PII) **MUST** be handled
...
+{
+  "signals": {
+    "buyer_ip": "203.0.113.42",
+    "user_agent": "Mozilla/5.0 ..."
+  }
+}
+```
+
+Signal fields may contain personally identifiable information
+(PII). Platforms **SHOULD** include only signals relevant to the current
```
raginpirate, line 1706: As a platform, how do I learn that set of minimum signals I need to provide?

Right now, you've proposed I learn this via the info messages, but that requires me to start a checkout/cart and get a first negotiated response. What if the business wants to instantly escalate for anyone missing certain signals like IP address, and provide that context as early as possible, like the well-known?

IMO this is why I think it makes sense for these signals to be a capability by default, but interested in your thoughts.

gsmith85: I'm left wondering if limiting `signals` to `cart` and `checkout` is too narrow, and whether it should be a broadly available across APIs.


Consider the [Discovery APIs](https://github.com/Universal-Commerce-Protocol/ucp/pull/55) for example, without risk signals:

1. Fraud models lose perspective into how a user interacted with the merchant prior to cart and checkout.  These models increasingly consider the whole shopping journey.
2. A merchant may be limited in their ability to detect anomalous behavior of a single user through a proxy that may violate their terms (e.g. bulk exfiltration of product data or reviews for unapproved model training).

source/schemas/shopping/types/signals.json
gsmith85: My interpretation of [EP: risk & abuse signals](https://github.com/Universal-Commerce-Protocol/ucp/issues/153) had me under the impression that signals *were* being handled via extension versus out-of-band.

Was this covered in the 2-27 TC [meeting](https://github.com/Universal-Commerce-Protocol/meeting-minutes/tree/main/tc/2026)?  I'd like to better understand the rationale.

docs/specification/overview.md
igrigorik: @gsmith85 the intent is to make it available on catalog as well. This PR does not do so because catalog schema is still in-flight and PR hasn't landed in draft -- I definitely plan on mirroring signals there.

igrigorik: @raginpirate earlier rounds of conversation also focused on early discovery, but as we iterated we identified that requesting signals dynamically — e.g. responding to changes in contents of the cart, behavior patterns, etc — as the key flow that we need to get right & enable. Current draft leans into that. 

We discussed capability vs core in TC syncs, and we agreed that the mechanism is universal and should be modelled as part of core: our job is to define the vessel / object that extensions can layer their signals into. If the checkout requires `dev.signal-provider.risk_signals` to proceed, that can be modelled via relevant extension that augments the `signals` object.

source/schemas/shopping/types/signals.json
igrigorik: @gsmith85 yep, see: https://github.com/Universal-Commerce-Protocol/ucp/pull/203#discussion_r2878982820

@raginpirate let's continue here: https://github.com/Universal-Commerce-Protocol/ucp/pull/203#issuecomment-3992498314

douglasborthwick-crypto:
Thanks Ilya — the language shift to "independently verifiable third-party attestations" is the right framing. The trust guarantee comes from cryptographic verifiability, not from who observed the data.

Two notes on the example shape in the updated spec:

**1. `kid` field alongside `provider_jwks`**

The `com.example.attestation` object should include a `kid` (key ID) alongside `provider_jwks`. This is standard JWKS practice (RFC 7517 §4.5) — it lets the business select the correct public key directly without fetching and iterating the full key set, and is necessary for key rotation to work cleanly.

**2. The signal needs the full signed payload to be self-verifying**

For the business to actually verify the signature locally, the signal must include the complete signed payload — not just `pass`. In our case the signed payload is `{"id", "pass", "results", "attestedAt"}` (fixed field order, Section 3.3 of our [attestation spec](https://insumermodel.com/state-attestation-spec)). Without `id`, `results`, and `attestedAt`, the business can't reconstruct the payload and verify `sig`. A signal carrying only `pass` + `signature` requires trusting the platform's relay — exactly the dependency the self-verifying property is supposed to eliminate.

Here's what a verifiable attestation signal looks like with reverse-domain naming:

```json
"com.insumermodel.attestation": {
  "provider_jwks": "https://api.insumermodel.com/v1/jwks",
  "kid": "insumer-attest-v1",
  "id": "ATST-A7C3E",
  "pass": true,
  "results": ["...per-condition evaluations with conditionHash for tamper evidence..."],
  "attestedAt": "2026-03-03T12:00:00Z",
  "sig": "MEYCIQDx...base64..."
}
```

The business verifies by: fetching JWKS from `provider_jwks`, selecting the key matching `kid`, reconstructing the signed payload from `{id, pass, results, attestedAt}`, and verifying `sig` over its SHA-256 digest. Zero trust in the platform required. The `conditionHash` in each result additionally lets them confirm exactly what on-chain condition was evaluated (contract, chain, threshold) without that being alterable post-signing.

Happy to contribute an extension schema for this once the PR lands.

docs/specification/overview.md
raginpirate: Thanks for the context, fair to push that responsibility out to an extension in the future!

source/schemas/shopping/types/signals.json
```diff
@@ -0,0 +1,22 @@
...
+  "$id": "https://ucp.dev/schemas/shopping/types/signals.json",
+  "title": "Signals",
+  "description": "Transaction environment data provided by the platform to support authorization decisions and abuse prevention. Values MUST NOT be buyer-asserted claims — platforms provide signals based on direct observation or independently verifiable third-party attestations. All signal keys MUST use reverse-domain naming to ensure provenance and prevent collisions when multiple extensions contribute to the shared namespace.",
+  "type": "object",
+  "propertyNames": {
+    "pattern": "^[a-z][a-z0-9]*(?:\\.[a-z][a-z0-9_]*)+$",
+    "description": "Reverse-domain identifier (e.g., dev.ucp.buyer_ip, com.example.device_id)."
+  },
+  "properties": {
+    "dev.ucp.buyer_ip": {
```
raginpirate, line 12: I appreciate applying consistency with the other rev-dns enforced fields 🚀

source/schemas/shopping/checkout.json
igrigorik: https://github.com/Universal-Commerce-Protocol/ucp/pull/203/changes/43fa00d9acd3b95719beaa1ba6c16b21d92ec5ab

PR #210, opened 2026-02-23
feat: Adding currency field to top level of Order

deinck6:
# Description

Adding a currency field to the top level of the Order entity, similar to Checkout. Adding this as an optional field for now but we may want to make it required in the next version update / breaking change.

## Type of change

### DIFF
--- a/hooks.py
+++ b/hooks.py
@@ -281,20 +281,30 @@ def on_page_markdown(markdown, page, config, files):
 
     # Rewrite relative links to assets/ to absolute URLs
     # pointing to served assets folder.
-    target_base = f"{base_path}assets/"
+    markdown = _root_pages_asset_link_rewrite(markdown, base_path)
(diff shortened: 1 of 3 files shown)


source/schemas/shopping/order.json
```diff
@@ -82,6 +82,10 @@
       },
       "description": "Append-only event log of money movements (refunds, returns, credits, disputes, cancellations, etc.) that exist independently of fulfillment."
     },
+    "currency": {
+      "type": "string",
+      "description": "ISO 4217 currency code reflecting the merchant's market determination. This needs to match the currency value used in during checkout."
```
igrigorik:
```suggestion
      "description": "ISO 4217 currency code. MUST match the currency from the originating checkout session."
```

igrigorik (approved):

PR #211, opened 2026-02-24
fix: Image (core-concepts diagram) and spec link URLs broken after si…

dwdii:
…te/spec version refactor.

# Description

Fixing issue with HTML img references in built site and adding Specification link to homepage nav tree

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] Documentation update

### DIFF
--- a/docs/documentation/core-concepts.md
+++ b/docs/documentation/core-concepts.md
@@ -24,7 +24,7 @@ a standardized common language and functional primitives.
 
 This document provides the detailed technical specification for UCP.
 For a complete definition of all data models and schemas, see the
-[Schema Reference](specification/reference).
+[Schema Reference](../specification/reference.md).
 
 Its primary goal is to enable:
 


docs/documentation/core-concepts.md
```diff
@@ -24,7 +24,7 @@ a standardized common language and functional primitives.
 
 This document provides the detailed technical specification for UCP.
 For a complete definition of all data models and schemas, see the
-[Schema Reference](../specification/reference.md).
+[Schema Reference](specification/reference).
```
wry-ry: Was this intentional?

mkdocs.yml
```diff
@@ -30,6 +30,7 @@ nav:
   - Overview:
       - Home: index.md
       - Core Concepts: documentation/core-concepts.md
+      - Specification: /latest/specification/overview
```
wry-ry, line 33: Can this just be specification/overview.md to match below? Or does that not work?

docs/documentation/core-concepts.md
dwdii: Reverted this.

mkdocs.yml
dwdii: Didn't work... keeping for now pending further refactor.

Issue #137, opened 2026-01-28
[Feat]: Allow tender-based discounts via payment instrument qualifiers in Checkout Session Update


ACSchil:
### Is your feature request related to a problem? Please describe.

I’m working on the REST binding + Checkout Update flow and I don’t see a place for a Platform to indicate “what tender subtype was selected” in a way that a Business can use for tender-based discounts.

Concretely: as the Business, I want to offer discounts based on a card subtype / BIN group (e.g., certain debit/commercial/store-card programs). In many tokenized flows, the Business may not have access to PAN/BIN, so the Platform (or handler) is the party that can detect the subtype early.

Today, `Checkout Update Request` includes `payment` as a `Payment Update Request`, but that object is basically just `selected_instrument_id` + `instruments[]`. The base card instrument fields (brand/last4/etc.) don’t include a “subtype” or “qualifier” concept. As a result, there isn’t a standard way for the Platform to signal “selected tender qualifies for discount bucket X”.

Refs: https://ucp.dev/specification/reference/ (Checkout Update Request, Payment Update Request, Card Payment Instrument)

### Describe the solution you'd like

Add a small, generic mechanism for the Platform to pass a coarse, business-defined indicator alongside the selected payment instrument during `PUT /checkout-sessions/{id}`.

One possible shape (example only):

- Add an optional field on `Payment Instrument` (or `Payment Instrument Base`) like:
  - `qualifiers: string[]` (namespaced values, coarse buckets only), or
  - `qualifiers: Array<{ issuer: string, value: string }>` for explicit namespacing.

Example usage (JSON payload):

```json
{
  "payment": {
    "selected_instrument_id": "pi_123",
    "instruments": [
      {
        "id": "pi_123",
        "handler_id": "com.example.pay",
        "type": "card",
        "brand": "visa",
        "last_digits": "4242",
        "qualifiers": [
          "com.target.tender.bin_group:debit_promo_a"
        ]
      }
    ]
  }
}
```

Important constraints I think we should document if we do this:
- This is a coarse eligibility indicator (no raw BINs, no PAN fragments).
- Values should be namespaced to avoid collisions.
- Business remains authoritative: qualifier is an input signal, not a guarantee. At the time of the `/complete` call, a business may reject the call if the tender does not qualify. 

[...]

PR #214, opened 2026-02-24
feat: add payment instrument qualifiers

ACSchil:
# Description

Context / Closes https://github.com/Universal-Commerce-Protocol/ucp/issues/137 

Adds an optional qualifiers: string[] field to the Payment Instrument schema and documents Payment Qualifiers in the Checkout overview. Qualifiers are opaque, reverse-domain namespaced hints whose meaning/derivation is agreed out-of-band between the Platform and Business, enabling the Business to apply expected benefits prior to completion. If benefits are applied based on qualifiers, the Business SHOULD fail checkout when the selected payment instrument at completion does not meet the hinted qualifications; unknown qualifiers are treated as no-ops.

## Type of change

- [x] Documentation update

---

No. Adds a new optional field.

## Checklist

### DIFF
--- a/source/schemas/shopping/types/payment_instrument.json
+++ b/source/schemas/shopping/types/payment_instrument.json
@@ -37,7 +37,6 @@
       "type": "array",
       "description": "Opaque, namespaced eligibility hint strings from the Platform that hint to the Business the benefits to apply at checkout time, based on the selected payment instrument.",
       "uniqueItems": true,
-      "maxItems": 99,
       "items": {
         "type": "string",
         "maxLength": 256


docs/specification/checkout-rest.md
```diff
@@ -694,6 +694,223 @@ Follow-up calls after initial `fulfillment` data to update selection.
...
 
+#### Update Payment
+
+Prior to completing checkout, the Platform may provide the Business with selected payment instrument hints. These hints allow the Business to apply to the checkout session the expected benefits the selected payment instruction qualifies the Buyer for.
+
+=== "Request"
+
+    ```json
+    PUT /checkout-sessions/{id} HTTP/1.1
+    UCP-Agent: profile="https://platform.example/profile"
```
maximenajim, line 714: I think we need to establish a clear trust model for who is allowed to assert qualifiers. Given qualifiers can directly influence pricing/benefits, this is a prime vector for benefit/discount injection by a compromised agent or untrusted client.

Add explicit normative language here (and/or reference the overview section) that businesses MUST only accept/apply qualifiers from authenticated/authorized platform identities that are allowlisted for those qualifier namespaces, and MUST treat qualifiers as untrusted hints unless they can be verified/bound to the completion credential. Consider requiring namespace allowlisting per Platform integration.

docs/specification/checkout-rest.md
```diff
@@ -694,6 +694,223 @@ Follow-up calls after initial `fulfillment` data to update selection.
     }
     ```
 
+#### Update Payment
+
+Prior to completing checkout, the Platform may provide the Business with selected payment instrument hints. These hints allow the Business to apply to the checkout session the expected benefits the selected payment instruction qualifies the Buyer for.
```
maximenajim: Typo/wording: “payment instrument hints… expected benefits the selected payment instruction qualifies…” should be “payment instrument.” Also consider tightening the language to consistently use “instrument” vs “handler” to avoid confusion in implementers.

docs/specification/checkout-rest.md
```diff
@@ -694,6 +694,223 @@ Follow-up calls after initial `fulfillment` data to update selection.
...
+      },
+      "line_items": [
+        {
+          "item": {
+            "id": "item_123",
+            "title": "Red T-Shirt",
+            "price": 2500
+          },
+          "id": "li_1",
+          "quantity": 2,
```
maximenajim: trailing comma

docs/specification/checkout-rest.md
```diff
@@ -694,6 +694,223 @@ Follow-up calls after initial `fulfillment` data to update selection.
...
+
+=== "Request"
+
+    ```json
+    PUT /checkout-sessions/{id} HTTP/1.1
+    UCP-Agent: profile="https://platform.example/profile"
+    Content-Type: application/json
+
+    {
+      "id": "chk_123456789",
```
maximenajim, line 718: The Update Payment example uses PUT /checkout-sessions/{id} and includes "id": "chk_123456789" in the body, but the spec section doesn’t state what happens if the path {id} and body id differ.  Add a normative rule: either (a) body id MUST match path {id} and mismatches MUST be rejected with a specific error, or (b) body id is ignored and the path is authoritative.

docs/specification/checkout-rest.md
```diff
@@ -694,6 +694,223 @@ Follow-up calls after initial `fulfillment` data to update selection.
...
+        "version": "2026-01-11",
+        "capabilities": {
+          "dev.ucp.shopping.checkout": [
+            {"version": "2026-01-11"}
+          ]
+        },
+        "payment_handlers": {
+          "com.shopify.shop_pay": [
+            {
+              "id": "shop_pay_1234",
```
maximenajim: In the response, ucp.payment_handlers advertises com.shopify.shop_pay with id: shop_pay_1234, but the selected instrument uses handler_id: gpay_1234.

docs/specification/checkout-rest.md
```diff
@@ -694,6 +694,223 @@ Follow-up calls after initial `fulfillment` data to update selection.
...
+              "version": "2026-01-11",
+              "config": {
+                "merchant_id": "shop_merchant_123"
+              }
+            }
+          ]
+        }
+      },
+      "id": "chk_1234567890",
```
maximenajim: The request id is chk_123456789 but the response returns chk_1234567890 (extra 0).

docs/specification/overview.md
```diff
@@ -1036,6 +1036,20 @@ within UCP: **Negotiation**, **Acquisition**, and **Completion**.
...
+### Payment Qualifiers
+
+Prior to completing checkout, the Platform **MAY** provide the Business with selected payment instrument hints. These hints allow the Business to apply to the checkout session the expected benefits the selected payment instrument qualifies the Buyer for.
+
+Payment instruments **MAY** include a `qualifiers` array: opaque, namespaced strings that indicate benefit eligibility associated with the selected instrument.  The meaning of qualifier values (and how they are derived) is communicated between the Business and Platform out of band (for example, via offline agreement on BIN ranges or program identifiers).
+
+Qualifiers are hints, not proofs:
+
+- Businesses **SHOULD** treat unknown qualifiers as no-ops.
+- When the Platform provides qualifiers and the Business applies benefits based on them, the Business **SHOULD** fail checkout if the selected payment instrument presented during checkout completion does not meet the hinted qualifications.
```
maximenajim: Add normative language defining the expected behavior on mismatch (e.g., return a specific error that forces the Platform to re-select payment, or return a repriced session requiring buyer confirmation).

docs/specification/overview.md
```diff
@@ -1036,6 +1036,20 @@ within UCP: **Negotiation**, **Acquisition**, and **Completion**.
 2. **Acquisition (Platform ↔ Payment Credential Provider):** The platform executes the handler's logic. This happens client-side or agent-side, directly with the payment credential provider (e.g., exchanging credentials for a network token). The business is not involved, ensuring raw data never touches the business's frontend API.
 3. **Completion (Platform → Business):** The platform submits the opaque credential (token) to the business. The business uses it to capture funds via their backend integration with the payment credential provider.
 
+### Payment Qualifiers
```
maximenajim: Please add explicit privacy / data minimization guidance: qualifiers must not encode or leak sensitive payment attributes (PAN/BIN ranges, full program identifiers that uniquely fingerprint a user, regulated benefit eligibility, etc.). Even if “opaque,” these strings are transferable and may be logged or persisted.
Add a MUST-level constraint like “qualifiers MUST NOT contain PAN/BIN/PII or be user-unique identifiers,” and recommend coarse-grained program identifiers.

docs/specification/overview.md
```diff
@@ -1036,6 +1036,20 @@ within UCP: **Negotiation**, **Acquisition**, and **Completion**.
...
+
+Payment instruments **MAY** include a `qualifiers` array: opaque, namespaced strings that indicate benefit eligibility associated with the selected instrument.  The meaning of qualifier values (and how they are derived) is communicated between the Business and Platform out of band (for example, via offline agreement on BIN ranges or program identifiers).
+
+Qualifiers are hints, not proofs:
+
+- Businesses **SHOULD** treat unknown qualifiers as no-ops.
+- When the Platform provides qualifiers and the Business applies benefits based on them, the Business **SHOULD** fail checkout if the selected payment instrument presented during checkout completion does not meet the hinted qualifications.
+- Businesses **SHOULD NOT** fail checkout solely because a qualifier is unrecognized.
+
+Qualifier strings **SHOULD** use reverse-domain naming to avoid collisions and make authority explicit.
```
maximenajim: The reverse-domain naming guidance is good, but the spec doesn’t state who has authority over a namespace. To prevent a third-party platform from asserting com.merchant.*-style qualifiers, add a recommendation/requirement that businesses only accept qualifiers from allowlisted namespaces per authenticated platform, and ideally that namespaces map to the asserting party (platform/handler/merchant).

source/schemas/shopping/types/payment_instrument.json
```diff
@@ -32,6 +32,13 @@
     "display": {
       "type": "object",
       "description": "Display information for this payment instrument. Each payment instrument schema defines its specific display properties, as outlined by the payment handler."
+    },
+    "qualifiers": {
```
maximenajim: Qualifiers is currently an unbounded array of unbounded strings. For a protocol-level schema used by autonomous agents, this is a straightforward DoS / log-injection / storage-amplification risk.

Add schema constraints, for example:
- maxItems (e.g., 10–50 depending on expected use)
- uniqueItems: true (if semantics are set-like)
- string constraints like maxLength (e.g., 128/256) and a conservative pattern (e.g., reverse-domain-ish allowed chars)

Even if implementations don’t enforce strictly, having constraints in the canonical schema sets a safe baseline.

source/schemas/shopping/types/payment_instrument.json
```diff
@@ -32,6 +32,13 @@
...
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "description": "Opaque, namespaced qualifier strings from the Platform that hint to the Business the benefits to apply at checkout time, based on the selected payment instruments."
```
maximenajim: The description says “based on the selected payment instruments” (plural). If split-tender is not supported, consider clarifying to “selected payment instrument.” If split-tender is supported, the spec should define whether multiple instruments can be selected: true concurrently and how qualifiers apply in that scenario.

docs/specification/checkout-rest.md
ACSchil: > Given qualifiers can directly influence pricing/benefits, this is a prime vector for benefit/discount injection by a compromised agent or untrusted client.

I can look at making the language clearer. The qualifiers are just hints. The Business MUST apply benefits based on the supplied payment instrument and its credential, and NOT based on the qualifiers. I was unsure if we wanted to take a stance on whether the Business should accept or reject a /complete call for a payment instrument that does not match the supplied qualifiers. I believe the Business MUST reject it, but I could see arguments for making it "SHOULD", depending on how the Business markets the benefits, and what guarantees exist around receiving the benefits.

docs/specification/checkout-rest.md
ACSchil: Great call-out. Because this PR feat did not introduce that potential conflict (i.e. there are other Update docs with this issue), I propose we create a separate issue to address that. Thoughts?

Thinking through the solution, it's probably to return a 400 or 422. And this would then be captured under https://ucp.dev/2026-01-23/specification/checkout-rest/#status-codes .

docs/specification/overview.md
ACSchil: What is the protocol's philosophy for defining error codes? It looks like the spec is lack in well defined error codes (see: https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/shopping/types/error_code.json#L6 ), but provides the guideline that the Business SHOULD provide accurate error messages, but does not define codes (see: https://ucp.dev/2026-01-23/specification/checkout/#business ).

I definitely see the argument for standardizing the exception handling and encouraging interoperability.

source/schemas/shopping/types/payment_instrument.json
ACSchil: How should we communicate the reverse-domain-ish characters? 

Shall we say items match the pattern of `^[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?(?:\.[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?)+\.[a-z][a-z0-9_-]*$` ? i.e. `{reverse-domain}.{service}.{qualifier}`, e.g. "com.target.red_card" or "co.uk.example.qualifier-abc123"

raginpirate: re: max items and string constraints; I feel like those needs to be applied across the board to UCP and are not today. We should discuss this in the TC; do we need a sweeping application of that to the spec, and should that be done in isolation of this PR?

I also disagree with enforcing reverse-dns here; from my POV the primary use case of this should be in-band by payment handlers to express how you negotiate qualfiier context that you want exposed by a payment handler for a specific instrument. Out-of-band use cases should not be dictated and require reverse-dns; they have their own solutions for identifying the relying party they negotiated with and interpreting those fields independently for those parties.

docs/specification/overview.md
ACSchil: Following https://github.com/Universal-Commerce-Protocol/ucp/pull/147, I've added a new error code. I think we should come up with a more maintainable approach for documenting error codes though; I can foresee this become a _long_ laundry list.

source/schemas/shopping/types/payment_instrument.json
```diff
@@ -32,6 +32,16 @@
     "display": {
       "type": "object",
       "description": "Display information for this payment instrument. Each payment instrument schema defines its specific display properties, as outlined by the payment handler."
+    },
+    "qualifiers": {
+      "type": "array",
+      "description": "Opaque, namespaced qualifier strings from the Platform that hint to the Business the benefits to apply at checkout time, based on the selected payment instrument."
```
raginpirate: Will this always come from the platform?
Lets chat out a hypothetical.
I'm using <wallet brand>, and using it's agentic mode it returns back to the platform a token encrypted for merchant 1234. As the platform, I have no ability to actually determine the qualifiers; that credential isn't transparent to me.
But, if the wallet exposes the selected instrument qualifier, then easy; we can pass it through to the merchant early, just like a shipping option event.

docs/specification/checkout-rest.md
```diff
@@ -694,6 +694,230 @@ Follow-up calls after initial `fulfillment` data to update selection.
     }
     ```
 
+#### Update Payment
+
+Prior to completing checkout, the Platform may provide the Business with selected payment instrument hints. These hints allow the Business to apply to the checkout session the expected benefits the selected payment instrument qualifies the Buyer for.
```
raginpirate, line 708: I think we should come at this explanation as a feature of payment handlers, NOT from the perspective of a black hole platforms can leverage for out-of-band solutions.

As in, my payment handler lets me negotiate available qualifiers, and exposes that context to the agent to submit to the platform. It just so happens to give platforms and businesses an easy path to do out-of-band context sharing, but IMO we shouldn't be authoring this from that perspective.

docs/specification/overview.md
```diff
@@ -1050,6 +1050,26 @@ within UCP: **Negotiation**, **Acquisition**, and **Completion**.
 2. **Acquisition (Platform ↔ Payment Credential Provider):** The platform executes the handler's logic. This happens client-side or agent-side, directly with the payment credential provider (e.g., exchanging credentials for a network token). The business is not involved, ensuring raw data never touches the business's frontend API.
 3. **Completion (Platform → Business):** The platform submits the opaque credential (token) to the business. The business uses it to capture funds via their backend integration with the payment credential provider.
 
+### Payment Qualifiers
+
+Prior to completing checkout, the Platform **MAY** provide the Business with selected payment instrument *hints*. These hints allow the Business to apply to the checkout session the expected benefits the selected payment instrument qualifies the Buyer for. This gives the user a more accurate preview of the final order total before they commit to the purchase.
```
gsmith85, line 1055: Suggestion: Add a few tangible examples of "benefits" to the description to improve comprehension for implementers. Including a brief list would clarify that the qualifiers array is intended to support things like:
- Co-branded Card Perks (e.g., instant discounts for using a specific brand's card).
- Issuer-Funded Promotions (e.g., statement credits).
- FX Fee Waivers for international transactions.
- Promotional APRs or specialized financing terms.
- Logistical Upgrades (e.g., expedited shipping for premium cardholders).

docs/specification/overview.md
```diff
@@ -1050,6 +1050,26 @@ within UCP: **Negotiation**, **Acquisition**, and **Completion**.
 2. **Acquisition (Platform ↔ Payment Credential Provider):** The platform executes the handler's logic. This happens client-side or agent-side, directly with the payment credential provider (e.g., exchanging credentials for a network token). The business is not involved, ensuring raw data never touches the business's frontend API.
 3. **Completion (Platform → Business):** The platform submits the opaque credential (token) to the business. The business uses it to capture funds via their backend integration with the payment credential provider.
 
+### Payment Qualifiers
```
gsmith85: Consideration: Is `qualifiers` descriptive enough? "Qualifiers" often implies a binding state or a restrictive limitation. Since the semantics explicitly define these as non-binding "hints" that require independent verification, `benefit_hints` or `eligibility_hints` might be a better fit. It signals that the field is a suggestion for the presentation and pre-checkout logic rather than a final, validated qualification for financial settlement.

docs/specification/overview.md
```diff
@@ -1050,6 +1050,26 @@ within UCP: **Negotiation**, **Acquisition**, and **Completion**.
...
+
+Prior to completing checkout, the Platform **MAY** provide the Business with selected payment instrument *hints*. These hints allow the Business to apply to the checkout session the expected benefits the selected payment instrument qualifies the Buyer for. This gives the user a more accurate preview of the final order total before they commit to the purchase.
+
+Payment instruments **MAY** include a `qualifiers` array: opaque, namespaced strings that hint benefit eligibility associated with the selected instrument.  The meaning of qualifier values, and how they are derived, is communicated between the Business and Platform out of band (for example, via offline agreement on BIN ranges or program identifiers).
+
+**Qualifier Strings:**
+
+- Qualifier strings **SHOULD** use reverse-domain naming to avoid collisions.
+- Qualifiers strings **MUST NOT** contain sensitive payment attributes such as PAN, BIN, PII, or user-unique identifiers.
```
gsmith85: Nit: We are not at all  consistent, but define acronyms like BIN (Bank Identification Number)upon their first mention? It helps ensure the spec remains accessible to engineers who might be new to the payments domain.

docs/specification/overview.md
```diff
@@ -1050,6 +1050,26 @@ within UCP: **Negotiation**, **Acquisition**, and **Completion**.
...
 3. **Completion (Platform → Business):** The platform submits the opaque credential (token) to the business. The business uses it to capture funds via their backend integration with the payment credential provider.
 
+### Payment Qualifiers
+
+Prior to completing checkout, the Platform **MAY** provide the Business with selected payment instrument *hints*. These hints allow the Business to apply to the checkout session the expected benefits the selected payment instrument qualifies the Buyer for. This gives the user a more accurate preview of the final order total before they commit to the purchase.
+
+Payment instruments **MAY** include a `qualifiers` array: opaque, namespaced strings that hint benefit eligibility associated with the selected instrument.  The meaning of qualifier values, and how they are derived, is communicated between the Business and Platform out of band (for example, via offline agreement on BIN ranges or program identifiers).
```
gsmith85: Consideration: Will this field exclusively be used for "benefits"?

While the current description focuses on perks like discounts, there are several
scenarios where a Platform might pass hints for regulatory or technical restriction
that aren't "benefits" to the user. For example:
- Product-Category Restrictions: Signaling a corporate or government card that is legally barred from purchasing specific items (e.g., alcohol, tobacco, or gaming).
- Surcharge Eligibility: Signaling a high-cost commercial card where a merchant is legally allowed (or required) to pass on a processing fee.
- KYC/Identity Status: Signaling a specific "Know Your Customer" verification level that might skip—or trigger—additional identity checks.
- Cross-Border Compliance: Signaling an instrument with geographic limitations that might affect tax or shipping legality.

If we anticipate these non-benefit use cases, would a more neutral term like `eligibility_hints` be more appropriate?

source/schemas/shopping/types/payment_instrument.json
ACSchil: I agree we have an opportunity to apply these across the board. I am okay adding the constraint here (as I have done), or removing it an referencing this feedback in a new issue that takes a broader look. 

> Out-of-band use cases should not be dictated and require reverse-dns

Without reverse domain naming, there's a significant chance of collision. Imagine two entities with `member_program`; it would be helpful for the program named be tied to that program owner, e.g. `com.target.member_program` so that another merchant can also have `com.merchant_b.member_program`.

source/schemas/shopping/types/payment_instrument.json
ACSchil: > we can pass it through to the merchant early

This comes from the Platform, correct? Even if the Platform is using a wallet that abstract the credential?

docs/specification/checkout-rest.md
maximenajim: Sounds good. Please create a separate issue.

docs/specification/checkout-rest.md
ACSchil: I understand this POV. See https://github.com/Universal-Commerce-Protocol/ucp/pull/214#issuecomment-3974835126 . Are you okay with us moving forward with this for now?

docs/specification/overview.md
ACSchil: Thanks for this input. I went with `eligibility_hints`.

docs/specification/overview.md
ACSchil: I agree. We have a glossary, but it's mostly for our own terms. Maybe separately we take a pass at capturing all acronyms there?

source/schemas/shopping/types/payment_instrument.json
```diff
@@ -32,6 +32,16 @@
     "display": {
       "type": "object",
       "description": "Display information for this payment instrument. Each payment instrument schema defines its specific display properties, as outlined by the payment handler."
+    },
+    "eligibility_hints": {
+      "type": "array",
+      "description": "Opaque, namespaced eligibility hint strings from the Platform that hint to the Business the benefits to apply at checkout time, based on the selected payment instrument.",
+      "uniqueItems": true,
+      "maxItems": 99,
```
amithanda: @maximenajim - I think we discussed in TC that we will not be adding maxItems (as it is hard to specify what is a good number and we have different places where we have arrays and we don't do this at other places), but rather provide a more generic implementation guidance for implementors to restrict to a reasonable size list for various use cases across UCP. Should we update this to reflect that?

ACSchil: Thanks for sharing this here! I've made that change.

docs/specification/overview.md
```diff
@@ -1050,6 +1050,34 @@ within UCP: **Negotiation**, **Acquisition**, and **Completion**.
...
+- `eligibility_hints` strings **MUST NOT** contain sensitive payment attributes such as PAN, BIN, PII, or user-unique identifiers.
+- `eligibility_hints` strings **SHOULD** be coarse-grained program identifiers.
+
+**eligibility_hints Semantics:**
+
+- `eligibility_hints` are hints and **MUST NOT** be treated as proof of eligibility.
+- The Business **MUST NOT** grant final or irreversible benefits solely due to `eligibility_hints`.
+- The Business **MUST** determine benefits eligibility from the completion payment instrument and credential, not from `eligibility_hints`.
+- The Business **SHOULD** return an error, using the [`invalid_eligibility_hint`](checkout.md#standard-errors), during checkout completion if the provided `eligibility_hints` do not match the final payment instrument's eligibility.
+- When receiving `invalid_eligibility_hint`, the Platform **SHOULD** update `eligibility_hints`, and **MUST** present the user with an opportunity to review benefits changes (e.g. discounts, totals, etc.).
```
amithanda, line 1079: So does this mean, the platform should ask the user to select a different payment instrument or remove the eligibility_hint from their request and use the same instrument or are there other options for the platform? 
I am trying to understand what are the valid responses from Platform once Business returns an error "invalid_eligibility_hint" and should we update the documentation to be clearer?

docs/specification/overview.md
gsmith85: Primarily as a smoke test to make sure that I had the CLA signed, I split out a central glossary (https://github.com/Universal-Commerce-Protocol/ucp/pull/241).  We can layer these acronyms in there once it's merged!

source/schemas/shopping/types/payment_instrument.json
```diff
@@ -32,6 +32,15 @@
     "display": {
       "type": "object",
       "description": "Display information for this payment instrument. Each payment instrument schema defines its specific display properties, as outlined by the payment handler."
+    },
+    "eligibility_hints": {
```
raginpirate, line 36: Nit: The phrase "eligibility_hints" feels like its conveying what this instrument can be used for (especially confusing for me after working on constraints) and I'm not sure it has any industry examples of the term being used. I preferred the original title, or some other title along the lines of "something_hint", but this is non-blocking.

docs/specification/overview.md
raginpirate: Great question Amit and I will actually drive this further while thinking through the full flow:

- As a platform, I submit instrument 1234 with a qualifier the business does not respect. I am expecting the business can actually silently ignore it, or optionally provide back a message to say "hey this qualifier isn't right so I removed it". The state can still be ready-for-submit immediately, not requiring platform action. The platform is informed regardless because the qualifier is **not** surfaced back.
- As a business, when I receive a credential for an instrument with qualifier 1234, but now the credential mismatches from that qualifier context, I can perform the same operation: remove the qualifier, optional non-blocking message back to the platform.
- If this were to occur during a complete call, and this actually impacts the totals, I should fail the complete. It should not be a requirement to fail though if the qualifier never meaningfully impacted the state of the checkout; that should be up to business discretion, and I am not sure needs to be reflected in the overview of this feature.

Is all of the above correct? If so, I'd love to see it (or the adjustments to that view) codified with the right MUST / SHOULD outlook.

source/schemas/shopping/types/payment_instrument.json
gsmith85: With the feedback from @raginpirate on naming and to push on the idea that these might not strictly be benefits (per https://github.com/Universal-Commerce-Protocol/ucp/pull/214#discussion_r2862047862) perhaps consider `provisions`?   Or, `provision_hints` if you want to more strictly communicate it's a hint.

In the `description`, and elsewhere we can note `provisions` may capture hints for "entitlements" (e.g. priority_fulfillment, travel_credit) and "constraints" (e.g. no_tobacco, domestic_only).

Also non-blocking, but obviously this modeling is easiest to change early.

docs/specification/checkout-rest.md
```diff
@@ -703,6 +703,230 @@ Follow-up calls after initial `fulfillment` data to update selection.
...
+            "id": "pi_gpay_5678",
+            "handler_id": "gpay_1234",
+            "type": "card",
+            "selected": true,
+            "display": {
+              "brand": "mastercard",
+              "last_digits": "5678",
+              "rich_text_description": "Google Pay •••• 5678"
+            },
+            "eligibility_hints": [
```
lemonmade, line 773: What do we think of using `context` for this instead, introduced in https://github.com/Universal-Commerce-Protocol/ucp/pull/62/changes? It feels like that slice of the object is meant to give us a home for these kinds of up front "hints", and its documentation specifically references custom pricing, and the higher-resolution fields in the API that would override the hinting. It feels a little awkward to require the partner to provide so much of the payment instrument, just to offer up a hint on what payment-related qualifiers will apply.

douglasborthwick-crypto:
On @lemonmade's context suggestion: if context is the right home for pre-checkout signals, it might be worth distinguishing in the spec between opaque hints the Business has no way to validate until completion, and signed attestations that carry a signature and key ID verifiable against a published JWKS. The latter don't require any trust assumption about the Platform.

That distinction also speaks to the unhappy path @raginpirate outlined. The credential mismatch scenario becomes verifiable rather than trust-based when the hint itself is signed by a third party -- the Business can check independently, without a round-trip to the Platform, before completion.

docs/specification/checkout-rest.md
raginpirate: Honestly great question @lemonmade. I think there is serious merit in just using context.

To argue the point of this PR though, I do see that qualifiers are tied to an instrument, so in the case of split payments it becomes harder to justify using a default checkout context bucket; which instrument should be visualized as contributing which qualifier(s)?

And the deeper reason I would prefer to push for qualifiers, is because I do want the primitive to be usable by payment handler developers, rather than it being this arbitrary bucket its being first introduced as here. Example: Use the `company_name`_pay sdk, and have it expose an event similar to address selection for method selection, which lets the platform push an instrument with qualifier pay_1234 to describe the matching bin range it was initialized with. If this is the future we want to support, the primitive living on the instrument does make sense to me, as its meant for handlers to interface with, rather than checkout's generic context bucket.

douglasborthwick-crypto:
@igrigorik this reframe is sharp — `context.eligibility` available across the full lifecycle solves the "too late" problem cleanly, and `provisional` + `eligibility` on `applied_discount` gives the Platform the attribution it currently has to guess at.

One thing the proposal leaves open is the verification mechanism at completion. For instrument-bound eligibility (RedCard BIN check), the Business can verify locally. But for eligibility that isn't derivable from the payment credential — on-chain token holdings, DAO memberships, NFT ownership — the Business needs an external attestation.

This is where the opaque-vs-signed distinction from my earlier comment becomes concrete under your model. A Platform can call an attestation provider and pass the signed result alongside the eligibility signal:

```json
{
  "context": {
    "eligibility": ["com.example.nft_holder"],
    "eligibility_attestations": {
      "com.example.nft_holder": {
        "code": "INSR-A7F2K",
        "expiresAt": "2026-03-06T14:30:00.000Z",
        "sig": "MEUCIQD...base64-P1363...",
        "kid": "insumer-attest-v1"
      }
    }
  }
}
```

That's the actual shape returned by our [`POST /v1/ucp/discount`](https://insumermodel.com/developers/commerce/) `verification` object. The `sig` is ECDSA P-256 over `{"verified":true,"totalDiscount":10,"code":"INSR-A7F2K","merchantId":"...","expiresAt":"..."}`. The Business verifies it against the JWKS at [`/.well-known/jwks.json`](https://insumermodel.com/.well-known/jwks.json) using `kid` to select the key — no round-trip to the Platform or the chain.

The `provisional` flag in your model still works as-is: the Business applies the discount as `provisional: true` at cart, then at completion verifies the signature and drops the flag. The unhappy path becomes deterministic — a failed sig check produces `invalid_eligibility_hint`, not a trust dispute.

This doesn't require any new schema beyond what you've already proposed. `eligibility_attestations` on `context` would be an optional companion to `eligibility` — Platforms that can supply a signed backing do so, Platforms that can't still send opaque hints. The Business decides its own policy on which it accepts as provisional vs. which it requires a signature for.

PR #128, opened 2026-01-28
feat(errors): spec error handling for UCP negotiation failures

igrigorik:
Current spec does not clearly specify how to handle discovery (profile fetch) failures vs result of incompatible set of features  after services + capability intersection logic is executed. This PR addresses that.

* Discovery failure → transport error (REST 400/422/424, JSON-RPC error)
* Negotiation failure → UCP response with messages (REST 200, JSON-RPC result)
* Business outcomes → UCP response with messages (REST 200, JSON-RPC result)

Status codes:
  - 400: INVALID_PROFILE_URL (malformed URL)
  - 424: PROFILE_UNREACHABLE (fetch failed - failed dependency)
  - 422: PROFILE_MALFORMED (invalid content)
  - 200: All business outcomes including NOT_FOUND, INSUFFICIENT_STOCK

---

## Checklist
- [x] New feature (non-breaking change which adds functionality)
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings

PR #216, opened 2026-02-24
feat(errors): business logic error response for checkout and cart

alexpark20:
# Description

## Background

UCP is transport-portable by design — the same business logic runs over REST, MCP (JSON-RPC), and embedded protocols. Delivering on this requires a crisp boundary between transport-level failures and application-level outcomes:

**Transport error** — the request could not be delivered to, parsed by, or routed to the UCP handler (auth, rate limit, malformed JSON, missing headers, discovery profile unreachable).

**Application outcome** — the UCP handler received a well-formed request, executed, and is reporting its result — including failures.

The boundary is between stages 4 and 5 in the execution pipeline:

| Stage                        | Example                             | Error Type      |
| ---------------------------- | ----------------------------------- | --------------- |
| 1. Delivery (TCP, TLS)       | Connection refused                  | Transport       |
| 2. Framing (JSON parse)      | Malformed JSON                      | Transport       |
| 3. Protocol validation       | Missing required headers            | Transport       |
| 4. Routing, Auth, Signatures | 401, 403, 429                       | Transport       |
| 5. **Handler execution**     | capability mismatch, not-found, ... | **Application** |

## Problem

Checkout and cart operations currently require every response to include `id` and other resource fields. When business logic prevents returning a valid resource, implementations have no valid path:

- Creating or returning an empty/meaningless resource violates business logic
- Using protocol-level errors (HTTP 4xx / JSON-RPC error) conflates transport failures with business outcomes

This affects any operation that returns a checkout or cart resource, not only creation. Some examples: 

- **Creation rejected**: all items out of stock, regional restriction, etc
- **Resource not found**: a valid checkout or cart no longer exists (get_checkout, get_cart)
- **Capability negotiation failure**: no intersection found (create_ , update_, or complete_)

> **Note:** Error response examples for these scenarios already existed in the
> specification docs (introduced in #141 and #128), but the JSON schemas had no
> formal support for this shape. This PR closes that gap by making the schema
> match what the docs have always described.

### Desired shape for each error scenario

From https://github.com/Universal-Commerce-Protocol/ucp/pull/216#pullrequestreview-3867721295 by @igrigorik 

| Scenario                                 | REST    | MCP                     | Response Shape                        |
| ---------------------------------------- | ------- | ----------------------- | ------------------------------------- |
| Infra: auth, rate limit, server error    | 4xx/5xx | JSON-RPC error          | Transport-native                      |
| Discovery: profile unreachable/malformed | 422/424 | JSON-RPC error (-32001) | Transport-native                      |
| **Handler ran — everything below**       | **200** | **JSON-RPC result**     | **UCP envelope**                      |
| Capabilities incompatible                | 200     | JSON-RPC result         | `{ucp, messages, continue_url}`       |
| Resource not found                       | 200     | JSON-RPC result         | `{ucp, messages, continue_url}`       |
| Creation rejected (all items fail)       | 200     | JSON-RPC result         | `{ucp, messages, continue_url}`       |
| Success                                  | 200     | JSON-RPC result         | `{ucp, ..., messages}`  (full object) |

## Solution

**ucp.status as discriminator**.

A status field ("success" | "error", default "success") is added to `ucp.json#/$defs/base`. When `ucp.status === "error"`, the response **must not** include a resource (checkout or cart) — only the error envelope. When `ucp.status === "success"`, the full resource is always present. 

This gives platforms a clean, unambiguous application-level signal. The field is constrained in new `error_response.json` via allOf to require "error".

**New file:** `source/schemas/shopping/types/error_response.json`

```json
{
  "ucp": { 
       "version": "2026-01-11",
       "status": "error" // <-- new field to ucp base
  },
  "messages": [
    {
      "type": "error",
      "code": "out_of_stock",
      "severity": "unrecoverable",
      "content": "All requested items are currently out of stock."
    }
  ],
  "continue_url": "https://merchant.com/"
}
```

**Transport schemas own the union**. 

Each transport defines how it signals business-logic failure in its response/result type:

- rest.openapi.json — adds checkout_response and cart_response components (oneOf: [resource, error_response]), used by all operation response schemas. Request body schemas continue to reference checkout / cart directly.
- mcp.openrpc.json — adds checkout_result and cart_result components with the same oneOf, used by all method result schemas. Param schemas unchanged.
- embedded.openrpc.json — no changes; ECP uses native JSON-RPC error objects for failures, the UCP error envelope does not apply there.

### Why `additionalProperties: false` on `error_response.json`

Extensions like fulfillment and buyer_consent compose onto checkout using allOf: [$ref: checkout.json, {extra fields}]. Per the JSON Schema spec, additionalProperties only recognises properties declared in the same schema object — it does not see properties defined in sibling allOf branches. So error_response.json's additionalProperties: false correctly rejects any error response that includes extension fields. This is intentional.

### Documentation changes

**Added "Error Response" tabs** to `create_checkout` (REST + MCP) and `create_cart` (REST + MCP) showing the no-resource error shape.

**Updated all error examples** across checkout and cart docs to include `ucp.status: "error"` and `severity: "unrecoverable"`.

**Fixed pre-existing cart doc examples**:

- `cart-mcp.md` — Not Found responses were wrapped under a `"cart"` key in `structuredContent`. Since no cart resource is created, the error envelope belongs at the root of `structuredContent` (consistent with the checkout binding and the new `error.json` schema).

### Alternatives rejected

**Optional `id` field** — weakens schema semantics, poor type safety for clients (must check every field even in success cases), doesn't solve the semantic problem.

**Sentinel/reserved ID values** (e.g. `"-1"`) — magic values, semantic confusion, database issues.

## Type of change

---

## Checklist

### DIFF
--- a/docs/specification/checkout.md
+++ b/docs/specification/checkout.md
@@ -123,12 +123,12 @@ is `"success"`, a resource is returned and severity indicates the
 recommended action. When `ucp.status` is `"error"`, no valid resource
 exists — severity is `unrecoverable`:
 
-| Severity                | Meaning                                          | Platform Action                   |
-| :---------------------- | :----------------------------------------------- | :-------------------------------- |
-| `recoverable`           | Platform can resolve by modifying inputs via API | Update resource and retry         |
-| `requires_buyer_input`  | Business requires input not available via API    | Hand off via `continue_url`       |
-| `requires_buyer_review` | Buyer review and authorization is required       | Hand off via `continue_url`       |
-| `unrecoverable`         | No resource exists to act on                     | Retry with new resource or inputs |
+| Severity                | Meaning                                          | Platform Action                                                   |
+| :---------------------- | :----------------------------------------------- | :---------------------------------------------------------------- |
+| `recoverable`           | Platform can resolve by modifying inputs via API | Update resource and retry                                         |
+| `requires_buyer_input`  | Business requires input not available via API    | Hand off via `continue_url`                                       |
+| `requires_buyer_review` | Buyer review and authorization is required       | Hand off via `continue_url`                                       |
+| `unrecoverable`         | No resource exists to act on                     | Retry with new resource or inputs, or hand off via `continue_url` |
 
 Errors with `requires_*` severity contribute to `status: requires_escalation`.
 Both result in buyer handoff, but represent different checkout states.
--- a/source/schemas/shopping/types/message_error.json
+++ b/source/schemas/shopping/types/message_error.json
@@ -43,7 +43,7 @@
         "requires_buyer_review",
         "unrecoverable"
       ],
-      "description": "Reflects the resource state and recommended action. 'recoverable': platform can resolve by modifying inputs and retrying via API. 'requires_buyer_input': merchant requires information their API doesn't support collecting programmatically (checkout incomplete). 'requires_buyer_review': buyer must authorize before order placement due to policy, regulatory, or entitlement rules 'unrecoverable': no valid resource exists to act on, retry with new resource or inputs. Errors with 'requires_*' severity contribute to 'status: requires_escalation'."
+      "description": "Reflects the resource state and recommended action. 'recoverable': platform can resolve by modifying inputs and retrying via API. 'requires_buyer_input': merchant requires information their API doesn't support collecting programmatically (checkout incomplete). 'requires_buyer_review': buyer must authorize before order placement due to policy, regulatory, or entitlement rules. 'unrecoverable': no valid resource exists to act on, retry with new resource or inputs. Errors with 'requires_*' severity contribute to 'status: requires_escalation'."
     }
   }
 }


source/schemas/shopping/cart.json
```diff
@@ -118,5 +94,29 @@
       "description": "Cart expiry timestamp (RFC 3339). Optional.",
       "ucp_request": "omit"
     }
+  },
+  "$defs": {
```
igrigorik: why did this get shuffled? diff noise.

alexpark20: Ah, wrong push. I removed this diff now.

richmolj (approved):

docs/specification/checkout.md
```diff
@@ -120,11 +120,12 @@ The `messages` array contains errors, warnings, and informational messages
...
-| `recoverable`           | Platform can fix via API                      | Resolve using Update Checkout |
-| `requires_buyer_input`  | Business requires input not available via API | Hand off via `continue_url`   |
-| `requires_buyer_review` | Buyer review and authorization is required    | Hand off via `continue_url`   |
+| Severity                | Meaning                                                          | Platform Action                     |
+| :---------------------- | :--------------------------------------------------------------- | :---------------------------------- |
+| `recoverable`           | Platform can resolve by modifying inputs and retrying via API    | Modify inputs and retry calling API |
+| `requires_buyer_input`  | Business requires input not available via API                    | Hand off via `continue_url`         |
+| `requires_buyer_review` | Buyer review and authorization is required                       | Hand off via `continue_url`         |
+| `unrecoverable`         | Terminal failure — cannot be recovered via API                   | Redirect via `continue_url`         |
```
jingyli: nit: How is `Redirect` different from `Hand off`?

Also given that it's not part of the guidelines that `continue_url` **MUST** be populated for `unrecoverable` errors, should we loosen the language here with a MAY?

source/schemas/shopping/types/error_response.json
```diff
@@ -0,0 +1,34 @@
...
+          "properties": { "status": { "const": "error" } },
+          "required": ["status"]
+        }
+      ],
+      "description": "UCP protocol metadata. Status MUST be 'error' for error response."
+    },
+    "messages": {
+      "type": "array",
+      "items": {
+        "$ref": "message.json"
```
jingyli, line 23: Just thinking out loud: Given this is very specific to the error use case, should we tighten this schema composition so only error messages can ever be present in the array (i.e. we should have `$ref` pointing to `message_error.json` directly)?

igrigorik: Good question, debated this as well and left it as intentionally open. A response can carry multiple messages, which can include info and warning messages alongside errors — e.g. error on invalid input, info for missing user_agent signal. The description ("messages describing why the operation failed") scopes the intent without over-constraining the schema.

igrigorik:
@jingyli good questions and feedback. I believe this should address it: https://github.com/Universal-Commerce-Protocol/ucp/pull/216/changes/d7bc7a0f6277b3eee1f3fa07f8fcd511b3fc8f5c

source/schemas/shopping/types/message_error.json
```diff
@@ -40,9 +40,10 @@
       "enum": [
         "recoverable",
         "requires_buyer_input",
-        "requires_buyer_review"
+        "requires_buyer_review",
+        "unrecoverable"
       ],
-      "description": "Declares who resolves this error. 'recoverable': agent can fix via API. 'requires_buyer_input': merchant requires information their API doesn't support collecting programmatically (checkout incomplete). 'requires_buyer_review': buyer must authorize before order placement due to policy, regulatory, or entitlement rules (checkout complete). Errors with 'requires_*' severity contribute to 'status: requires_escalation'."
+      "description": "Reflects the resource state and recommended action. 'recoverable': platform can resolve by modifying inputs and retrying via API. 'requires_buyer_input': merchant requires information their API doesn't support collecting programmatically (checkout incomplete). 'requires_buyer_review': buyer must authorize before order placement due to policy, regulatory, or entitlement rules 'unrecoverable': no valid resource exists to act on, retry with new resource or inputs. Errors with 'requires_*' severity contribute to 'status: requires_escalation'."
```
jingyli: Nit: Looks like the `(checkout complete).` part is truncated by accident here?

docs/specification/checkout.md
```diff
@@ -118,13 +118,17 @@ platform receives messages indicating what's needed to progress.
...
+is `"success"`, a resource is returned and severity indicates the
+recommended action. When `ucp.status` is `"error"`, no valid resource
+exists — severity is `unrecoverable`:
+
+| Severity                | Meaning                                          | Platform Action                   |
+| :---------------------- | :----------------------------------------------- | :-------------------------------- |
+| `recoverable`           | Platform can resolve by modifying inputs via API | Update resource and retry         |
+| `requires_buyer_input`  | Business requires input not available via API    | Hand off via `continue_url`       |
+| `requires_buyer_review` | Buyer review and authorization is required       | Hand off via `continue_url`       |
+| `unrecoverable`         | No resource exists to act on                     | Retry with new resource or inputs |
```
jingyli: Nit: may also want to add "OR hand off via `continue_url`" as another platform action here for `unrecoverable` errors to be consistent with the processing algorithm below?

source/schemas/shopping/types/error_response.json
jingyli: Ack, the expectation & example makes sense.

The only nit-pick follow-up thought I have is why can't those info & warning messages be folded into the `path` & `content` fields of `message_error.json` if it's further explaining on why the operation failed..? But I guess an edge case I can think of is if business want to force platform display of the error via warning messages.

source/schemas/shopping/types/message_error.json
igrigorik: Good catch. Deliberate on the parens but messed the sentence split. Fixed.

PR #217, opened 2026-02-25
fix: update 2026-01-11 to use ucp-schema and latest version of main.py

wry-ry:
# Description

Update 2026-01-11 to use ucp-schema and latest version of main.py for consistency across versions.

## Type of change

      or fields**)
- [x] Documentation update

### DIFF
--- a/docs/specification/fulfillment.md
+++ b/docs/specification/fulfillment.md
@@ -52,7 +52,7 @@ method.
 
 ### Properties
 
-{{ extension_fields('fulfillment_resp', 'fulfillment') }}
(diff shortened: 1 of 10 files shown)


igrigorik:
\o/

docs/specification/fulfillment.md
jingyli, line 61: Would these also need to be updated? I think we have more of these annotations in the specification doc folder.

wry-ry: schema_fields macro works differently than extension_fields. schema_fields still expects _resp or _req to understand the direction of the schema to be rendered.

PR #226, opened 2026-02-27
docs: fix broken links and add dynamic versioning

wry-ry:
# Description

Fix broken links in draft, make sure all links to specifications or schemas point to a dated version, and dynamically replace versions within json examples.

See staged site here: https://wry-ry.github.io/ucp/

## Type of change

      or fields**)
- [x] Documentation update

### DIFF
--- a/docs/specification/checkout-a2a.md
+++ b/docs/specification/checkout-a2a.md
@@ -245,7 +245,7 @@ checkout object containing an `order` attribute with `id` and `permalink_url`.
         "kind": "data",
         "data": {
           "a2a.ucp.checkout.payment": {
-            ...paymentDataObject
+            ...paymentObject
           },
           "a2a.ucp.checkout.risk_signals":{...content}
         }
@@ -327,11 +327,11 @@ public keys.
 
 When the user confirms the payment on a platform, the user signed checkout and
 payment mandate objects must be sent as `DataPart`s to the business agent for
-completing checkout. The `payment_data` which includes the payment mandate must
+completing checkout. The `payment` which includes the payment mandate must
 be submitted as part of a `DataPart` with attribute name
-`a2a.ucp.checkout.payment_data`. Signed checkout mandate must be specified in
+`a2a.ucp.checkout.payment`. Signed checkout mandate must be specified in
 the `DataPart` as `ap2.checkout_mandate`. The `token` attribute of
-`payment_data` contains the payment mandate. Refer to
+`payment` contains the payment mandate. Refer to
 [AP2 Mandates Extension](ap2-mandates.md) documentation for more details about
 verification and processing of the mandates to complete the checkout.
 
--- a/.github/workflows/linter.yaml
+++ b/.github/workflows/linter.yaml
@@ -66,3 +66,4 @@ jobs:
           VALIDATE_GITHUB_ACTIONS_ZIZMOR: false
           VALIDATE_JSCPD: false
           VALIDATE_SPELL_CODESPELL: false  # Using cspell (spellcheck.yaml)
+          VALIDATE_OPENAPI: false
--- a/docs/specification/catalog/mcp.md
+++ b/docs/specification/catalog/mcp.md
@@ -29,27 +29,27 @@ Businesses advertise MCP transport availability through their UCP profile at
 ```json
 {
   "ucp": {
-    "version": "2026-01-11",
+    "version": "{{ ucp_version }}",
     "services": {
       "dev.ucp.shopping": {
-        "version": "2026-01-11",
-        "spec": "https://ucp.dev/specification/overview",
+        "version": "{{ ucp_version }}",
+        "spec": "https://ucp.dev/{{ ucp_version }}/specification/overview",
         "mcp": {
-          "schema": "https://ucp.dev/services/shopping/mcp.openrpc.json",
+          "schema": "https://ucp.dev/{{ ucp_version }}/services/shopping/mcp.openrpc.json",
           "endpoint": "https://business.example.com/ucp/mcp"
         }
       }
     },
     "capabilities": {
       "dev.ucp.shopping.catalog.search": [{
-        "version": "2026-01-11",
-        "spec": "https://ucp.dev/specification/catalog/search",
-        "schema": "https://ucp.dev/schemas/shopping/catalog_search.json"
+        "version": "{{ ucp_version }}",
+        "spec": "https://ucp.dev/{{ ucp_version }}/specification/catalog/search",
+        "schema": "https://ucp.dev/{{ ucp_version }}/schemas/shopping/catalog_search.json"
       }],
       "dev.ucp.shopping.catalog.lookup": [{
-        "version": "2026-01-11",
-        "spec": "https://ucp.dev/specification/catalog/lookup",
-        "schema": "https://ucp.dev/schemas/shopping/catalog_lookup.json"
+        "version": "{{ ucp_version }}",
+        "spec": "https://ucp.dev/{{ ucp_version }}/specification/catalog/lookup",
+        "schema": "https://ucp.dev/{{ ucp_version }}/schemas/shopping/catalog_lookup.json"
       }]
     }
   }
@@ -100,15 +100,19 @@ version compatibility checking and capability negotiation.
 
 Maps to the [Catalog Search](search.md) capability.
 
-#### Request
+#### Search Request
 
-{{ extension_schema_fields('catalog_search.json#/$defs/search_request', 'catalog-mcp') }}
+{{ extension_schema_fields(
+  'catalog_search.json#/$defs/search_request', 'catalog/mcp'
+) }}
 
-#### Response
+### Search Response
 
-{{ extension_schema_fields('catalog_search.json#/$defs/search_response', 'catalog-mcp') }}
+{{ extension_schema_fields(
+  'catalog_search.json#/$defs/search_response', 'catalog/mcp'
+) }}
 
-#### Example
+#### Search Example
 
 === "Request"
 
@@ -156,10 +160,10 @@ Maps to the [Catalog Search](search.md) capability.
       "result": {
         "structuredContent": {
           "ucp": {
-            "version": "2026-01-11",
+            "version": "{{ ucp_version }}",
             "capabilities": {
               "dev.ucp.shopping.catalog.search": [
-                {"version": "2026-01-11"}
+                {"version": "{{ ucp_version }}"}
               ]
             }
           },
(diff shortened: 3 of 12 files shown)


igrigorik (approved):

damaz91 (approved):

docs/documentation/schema-authoring.md
```diff
@@ -285,9 +285,9 @@ A capability schema defines both payload structure and declaration variants:
 ```json
 {
-  "$id": "https://ucp.dev/schemas/shopping/checkout.json",
+  "$id": "https://ucp.dev/{{ ucp_version }}/schemas/shopping/checkout.json",
   "name": "dev.ucp.shopping.checkout",
-  "version": "2026-01-11",
+  "version": "{{ ucp_version }}",
```
jingyli, line 315: Not related to this file per se - but looks like in the hosted schema, somehow this `version` value is tagged on at the very bottom vs. immediately after `name`. An example: https://ucp.dev/draft/schemas/shopping/fulfillment.json (if you scroll all the way to the bottom you would see `version`).

docs/specification/examples/encrypted-credential-handler.md
```diff
@@ -353,5 +353,5 @@ Content-Type: application/json
 
 ## References
 
-* **Identity Schema:** `https://ucp.dev/schemas/shopping/types/payment_identity.json`
-* **Instrument Schema:** `https://ucp.dev/schemas/shopping/types/card_payment_instrument.json`
+* **Identity Schema:** `https://ucp.dev/{{ ucp_version }}/schemas/shopping/types/payment_identity.json`
```
jingyli: nit: Should we remove these addition ` formatters (here and below) so these references can behave like proper hyperlinks?

docs/specification/ap2-mandates.md
```diff
@@ -420,7 +420,7 @@ checkout.
 
 ## Schema
 
-### Business Authorization
+### Business Authorization {: #merchant-authorization }
```
jingyli, line 423: Is this an intended edit here?

docs/specification/cart-rest.md
```diff
@@ -28,29 +28,29 @@ Businesses advertise REST transport availability through their UCP profile at
...
+    "version": "{{ ucp_version }}",
     "services": {
       "dev.ucp.shopping": {
-        "version": "2026-01-15",
-        "spec": "https://ucp.dev/specification/overview",
+        "version": "{{ ucp_version }}",
+        "spec": "https://ucp.dev/{{ ucp_version }}/specification/overview",
         "rest": {
-          "schema": "https://ucp.dev/services/shopping/openapi.json",
+          "schema": "https://ucp.dev/{{ ucp_version }}/services/shopping/rest.openapi.json",
```
jingyli, line 37: Also a naive QQ on this hosted file - in https://ucp.dev/draft/services/shopping/rest.openapi.json, there is a hardcoded `version` date (`"2026-03-03"`) instead of `"draft"` that is consistently being used throughout the documentation here, is this intended?

docs/specification/cart.md
```diff
@@ -171,9 +175,13 @@ consistent data structures when converting a cart to a checkout session.
 
 {{ schema_fields('types/line_item_update_req', 'checkout') }}
 
-#### Line Item Response
+#### Line Item
+
+{{ schema_fields('types/line_item_resp', 'cart') }}
```
jingyli, line 200: Is it expected that the hyperlinks (i.e. if I click on `Item`) within this schema go to the `/specification/reference/#item` instead of staying on the `cart.md` page here?

docs/specification/checkout-a2a.md
```diff
@@ -248,8 +244,8 @@ checkout object containing an `order` attribute with `id` and `permalink_url`.
       {
         "kind": "data",
         "data": {
-          "a2a.ucp.checkout.payment": {
-            ...paymentObject
+          "a2a.ucp.checkout.payment_data": {
```
jingyli: This change should be reverted, given the expected field in `complete_checkout` should be `payment` instead of `payment_data`?

ditto on the section below where it's referencing `payment_data` still..

docs/specification/checkout.md
jingyli: With the addition of "Selected Payment Instrument" below, this section can be removed? I don't think we are using this entity anymore?

docs/specification/embedded-checkout.md
jingyli: This shows up as `No properties defined` in the staged repo (maybe not related to this change per se given it's consistently showing this issue even in the current state).

docs/specification/order.md
```diff
@@ -391,10 +391,10 @@ zero-downtime key rotation procedures.
 
 {{ extension_schema_fields('capability.json#/$defs/response_schema', 'order') }}
 
-### Total Response
+### Total
 
 {{ schema_fields('types/total_resp', 'order') }}
 
-### UCP Response Order
+### UCP Response Order Schema <span id="ucp"></span> {: #ucp-response-order-schema }
```
jingyli, line 398: Naive QQ: why is this <span> element needed here?

docs/specification/overview.md
```diff
@@ -1556,11 +1556,11 @@ Initiation comes through a `continue_url` that is returned by the business.
 
 UCP defines a set of standard capabilities:
```
jingyli, line 1668: I just realize that `cart` capability is missing from the table here, is it possible to quickly add in this documentation tweak in the PR?

.linkignore
jingyli, line 1: Any reason this file is needed?  Links here looks to be pointing correctly to the older reference versions (or is it just a temporary ignore that will be cleaned up)?

docs/documentation/schema-authoring.md
wry-ry: Yeah the actual source files don't have version, it gets added when the file is processed. Is it a problem that it's at the bottom?

docs/specification/examples/encrypted-credential-handler.md
wry-ry: Good point, updated.

docs/specification/ap2-mandates.md
wry-ry: Yes, it's adding a tag so that the hyperlink anchors from generated schemas land at the right place.

docs/specification/cart-rest.md
wry-ry: AFAIU it's a standard practice for JSON schema publishing to use the current date instead of "draft" for schema validation tools and also so developers can easily tell if the version they are using is different than the published one. I was thinking having "draft" in the docs was fine for human readers but open to keeping it consistent between the two if you feel that makes more sense.

docs/specification/cart.md
wry-ry: Kinda intended but admittedly not the most elegant solution. The problem is that I think we don't consistently include referenced types in each file (so cart.md may or may not have Item schema included), but we should have all of them in reference, so I just had the script set the link to reference/#item. Alternatively, we could have it point to the same page, and just make sure all referenced schemas are on the page instead.

docs/specification/checkout-a2a.md
wry-ry: Thanks good catch, I think Gemini updated the schema referenced based on the descriptive text. I fixed the actual text.

docs/specification/checkout.md
wry-ry: Removed.

docs/specification/order.md
wry-ry: It's because there is a sequence of references for the UCP object in the webhook request body in services/shopping/rest.openapi.json: schemas/order.json -> shopping/ucp.json#/$defs/response_order_schema. So we generate a ucp-response-order-schema anchor link, and the tag needs to be there.
But in the webhook response body in services/shopping/rest.openapi.json, it references schemas/ucp.json, so it's looking for the UCP object with a ucp anchor link, so we need to add another tag, which has to be done using a span because markdown only supports one tag.

Alternatively, maybe we could update services/shopping/rest.openapi.json so the ucp ref in responses for orderEvent also points to ucp.json#/$defs/response_order_schema? Not sure if that will break anything?

.linkignore
wry-ry: This is so the new link checker script doesn't flag these links. Because it's used as the URI in released versions for A2A, we can't change it.

docs/specification/cart.md
wry-ry: Actually it should just reference the same page so I updated the script logic.

docs/specification/embedded-checkout.md
wry-ry: This is fixed now.

docs/specification/overview.md
wry-ry: Added to the table, let me know if you want to tweak the langauge.

docs/documentation/schema-authoring.md
jingyli: Nope, no problem from a schema perspective! Was just thinking out loud to see if there is an easy fix to put the `version` param right on the top (but per your comment above, it looks like inlining will not be easy since it is appended as part of post-processing).

docs/specification/cart-rest.md
jingyli: Ack, thanks for the explanation! I don't think it's a top problem to be considered right now (given the URL still has draft), but one thought is: if business is using hosted schema to build their profile/UCP metadata, I wonder if we would run into some confusion as they may accidentally use this "placeholder" date as their transport version.

docs/specification/checkout-a2a.md
```diff
@@ -249,7 +245,7 @@ checkout object containing an `order` attribute with `id` and `permalink_url`.
         "kind": "data",
         "data": {
           "a2a.ucp.checkout.payment": {
-            ...paymentObject
+            ...paymentDataObject
```
jingyli: nit: Also revert this back to `paymentObject` here?

docs/specification/checkout-a2a.md
```diff
@@ -330,18 +325,17 @@ checkout payload against the business's public keys.
...
-must be specified in the `DataPart` as `ap2.checkout_mandate`. The `token`
-attribute of `payment.instruments[*].credential` contains the payment mandate.
-Refer to [AP2 Mandates Extension](ap2-mandates.md) documentation for more
-details about verification and processing of the mandates to complete the
-checkout.
+When the user confirms the payment on a platform, the user signed checkout and
+payment mandate objects must be sent as `DataPart`s to the business agent for
+completing checkout. The `payment_data` which includes the payment mandate must
+be submitted as part of a `DataPart` with attribute name
+`a2a.ucp.checkout.payment_data`. Signed checkout mandate must be specified in
```
jingyli: nit: This section should also be reverted back?

docs/specification/overview.md
jingyli: LGTM, thanks!

docs/specification/checkout-a2a.md
wry-ry: Reverted.

docs/specification/checkout-a2a.md
wry-ry: Reverted.

PR #244, opened 2026-03-07
feat!: Embedded protocol transport binding for cart capability + reauth mechanism in ECP

jingyli:
Reopening the PR to keep commit history clean (original [PR](https://github.com/Universal-Commerce-Protocol/ucp/pull/116)) and hopefully third-time is the charm. This will also cross-reference/carry-over comments from the old PR.

## Summary

Cart capability (`dev.ucp.shopping.cart`) is being introduced in https://github.com/Universal-Commerce-Protocol/ucp/pull/73. This adds an additional transport binding (EP) on the capability (building directly upon the referenced PR), on top of existing REST & MCP bindings. It follows the main UCP principle that **"Embedded bindings should retain symmetry regardless of the capability it’s implemented for"** (and hence why certain special mechanisms like auth are also being introduce to ECP to maintain symmetry).

## Motivation

Businesses need a way to embed their cart building UI into eligible platforms, especially when complex experiences (i.e. item recommendations & upsells) are involved during cart building. Embedded Cart Protocol (ECaP) addresses this need.

### Goals

* Establish a new auth exchange mechanism for embedded transport binding that utilizes postMessage mechanisms.
* Maintain general schema & design consistency with existing ECP (i.e. reuse naming conventions, message types, etc.).
* Introduce a way for businesses to advertise exactly for which capability is EP active for.

### Non-Goals

## Detailed Design

| Category          | Communication Direction | Purpose                                                                   | Pattern                | Core Messages                                                                             |
| :---------------- | :---------------------- | :------------------------------------------------------------------------ | :--------------------- | :---------------------------------------------------------------------------------------- |
| **Handshake**     | Embedded Cart -> Host   | Establish connection between host and Embedded Cart.                      | Request                | `ep.cart.ready`                                                                           |
| **Authentication**| Embedded Cart -> Host   | Communicate auth data exchanges between Embedded Cart and host.           | Request                | `ep.cart.auth`                                                                            |
| **Lifecycle**     | Embedded Cart -> Host   | Inform of cart state in Embedded Cart.                                    | Notification           | `ep.cart.start`, `ep.cart.complete`                                                       |
| **State Change**  | Embedded Cart -> Host   | Inform of cart field changes.                                             | Notification           | `ep.cart.line_items.change`, `ep.cart.buyer.change`, `ep.cart.messages.change`            |

## Risks and Mitigations

* **Security:** Introduced new request mechanism  (`ep.cart.auth`) to exchange required authorization data instead of adding them as query parameters in `continue_url` to avoid hijacking attacks & support reauth scenarios.

## Graduation Criteria

**Candidate → Stable:**

### Implementation History

TBD


docs/specification/embedded-checkout.md
```diff
@@ -648,6 +648,62 @@ information:**
...
+
+The `ec.auth` message is a request, which means that host
+**MUST** respond to exchange the authorization.
+
+- **Direction:** host → Embedded Checkout
+- **Type:** Response
+- **Result Payload:**
+    - `authorization` (string, **REQUIRED**): The requested authorization data,
+    can be in the form of an OAuth token, JWT, API keys, etc.
+    - `checkout` (object, **REQUIRED**): An optional checkout holding the last known state to the host.
```
lemonmade: Why is this needed? The embedded checkout should know the last `checkout` it sent before this message.

docs/specification/embedded-checkout.md
```diff
@@ -648,6 +648,62 @@ information:**
...
+
+**Example Message:**
+
+```json
+{
+    "jsonrpc": "2.0",
+    "id": "auth_1",
+    "method": "ec.auth",
+    "params": {
+        "type": "oauth"
```
lemonmade, line 698: What are the allowed `type`s here?

docs/specification/embedded-checkout.md
```diff
@@ -648,6 +648,62 @@ information:**
...
+}
+```
+
+The `ec.auth` message is a request, which means that host
+**MUST** respond to exchange the authorization.
+
+- **Direction:** host → Embedded Checkout
+- **Type:** Response
+- **Result Payload:**
+    - `authorization` (string, **REQUIRED**): The requested authorization data,
```
lemonmade: IMO, `authorization` reads a little weirdly for a message called `ec.auth`. What about using `token` here, or `credential`, which mirrors the name for a similar concept in payment instruments?

docs/specification/embedded-checkout.md
```diff
@@ -648,6 +648,62 @@ information:**
...
+    "jsonrpc": "2.0",
+    "id": "auth_1",
+    "result": {
+        "authorization": "fake_identity_linking_oauth_token"
+    }
+}
+```
+
+If the ingestion of the authorization is not successful, Embedded Checkout **MAY**
+re-initiate this request with the host again.
```
lemonmade: Can we should how a Host would indicate failure, and what an Embedded Checkout should do with failures that come up?

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,702 @@
...
+}
+```
+
+### State Change Messages
+
+State change messages inform the host of changes that have already occurred
+in the cart interface. These are informational only. The cart has
+already applied the changes and rendered the updated UI.
+
+#### `ect.line_items.change`
```
lemonmade: I think we are missing documentation for `ect.line_items.change_request`.

lemonmade (commented):
Left a few small nits and questions, but I am really liking how this has shaped up, especially the symmetry with Embedded Checkout 👍 This PR does still have one Cart-only message, `ect.line_items.change_request`; do we think this same feature makes sense in embedded checkout?

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,702 @@
...
+- **Type:** Notification
+- **Payload:**
+    - `cart`: The latest state of the cart
+
+**Example Message:**
+
+```json
+{
+    "jsonrpc": "2.0",
+    "method": "ec.messages.change",
```
igrigorik: Should be `ect.messages.change`?

docs/specification/embedded-cart.md
igrigorik: Echoing @lemonmade: `ect.line_items.change_request` is defined in `embedded.openrpc.json` but is not documented. Also: `ect.context.change` is defined in the OpenRPC but missing from the spec.

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,702 @@
...
+
+The `ect.auth` message is a request, which means that host
+**MUST** respond to exchange the authorization.
+
+- **Direction:** host → Embedded Cart
+- **Type:** Response
+- **Result Payload:**
+    - `authorization` (string, **REQUIRED**): The requested authorization data,
+    can be in the form of an OAuth token, JWT, API keys, etc.
+    - `cart` (object, **REQUIRED**): An optional cart holding the last known state to the host.
```
igrigorik: Contradiction: `cart` is declared `(object, **REQUIRED**)` in the payload description, but the text says "An optional cart holding the last known state."

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,702 @@
...
+            ]
+        },
+        "capabilities": {...},
+        "payment_handlers": {...}
+    }
+    // ...other cart fields...
+}
+```
+
+### Loading an Embedded Checkout URL
```
igrigorik: Should be "Loading an Embedded Cart URL"?

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,702 @@
...
+
+Indicates completion of cart building process and buyer now is ready to be transitioned to
+the next stage of their purchase journey.
+
+This marks the completion of Embedded Cart. If `dev.ucp.shopping.checkout` is part of the negotiated
+capabilities during discovery, host **MAY** proceed to initiate a checkout session based on the
+completed cart.
+
+- **Direction:** Embedded Cart → host
+- **Type:** Request
```
igrigorik: `ect.complete` is documented as a **Request** (with `id: "transition_1"`), but in `embedded.openrpc.json` it has no `result` schema — which is the notification pattern.

For comparison, `ec.complete` in ECP is a notification. If ECaP's `ect.complete` should be a request (which makes sense for the cart→checkout transition handshake), the OpenRPC needs a `result` schema defining what the host returns (e.g., acknowledgment, checkout initiation data).

source/services/shopping/embedded.openrpc.json
```diff
@@ -51,6 +51,38 @@
...
+    {
+      "name": "ec.auth",
+      "summary": "Authorization exchange between host & business.",
+      "description": "Augments Embedded Checkout Protocol by doing auth exchanges per business requirement. Can be used by business to request initial authorization during handshake or refresh a previously exchanged authorization.",
+      "params": [
+        {
+          "name": "type",
+          "description": "The type of authorization business is requesting from the host.",
+          "type": "string",
+          "enum": ["api_key", "oauth"]
```
igrigorik: `["api_key", "oauth"]` here vs `["api_key", "oauth", "jwt"]` in `ect.auth` - intentional?

Should this be an enum or list of well-known? We typically avoid enums.

source/services/shopping/embedded.openrpc.json
```diff
@@ -51,6 +51,38 @@
...
+          "type": "object",
+          "description": "Auth response from host containing the requested authorization and optional checkout state.",
+          "required": ["authorization"],
+          "properties": {
+            "authorization": {
+              "type": "string",
+              "description": "Requested authorization. Some common examples include API key and OAuth token."
+            },
+            "cart": {
+              "$ref": "../../schemas/shopping/checkout.json",
```
igrigorik: Same question as earlier: why does the embedded checkout need to receive checkout state via the auth response when it already has its own state?

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,702 @@
...
+
+- `ect_version` (string, **REQUIRED**): The UCP version for this session
+    (format: `YYYY-MM-DD`). Must match the version from service discovery.
+- `ect_auth` (string, **OPTIONAL**): Authentication token in business-defined
+    format.
+- `ect_delegate` (string, **OPTIONAL**): Comma-delimited list of delegations
+    the host wants to handle. **MAY** be empty if no delegations are needed.
+    **SHOULD** be a subset of `config.delegate` from the embedded service binding.
+
+## Transport & Messaging
```
igrigorik, line 154: This is verbatim duplicate of checkout content? Let's extract it into a reusable reference that both can link to. I want to prevent any subtle bugs or deviations here.

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,702 @@
...
+| **Handshake**     | Embedded Cart -> Host   | Establish connection between host and Embedded Cart.                      | Request                | `ect.ready`                                                                               |
+| **Authentication**| Embedded Cart -> Host   | Communicate auth data exchanges between Embedded Cart and host.           | Request                | `ect.auth`                                                                                |
+| **Lifecycle**     | Embedded Cart -> Host   | Inform of cart state in Embedded Cart.                                    | Notification           | `ect.start`, `ect.complete`                                                               |
+| **State Change**  | Embedded Cart -> Host   | Inform of cart field changes.                                             | Notification           | `ect.line_items.change`, `ect.buyer.change`, `ect.messages.change`                        |
+
+### Handshake Messages
+
+#### `ect.ready`
```
igrigorik: Ditto, this is same handshake as ec.ready: can we extract this and just note the name difference?

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,702 @@
...
+
+When the host responds with an `upgrade` object, the Embedded Cart **MUST**
+discard any other information in the message, send a new `ect.ready` message
+over the upgraded communication channel, and wait for a new response. All
+subsequent messages **MUST** be sent only over the upgraded communication
+channel.
+
+### Authentication
+
+#### `ect.auth`
```
igrigorik: > 2. Reauth: Certain authentication methods (i.e. OAuth token) have strict expiration timestamps.
If a session lasted longer than the allowed duration, business can request for a refreshed
authorization to be provided by the host before the session continues

True but practically, is this.. a problem we need to solve? What is our expected TTL on these tokens?

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,702 @@
...
+                    "severity": "recoverable"
+                }
+            ]
+            // ...
+        }
+    }
+}
+```
+
+### Security for Web-Based Hosts
```
igrigorik, line 676: Ditto, let's not replicate this. Extract to shared resource.

docs/specification/embedded-checkout.md
igrigorik: Same question. Omit?

source/services/shopping/embedded.openrpc.json
```diff
@@ -51,6 +51,38 @@
...
+        "schema": {
+          "type": "object",
+          "description": "Auth response from host containing the requested authorization and optional checkout state.",
+          "required": ["authorization"],
+          "properties": {
+            "authorization": {
+              "type": "string",
+              "description": "Requested authorization. Some common examples include API key and OAuth token."
+            },
+            "cart": {
```
igrigorik: checkout?

igrigorik (changes requested):
Overall  👍 on the feature. A couple authoring and technical flags to jam through...

## Why are adding ec.auth / ect.auth?
I understand the reauth scenario (token expiry mid-session) but this feels unlikely and marginal. Why was this not an issue for checkout and problem for cart? I would separate this discussion and functionality out from this PR.

## Do we really need config.capabilities?
UCP already has a capability negotiation model: discovery (ucp.capabilities) → per-session intersection (response envelope) → per-message handshake. This PR adds a fourth layer — a capability registry nested inside the embedded transport config.

I think the per-response pattern is sufficient here. When a host creates a cart, it gets back `ucp.services[embedded]` in the response: it knows ECaP is available for **this cart**. When you're ready to transition to checkout, checkout response will tell you if embedded checkout is available — all of this works already. The response context disambiguates which capability the embedded transport applies to. 

I don't think we need another discovery mechanism here. More on this below...

## Complete semantics are underspecified

`ec.complete` in checkout means "order placed" — it's terminal. `ect.complete` in cart means "buyer wants to proceed to checkout", which is a transition signal. Same name, different semantics.  

Conceptually what we're modelling is closer to `ec.payment.credential_request`: buyer taps "pay now," business requests what it needs from the host, then proceeds. I think complete is workable but something to think about.

Should we model transition explicitly? As a napkin sketch...
```json
  {
    "method": "ec.cart.complete",
    "params": {
      "cart": { "id": "cart_123", ... },
      "transition": {   // or navigation, ..., whichever
        "checkout": {
          "url": "https://business.com/checkout/from-cart/cart_123",
          "delegate": ["payment.credential", "fulfillment.address_change"]
        }
      }
    }
  }
```

If `transition.checkout` is present: business is advertising that it offers embedded checkout at this URL with these delegations. In this contract, host is always in control. It can use the offered transition URL, append `ec_*` params, navigate current or new frame/webview, or ignore the offer and run create checkout via API call +  use native flow. Business offers, host chooses. 

This also allows host to control whether it wants to reuse current frame/webview vs. spin up new one -- for example, if it wants to control transition between states. It also eliminates the need for config.capabilities: the transition advertisement IS the per-session embedded checkout signal, delivered at exactly the moment it matters.

## There is now a lot of duplicated boilerplate across EC/ECT
Let's think how we can extract and refactor duplicated spec content and schema to ensure we have consistency across these protocols. 

To that end, why are redefining same and common mechanisms (e.g. auth parameters) with different names? As an example, `ec_auth` should be consistent and reusable primitive across all flows; `ec_auth` is the "embeded commerce auth token". URL params configure the embedded session; they don't care whether you're embedding a cart or a checkout. We shouldn't need to redefine these and create opportunities to deviate across capabilities. This simplifies the implementation dramatically as well. Same argument applies to `delegate`, `color_scheme`, etc.

Let's extract the common set of in a shared spec. Then checkout, cart, and whatever other embedded modes we invent, can build on top with just the unique bits they need.

## ECT... 😅 
Bikeshed: why "T", it's never explained?

In retrospect, maybe we should have namespaced (`ec.{capability}.{action}`), and should we consider that here and now? Today we have checkout and adding cart, next we'll invent orders (eo?). Instead of proliferating opaque prefixes: `ec.cart.{...}, ...`?

> Embedded Checkout Protocol (ECP) is a checkout-specific implementation of UCP's Embedded Protocol (EP) transport binding that enables a host to embed a business's checkout interface

This is a mess. Here's how I'd clean it up...

> Embedded **Commerce** Protocol (ECP) is a transport binding that enables a host to embed a business capability (e.g. cart, checkout), receive events as buyer interacts, and request and process delegated buyer actions.

In other words, "ECP" is the embedded protocol of UCP. It supports multiple capabilities, and provides shared primitives (auth, delegate, etc), on which capabilities scaffold their specialized request/response semantics.

---

I recognize that I'm pointing to a larger refactor, but *now* is the right time to do it as we go from n=1 to multiple embedded capabilities. Happy to help with the refactor.

source/services/shopping/embedded.openrpc.json
```diff
@@ -237,6 +269,217 @@
...
+                  "description": "Cart line items after passing host validations."
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+
+    {
```
gsmith85, line 456: Idea: Consolidate to a single `ect.change` method.

Right now, the host has to run a JSON diff to figure out what triggered the event for several of these methods (e.g. `ect.line_items.change`). If the contract requires an array of mutated JSON paths alongside the full cart snapshot, we keep the idempotency of the snapshot pattern but eliminate the diffing overhead for the host.

Further, we can continue to be specific about which properties `MUST` provide change notification now, while providing a flexible API that will avoid needing to add many `change*` notification RPCs going forward.

```
{
  "method": "ect.change",
  "params": {
    "changed_paths": [
      "$.line_items[1].quantity", 
      "$.totals"
    ],
    "cart": { ... full snapshot ... }
  }
}
```

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,702 @@
...
+
+1. Optionally complete authentication mechanisms (i.e. identity linking)
+   if required by the business
+
+To initiate the session, the host **MUST** augment the `continue_url` with ECaP
+query parameters using the `ect_` prefix.
+
+All ECaP parameters are passed via URL query string, not HTTP headers, to ensure
+maximum compatibility across different embedding environments. Parameters use
+the `ect_` prefix to avoid namespace pollution and clearly distinguish ECaP
```
gsmith85: Per this observation from @igrigorik: https://github.com/Universal-Commerce-Protocol/ucp/pull/244#pullrequestreview-3949100688 :

> In retrospect, maybe we should have namespaced (ec.{capability}.{action}), and should we consider that here and now? Today we have checkout and adding cart, next we'll invent orders (eo?). Instead of proliferating opaque prefixes: ec.cart.{...}, ...?

Even with the inconsistency it will introduce, I would take this opportunity to use more descriptive naming for the embedded APIs, rather than forward propagating usage of abbreviations.
1. Anecdotally, it is noticeably harder to parse this naming scheme in the documentation.
2. Abbreviated names will reduce readability in implementation code.
3. Specifically, `ect` is so visually similar to `ec` that it is begging for transposition typos 🙂.

source/services/shopping/embedded.openrpc.json
gsmith85: An example of this: https://github.com/Universal-Commerce-Protocol/ucp/commit/040684cbed60031fbeb4b6d89c926398db82612f

This is a reminder to add a blurb about avoiding enums, for forward compatibility in `schema-authoring.md`, I can create a PR.

source/services/shopping/embedded.openrpc.json
```diff
@@ -237,6 +269,217 @@
           }
         }
       }
+    },
+
+    {
+      "name": "ect.ready",
+      "summary": "Handshake from business to host",
+      "description": "Initiates the Embedded Cart Protocol. Business declares which delegation it supports, host responds with optional channel upgrade and initial cart state.",
```
gsmith85: The current initialization flow requires two round trips for initialization.  Is there a reason we can't support proactive auth and reduce it to one?

If the business includes a `requested_auth type` the Host can optionally return the token in the `readyResult`. This reduces UI flicker and bridge latency, especially on mobile where native-to-web serialization is expensive.

If we want to maintain symmetry this could be added to `ec.ready` as well?

sinhanurag (changes requested):
@jingyli there are some good open comments here needing resolution. Can you please address these ASAP and then I will do another pass.

docs/specification/embedded-cart.md
jingyli: My thoughts:
- `checkout` is lower-funnel with stronger user intent to proceed with the purchase vs. `cart` has that potential basket building exercise involved (and as a buyer, I may not commit immediately with my cart).
- If the authorization method is OAuth access token (a by-product of identity linking), these tokens can expire quite quickly per business configuration.
- Even if 1% of user falls in the world where they are in the middle of cart building (or even in the middle of transitioning to checkout) and their access token expires, it can be quite a bad experience without proper reauth mechanisms (with our current state, the session would not be able continue unless an out-of-band token refresh mechanism is initiated).

I'd like for us to design with this edge case in scope, but feel free to flag if you have any other thoughts/concerns in mind!

docs/specification/embedded-checkout.md
jingyli: Originally I was thinking to make this optional to maintain consistency with all other request messages support  in embedded. However, thinking this over, omitting it entirely makes more sense since this is business requesting extra data on top of the last known state.

docs/specification/embedded-checkout.md
jingyli: The actual allowed values are specified in the schema: `oauth`, `api_key`.

There are also some feedback around open-string values, of which I'm leaning towards making these values as `examples` but allow for further extensions.

docs/specification/embedded-checkout.md
jingyli: Good suggestion, updated to `credential`!

docs/specification/embedded-cart.md
jingyli: We decided to remove `ect.line_items.change_request` from the scope of this PR - the definition in `embedded.openrpc.json` was a mistake from a previous rebase operation so have removed it from the schema.

We will follow-up with another PR if the use case come up on item selection delegation.

jingyli: Made 2 updates:
1) Removed `ect.line_items.change_request` due to a rebase error, it should be removed from this PR scope.
2) Removed `ect.context.change` in `embedded-cart.md` (I don't think we have a strong enough use case for this state message change in the embedded scenario - need to give some more thoughts on how would embedded cart/checkout get the signal inputs from buyer).

docs/specification/embedded-cart.md
jingyli: Thanks for catching the typo! Combined with feedback above, we will remove this entire object from the response payload.

source/services/shopping/embedded.openrpc.json
jingyli: This is a good callout, originally I was thinking that providing an exhaustive list of auth types that we can prescribe here would make it simpler on platform support.

But following principles set throughout the protocol, it's probably better to keep this as an open string, have explicit examples based on the these 3 types, and allow platforms to throw an error if an unrecognized/unsupported type is requested by embedded cart/checkout.

docs/specification/embedded-checkout.md
jingyli: Added an example for error message to represent host-side failures when responding to the request.

> what an Embedded Checkout should do with failures that come up?

In my mind, it needs to be 2 fold:
- Embedded Checkout/Cart **may** retry sending this request to the host again if the error is something retry-able (this is already mentioned in the guideline here).
- More holistically, I feel we should have some mechanism for Embedded Checkout/Cart to communicate errors back to the host when they are based on the result payload returned by the host. Right now we allow hosts to respond back with an error but not the other way around <- would love to think about this a bit more and follow-up with another PR if needed.

source/services/shopping/embedded.openrpc.json
jingyli: Thanks! I think this is an interesting idea, but I'd like to leave it out of the main scope of this PR so we keep this one focused on core changes needed to introduce `Embedded Cart`.

Also specifically on this idea, some initial reactions:
- `what triggered the event for several of these method` -> naive QQ: but wouldn't this be clear from the method name construction, where the `domain` part of `ec.{domain}.change` should indicate the expectation of the event source?
- I'm also unsure if the host really needs to run a JSON diff. In my opinion, these state change messages act like the response of some `Update` operations applied on the capability (they just happen to occur within the iframe). From host's perspective, this new `cart` object is now the latest known/agreed state between buyer & business, so similar to the REST transport world, this should be treated as a full replacement.

gsmith85: I definitely hear you on keeping the scope tight!  My only concern with sticking to a full replacement model is that it might create a bit of an observability gap for the host as we scale.

In a REST world, full replacement is fine because the goal is just state convergence.  Our use case is distinct: this mechanism is used to build reactive UIs across a bridge. In a client/host world, the host needs to know the intent of the change to trigger the right UI animations, native toasts, or agentic logic without page flickering.

This risk of omitting this affordance is when you verge into larger carts. Forcing the host to diff a 10KB+ JSON object to find one quantity change turns a simple notification into a heavy computation task. On less performant hardware, moving from an $O(N)$ brute-force search to a targeted $O(1)$ update via changed_paths is the difference between a UI that feels slow vs. one that feels native and responsive.

I agree we can add this later, but we have a unique window early with the protocol to make UCP performant by default. If we need to do this as a follow-up to keep this PR moving, I'm happy to take a stab at the proposal and partner with you to get it across the line!

source/services/shopping/embedded.openrpc.json
jingyli: The mentioned design pattern was considered (and even initially proposed as part of https://github.com/Universal-Commerce-Protocol/ucp/pull/109/changes#diff-78ca159ac692acb0be12b552ed811e6e47c61726fcfceed4681b49b7cf4d5203). I ultimately decided against it for 2 reasons:
1) I find it difficult to prescribe the error handling flows when host responds with an error due to auth related operations (Will the host reject the entire `xx.ready` request - i.e. not even return channel upgrade? Will we ask embedded cart/checkout to issue another `xx.ready` request?).
2) Handling of authorization data is no longer consolidated in one postMessage, so if it ever needs to be expanded/tweaked in the future, we would need to update multiple places and ensure they are kept in sync (imo this would be a maintenance burden from both protocol growth & implementation angles).

docs/specification/embedded-cart.md
lemonmade: I think this concern is still valid. What if we made the prefixes `ep_` (for Embed Protocol), with the understanding that we will apply the same to Checkout in a future breaking change, and to any future embeddable capabilities?

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,705 @@
...
+    "params": {
+        "delegate": []
+    }
+}
+```
+
+The `ect.ready` message is a request, which means that the host **MUST** respond
+to complete the handshake.
+
+- **Direction:** host → Embedded Cart
```
lemonmade:
```suggestion
- **Direction:** Host → Embedded Cart
```

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,705 @@
...
+
+### Authentication
+
+#### `ect.auth`
+
+Embedded cart **MAY** request authorization from the host in the following scenarios:
+
+1. **Initial handshake**: When `ect_auth` URL param is neither sufficient nor applicable due
+to additional considerations, business can request for authorization to be exchanged
+through this mechanism before the session starts.
```
lemonmade: This part seems like it could fit well into the `ect.ready` message instead, especially since decoupling them means the Embedded Cart will need to handle some weird initial loading states where the handshake is complete but authentication is invalid. Did you consider moving the "initial auth is no good" handling to that first request instead?

source/services/shopping/embedded.openrpc.json
```diff
@@ -255,28 +272,160 @@
...
+      "params": [
+        {
+          "name": "type",
+          "description": "The type of authorization data business is requesting from the host.",
+          "type": "string",
+          "examples": [
+            "oauth",
+            "api_key",
+            "jwt"
+          ]
```
lemonmade: Is this an exhaustive list that covers every platform/ business? Changing this will require a breaking API change, so I am a little nervous about making this a strict enum.

docs/specification/embedded-checkout.md
lemonmade: Ah, I left a comment about this elsewhere. This definitely felt like it would be hard to make into an exhaustive enum, so I would also favour an open string with examples 👍

lemonmade (approved):
Just a few nits left from me. Great job, and thank you for trimming this back to a smaller core (it was a lot easier to review this time around).

docs/specification/embedded-cart.md
jingyli: Per https://github.com/Universal-Commerce-Protocol/ucp/pull/244#issuecomment-4164956705, this is actually a typo and will be corrected back to the notification type.

docs/specification/embedded-cart.md
jingyli: Thanks both for the feedback! Per https://github.com/Universal-Commerce-Protocol/ucp/pull/244#issuecomment-4164956705, I'm updating all usages of `ect` to `ep.cart`, with a future vision that we will draft a breaking change to update `ec` on Embedded Checkout side to `ep.checkout`.

docs/specification/embedded-cart.md
jingyli: Some more related discussion here - https://github.com/Universal-Commerce-Protocol/ucp/pull/244#discussion_r3017438479 (tl;dr - I did consider both design patterns 😃).

> some weird initial loading states where the handshake is complete

Even today, if there are multiple roundtrips involved (i.e. communication channel upgrade), depending on latency, it's possible this `loading state` will be visible to buyers. Acknowledge that it will now be amplified given we have this auth message, but I don't think it's a net new problem we are introducing?

source/services/shopping/embedded.openrpc.json
jingyli: Nope, given the other feedback, I have updated this as some common examples and allowing hosts to reject request on authorization types they cannot handle.

docs/specification/embedded-cart.md
jingyli: Per https://github.com/Universal-Commerce-Protocol/ucp/pull/244#issuecomment-4164956705, I'd like to follow-up with this doc refactoring in a separate RP so we can really overhaul the common pieces and think about how the documentation should be shown in our current nav bar.

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,704 @@
...
+
+When a host receives a cart response with a `continue_url` from a business
+that advertises ECaP support, it **MAY** initiate an ECaP session by loading the
+URL in an embedded context.
+
+**Example:**
+
+```text
+https://example.com/cart/cart123?ect_version=2026-01-23...
+```
```
igrigorik, line 125: Introducing `ect_version` diverges from what we have in EP checkout: `ec_version` is defined as "The UCP version for this session... must match the version from the checkout response" — where the checkout response carries a single `ucp.version` that pins both. 

Technicaly, you need _both_ the UCP EP protocol version *and* the capability version. Practically today they're pinned together, but _can_ be decoupled in the future. Given that, I think the thing we need to here is negotiate `ep_version` which (equivalent of ect_version, which we can fold into it), and can defer capability version until later.

source/services/shopping/embedded.openrpc.json
```diff
@@ -237,6 +269,162 @@
...
+          }
+        }
+      }
+    },
+
+    {
+      "name": "ep.cart.start",
+      "summary": "Cart visible to buyer",
+      "description": "Business notifies host that checkout UI is visible and ready for interaction.",
+      "params": [
```
igrigorik, line 428: Should be "cart UI".

source/services/shopping/embedded.openrpc.json
```diff
@@ -51,6 +51,38 @@
...
+    {
+      "name": "ec.auth",
+      "summary": "Authorization exchange between host & business.",
+      "description": "Augments Embedded Checkout Protocol by doing auth exchanges per business requirement. Can be used by business to request initial authorization during handshake or refresh a previously exchanged authorization.",
+      "params": [
+        {
+          "name": "type",
+          "description": "The type of authorization data business is requesting from the host.",
+          "type": "string",
+          "examples": [
```
igrigorik: Structural bug here and in `ep.cart.auth` below (~line 326): the `type` param has `"type": "string"` at the parameter level, but OpenRPC requires type info inside a `schema` object. Every other param in this file correctly uses `"schema": { "type": "..." }`. Validators and codegen will silently ignore the type constraint as-is.

Should be:
```json
{
  "name": "type",
  "description": "...",
  "schema": {
    "type": "string",
    "examples": ["oauth", "api_key", "jwt"]
  }
}
```
`ec.auth` is pre-existing but worth fixing in the same pass since `ep.cart.auth` copies it.

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,704 @@
...
+    "error": {
+        "code": "internal_error",
+        "message": "Something went wrong when fetching the required authorization data."
+    }
+}
+```
+
+If the ingestion of the authorization is not successful, Embedded Cart **MAY**
+re-initiate this request with the host again.
+
```
igrigorik, line 454: The failure path is underspecified. "MAY re-initiate" implies retry is meaningful, but auth failure is not a transient condition — the host either has the credentials or it doesn't? 

What happens when auth is not accepted, what's the error recovery or fallback? Do we expect host to tear down the iframe, use continueUrl, etc?

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,704 @@
...
+            "totals": [/* ... */],
+            "line_items": [/* ... */],
+            "buyer": {/* ... */},
+        }
+    }
+}
+```
+
+#### `ep.cart.complete`
+
```
igrigorik, line 487: @jingyli I don't follow concern over fulfillment/payment, they're optional fields negotiated later. So the business can create the checkout from the terminal cart state at complete time, no blocker there.

I think we need to model `transition` now, not defer it. Multiple reasons:

1. **Per-session embedded checkout signal** — the host has no structured way to know if embedded checkout is available for *this specific cart*. `transition.checkout` delivers that signal at exactly the right moment — the same reason we just removed `config.capabilities`.
2. **Delegation hint** — without `delegate` in the payload, host either makes an extra checkout create API call to read `config.delegate`, or offers delegations blindly.
3. **Semantic clarity** — `continue_url` on the cart object means cart handoff/recovery; at complete time the meaning shifts. `transition.checkout.url` is unambiguous.

Rough sketch from before:
```json
{
  "method": "ep.cart.complete",
  "params": {
    "cart": { "..." },
    "transition": {
      "checkout": {
        "url": "https://business.com/checkout/from-cart/cart_123",
        "delegate": ["payment.credential", "fulfillment.address_change"]
      }
    }
  }
}
```

docs/specification/embedded-cart.md
igrigorik: Ok, non-blocking, but let's please scrub and make sure there are no deviations.

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,704 @@
...
+**Success Response:**
+
+```json
+{ "jsonrpc": "2.0", "id": "...", "result": {...} }
+```
+
+**Error Response:**
+
+```json
+{ "jsonrpc": "2.0", "id": "...", "error": {...} }
```
igrigorik, line 196: There is a separate PR to update error shapes, which will need to update this.

docs/specification/embedded-cart.md
igrigorik: These are very expensive roundtrips (platform to remote service), I agree with Chris that we should be trying to negotiate auth in the handshake not after. For re-auth, OK, we need a distinct message, but this should be AND not OR.

Same feedback as https://github.com/Universal-Commerce-Protocol/ucp/pull/244/changes#r2938657812

source/services/shopping/embedded.openrpc.json
igrigorik: > I find it difficult to prescribe the error handling flows when host responds with an error due to auth related operations (Will the host reject the entire xx.ready request - i.e. not even return channel upgrade?

Exactly that. The precondition failed, upgrade not allowed. 

I think re-auth is more of a hypothetical than real concern at this point. My proposal is to solve the init auth and defer re-auth as followup.

igrigorik (changes requested):
@jingyli ty, great progress on this! There are still few blockers that I'd like to address before we land this, see comments above.

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,704 @@
...
+            },
+            {
+                "version": "{{ ucp_version }}",
+                "transport": "mcp",
+                "schema": "https://ucp.dev/services/shopping/mcp.openrpc.json",
+                "endpoint": "https://merchant.example.com/ucp/mcp"
+            },
+            {
+                "version": "{{ ucp_version }}",
+                "transport": "embedded",
```
westeezy, line 71: With the inclusion of this change, a host would see an embedded transport in discovery but has no way to know whether the business supports embedded cart, embedded checkout, or both. Is it intentional that the the host only learns this from per-response config after creating a cart or checkout?

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,704 @@
...
+### Loading an Embedded Cart URL
+
+When a host receives a cart response with a `continue_url` from a business
+that advertises ECaP support, it **MAY** initiate an ECaP session by loading the
+URL in an embedded context.
+
+**Example:**
+
+```text
+https://example.com/cart/cart123?ect_version=2026-01-23...
```
westeezy: Trying to follow along the renaming that already occurred but should this now be: `ep_cart_version`?

source/services/shopping/embedded.openrpc.json
jingyli: > In a client/host world, the host needs to know the intent of the change to trigger the right UI animations, native toasts, or agentic logic without page flickering.

I'm having some trouble processing this statement here - In EP, the embedded cart/checkout actually controls all UI rendering (with the exception of delegation use cases - but that's out-of-scope for this discussion here). These state change messages (`xx.{domain}.change`) are purely informational and would only get triggered after the actual UI update occurs (i.e. host would receive `ep.cart.line_item.change` **after** the new item is added & reflected in the embedded cart UI). Relevant embedded checkout language can be found here: https://ucp.dev/draft/specification/embedded-checkout/#state-change-messages

Therefore I still fail to see why host needs to run any diffs to discover deltas vs. processing the entire object as-is..

docs/specification/embedded-cart.md
jingyli: Good point - I revisited the language in the parameter and agree that `ep_version` makes more sense here. I also updated some surrounding language here since it's not always the case `ec_cart` will be the applied prefix here.

docs/specification/embedded-cart.md
jingyli: Noted, resolving for now to help me track remaining actionable open threads, thanks!

source/services/shopping/embedded.openrpc.json
jingyli: I fixed both - both `ec.auth` and `ep.cart.auth` are being introduced as part of this PR. I left the `ec` namespace to maintain consistency and we can clean it up together with other methods.

docs/specification/embedded-cart.md
jingyli: > concern over fulfillment/payment

To clarify what I mean here: It's possible that in order to create a checkout session, business may require fulfillment or payment info (maybe unlikely, but nothing is stopping this scenario). In the REST/MCP world, they can reject this `create_checkout` request by returning a business error with no resource attached. However, with the `transition` model we have here, if they fall in this bucket, then they would never be able to find value in using `transition` directly.

In my mind, `ep.cart.complete` should be the terminal notification for cart (this maintains symmetry with ECP shape and also gives host/platform an opportunity to run any processes that need to take place prior to checkout creations). If cart to checkout transition occurs, then platform should make a `create_checkout` API call with business again using the `cart_id` (and this is where the per-session embedded checkout signal & available delegates can be received). <- I will expand the language here a bit to clarify this ambiguity.

docs/specification/embedded-cart.md
jingyli: Updated to `ep_version` here.

docs/specification/embedded-cart.md
jingyli: > Is it intentional that the the host only learns this from per-response config after creating a cart or checkout?

Yes! The way we designed EP means the per-response `ucp` object will be able to offer the availability of ECP or ECaP at a per-session level. It's pointed out in https://github.com/Universal-Commerce-Protocol/ucp/pull/244#pullrequestreview-3949100688's `Do we really need config.capabilities?` section.

docs/specification/embedded-cart.md
jingyli: Scrubbed common contents (dd02dfa) - seems like there is a minor typo in ECP side? There was a reference to a non-existent `checkout_url` in the sentence "...they **MUST** validate the origin matches the `checkout_url` used to start the checkout." (corrected to `continue_url`).

docs/specification/embedded-cart.md
westeezy: Whoops I should have read everything first 😆 ty!

docs/specification/embedded-cart.md
jingyli: Ack, depending on the merge sequence, will keep this thread open for tracking in case action is needed on this PR.

docs/specification/embedded-cart.md
jingyli: Added some more contents using https://github.com/Universal-Commerce-Protocol/ucp/pull/325 as a reference.

Specifically I think allowing `error_response` to be returned by embedded cart/checkout's `xx.messages.change` notification to indicate any unrecoverable errors (on top of just host-initiated delegation results) throughout the session may be useful (i.e. what happens when all items goes OOS while the buyer is in the embedded context?).

docs/specification/embedded-checkout.md
jingyli: This is now addressed per https://github.com/Universal-Commerce-Protocol/ucp/pull/244#discussion_r3025631266.

source/services/shopping/embedded.openrpc.json
jingyli: Ack on all the concerns, have made this update in https://github.com/Universal-Commerce-Protocol/ucp/pull/244/commits/073a04b3af27992b275be68e15cefb22026e3d8f

I still would like to keep re-auth in scope per https://github.com/Universal-Commerce-Protocol/ucp/pull/244#discussion_r3013014621.

jingyli:
@igrigorik @westeezy @gsmith85 Thanks all for the feedback, the comments should all be addressed now so hoping both can take another pass at the increments! @sinhanurag also if you can help to review this PR.

@lemonmade FYI that some of your concerns before your approval are now addressed as well.

source/services/shopping/embedded.openrpc.json
```diff
@@ -276,7 +313,183 @@
...
+        "examples": [
+          "oauth",
+          "api_key",
+          "jwt"
+        ]
+      },
+      "auth": {
+        "type": "object",
+        "properties": {
+          "type": "#/components/schemas/type"
```
igrigorik: Bug.. 

```suggestion
          "type": { "$ref": "#/components/schemas/type" }
```

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,691 @@
...
+    "id": "auth_1",
+    "result": {
+        "credential": "fake_identity_linking_oauth_token"
+    }
+}
+```
+
+**Example Error Message:**
+
+```json
```
igrigorik, line 405: With #325, I believe should this be... cc @westeezy 

```
  {
      "jsonrpc": "2.0",
      "id": "auth_1",
      "result": {
          "ucp": { "version": "{{ ucp_version }}", "status": "error" },
          "messages": [
              {
                  "type": "error",
                  "code": "timeout_error",
                  "content": "An internal service timed out when fetching the required authorization data.",
                  "severity": "recoverable"
              }
          ]
      }
  }
```

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,691 @@
...
+Embedded Cart **MAY** re-initiate this request with the host again. Otherwise, Embedded Cart
+**MUST** issue a state change message containing an `unrecoverable` error response. This response
+**SHOULD** also contain a `continue_url` to allow buyer handoff.
+
+**Example Error Response Message Through ep.cart.messages.change:**
+
+```json
+{
+    "jsonrpc": "2.0",
+    "method": "ep.cart.messages.change",
```
igrigorik: Is this right tool for the job?  cc @westeezy 

ep.cart.messages.change is a cart-state notification — designed for business-level cart messages (out-of-stock, invalid quantity, etc.). A session-level auth failure is categorically different, and the params shape confirms it: the current example sends ucp + messages + continue_url flat at the params level, which doesn't match the OpenRPC definition that expects a cart object.

Do we need a dedicated `ep.cart.error` notification? If we do, then current shape fits / works.

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,691 @@
...
+    }
+}
+```
+
+The `ep.cart.ready` message is a request, which means that the host **MUST** respond
+to complete the handshake.
+
+- **Direction:** Host → Embedded Cart
+- **Type:** Response
+- **Result Payload:**
```
igrigorik, line 304: Can we inline the requirements to consolidate?

```
      - `upgrade` (object, **OPTIONAL**): An object describing how the Embedded
          Cart should update the communication channel it uses to communicate
          with the host. When present, host **SHOULD NOT** include `credential`
          — the channel will be re-established and any credential sent here
          will be discarded.
      - `credential` (string, **CONDITIONAL**): The requested authorization data,
          in the form of an OAuth token, JWT, API key, etc.
          **MUST** be set if `auth` was present in the request and `upgrade` is
          absent. **MUST NOT** be set if `upgrade` is present.
```

With that we can trim 328/329.

Separately, I think it's worth adding an implementation note that hosts should prefer ep_cart_auth URL param to provide the initial token — this cuts a full RTT of latency in the (critical) bootstrap process. The ep.cart.ready auth mechanism should be defined as the fallback for cases where the URL token is insufficient or unavailable, not the default path.

docs/specification/embedded-cart.md
```diff
@@ -0,0 +1,691 @@
...
+ECaP query parameters.
+
+All ECaP parameters are passed via URL query string, not HTTP headers, to ensure
+maximum compatibility across different embedding environments. Parameters **SHOULD**
+use the `ep_cart` prefix to avoid namespace pollution and clearly distinguish ECaP
+parameters from business-specific query parameters:
+
+- `ep_version` (string, **REQUIRED**): The UCP version for this session
+    (format: `YYYY-MM-DD`). Must match the version from service discovery.
+- `ep_cart_auth` (string, **OPTIONAL**): Authentication token in business-defined
```
igrigorik: `ep_cart_auth` and `ep_cart_color_scheme` have the same shape problem as (now fixed, ty!) `ep_cart_version`: all three are capability-scoped names carrying session-level concerns that are shared across all EP capabilities. Let's complete the arc...

- `ep_version` — UCP protocol version applies equally to cart and checkout; one version negotiates both today, capability-scoped versions are additive if they ever diverge.
- `ep_auth` — the identity-linking credential is scoped to the user+merchant session, not the capability. Cart and checkout within the same flow share the same token. Checkout calls it `ec_auth`, cart calls it `ep_cart_auth`; both should be `ep_auth`.
- `ep_color_scheme` — display preference is capability-agnostic by definition. The shared `embedded_config.json` schema already models this as a capability-agnostic property; its description just needs to stop hardcoding `ec_color_scheme`.

Tactically, suggestions:
 - ep_cart_auth -> ep_auth -- we'll converge checkout to this later too.
 - ep_cart_color_scheme -> ep_color_scheme -- ditto.

docs/specification/embedded-cart.md
igrigorik: The fulfillment/payment concern doesn't hold up: `ep.cart.complete` is fired by the business. They control when to send it / when to allow buyer to hit the complete button that emits this message. If the cart can't transition to checkout due to required missing info, the business simply must not allow and must not fire `ep.cart.complete` yet -- if they do, that's a logic error on their end.

The core value of `transition` is eliminating the need for an explicit `create_checkout` call just to discover whether embedded checkout is available and with what delegations. Without it, the host is forced into a round trip whose sole purpose is capability discovery, and the purpose of `transition.checkout` is to deliver that signal and eliminate that (very costly) RTT. Remember, it's not just the network RTT but also waiting for the full checkout negotiation to execute, just to get back signal that you can (re)load the same checkout (and rerun the negotiation in the process once over!) in embedded mode -- two checkout negotiations, in addition to two RTTs instead of one.

`transition` is optional in that not all businesses support embedded checkout — those that don't simply omit it. But for those that do, `ep.cart.complete` is the right moment to deliver that signal, not force a full checkout negotiation round trip just to discover availability.

igrigorik (changes requested):
@jingyli we're close! Another batch..

docs/specification/embedded-cart.md
jingyli: Inlined 328/329 into the payload parameter description here.

On the implementation note: The exact content is present in the payload parameter description for `auth` in the request above. I don't feel we need to repeat it again separately.

docs/specification/embedded-cart.md
jingyli: > the current example sends ucp + messages + continue_url flat at the params level, which doesn't match the OpenRPC definition that expects a cart object.

Getting the "easier" feedback out of the way first 😃: that's also why `embedded.openrpc.json` was updated for `ep.cart.messages.change` (it will return either a `cart` object OR an `error_response` object). This is consistent with our MCP OpenRPC modelling approach.

> ep.cart.messages.change is a cart-state notification — designed for business-level cart messages (out-of-stock, invalid quantity, etc.). A session-level auth failure is categorically different

I think this is the more fundamental question we should answer and in my mind, I'm aligned that it feels categorically different now that it's explicitly pointed out. Introducing `ep.cart.error` message makes sense. Though I think the change to `embedded.openrpc.json` to allow`ep.cart.messages.change` also return `error_response` should also be kept - this should bridge gaps for errors like `out_of_stock` where the resource may no longer exist on business side (following the principle of transport agnostic).

docs/specification/embedded-cart.md
jingyli: >  They control when to send it / when to allow buyer to hit the complete button that emits this message. If the cart can't transition to checkout due to required missing info, the business simply must not allow and must not fire ep.cart.complete yet -- if they do, that's a logic error on their end.

I'm having some trouble processing this particular point.. I wonder if withholding `ep.cart.complete` for this purpose is correct - this feels like it's blurring the lines on why UCP decided to define separate capabilities for cart & checkout. By design, cart holds a subset of signals available for checkout, where fulfillment (at least for now) & payment are not part of what it contains.

I also find it strange that in every other transport combination (`cart-only`, `checkout-only`, `REST/MCP cart -> REST/MCP checkout`, `REST/MCP cart -> embedded checkout`, `embedded cart -> REST/MCP checkout`), we do not enforce this requirement but `embedded cart -> embedded checkout`: a business continues to have the flexibility to require additional unique to checkout signals (like shipping address, payment metadata) during `create_checkout` and platform can either supply the relevant data in the request or if it's missing, business can notify the platform through `messages[]`.

Though do want to call out: I'm definitely not saying there is no value for `transition` (100% ack the RTT elimination!), I just want to think it through to understand what does it truly mean for platforms & businesses.

igrigorik: > I wonder if withholding ep.cart.complete for this purpose is correct.

The business controls the buyer interaction that triggers `ep.cart.complete` — concretely, the "Checkout" button or equivalent action. If business rules require a shipping address or payment info before checkout can be entered, the business must not enable that interaction until those conditions are met. Firing `ep.cart.complete` on a cart that can't become a checkout would be a business logic error. The buyer clicked "Checkout" but checkout creation fails. The implicit contract of `ep.cart.complete` is that the business is ready to negotiate a checkout successfully.

Also, I'm not suggesting we force the host into embedded checkout. When the host receives `ep.cart.complete`, it has the cart and can choose its own adventure: save for later, call `create_checkout` over REST/MCP, initiate embedded checkout session.

What I'm trying to avoid is a **requirement** to call `create_checkout` over REST/MCP just to answer one question: *is embedded checkout available for this cart?* If the answer is yes, the host then has to load the embedded checkout and renegotiate, which is two network round trips and two full checkout negotiations. In practice, for a buyer clicking "Checkout" and expecting an instant transition, that's potentially multiple seconds of added (unnecessary) latency.

`transition.checkout` (or whatever we want to call it) answers that question inline as part of complete notification. If the business does not support embedded mode for checkout, they omit it and host falls back to other mechanisms. Instant and fast transition will be the *dominant* flow.. I want to make sure we don't force unnecessary latency tax on it.

igrigorik (changes requested):
@jingyli ty, so close! 

1) Pushed a couple of minor fixes, please sanity check.
2) Main outstanding thing to resolve is the transition / signaling for efficient embedded cart to checkout flow, see my followup reply in that thread.

With #325 as parallel PR, there is a fair amount of surgery that we'll have to do on this PR. We can either land 325 and update here, or land this and then have 325 do a full sweep. For reference..

**OpenRPC schema fixes:**
- `ec.auth`, `ep.cart.auth`, `ep.cart.ready` results — add `oneOf` with correct `embedded_error_response.json` and `ucp.status: "success"` success branch
- `ep.cart.messages.change` / `line_items.change` / `buyer.change` — align notification error branch to `embedded_error_response.json` (or revert to plain `$ref`)
- `ec.error` / `ep.cart.error` params — swap `message.json` items for `embedded_message_error.json`
- Add `timeout_error` to `embedded_error_code.json`

**Spec prose fixes:**
- Response Handling section in both specs — replace `"error": {...}` 
- Auth response prose in both specs — "respond with either an error" → `result`-based language, document both success and error payload shapes

igrigorik (approved):
@jingyli ty for the thorough work and patience to land this one! 🙇🏻 

Per TC review, approving 👍 and we'll split (a) error shape refactor (#325 WIP) and (b) fast path ([this thread](https://github.com/Universal-Commerce-Protocol/ucp/pull/244/changes#r3022718182)) into followup PRs.

sinhanurag (approved):

PR #246, opened 2026-03-09
feat: extend discount capability to cart

igrigorik:
We landed discount capability before cart and forgot to update discount capability. This PR fixes that:

  - Add dev.ucp.shopping.cart extension in discount schema (create/update)
  - Update discovery example to show cart + checkout extends
  - Add cart-to-checkout continuity requirement (MUST carry forward codes)
  - Add cart discount example to specification docs

Backwards compatible. It is up to the business whether they extend and support discounts on cart by using built-in `extends` mechanism on the capability, this change wires up the right schema shapes.

---

## Checklist
- [x] Documentation update

### DIFF
--- a/docs/specification/discount.md
+++ b/docs/specification/discount.md
@@ -41,14 +41,14 @@ extend cart, checkout, or both:
 ```json
 {
   "ucp": {
-    "version": "2026-01-11",
+    "version": "{{ ucp_version }}",
     "capabilities": {
       "dev.ucp.shopping.discount": [
         {
-          "version": "2026-01-11",
+          "version": "{{ ucp_version }}",
           "extends": ["dev.ucp.shopping.cart", "dev.ucp.shopping.checkout"],
-          "spec": "https://ucp.dev/specification/discount",
-          "schema": "https://ucp.dev/schemas/shopping/discount.json"
+          "spec": "https://ucp.dev/{{ ucp_version }}/specification/discount",
+          "schema": "https://ucp.dev/{{ ucp_version }}/schemas/shopping/discount.json"
         }
       ]
     }
--- a/.cspell/custom-words.txt
+++ b/.cspell/custom-words.txt
@@ -70,6 +70,7 @@ inlinehilite
 lifecycles
 ligeras
 linenums
+linkignore
 llms
 llmstxt
 mastercard
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -110,9 +110,19 @@ jobs:
           echo "Set SITE_URL to $PAGES_URL"
           echo "SITE_URL=$PAGES_URL" >> $GITHUB_ENV
 
-      - name: Build Documentation (PR Check)
-        if: github.event_name == 'pull_request'
-        run: uv run mkdocs build --strict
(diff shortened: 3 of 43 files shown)


docs/specification/discount.md
```diff
@@ -130,6 +143,10 @@ Discount codes are submitted via standard checkout create/update operations.
 - Rejected codes communicated via `messages[]` (see below)
 - Discount amounts reflected in `totals[]` and `line_items[].discount`
 
+**Cart-to-checkout continuity:** When a cart is converted to a checkout via the
+cart capability's `cart_id` field, businesses MUST carry forward any discount
```
amithanda, line 141: This adds a sort of state-management requirement for implementations. Do we need to add explicit guidance on how to handle codes that might become invalid during that transition (e.g., a code that works for "Cart Exploration" but is restricted during "Checkout") e.g first time user discount code shown at cart time, which is invalid when the user signs in and merchant finds out that the user is an existing user?

docs/specification/discount.md
```diff
@@ -206,220 +223,290 @@ subtractive (e.g., "-$13.99").
 
 ## Examples
 
+### Cart with discount codes
+
+Discount codes applied during cart exploration. The cart response includes
```
amithanda, line 224: The schema uses the same applied_discount object for both Cart (estimated) and Checkout (final).Do we need something in the documentation to clarify that cart-level discounts are non-binding/provisional, whereas checkout-level discounts are typically locked in.

docs/specification/discount.md
```diff
@@ -31,18 +31,25 @@ shared between the platform and the business.
...
+Businesses advertise discount support in their profile. The capability can
+extend cart, checkout, or both:
 
 ```json
 {
   "ucp": {
     "version": "2026-01-11",
     "capabilities": {
       "dev.ucp.shopping.discount": [
+        {
```
jingyli, line 47: Naive fly-by QQ: Should we be using the [multi-parent `extends` approach](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/capability.json#L20) here instead of having these 2 parent capabilities listed separately?

Else I feel we should eliminate `type: array` pattern in the `extends` field entirely to keep definitions simpler and ensure consistency..

docs/specification/discount.md
igrigorik: @amithanda yep, updated in https://github.com/Universal-Commerce-Protocol/ucp/pull/246/changes/7abc6951c4f6dd7d09e1efd7edc63a7ebfe357cc.

docs/specification/discount.md
igrigorik: updated

docs/specification/discount.md
igrigorik: I think this contract is implicit due to nature of cart and mandatory transition to checkout. 

To your point, some discounts cannot be computed in cart stage due to missing facts about the buyer or transaction. The "fidelity" also depends on the implementation of discounting engine in cart vs checkout. There are differences across businesses how this is handled, and I don't think we can or should mandate specific (e.g. "estimate") behavior because some implementations (e.g. Shopify) run the same full-resolution discount engine logic in both places.

PR #250, opened 2026-03-10
feat: eligibility claims & verification contract

### LINKED ISSUE #137: [Feat]: Allow tender-based discounts via payment instrument qualifiers in Checkout Session Update

igrigorik:
Context: https://github.com/Universal-Commerce-Protocol/ucp/pull/214#issuecomment-4012461565. Supersedes #214, closes #137.

Introduce `context.eligibility` — buyer claims about eligible benefits (loyalty membership, payment instrument perks, etc.) that Businesses can act on across the shopping lifecycle.

Processing model:
  - Platform provides claims via context.eligibility on any request
  - Business MAY act on recognized claims (adjust pricing, product access, provisional discounts); MUST ignore unrecognized claims
  - At checkout completion, all claims that influenced the checkout MUST be resolved: verified against proof, or rescinded by Platform
  - Unresolved claims block completion (invalid_eligibility error)
  - Business MUST NOT mutate checkout on verification failure

Layering:
  - context.json: eligibility array with reverse-domain $ref validation
  - checkout.md: normative verification contract (core obligation)
  - discount.json: provisional + eligibility fields on applied_discount for structured attribution when extension is active
  - catalog/index.md: MAY adjust price/list_price for eligible claims
  - error_code.json: invalid_eligibility standard error

Eligibility lives on context (not PaymentInstrument) enabling full-funnel coverage from catalog through checkout. Verification is a core checkout concern; the discount extension adds attribution.

---

## Checklist

### DIFF
--- a/source/schemas/shopping/discount.json
+++ b/source/schemas/shopping/discount.json
@@ -64,7 +64,7 @@
         "provisional": {
           "type": "boolean",
           "default": false,
-          "description": "True if this discount requires additional verification. Platforms MUST communicate provisional status to the buyer."
+          "description": "True if this discount requires additional verification."
         },
         "eligibility": {
           "$ref": "types/reverse_domain_name.json",


docs/specification/checkout.md
```diff
@@ -249,6 +250,71 @@ messages or deferring to checkout completion.
...
+Platforms provide `context.eligibility` — buyer claims about eligible benefits
+such as loyalty membership, payment instrument perks, and similar. These are
+claims, not verified facts. Businesses **MAY** act on recognized claims during
+the session (adjusting pricing, granting product access, applying provisional
+discounts), but all claims that influenced the checkout **MUST** be resolved
+before the transaction can complete.
+
+Unrecognized or inapplicable claims **MUST NOT** block the checkout.
+Businesses **SHOULD** notify the buyer via `messages` with `type: "warning"`
+when a claim is not applied. At completion, unverified claims **MUST**
```
ACSchil:
```suggestion
when a claim is not applied. At completion, unverified claims that influenced the checkout **MUST**
```

Perhaps this is a nit, but unrecognized and inapplicable claims could also be considered "unverified." To me, what we are trying to communicate is that the completion should be rejected if a claim has altered the checkout state, but could not be verified.

docs/specification/checkout.md
```diff
@@ -249,6 +250,71 @@ messages or deferring to checkout completion.
...
+
+* **Verified**: The Business confirms the claim against a proof provided at
+  completion time. UCP does not prescribe how verification occurs — proof
+  may come from the payment credential, an identity verification capability,
+  or any other mechanism negotiated between Platform and Business.
+* **Rescinded**: The Platform removes the claim from `context.eligibility`
+  before completion (e.g., buyer changes payment method, withdraws a
+  membership claim). Once removed, the Business recalculates without it.
+
+Businesses **MUST NOT** complete a transaction with unresolved eligibility
```
ACSchil, line 287: Given: ["Unrecognized or inapplicable claims **MUST NOT** block the checkout."](https://github.com/Universal-Commerce-Protocol/ucp/pull/250/changes#diff-6c5ffde9bc6fd245f3e1313019b294f335d4fdb423207297690ed532c8ce6db7R263)

Should we say Platforms **MUST** rescind claims when the Business provides a [warning](https://github.com/Universal-Commerce-Protocol/ucp/pull/250/changes#diff-6c5ffde9bc6fd245f3e1313019b294f335d4fdb423207297690ed532c8ce6db7R263) of a claim that was not applied? Or do we need a third state, something like **inapplicable**?

~~Related to this, shall we add normative language for what a Platform MUST do when it receives an `eligibility_invalid`? I'm concerned we've put to Business and Platform in a place where the Buyer cannot proceed.~~ I see this is addressed under "When verification fails".

docs/specification/checkout.md
```diff
@@ -249,6 +250,71 @@ messages or deferring to checkout completion.
...
+  "id": "checkout_abc",
+  "status": "ready_for_complete",
+  "line_items": [ "..." ],
+  "totals": [ "..." ],
+  "messages": [
+    {
+      "type": "error",
+      "code": "eligibility_invalid",
+      "severity": "recoverable",
+      "content": "Payment credential does not match the claimed store card benefit."
```
ACSchil: I think we should include path to point to which claim produced the error, so the Platform can act on the correct claim if it decides to remove it. 

```suggestion
      "type": "error",
      "code": "eligibility_invalid",
      "severity": "recoverable",
      "content": "Payment credential does not match the claimed store card benefit.",
      "path": "$.context.eligibility[0]"
```

docs/specification/checkout.md
igrigorik: We could, but I don't think that's necessary. 

The partition here is 'accepted' vs 'not accepted': if the business does not accept a claim, it SHOULD provide feedback to the platform but otherwise the state of checkout (totals, products, etc) are all accurate and "completeable" — i.e. if platform calls complete_checkout with an instrument the business will accept and process because there are no provisional claims that it extended previously and still needs to verify. As such, no third state needed — unaccepted claims are inert.

ACSchil: Ah yes,

> there are no provisional claims that it extended previously and still needs to verify.

is the key. Thanks!

docs/specification/checkout.md
```diff
@@ -249,6 +250,72 @@ messages or deferring to checkout completion.
 Example: `out_of_stock` requires specific upfront UX, whereas
 `payment_required` can be handled generically at submission.
 
+#### Eligibility Verification at Completion
+
+Platforms provide `context.eligibility` — buyer claims about eligible benefits
+such as loyalty membership, payment instrument perks, and similar. These are
+claims, not verified facts. Businesses **MAY** act on recognized claims during
+the session (adjusting pricing, granting product access, applying provisional
+discounts), but all claims that influenced the checkout **MUST** be resolved
```
maximenajim: The contract hinges on “all claims that influenced the checkout MUST be resolved”, but “influenced” (and even “recognized”) isn’t defined precisely enough to guarantee interoperable behavior.
In practice, platforms need a deterministic way to know which specific context.eligibility[] entries the Business actually used (pricing, access gating, etc.) so they can remediate without guesswork. Right now the only guidance is “SHOULD notify… when a claim is not applied,” which doesn’t cover the applied case.

Suggestion: define “influenced” as “reflected in the Business response” and require the Business to explicitly echo applied/active claims (e.g., a new response field like eligibility_applied: [...] / eligibility_pending_verification: [...], or a standardized messages entry per accepted claim). For discount-related effects, discounts.applied[].eligibility covers this, but there’s no equivalent for non-discount effects like product access or catalog price/list_price adjustments.

docs/specification/checkout.md
```diff
@@ -249,6 +250,72 @@ messages or deferring to checkout completion.
...
+before the transaction can complete.
+
+Unrecognized or inapplicable claims **MUST NOT** block the checkout.
```
maximenajim, line 262: There’s a spec-level ambiguity that can lead to accidental checkout blocking: this section says “Unrecognized or inapplicable claims MUST NOT block the checkout,” but then says “At completion, unverified claims MUST result in … eligibility_invalid.”
We should explicitly state that only claims that the Business applied / relied on are subject to resolution/verification at completion, and that claims that were ignored (unrecognized or inapplicable) MUST NOT produce eligibility_invalid.
Possible wording tweak: “At completion, any applied eligibility claim that remains unverified MUST …” (or define a term like “active eligibility claim”).

docs/specification/checkout.md
```diff
@@ -249,6 +250,72 @@ messages or deferring to checkout completion.
...
+  before completion (e.g., buyer changes payment method, withdraws a
+  membership claim). Once removed, the Business recalculates without it.
+
+Businesses **MUST NOT** complete a transaction with unresolved eligibility
+claims. Unverified claims may result in incorrect pricing or unauthorized
+access to restricted products.
+
+**When verification fails:**
+
+The Business **MUST NOT** mutate the checkout and **MUST** return an error
```
maximenajim: “Business MUST NOT mutate the checkout” is underspecified. Adding the messages[] error itself is technically a mutation, and implementers will vary on whether they’re allowed to (a) change totals, (b) drop provisional discounts, (c) change status, etc.
Given this is a financially sensitive flow, we should be explicit about which parts must remain stable on verification failure: e.g., “MUST NOT change line_items, pricing/totals, discounts, fulfillment selection, or inventory commitments; MAY append messages; SHOULD keep status unchanged.”

docs/specification/catalog/index.md
```diff
@@ -69,6 +69,11 @@ the presentment currency differs, businesses SHOULD convert before applying
 (see [Price Filter](search.md#price-filter)). Response prices include
 explicit currency codes confirming the resolution.
 
+When `context.eligibility` claims are present, Businesses that accept them
+**MAY** adjust `price` / `list_price` directly for strikethrough display and
```
maximenajim, line 78: Maybe we should add explicit language that catalog price adjustments based on context.eligibility are provisional and subject to verification at checkout completion, and platforms SHOULD surface that to buyers. And maybe we add caching guidance: responses MAY vary by context.eligibility; intermediaries/platforms MUST avoid reusing eligibility-priced catalog responses for other users/sessions.

docs/specification/checkout.md
igrigorik: Clarified

docs/specification/checkout.md
igrigorik: Good catch, clarified.

docs/specification/checkout.md
igrigorik: Good observation. Tightened "influenced" to "accepted" in the latest revision — the partition is accepted vs. not-accepted, and Businesses SHOULD warn when a claim is not accepted.

The checkout state itself is the effect of accepted claims — pricing, availability, line items all reflect what the Business did with them. When a structured extension exists (like the discount extension's applied[].eligibility), it provides explicit attribution — I would keep this contract, instead of trying to model a generic `eligibility_applied`. For the general case, the Business MAY use messages with type: "info" to explain the effects of accepted claims — integrated into the spec text.

docs/specification/checkout.md
```diff
@@ -249,6 +250,74 @@ messages or deferring to checkout completion.
...
+Platforms provide `context.eligibility` — buyer claims about eligible benefits
+such as loyalty membership, payment instrument perks, and similar. These are
+claims, not verified facts. Businesses **MAY** act on recognized claims during
+the session (adjusting pricing, granting product access, applying provisional
+discounts), but all accepted claims **MUST** be resolved before the
+transaction can complete.
+
+Unrecognized or inapplicable claims **MUST NOT** block the checkout.
+Businesses **SHOULD** notify the buyer via `messages` with `type: "warning"`
+when a claim is not accepted, and **MAY** use `type: "info"` to explain
```
lemonmade, line 264: Should the warning message also use `code: eligibility_invalid`? And, should there be a suggested code for the `info` message described here?

source/schemas/shopping/discount.json
```diff
@@ -61,6 +61,15 @@
...
         },
+        "provisional": {
+          "type": "boolean",
+          "default": false,
+          "description": "True if this discount requires additional verification. Platforms MUST communicate provisional status to the buyer."
+        },
+        "eligibility": {
+          "$ref": "../ucp.json#/$defs/reverse_domain_name",
+          "description": "The eligibility claim accepted by the Business for this discount. Corresponds to a value from context.eligibility. Omitted for code-based and non-eligibility automatic discounts."
+        },
```
lemonmade, line 72: Something doesn’t quite sit right with me about this. I think it’s that we are providing details about the source of one kind of automatic discount with a top-level field. If we wanted to indicate the source of other automatic discounts (from identity linking, based on the line items, etc), we would need additional fields. Did you consider any design for this that broadened this field a bit to leave space for more detail on other discount types?

igrigorik: Good prompt. There is potential for "provenance proliferation" here, but so far I haven't spotted a generic schema. If you squint, codes and eligibility are both string[] — but they carry different lifecycle contracts that you'd want to capture and distinguish, so a generic source would still need type-specific semantics. All we'd have done is push the modeling complexity down one layer.

Open to suggestions.

docs/specification/checkout.md
```diff
@@ -249,6 +250,82 @@ messages or deferring to checkout completion.
...
+  "status": "ready_for_complete",
+  "line_items": [ "..." ],
+  "totals": [ "..." ],
+  "messages": [
+    {
+      "type": "error",
+      "code": "eligibility_invalid",
+      "severity": "recoverable",
+      "content": "Payment credential does not match the claimed store card benefit.",
+      "path": "$.context.eligibility[0]"
```
jingyli, line 318: `context` right now is [omitted from checkout responses](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/shopping/checkout.json#L60), making this kind of redirection back for error reference impossible.

Should we upgrade the field to an optional field in business responses as well to facilitate this?

source/schemas/shopping/discount.json
```diff
@@ -61,6 +61,15 @@
           "minimum": 1,
           "description": "Stacking order for discount calculation. Lower numbers applied first (1 = first)."
         },
+        "provisional": {
+          "type": "boolean",
+          "default": false,
+          "description": "True if this discount requires additional verification. Platforms MUST communicate provisional status to the buyer."
```
jingyli: This now introduces a separate source of which platform must look at to consider communication back to buyers. Curious whether we have considered:
1) If we want to force display/communication back to buyer, we can have business return warning messages to reference back at provisional discounts
**OR**
2) We loosen the MUST to SHOULD so platforms have more flexibility or control

Personally I think 1) makes more sense and maintains some consistent expectation on platforms where everything that requires communication back to buyers are organized in one place. If we do want to keep the design as-is, then at the very least this **MUST** requirement should show up in `discount.md` so it's explicitly called out?

docs/specification/checkout.md
igrigorik: Yep: https://github.com/Universal-Commerce-Protocol/ucp/pull/203/changes/43fa00d9acd3b95719beaa1ba6c16b21d92ec5ab

source/schemas/shopping/discount.json
igrigorik: https://github.com/Universal-Commerce-Protocol/ucp/pull/250/changes/f3fc01b8afafdc6094f6b88fd17d2ac78cfca9f7

PR #251, opened 2026-03-11
feat: Add basic schema for loyalty extension

ziwuzhou-google:
## Summary

A new loyalty extension is introduced to facilitate high-fidelity loyalty experiences across various commerce journeys, ensuring that shoppers can for example sign-up for membership, access personalized benefits, redeem rewards, and manage memberships seamlessly across various existing and future new capabilities.

## Motivation

Loyalty is a core concept in Commerce, serving as a primary driver for customer retention and long-term business growth. UCP in its current format can provide minimal and to some extent "hacky" support of loyalty in checkout for example. However, they are far from ideal and lack the generality. As Loyalty is a construct that spans across all verticals (retail, lodging, transportation, etc), the design is meant to encompass and capture the core common semantics of different capabilities and scenarios.

## Design Details

Four core concepts are introduced in the schema and structured in a hierarchy:
- Memberships: the overarching framework of a loyalty program, as well as the specific enrollment status and standing of a customer within it.
- Tiers: progressive achievement levels within a membership that unlock increasing value based on activity or spend/fee payment.
- Benefits: ongoing perks and privileges granted to a customer based on their current tier or membership status.
- Rewards: the accumulated balances and/or stored value available for the customer to redeem on transactions
```json
"memberships": [
  {
    "tiers": [
      {
        "benefits": [
          {
            ...
          }
        ]
      }
    ],
    "rewards": [
      {
        ...
      }
    ]
  }
]
```

## Example use cases

With the help of these four building blocks, one can support use cases that largely cover the lifecycle of a membership.

### Loyalty benefits discovery and upsell

Potential savings from member-only benefits can be shown in the upper-funnel discovery stage. No buyer membership is known or claimed in the request and business responses with the saving and program info for upsell.

=== "Request"

```json
    {
      "line_items": [
        {
          "item": {
            "id": "prod_1",
            "quantity": 1,
            "title": "T-Shirt",
            "price": 4000
          }
        }
      ]
    }
```

=== "Response"

```json
    {
      "line_items": [
        {
          "id": "li_1",
          "item": {
            "id": "prod_1",
            "quantity": 1,
            "title": "T-Shirt",
            "price": 4000
          },
          "totals": [
            {"type": "subtotal", "amount": 4000},
            {"type": "total", "amount": 4000}
          ]
        }
      ],
      "loyalty": {
        "memberships": [
          {
            "id": "membership_1",
            "name": "My Loyalty Program",
            "tiers": [
              {
                "id": "tier_1",
                "name": "GOLD",
                "level": 1,
                "benefits": [
                  {
                    "id": "BEN_001",
                    "title": "Members save $5",
                    "description": "Members receive $5 discount on eligible products"
                  },
                  {
                    "id": "BEN_002",
                    "title": "Early access",
                    "description": "24-hour early access to seasonal sales",
                  }
                ],
                "enrollment_conditions": [
                  {
                    "id": "CON_001",
                    "description": "Free to join"
                  }
                ],
                "links": [
                  {
                    "type": "terms_of_service",
                    "url": "loyalty_store.com/terms",
                    "title": "Sign-up Terms"
                  }
                ]
              }
            ]
          }
        ]
      },
      "totals": [
        {"type": "subtotal", "display_text": "Subtotal", "amount": 4000},
        {"type": "total", "display_text": "Estimated Total", "amount": 4000}
      ]
    }
```

### Membership sign-up

Buyer requests for membership sign-up after learning the benefits and business successfully register the buyer. Note that membership sign-up flow may or may not involve any product (i.e. may happen in early checkout phase, or in a pure account-linking flow), and thus the associated capability that loyalty extension attaches to is open.

=== "Request"

```json
    {
      "context": {
        "eligibility": ["com.example.loyalty"]
      },
      "loyalty": {
        "memberships": [
          {
            "id": "membership_1",
            "activated_tiers": ["tier_1"]
          }
        ]
      }
    }
```

=== "Response"

```json
    {
      "loyalty": {
        "memberships": [
          {
            "id": "membership_1",
            "member_id": "user_id_1",
            "name": "My Loyalty Program",
            "enrollment_date": "2026-01-12T00:00:00Z",
            "last_activity_date": "2026-01-12T00:00:00Z",
            "end_date": "2036-01-12T00:00:00Z",
            "tiers": [
              {
                "id": "tier_1",
                "name": "GOLD",
                "level": 1,
                "benefits": [
                  {
                    "id": "BEN_001",
                    "title": "Members save $5",
                    "description": "Members receive $5 discount on eligible products",
                  },
                  {
                    "id": "BEN_002",
                    "title": "Early access",
                    "description": "24-hour early access to seasonal sales",
                  }
                ]
              }
            ],
            "activated_tiers": ["tier_1"],
            "rewards": [
              {
                "currency": {
                  "name": "LoyaltyStars",
                  "code": "LST",
                },
                "balance": {
                  "available": 0,
                }
              }
            ],
            "provisional": false,
            "eligibility": "com.example.loyalty"
          }
        ]
      }
    }
```

### Member pricing

Buyer, as a loyalty member, receives a special lower price on eligible items. Business immediately runs the verification to check that the buyer is a valid member and the member pricing discount is no longer provisional.

=== "Request"

```json
    {
      "context": {
        "eligibility": ["com.example.loyalty"]
      },
      "line_items": [
        {
          "item": {
            "id": "prod_shirt",
            "quantity": 1,
            "title": "T-Shirt",
            "price": 4000
          }
        }
      ]
    }
```

=== "Response"

```json
    {
      "line_items": [
        {
          "id": "li_1",
          "item": {
            "id": "prod_1",
            "quantity": 1,
            "title": "T-Shirt",
            "price": 4000
          },
          "totals": [
            {"type": "subtotal", "amount": 4000},
            {"type": "items_discount", "amount": -500},
            {"type": "total", "amount": 3500}
          ]
        }
      ],
      "discounts": {
        "applied": [
          {
            "title": "Members save $5",
            "amount": 500,
            "automatic": true,
            "provisional": false,
            "eligibility": "com.example.loyalty",
            "allocations": [
              {"path": "$.line_items[0]", "amount": 500}
            ]
          }
        ]
      },
      "loyalty": {
        "memberships": [
          {
            "id": "membership_1",
            "member_id": "user_id_1",
            "name": "My Loyalty Program",
            "enrollment_date": "2026-01-12T00:00:00Z",
            "last_activity_date": "2026-01-12T00:00:00Z",
            "end_date": "2036-01-12T00:00:00Z",
            "tiers": [
              {
                "id": "tier_1",
                "name": "GOLD",
                "level": 1,
                "benefits": [
                  {
                    "id": "BEN_001",
                    "title": "Members save $5",
                    "description": "Members receive $5 discount on eligible products",
                    "applies_on": "$.discounts.applied[0]"
                  },
                  {
                    "id": "BEN_002",
                    "title": "Early access",
                    "description": "24-hour early access to seasonal sales",
                  }
                ]
              }
            ],
            "activated_tiers": ["tier_1"],
            "rewards": [
              {
                "currency": {
                  "name": "LoyaltyStars",
                  "code": "LST",
                },
                "balance": {
                  "available": 1000,
                }
              }
            ],
            "provisional": false,
            "eligibility": "com.example.loyalty"
          }
        ]
      },
      "totals": [
        {"type": "subtotal", "display_text": "Subtotal", "amount": 4000},
        {"type": "items_discount", "display_text": "Member Discounts", "amount": -500},
        {"type": "total", "display_text": "Estimated Total", "amount": 3500}
      ]
    }
```

### Point redemption

Buyer accumulates rewards through past transactions and they can be redeemed for total cost reduction. Rewards redemption also requires valid membership.

=== "Request"

```json
    {
      "context": {
        "eligibility": ["com.example.loyalty"]
      },
      "line_items": [
        {
          "item": {
            "id": "prod_shirt",
            "quantity": 1,
            "title": "T-Shirt",
            "price": 4000
          }
        }
      ],
      "loyalty": {
        "memberships": [
          {
            "id": "membership_1",
            "member_id": "member_id_1",
            "rewards": [
              {
                "currency": {
                  "code": "LST"
                },
                "redemption": {
                  "amount": 500
                }
              }
            ]
          }
        ]
      }
    }
```

=== "Response"

```json
    {
      "line_items": [
        {
          "id": "li_1",
          "item": {
            "id": "prod_1",
            "quantity": 1,
            "title": "T-Shirt",
            "price": 4000
          },
          "totals": [
            {"type": "subtotal", "amount": 4000},
            {"type": "items_discount", "amount": -500},
            {"type": "total", "amount": 3500}
          ]
        }
      ],
      "discounts": {
        "applied": [
          {
            "title": "Point Redemption",
            "amount": 500,
            "automatic": true,
            "provisional": false,
            "eligibility": "com.example.loyalty",
            "allocations": [
              {"path": "$.line_items[0]", "amount": 500}
            ]
          }
        ]
      },
      "loyalty": {
        "memberships": [
          {
            "id": "membership_1",
            "member_id": "user_id_1",
            "name": "My Loyalty Program",
            "enrollment_date": "2026-01-12T00:00:00Z",
            "last_activity_date": "2026-01-12T00:00:00Z",
            "end_date": "2036-01-12T00:00:00Z",
            "tiers": [
              {
                "id": "tier_1",
                "name": "GOLD",
                "level": 1,
                "benefits": [
                  {
                    "id": "BEN_001",
                    "title": "Early access",
                    "description": "24-hour early access to seasonal sales",
                  }
                ]
              }
            ],
            "activated_tiers": ["tier_1"],
            "rewards": [
              {
                "currency": {
                  "name": "LoyaltyStars",
                  "code": "LST"
                },
                "balance": {
                  "available": 500
                },
                "redemption": {
                  "amount": 500,
                  "applies_on": "$.discounts.applied[0]"
                }
              }
            ],
            "provisional": false,
            "eligibility": "com.example.loyalty"
          }
        ]
      },
      "totals": [
        {"type": "subtotal", "display_text": "Subtotal", "amount": 4000},
        {"type": "items_discount", "display_text": "Point Redemption", "amount": -500},
        {"type": "total", "display_text": "Estimated Total", "amount": 3500}
      ]
    }
```

## Type of change

      or fields**)
- [ ] Documentation update

---

## Checklist

### DIFF
--- a/docs/specification/loyalty.md
+++ b/docs/specification/loyalty.md
@@ -1,0 +1,1012 @@
+<!--
+   Copyright 2026 UCP Authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+
+# Loyalty Extension
+
+## Overview
+
+The loyalty extension is designed to facilitate high-fidelity loyalty experiences across various commerce journeys, ensuring that shoppers can for example sign-up for membership, access personalized benefits, redeem rewards, and manage memberships seamlessly across various capabilities.
+
+In its current format, it can be used to decorate the following Capabilities for example:
+
+* **Catalog**: For discovering potential immediate (e.g. price discounts) or delayed (e.g. cash back for future use) savings from being a member
+* **Checkout**: For applying member-specific discounts/fulfillment benefits, calculating point earnings, or initiating reward redemptions, in conjunction with discount extension to specify the monetary effect of applicable loyalty benefits
+* **Cart**: For displaying the loyalty benefits for potential usage on subsequent checkout (e.g. redeemable balance, tier evaluations), with or without the help of discount extension to upsell potential savings and non-monetary benefits from membership
+
+In the future, for membership lifecycle management by agents (e.g. membership status fetching, program sign-up, balance transfer, etc.), it is possible that loyalty could be considered an independent capability. The other direction is fitting them within higher-order account and profile management capabilities that allow for broader operations, with loyalty being a subset and hence potentially represented by the same extension as it. This is not yet finalized among community discussion and pending future updates to achieve.
+
+## Key Concepts
+
+Loyalty has four main components:
+
+**Memberships**: the overarching framework of a loyalty program, as well as the specific enrollment status and standing of a customer within it
+
+* Usually represents a customer lifecycle solution the businesses have, which offer a tiered ecosystem of access, convenience, and identity.
+* Encompasses the metadata of the concrete offerings as well as the customer level information associated with (if applicable).
+* Customers can have multiple memberships that are relevant to the business (e.g. self-owned program and third-party program).
+
+**Tiers**: progressive achievement levels within a membership that unlock increasing value based on activity or spend/fee payment
+
+* Defines achievable levels within a membership for a customer, with various types of qualification methods and conditions.
+* Customers can simultaneously be activated on and eligible for multiple tiers if the program allows.
+
+**Benefits**: ongoing perks and privileges granted to a customer based on their current tier or membership status
+
+* Contains both delayed (e.g. “Members have access to dedicated customer service”) and immediate-value (e.g. “Members get 5% off”) benefits.
+
+**Rewards**: the accumulated balances and/or stored value available for the customer to redeem on transactions
+
+* One membership can offer multiple types of accumulable/collectable rewards, each having its own usage and redemption rules.
+
+## Discovery
+
+Businesses can follow standard advertisement mechanism to advertise loyalty support in the Business profile
+
+```json
+{
+ "ucp": {
+   "version": "2026-01-23",
+   "capabilities": {
+     "dev.ucp.common.loyalty": [
+       {
+         "version": "2026-01-23",
+         "extends": "dev.ucp.shopping.checkout",
+         "spec": "https://ucp.dev/2026-01-23/specification/loyalty",
+         "schema": "https://ucp.dev/2026-01-23/schemas/common/loyalty.json"
+       }
+     ]
+   }
+ }
+}
+```
+
+## Schema
+
+### Entities
+
+#### Loyalty
+
+{{ schema_fields('loyalty', 'loyalty') }}
+
+#### Loyalty Membership
+
+{{ schema_fields('loyalty_membership', 'loyalty') }}
+
+#### Membership Tier
+
+{{ schema_fields('membership_tier', 'loyalty') }}
+
+#### Membership Tier Condition
+
+{{ schema_fields('membership_tier_condition', 'loyalty') }}
+
+#### Membership Tier Benefit
+
+{{ schema_fields('membership_tier_benefit', 'loyalty') }}
+
+#### Membership Reward
+
+{{ schema_fields('membership_reward', 'loyalty') }}
+
+#### Balance Currency
+
+{{ schema_fields('balance_currency', 'loyalty') }}
+
+#### Membership Balance
+
+{{ schema_fields('membership_balance', 'loyalty') }}
+
+#### Expiring Balance
+
+{{ schema_fields('expiring_balance', 'loyalty') }}
+
+#### Balance Redemption
+
+{{ schema_fields('balance_redemption', 'loyalty') }}
+
+## Eligibility Claims
+
+Given almost everything related to loyalty is provisional and requires eligible membership before benefits can be applied, [Context](checkout.md#context) naturally fits here where it allows buyers to claim eligibility for loyalty benefits. When loyalty extension is active and the request contains eligibility claims about loyalty, businesses that choose to accept eligibility claims MUST surface that as an indicator of buyer’s membership status, and if applicable, effects on pricing coming from monetary benefits. Platforms MUST display those provisional loyalty discounts to the buyer.
+
+### Loyalty behavior
+
+In the case platforms run in a “discovery” mode where buyer’s loyalty membership is unknown and there is no interest to request membership sign-up, platforms MAY not pass any loyalty related info in `context.eligibility`. Business MAY respond with a loyalty extension for upsell, if applicable and available. The responded loyalty extension normally contains program information/sign-up conditions/benefit descriptions to entice the buyer and prompt for a sign-up action.
+
+```json
+{
+  "loyalty": {
+    "memberships": [
+      {
+        "id": "membership_1",
+        "name": "My Loyalty Program",
+        "tiers": [
+          {
+            "id": "tier_1",
+            "name": "GOLD",
+            "level": 1,
+            "benefits": [
+              {
+                "id": "BEN_001",
+                "title": "Members save $5",
+                "description": "Members receive $5 discount on eligible products"
+              }
+            ],
+            "enrollment_conditions": [
+              {
+                "id": "CON_001",
+                "description": "Free to join"
+              }
+            ],
+            "links": [
+              {
+                "type": "terms_of_service",
+                "url": "loyalty_store.com/terms",
+                "title": "Sign-up Terms"
+              }
+            ]
+          }
+        ]
+      }
+    ]
+  }
+}
+```
+
+In the case platforms have prior knowledge about buyer’s membership info, or wish to request membership sign up, they MUST send buyer loyalty membership claims via `context.eligibility`. Businesses MAY run required verification for the membership claim and communicate back the result via `provisional` and `eligibility` fields under membership. The `eligibility` sitting under the membership holds the claim for the general loyalty membership. In case businesses run the verification and the claim is valid/sign up is successful, businesses MUST additionally populate `activated_tiers` on top of `provisional` and `eligibility` fields to communicate the tiers the buyer has. When verification runs but fails, businesses MUST communicate the failure via a recoverable error message that platforms can choose to either remove the claim on the membership entirely and proceed with no member benefits applied or keep the claim but proceed with the membership activation/sign-up flow.
+
+=== "Request with known buyer membership claim"
+
+    ```json
+    {
+      "context": {
+        "eligibility": ["com.example.loyalty"]
+      }
+    }
+    ```
+
+=== "Request"
+
+    ```json
+    {
+      "context": {
+        "eligibility": ["com.example.loyalty"]
+      },
+      "loyalty": {
+        "memberships": [
+          {
+            "id": "membership_1",
+            "activated_tiers": ["tier_1"]
+          }
+        ]
+      }
+    }
+    ```
+
+=== "Response without verification"
+
+    ```json
+    {
+      "loyalty": {
+        "memberships": [
+          {
+            "id": "membership_1",
+            "name": "My Loyalty Program",
+            "tiers": [
+              {
+                "id": "tier_1",
+                "name": "GOLD",
+                "level": 1
+              }
+            ],
+            "provisional": true,
+            "eligibility": "com.example.loyalty"
+          }
+        ]
+      }
+    }
+    ```
+
+=== "Response with verification and valid membership"
+
+    ```json
+    {
+      "loyalty": {
+        "memberships": [
+          {
+            "id": "membership_1",
+            "user_id": "user_id_1",
+            "name": "My Loyalty Program",
+            "enrollment_date": "2022-11-12T00:00:00Z",
+            "last_activity_date": "2026-01-12T00:00:00Z",
+            "end_date": "2036-01-12T00:00:00Z",
+            "tiers": [
+              {
+                "id": "tier_1",
+                "name": "GOLD",
+                "level": 1
+              }
+            ],
+            "activated_tiers": ["tier_1"],
+            "provisional": false,
+            "eligibility": "com.example.loyalty"
+          }
+        ]
+      }
+    }
+    ```
+
+=== "Response with verification and invalid membership"
+
+    ```json
+    {
+      "loyalty": {
+        "memberships": [
+          {
+            "id": "membership_1",
+            "name": "My Loyalty Program",
+            "tiers": [
+              {
+                "id": "tier_1",
+                "name": "GOLD",
+                "level": 1
+              }
+            ],
+            "provisional": false,
+            "eligibility": "com.example.loyalty"
+          }
+        ]
+      },
+      "messages": [
+        {
+          "type": "error",
+          "severity": "recoverable",
+          "code": "eligibility_invalid",
+          "path": "$.loyalty.memberships[0]",
+          "content": "Buyer is not a valid loyalty member"
+        }
+      ]
+    }
+    ```
+
+### Loyalty benefits behavior
+
+An eligible membership is sometimes only a preliminary prerequisite of a member-only benefit and a verified claiming of loyalty membership does not necessarily result in a valid claim of associated member-only benefit. For example, in the event of a $50 order, the member-only benefit "Free shipping for all orders" applies while the other member-only discount "Save $10 with $100+ purchase" does not, assuming the buyer is an eligible member. As such, business MUST additionally surface all monetary price impacting benefits as provisional discounts using the `provisional` and `eligibility` fields within the `discounts.applied` object (referenced by the `applies_on` field within `benefit`). When membership is valid but there are benefits that are inapplicable, businesses MUST communicate this via the message[] array. Depending on the type of inapplicable benefits (e.g. they are affecting the order totals), businesses can choose the message type between warning or info to get them surfaced.
+
+=== "Response without verification"
+
+    ```json
+    {
+      "discounts": {
+        "applied": [
+          {
+            "title": "Free shipping",
+            "amount": 200,
+            "method": "each",
+            "provisional": true,
+            "eligibility": "com.example.loyalty",
+            "allocations": [
+              {"path": "$.line_items[0]", "amount": 200}
+            ]
+          },
+          {
+            "title": "Save $10 with $100+ order",
+            "amount": 100,
+            "method": "each",
+            "provisional": true,
+            "eligibility": "com.example.loyalty",
+            "allocations": [
+              {"path": "$.line_items[0]", "amount": 100}
+            ]
+          }
+        ]
+      },
+      "loyalty": {
+        "memberships": [
+          {
+            "id": "membership_1",
+            "name": "My Loyalty Program",
+            "tiers": [
+              {
+                "id": "tier_1",
+                "name": "GOLD",
+                "level": 1,
+                "benefits": [
+                  {
+                    "id": "BEN_001",
+                    "title": "Free shipping",
+                    "description": "Complimentary standard shipping on all orders",
+                    "applies_on": "$.discounts.applied[0]"
+                  },
+                  {
+                    "id": "BEN_002",
+                    "title": "Early access",
+                    "description": "24-hour early access to seasonal sales"
+                  },
+                  {
+                    "id": "BEN_003",
+                    "title": "Save $10 with $100+ order",
+                    "description": "Get $10 discount when purchase order is $100+",
+                    "applies_on": "$.discounts.applied[1]"
+                  }
+                ]
+              }
+            ],
+            "provisional": true,
+            "eligibility": "com.example.loyalty"
+          }
+        ]
+      }
+    }
+    ```
+
+=== "Response with valid membership and benefits"
+
+    ```json
+    {
+      "discounts": {
+        "applied": [
+          {
+            "title": "Free shipping",
+            "amount": 200,
+            "method": "each",
+            "provisional": false,
+            "eligibility": "com.example.loyalty",
+            "allocations": [
+              {"path": "$.line_items[0]", "amount": 200}
+            ]
+          },
+          {
+            "title": "Save $10 with $100+ order",
+            "amount": 100,
+            "method": "each",
+            "provisional": false,
+            "eligibility": "com.example.loyalty",
+            "allocations": [
+              {"path": "$.line_items[0]", "amount": 100}
+            ]
+          }
+        ]
+      },
+      "loyalty": {
+        "memberships": [
+          {
+            "id": "membership_1",
+            "user_id": "user_id_1",
+            "name": "My Loyalty Program",
+            "enrollment_date": "2022-11-12T00:00:00Z",
+            "last_activity_date": "2026-01-12T00:00:00Z",
+            "end_date": "2036-01-12T00:00:00Z",
+            "tiers": [
+              {
+                "id": "tier_1",
+                "name": "GOLD",
+                "level": 1,
+                "benefits": [
+                  {
+                    "id": "BEN_001",
+                    "title": "Free shipping",
+                    "description": "Complimentary standard shipping on all orders",
+                    "applies_on": "$.discounts.applied[0]"
+                  },
+                  {
+                    "id": "BEN_002",
+                    "title": "Early access",
+                    "description": "24-hour early access to seasonal sales"
+                  },
+                  {
+                    "id": "BEN_003",
+                    "title": "Save $10 with $100+ order",
+                    "description": "Get $10 discount when purchase order is $100+",
+                    "applies_on": "$.discounts.applied[1]"
+                  }
+                ]
+              }
+            ],
+            "activated_tiers": ["tier_1"],
+            "provisional": false,
+            "eligibility": "com.example.loyalty"
+          }
+        ]
+      }
+    }
+    ```
+
+=== "Response with valid membership and partially valid benefits"
+
+    ```json
+    {
+      "discounts": {
+        "applied": [
+          {
+            "title": "Free shipping",
+            "amount": 200,
+            "method": "each",
+            "provisional": false,
+            "eligibility": "com.example.loyalty",
+            "allocations": [
+              {"path": "$.line_items[0]", "amount": 200}
+            ]
+          }
+        ]
+      },
+      "loyalty": {
+        "memberships": [
+          {
+            "id": "membership_1",
+            "user_id": "user_id_1",
+            "name": "My Loyalty Program",
+            "enrollment_date": "2022-11-12T00:00:00Z",
+            "last_activity_date": "2026-01-12T00:00:00Z",
+            "end_date": "2036-01-12T00:00:00Z",
+            "tiers": [
+              {
+                "id": "tier_1",
+                "name": "GOLD",
+                "level": 1,
+                "benefits": [
+                  {
+                    "id": "BEN_001",
+                    "title": "Free shipping",
+                    "description": "Complimentary standard shipping on all orders",
+                    "applies_on": "$.discounts.applied[0]"
+                  },
+                  {
+                    "id": "BEN_002",
+                    "title": "Early access",
+                    "description": "24-hour early access to seasonal sales"
+                  },
+                  {
+                    "id": "BEN_003",
+                    "title": "Save $10 with $100+ order",
+                    "description": "Get $10 discount when purchase order is $100+",
+                  }
+                ]
+              }
+            ],
+            "activated_tiers": ["tier_1"],
+            "provisional": false,
+            "eligibility": "com.example.loyalty"
+          }
+        ]
+      },
+      "messages": [
+        {
+          "type": "info",
+          "code": "membership_benefit_ineligible",
+          "path": "$.loyalty.memberships[0].tiers[0].benefits[2]",
+          "content": "Member benefit is not eligible for the buyer"
+        }
+      ]
+    }
+    ```
+
+## Use Cases and Examples
+
+### Loyalty benefits discovery and upsell
+
+Potential savings from member-only benefits can be shown in the upper-funnel discovery stage. No buyer membership is known or claimed in the request and business responses with the saving and program info for upsell.
+
+=== "Request"
+
+    ```json
+    {
+      "line_items": [
+        {
+          "item": {
+            "id": "prod_1",
+            "quantity": 1,
+            "title": "T-Shirt",
+            "price": 4000
+          }
+        }
+      ]
+    }
+    ```
+
+=== "Response"
+
+    ```json
+    {
+      "line_items": [
+        {
+          "id": "li_1",
+          "item": {
+            "id": "prod_1",
+            "quantity": 1,
+            "title": "T-Shirt",
+            "price": 4000
+          },
+          "totals": [
+            {"type": "subtotal", "amount": 4000},
+            {"type": "total", "amount": 4000}
+          ]
+        }
+      ],
+      "loyalty": {
+        "memberships": [
+          {
+            "id": "membership_1",
+            "name": "My Loyalty Program",
+            "tiers": [
+              {
+                "id": "tier_1",
+                "name": "GOLD",
+                "level": 1,
+                "benefits": [
+                  {
+                    "id": "BEN_001",
+                    "title": "Members save $5",
+                    "description": "Members receive $5 discount on eligible products"
+                  },
+                  {
+                    "id": "BEN_002",
+                    "title": "Early access",
+                    "description": "24-hour early access to seasonal sales",
+                  }
+                ],
+                "enrollment_conditions": [
+                  {
+                    "id": "CON_001",
+                    "description": "Free to join"
+                  }
+                ],
+                "links": [
+                  {
+                    "type": "terms_of_service",
+                    "url": "loyalty_store.com/terms",
+                    "title": "Sign-up Terms"
+                  }
+                ]
+              }
+            ]
+          }
+        ]
+      },
+      "totals": [
+        {"type": "subtotal", "display_text": "Subtotal", "amount": 4000},
+        {"type": "total", "display_text": "Estimated Total", "amount": 4000}
+      ]
+    }
+    ```
+
+### Membership sign-up
+
+Buyer requests for membership sign-up after learning the benefits and business successfully register the buyer. Note that membership sign-up flow may or may not involve any product (i.e. may happen in early checkout phase, or in a pure account-linking flow), and thus the associated capability that loyalty extension attaches to is open.
+
+=== "Request"
+
+    ```json
+    {
+      "context": {
+        "eligibility": ["com.example.loyalty"]
+      },
+      "loyalty": {
+        "memberships": [
+          {
+            "id": "membership_1",
+            "activated_tiers": ["tier_1"]
+          }
+        ]
+      }
+    }
+    ```
+
+=== "Response"
+
+    ```json
+    {
+      "loyalty": {
+        "memberships": [
+          {
+            "id": "membership_1",
+            "member_id": "user_id_1",
+            "name": "My Loyalty Program",
+            "enrollment_date": "2026-01-12T00:00:00Z",
+            "last_activity_date": "2026-01-12T00:00:00Z",
+            "end_date": "2036-01-12T00:00:00Z",
+            "tiers": [
+              {
+                "id": "tier_1",
+                "name": "GOLD",
+                "level": 1,
+                "benefits": [
+                  {
+                    "id": "BEN_001",
+                    "title": "Members save $5",
+                    "description": "Members receive $5 discount on eligible products",
+                  },
+                  {
+                    "id": "BEN_002",
+                    "title": "Early access",
+                    "description": "24-hour early access to seasonal sales",
+                  }
+                ]
+              }
+            ],
+            "activated_tiers": ["tier_1"],
+            "rewards": [
+              {
+                "currency": {
+                  "name": "LoyaltyStars",
+                  "code": "LST",
+                },
+                "balance": {
+                  "available": 0,
+                }
+              }
+            ],
+            "provisional": false,
+            "eligibility": "com.example.loyalty"
+          }
+        ]
+      }
+    }
+    ```
+
+### Member pricing
+
+Buyer, as a loyalty member, receives a special lower price on eligible items. Business immediately runs the verification to check that the buyer is a valid member and the member pricing discount is no longer provisional.
+
+=== "Request"
+
+    ```json
+    {
+      "context": {
+        "eligibility": ["com.example.loyalty"]
+      },
+      "line_items": [
+        {
+          "item": {
+            "id": "prod_shirt",
+            "quantity": 1,
+            "title": "T-Shirt",
+            "price": 4000
+          }
+        }
+      ]
+    }
+    ```
+
+=== "Response"
+
+    ```json
+    {
+      "line_items": [
+        {
+          "id": "li_1",
+          "item": {
+            "id": "prod_1",
+            "quantity": 1,
+            "title": "T-Shirt",
+            "price": 4000
+          },
+          "totals": [
+            {"type": "subtotal", "amount": 4000},
+            {"type": "items_discount", "amount": -500},
+            {"type": "total", "amount": 3500}
+          ]
+        }
+      ],
+      "discounts": {
+        "applied": [
+          {
+            "title": "Members save $5",
+            "amount": 500,
+            "automatic": true,
+            "provisional": false,
+            "eligibility": "com.example.loyalty",
+            "allocations": [
+              {"path": "$.line_items[0]", "amount": 500}
+            ]
+          }
+        ]
+      },
+      "loyalty": {
+        "memberships": [
+          {
+            "id": "membership_1",
+            "member_id": "user_id_1",
+            "name": "My Loyalty Program",
+            "enrollment_date": "2026-01-12T00:00:00Z",
+            "last_activity_date": "2026-01-12T00:00:00Z",
+            "end_date": "2036-01-12T00:00:00Z",
+            "tiers": [
+              {
+                "id": "tier_1",
+                "name": "GOLD",
+                "level": 1,
+                "benefits": [
+                  {
+                    "id": "BEN_001",
+                    "title": "Members save $5",
+                    "description": "Members receive $5 discount on eligible products",
+                    "applies_on": "$.discounts.applied[0]"
+                  },
+                  {
+                    "id": "BEN_002",
+                    "title": "Early access",
+                    "description": "24-hour early access to seasonal sales",
+                  }
+                ]
+              }
+            ],
+            "activated_tiers": ["tier_1"],
+            "rewards": [
+              {
+                "currency": {
+                  "name": "LoyaltyStars",
+                  "code": "LST",
+                },
+                "balance": {
+                  "available": 1000,
+                }
+              }
+            ],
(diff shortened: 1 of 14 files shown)


source/schemas/common/types/balance_currency.json
```diff
@@ -0,0 +1,18 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/common/types/balance_currency.json",
+  "title": "Expiring Balance",
```
amithanda: Balance Currency?

source/schemas/common/loyalty.json
```diff
@@ -0,0 +1,34 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/common/loyalty.json",
+  "name": "dev.ucp.common.loyalty",
+  "title": "Loyalty Extension",
+  "description": "Extends various Capabilities with loyalty support using memberships info.",
+  "$defs": {
+    "loyalty": {
+      "$ref": "types/loyalty.json"
+    },
+    "dev.ucp.shopping.checkout": {
```
amithanda, line 32: Should Loyalty benefits apply to Cart as well? Customers may expect to see their accrued loyalty points, tier evaluations, and redeemable balances within their Cart before proceeding to Checkout.

source/schemas/common/types/balance_currency.json
ziwuzhou-google: Nice catch! Will update this in my next commit.

ziwuzhou-google:
> Great start on this extension! However, Is there an issue with how request payloads are mapped when a platform attempts to redeem rewards or activate tiers? Currently, the schema explicitly sets `"ucp_request": "omit"` on critical identifier fields:
> 
> * `id` in source/schemas/common/types/loyalty_membership.json
> * `currency` in source/schemas/common/types/membership_reward.json
> 
> **The Problem:** If a customer has multiple different memberships (e.g., standard vs credit cardholder) or a single membership with multiple reward currencies (e.g., Points vs Miles), a platform needs a way to explicitly address exactly _which_ membership or reward balance they are redeeming during a `Checkout update` or `Checkout complete` request. Because `id` and `currency` are `omit`, the client is technically forbidden from sending them in the update payload. If a client sends this:
> 
> ```json
> "loyalty": {
>   "memberships": [
>     {
>       "activated_tiers": ["T_1"],
>       "rewards": [ 
>         { "redemption": { "amount": 100 } } 
>       ]
>     }
>   ]
> }
> ```
> 
> The business server has no stable identifier to know which membership ID `T_1` belongs to, or which currency `100` refers to, except by relying on brittle array indexing which changes between requests.

Both are correct. These identifiers should not have `omit`. See the updated example in https://github.com/Universal-Commerce-Protocol/ucp/pull/251#issue-4055515507 for how platform can request redemption with multiple types of award & sign-up for multiple memberships.

docs/specification/loyalty.md
```diff
@@ -0,0 +1,239 @@
...
+   limitations under the License.
+-->
+
+# Loyalty Extension
+
+## Overview
+
+The loyalty extension is designed to facilitate high-fidelity loyalty experiences across various commerce journeys, ensuring that shoppers can for example sign-up for membership, access personalized benefits, redeem rewards, and manage memberships seamlessly across various capabilities.
+
+A few examples on how it might be used when decorating existing Capabilities:
```
sinhanurag: We should add a Cart example as well. Cart is currently in draft stage but this will apply there too. I would hold off on other capabilities for now.

source/schemas/common/types/loyalty_membership.json
```diff
@@ -0,0 +1,56 @@
...
+  "required": ["id", "name", "tiers", "activated_tiers"],
+  "properties": {
+    "id": {
+      "type": "string",
+      "description": "Unique loyalty membership identifier."
+    },
+    "membership_id": {
+      "type": "string",
+      "description": "User's loyalty membership ID (e.g. loyalty card #, membership #).",
+      "ucp_request": "omit"
```
sinhanurag: Technically we will need this in the incoming request too if the platform already has the membership id and needs to communicate that to the business on behalf of the user.

source/schemas/common/types/loyalty_membership.json
```diff
@@ -0,0 +1,56 @@
...
+  "title": "Loyalty Membership",
+  "description": "Loyalty membership that the business offers or associated with.",
+  "type": "object",
+  "required": ["id", "name", "tiers", "activated_tiers"],
+  "properties": {
+    "id": {
+      "type": "string",
+      "description": "Unique loyalty membership identifier."
+    },
+    "membership_id": {
```
sinhanurag: To make it clearer maybe rename to "member_id"?

source/schemas/common/types/membership_tier.json
```diff
@@ -0,0 +1,44 @@
...
+    "name": {
+      "type": "string",
+      "description": "The human-readable name of the tier (e.g., “Platinum”)."
+    },
+    "level": {
+      "type": "integer",
+      "minimum": 1,
+      "description": "The relative ranking or numerical level of the tier within the loyalty program."
+    },
+    "conditions": {
```
sinhanurag: Can we call it "enrollment_conditions" ? or something similar?

docs/specification/loyalty.md
```diff
@@ -0,0 +1,239 @@
...
+        },
+        {
+          "id": "tier_2",
+          "name": "PLATINUM",
+          "level": 2,
+          "benefits": [
+            {
+              "id": "BEN_001",
+              "title": "Free shipping",
+              "description": "Complimentary standard shipping on all orders"
```
sinhanurag: We will need some examples of cart, checkout, order objects (starting with checkout is fine) which have this extension as well as the discount object clearly displaying the benefit being applied and the discount object enumerating the benefit.

Like in this case "BEN_001" will apply to discounts.applied.allocations (path = fulfillment-options.total , amount = xx) etc.

Ref: https://ucp.dev/latest/specification/discount/#discounts-object

docs/specification/loyalty.md
```diff
@@ -0,0 +1,239 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+
+# Loyalty Extension
+
+## Overview
+
+The loyalty extension is designed to facilitate high-fidelity loyalty experiences across various commerce journeys, ensuring that shoppers can for example sign-up for membership, access personalized benefits, redeem rewards, and manage memberships seamlessly across various capabilities.
```
sinhanurag, line 21: Can we add some of our high level thinking around how this extension is specifically to contextualize loyalty in the context of existing UCP capabilities? And that there will be more PRs and discussions around managing loyalty and orthogonal account operations in a separate PR/forum.

docs/specification/loyalty.md
```diff
@@ -0,0 +1,239 @@
...
+
+#### Balance Redemption
+
+{{ schema_fields('types/balance_redemption', 'loyalty') }}
+
+### Example
+
+```json
+{
+  "memberships": [
```
sinhanurag: Can we also add a request/response example of how an instrument based discount will look like?

I am thinking of cases when there is a co branded instrument used for purchase. At that point in time the agent might or might not be aware of the associated discount. The agent will send the chosen FOP or identifier as a signal. This signal can be part of the loyalty extension in the incoming checkout/cart request or passed in a different placeholder. The response should contain a loyalty extension with the applied discount based on the instrument.

sinhanurag (commented):
Good start based on all our discussions! Left some comments. Let's iterate on it more and discuss further.

docs/specification/loyalty.md
sinhanurag: I am actually fine reconciling it with the PR raised here - https://github.com/Universal-Commerce-Protocol/ucp/pull/250

-cc @igrigorik 

So the ingress signal for loyalty identification can come from context.eligibility but then the response can contain the loyalty extension decorating the objects.

source/schemas/common/loyalty.json
ziwuzhou-google: Yes it should. We didn't add it in the first place because we didn't see precedence of extension decorating cart  capability. However given that discount extension can do that now, we will add this for loyalty extension. Fixing that in my next commit.

docs/specification/loyalty.md
ziwuzhou-google: SG! Updated the example to just include checkout and cart (to stay consistent with the .json definition). Can always can back to update the document when new capability is supported.

source/schemas/common/types/loyalty_membership.json
ziwuzhou-google: Done.

source/schemas/common/types/loyalty_membership.json
ziwuzhou-google: Updated.

source/schemas/common/types/membership_tier.json
ziwuzhou-google: Updated to `enrollment_conditions`.

docs/specification/loyalty.md
ziwuzhou-google: Added some new wordings around this.

docs/specification/loyalty.md
ziwuzhou-google: Added an example here for shipping but I don't think the `applies_on` needs to go all the way to `allocations` level as the shipping benefit applies on the entire order.

tpindel (commented):
Great work on this extension  @ziwuzhou-google 👏 the hierarchical memberships → tiers → benefits → rewards model is well-designed, and placing it in common/ for cross-vertical use is the right call.

We've been building a reference implementation of incentives + loyalty on UCP (MCP transport, Shopify + Voucherify API/MCP, ~22 ucp_* tools covering the full shopping lifecycle). Based on real  agent-customer interactions, there are a few capabilities we've found essential that the current schema doesn't cover yet. Happy to contribute PRs for any of these if there's interest.

**1. Earning Forecast - "What will I earn from this purchase?"**

This is the single most impactful missing piece for agent-driven commerce. In our implementation, showing "You'll earn 120 Loyalty Stars/Points with this order, bringing your balance to 4,620" before the customer commits is one of the strongest purchase motivators. It also helps agents handle price objections - points earning becomes additional value on top of any discount.

The current schema has balance (what you have) and redemption (what you're spending), but no way to show what the customer gains from the transaction.

Proposed addition - a new earning_forecast field on membership_reward, response-only:

```
"earning_forecast": {
    "type": "object",
    "description": "Preview of rewards to be earned from the current transaction.",
    "ucp_request": "omit",
    "properties": {
      "amount": {
        "type": "integer",
        "minimum": 0,
        "description": "Total points to be earned if the transaction completes."
      },
      "rules": {
        "type": "array",
        "items": {
          "type": "object",
          "required": ["id", "amount"],
          "properties": {
            "id": { "type": "string", "description": "Earning rule identifier." },
            "description": { "type": "string", "description": "Human-readable explanation (e.g. '2x points on footwear')." },
            "amount": { "type": "integer", "minimum": 0, "description": "Points earned from this rule." }
          }
        },
        "description": "Breakdown of earning rules contributing to the total."
      },
      "projected_balance": {
        "type": "integer",
        "minimum": 0,
        "description": "Projected available balance after earning and any redemption from this transaction."
      }
    }
}
```

This fits naturally into the existing membership_reward structure - each reward currency gets its own earning forecast. The per-rule breakdown gives agents transparency to explain why the customer is earning (e.g. "2x points on footwear" vs "base earning on all purchases"), which makes loyalty feel tangible rather than a black box.

**2. Tier Progression - "How close am I to the next tier?"**

The current schema shows which tiers exist and which are activated, but doesn't tell the agent how close the customer is to qualifying for a higher tier. In practice this drives natural upsell: "You're only 200 points from Platinum - this purchase would get you there!"

Proposed addition to membership_tier:

```
"progression": {
    "type": "object",
    "description": "Customer's progress toward qualifying for this tier.",
    "ucp_request": "omit",
    "properties": {
      "qualified": {
        "type": "boolean",
        "description": "Whether the customer currently qualifies for this tier."
      },
      "progress_amount": {
        "type": "integer", "minimum": 0,
        "description": "Current progress value toward qualification threshold."
      },
      "threshold_amount": {
        "type": "integer", "minimum": 0,
        "description": "Required value to qualify for this tier."
      },
      "remaining_amount": {
        "type": "integer", "minimum": 0,
        "description": "Gap between current progress and threshold."
      }
    }
}
```

This is also response-only. For tiers the customer already has, qualified: true. For higher tiers, the agent can use remaining_amount to suggest cart additions that would push the customer over the threshold.

**3. Agent Journey Example - When to Present Loyalty Context**

Regarding PR #250 (eligibility claims) - in our implementation the agent flow looks like:

1. Identify - customer provides email or loyalty card → agent looks up member profile (memberships, tiers, rewards balance)
2. Discover deals - agent checks what promotions + loyalty benefits apply to the current cart, including tier-specific benefits
3. Present - agent shows price with discount + "you'll earn X points" + "you're Y away from next tier"
4. Redeem (optional) - customer chooses to spend points → redemption.amount on the relevant reward, linked via applies_on to discount extension
5. Complete - earning is confirmed in the response

A concrete request/response example in the spec doc showing this full flow - especially the interplay between loyalty benefits, discount applies_on references, beast deals, and earning in the response - would be very helpful for agent developers building on this extension.

**Context**
We have a working MCP-based UCP server implementing these patterns - earning forecast and the full incentives negotiation lifecycle (search best deal → validate → lock → redeem). Happy to share more details or contribute schema additions as a follow-up PR if any of these proposals are useful.

docs/specification/loyalty.md
jingyli: A thought on this point:
> So the ingress signal for loyalty identification can come from context.eligibility but then the response can contain the loyalty extension decorating the objects.

Given `context.eligibility` uses reverse domain names, I think one simple way to reconcile the 2 approaches would be to either have `memberships[].id` to also be reverse domain names OR add another field to reference back to the corresponding `eligibility` (similar to the provisional discount).

What I feel also fits well between the 2 PR proposals is that in the case any generic (not necessarily instrument-based discount) loyalty eligibility verification fails (i.e. buyer is not actually properly signed up to the loyalty program), this extension offers a mechanism for platform to request for sign-ups and renegotiate for that verification.

lemonmade: > add another field to reference back to the corresponding eligibility (similar to the provisional discount).

This is what I was expecting, for symmetry with discounts. From #250, both `provisional` and `eligibility` from the applied discount type feel appropriate to include on each membership, too.

docs/specification/loyalty.md
```diff
@@ -0,0 +1,243 @@
...
+        {
+          "id": "tier_2",
+          "name": "PLATINUM",
+          "level": 2,
+          "benefits": [
+            {
+              "id": "BEN_001",
+              "title": "Free shipping",
+              "description": "Complimentary standard shipping on all orders",
+              "applies_on": ["$.discounts.applied[0]"]
```
lemonmade: It seems like a lot of the display information for the referenced applied discount information is likely to be duplicated on this object. Is there any way to deduplicate these two (e.g., by splitting `applied_benefits: ["discounts.keypath"]` from `available_benefits`), or do we need the full-fidelity information on the membership itself for the host to render this on the business's behalf?

source/schemas/common/types/membership_tier_condition.json
```diff
@@ -0,0 +1,28 @@
...
+  "properties": {
+    "id": {
+      "type": "string",
+      "description": "Unique identifier for the tier condition."
+    },
+    "description": {
+      "type": "string",
+      "description": "Short and succinct explanation of the tier participation condition (e.g. “Free to join”)."
+    },
+    "condition_type": {
```
lemonmade, line 17: Why does the platform need to know this? I feel like this list is probably not comprehensive, but will require breaking changes to add to in the future. What will a host do with this information, beyond just printing `description`?

docs/specification/loyalty.md
lemonmade: If we leave the `applies_on` here, when could it point to more than one resource in checkout?

docs/specification/loyalty.md
lemonmade, line 1: There are some parts of this document that read a little more mechanical than other specification docs in UCP. I’d recommend putting this new spec and its siblings, like fulfillment, discount, and checkout, through your favorite LLM. I would at least expect to see a bit more guidance using the MUST/ SHOULD/ MAY structure of the other specifications.

docs/specification/loyalty.md
```diff
@@ -0,0 +1,243 @@
...
+         "version": "2026-01-23",
+         "extends": "dev.ucp.shopping.checkout",
+         "spec": "https://ucp.dev/2026-01-23/specification/loyalty",
+         "schema": "https://ucp.dev/2026-01-23/schemas/common/loyalty.json"
+       }
+     ]
+   }
+ }
+}
+```
```
lemonmade, line 74: There's no mention/ link to identity linking or `context.eligibility` in here; should there be? I assume most of the application logic in here will hinge on providing one of those two signals to checkout.

source/schemas/common/types/loyalty.json
```diff
@@ -0,0 +1,19 @@
...
+  "properties": {
+    "memberships": {
+      "type": "array",
+      "items": { "$ref": "loyalty_membership.json" },
+      "description": "List of memberships that the business offers or associated with."
+    },
+    "activated_memberships": {
+      "type": "array",
+      "items": { "type": "string" },
+      "description": "List of memberships that the customer is activated with, identified by the `id` of the Loyalty Membership object."
```
lemonmade: Should this be `ucp_request: omit`?

source/schemas/common/types/loyalty.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/common/types/loyalty.json",
+  "title": "Loyalty",
+  "description": "Container for loyalty memberships info.",
+  "type": "object",
+  "properties": {
+    "memberships": {
```
lemonmade, line 9: Should this field be required?

source/schemas/common/types/loyalty_membership.json
```diff
@@ -0,0 +1,55 @@
...
+      "items": {
+        "$ref": "membership_tier.json"
+      },
+      "description": "List of tiers associated with the loyalty membership.",
+      "ucp_request": "omit"
+    },
+    "activated_tiers": {
+      "type": "array",
+      "items": { "type": "string" },
+      "description": "List of tiers that the customer is activated for or wish to activate, identified by the `id` of the Membership Tier object."
```
lemonmade, line 45: Should this be `ucp_request: omit`?

docs/specification/loyalty.md
lemonmade: I would love to see an input example for specifying an reward balance redemption, though — it seems like that is brand new input as part of this capability.

lemonmade (commented):
Learned a lot about the domain, and this generally hung together for me as someone unfamiliar with the ins and outs of loyalty.

source/schemas/common/types/membership_tier_condition.json
ziwuzhou-google: As you pointed out, the purpose is really for displaying the condition. In the flow of membership sign-up, platform needs to at least display some short excerpt of sign up requirement (both due to legal requirement and as an upsell), and provide a full link of terms & conditions to fully explain the nitty-gritty in case user wants to learn more.

source/schemas/common/types/loyalty.json
ziwuzhou-google: Request can set this field to request for membership sign up like below (multiple memberships in hypothetical case):

```json
"loyalty": {
  "memberships": [
    {
      "id": "membership_1",
      "activated_tiers": ["tier_1"]
    },
    {
      "id": "membership_2",
      "activated_tiers": ["tier_2"]
    }
  ],
  "activated_memberships": ["membership_1", "membership_2"]
}
```

Maybe the argument is: with `activated_tiers` + `id` set in the request, do we still need to set `activated_memberships` (as it seems redundant with the `id` field). My thought is yes / no. Having it available in the request can help the business to cross-check the platform is indeed requesting for `"membership_1", "membership_2"`, and this field is optional in essence (correct me if I'm wrong here as we don't have it in the `required` field for the entire `loyalty` object).

source/schemas/common/types/loyalty_membership.json
ziwuzhou-google: See one of the above comments. This should not as it will allow the platform to request for membership sign-up, and in the request it needs to specify which tier they are trying to sign up for.

source/schemas/common/types/loyalty.json
ziwuzhou-google: Good catch. It should!

docs/specification/loyalty.md
ziwuzhou-google: 1). I also share the same feeling about duplicated message (i.e. `title` here v.s. `title` within the referenced `discounts.applied` field). My thinking is we can put some guidance in our spec documentation that asks platform MUST use one and only one of it in this case (and I prefer the one referenced applied discount).  Splitting that referencing achieved by `applies_on` out of `benefit` makes it hard to understand the linkage between the referenced object and benefit within membership. In the meantime, I do think having the full-fidelity information on the membership side will help platform as they can choose to display the corresponding tier name for clarity for example. I just don't see enough damage/confusion caused by this nesting that worth the effort to move it out and lose the "connection".

2). That is a very good point about about `applies_on` being an array. It does not need to be actually. If there is a benefit that touches more than one resource in checkout. it should always be possible to split the benefit into multiple non-overlapping ones. Updated!

docs/specification/loyalty.md
ziwuzhou-google: Thanks for the discussion here. I agree that we should add both `provisional` and `eligibility`, to handle monetary benefits (e.g. member pricing, points redemption) that links to a discount object (more specifically, `discounts.applied` field), AND non-monetary benefits (e.g. points/cashback earning based on transaction amount - 20 pts per dollar spent, 5 miles per dollar in base fare etc), as they all need verification from business side to be honored.

Now the question is where to, at `membership` level, `tier` level, or even `benefit` level? Given eligibility claims for loyalty are verified at a tier level (not membership level as businesses not only need to check if buyer is a member of the program but also which specific tier buyer is in to give corresponding benefits; also not benefit level as normally eligible for a tier means all benefits under that tier are available for the buyer), I personally felt both `provisional` and `eligibility` should be added under `tier`.

Concretely, a request/response would look like this for point redemption:

=== "Request"
```json
    {
      "context": {
        "eligibility": ["com.example.loyalty_tier_1"]
      },
      "line_items": [
        {
          "item": {
            "id": "prod_shirt",
            "quantity": 2,
            "price": 2500
          }
        }
      ],
      "loyalty": {
        "memberships": [
          {
            "id": "membership_1",
            "member_id": "member_id_1",
            "rewards": [
              {
                "currency": {
                  "code": "LST"
                },
                "redemption": {
                  "amount": 100
                }
              }
            ]
          }
        ]
      }
    }
```

=== "Response"

```json
    {
      "discounts": {
        "applied": [
          {
            "title": "Point Redemption",
            "amount": 250,
            "automatic": true,
            "provisional": true,
            "eligibility": "com.example.loyalty_tier_1",
            "priority": 1,
            "method": "each",
            "allocations": [
              {"path": "$.line_items[0]", "amount": 250}
            ]
          }
        ]
      },
      "loyalty": {
        "memberships": [
          {
            "id": "membership_1",
            "member_id": "member_id_1",
            "name": "My Loyalty Program",
            "tiers": [
              {
                "id": "tier_1",
                "name": "GOLD",
                "level": 1,
                "provisional": true,
                "eligibility": "com.example.loyalty_tier_1",
              }
            ],
            "activated_tiers": ["tier_1"],
            "rewards": [
              {
                "currency": {
                  "name": "LoyaltyStars",
                  "code": "LST"
                },
                "balance": {
                  "available": 4500
                },
                "redemption": {
                  "amount": 100,
                  "applies_on": "$.discounts.applied[0]"
                }
              }
            ]
          }
        ],
        "activated_memberships": ["membership_1"]
      },
      "totals": [
        {"type": "subtotal", "display_text": "Subtotal", "amount": 5000},
        {"type": "items_discount", "display_text": "Point Redemption", "amount": -250},
        {"type": "total", "display_text": "Total", "amount": 4750}
      ]
    }
```

@lemonmade @jingyli @sinhanurag does this make sense?

source/schemas/common/types/loyalty.json
lemonmade: I don't think I understood that you intended to have this be mutable at all. To me, it feels odd that the platform would activate memberships on the business by way of a `update_checkout` mutation. I don’t think there are other mutations you can perform on checkout today that would have a side effectin this same way. Is this really an operation we expect to perform mid-checkout, without any additional confirmation/ payment/ follow-up beyond what can fit in a successful checkout update response?

ziwuzhou-google: I might miss some of the setup/syntax in schema definition but this is never intended to be bind with checkout (yes loyalty extension can decorate checkout but not a must). My thought was the loyalty sign-up happened in the upper funnel discovery phase: 1). platform advertise the business loyalty program, potential benefits and there are also info provided about the sign-up requirement; 2). then platform can ask if user want to sign up a particular tier; 3). user said yes with platform sending above mentioned request to business. 

Later in the real checkout flow after use has signed up, platform just send a request like this and it's up to business to verify that eligibility claim:

```json
    {
      "context": {
        "eligibility": ["com.example.loyalty_tier_1"]
      },
      "line_items": [
        {
          "item": {
            "id": "prod_shirt",
            "quantity": 2,
            "price": 2000
          }
        }
      ]
    }
```

lemonmade: OK, I was also expecting that sign up for layalty would happen pre-checkout, so it seems odd that sign up would ever happen through Checkout operations, by specifying the above input. Shouldn’t there be dedicated operations for loyalty membership sign up/ updates?

ziwuzhou-google: Correct there should be one. I think there are lots of open discussion about loyalty membership sign up/ updates around how we should model it. One possibility is we have a high level account management capability (beyond membership specific updates) and loyalty extension is used to decorate it; the other is Loyalty/Membership management has its own capability for all the operations. I don't think this part has been fully nailed down (and it is not the plan of this PR).

douglasborthwick-crypto:
Great work on this extension — the Memberships → Tiers → Benefits → Rewards hierarchy is clean and the `applies_on` JSONPath linkage to discounts is elegant.

Two observations on how this connects to verifiable on-chain state, which I think strengthens the extension:

### 1. `condition_type` should support verifiable on-chain conditions

The current `membership_tier_condition.condition_type` enum is `["free", "fee", "spend", "brand_card"]`. These are all payment-history conditions. But tier eligibility increasingly depends on verifiable state that lives outside the merchant's system. A few examples:

**A Nike employee buys from a partner retailer.** Nike issues an on-chain employment credential to staff wallets. The partner merchant defines a tier condition: "holds a valid Nike employee attestation." At checkout, the wallet auth check confirms the credential exists, hasn't been revoked, and was issued by Nike's attester — the employee gets 20% off without showing a badge, without the merchant seeing any HR data, and without Nike operating a discount code system.

**A liquor store verifies age at checkout.** A trusted identity provider (Coinbase, a government agency, a KYC service) has attested that a wallet belongs to someone over 21. The merchant's enrollment condition requires this credential. At checkout, the signed attestation proves age eligibility — boolean pass/fail, no date of birth exposed, no ID photographed. The same pattern works for age-restricted content, cannabis dispensaries, or any regulated sale.

**A fan token project rewards holders with tiered discounts.** A sports club defines tiers on their token: hold 100 → Bronze (5% off merch), 1,000 → Silver (10%), 10,000 → Gold (20%). At checkout, the wallet balance is checked against ascending thresholds and the highest qualifying tier is returned. This maps directly to your `membership_tier` → `enrollment_conditions` → `benefits` hierarchy.

**An NFT collection grants access to a partner network.** A merchant lists a specific NFT collection — any holder qualifies for a flat benefit. No threshold, just ownership. Membership cards, event tickets, loyalty passes all work this way.

**These compose at checkout.** A shopper might qualify for Gold tier on the merchant's token *and* hold an NFT from a partner collection *and* carry a verified-age credential — all contributing benefits simultaneously. The merchant controls how they stack: take the highest single discount, sum them, or sum with a cap.

All of these conditions share a property: they can be verified cryptographically and the result expressed as a signed attestation. Suggesting either extending `condition_type` with values like `"token_holding"`, `"nft_ownership"`, `"credential"` — or replacing the closed enum with an extensible pattern, since the set of verifiable conditions will only grow. @lemonmade raised the same extensibility concern on the current enum.

### 2. Eligibility claims can carry cryptographic proof

Right now `context.eligibility` is a self-asserted string and `provisional` is a boolean the business sets after internal verification. For on-chain state conditions, the eligibility claim can be backed by a signed attestation — an ES256 JWT verifiable via JWKS — so that `provisional: false` means "here is a cryptographic proof any party can verify offline" rather than "the business checked internally."

This is what we're proposing in #264 (attestation extension) — an `attestations` map on the cart/checkout object keyed by the same reverse-domain eligibility claim, carrying `provider_jwks`, `kid`, `payload`, and `sig`. The business fetches the JWKS, verifies the signature, and reads `payload.pass` — no callback required.

Concretely, a loyalty flow with on-chain tier verification would look like:

```json
{
  "context": {
    "eligibility": ["com.example.loyalty_gold"]
  },
  "attestations": {
    "com.example.loyalty_gold": {
      "provider_jwks": "https://verifier.example/.well-known/jwks.json",
      "kid": "attest-v1",
      "payload": { "pass": true, "tier": "Gold", "attestedAt": "2026-03-26T..." },
      "sig": "base64...",
      "expires_at": "2026-03-26T20:00:00Z"
    }
  },
  "loyalty": {
    "memberships": [{
      "id": "membership_1",
      "activated_tiers": ["tier_gold"],
      "provisional": false,
      "eligibility": "com.example.loyalty_gold"
    }]
  }
}
```

The loyalty extension defines *what tiers and benefits exist*. The attestation extension proves *that the buyer qualifies*. They compose naturally without either depending on the other.

Happy to coordinate on how #264 and this extension reference each other — they're complementary.

source/schemas/common/types/loyalty_membership.json
```diff
@@ -0,0 +1,66 @@
...
+      "description": "Membership enrollment date (RFC 3339).",
+      "ucp_request": "omit"
+    },
+    "end_time": {
+      "type": "string",
+      "format": "date-time",
+      "description": "Membership end/expiration date (RFC 3339).",
+      "ucp_request": "omit"
+    },
+    "tiers": {
```
maximenajim, line 34: Tier model assumes linear hierarchy.  The tiers array with level: integer (min 1) and activated_tiers forces a ranked ladder model. Target Circle members can simultaneously hold Free membership or Circle Card or Circle 360 or a combination of these— these composable membership types, not levels 1/2/3.

Suggested change: Add a type field to membership or tier to distinguish program structures:

```
{
  "membership_type": {
    "type": "string",
    "enum": ["program", "payment_instrument", "subscription", "verification"],
    "description": "The category of this membership. Programs may be composed together."
  },
  "composable": {
    "type": "boolean",
    "default": false,
    "description": "Whether this membership can be held concurrently with other memberships."
  }
}

```
Alternatively, make tiers optional (it's currently required) and support flat membership structures that don't use tiers at all.

source/schemas/common/types/membership_balance.json
```diff
@@ -0,0 +1,37 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/common/types/membership_balance.json",
+  "title": "Membership Balance",
+  "description": "The overall balance information of the membership.",
+  "type": "object",
+  "required": ["available"],
+  "properties": {
+    "available": {
+      "type": "integer",
```
maximenajim, line 10: Is integer is too restrictive?  The schema uses type: integer for all balance fields but balance_currency.json has no decimal_places or conversion_factor to communicate the unit.
Suggested change: Either change to type: number or add a decimal_places field to balance_currency.json:

source/schemas/common/types/loyalty_membership.json
ziwuzhou-google: Sorry I might missed something, but even with the current proposal it can still support buyer having multi-tier (and getting the benefits of the different tiers simutaneously)? In the last example in the documentation:

```json
      "loyalty": {
        "memberships": [
          {
            "id": "membership_1",
            "member_id": "user_id_1",
            "name": "My Loyalty Program",
            "enrollment_date": "2026-01-12T00:00:00Z",
            "last_activity_date": "2026-01-12T00:00:00Z",
            "end_date": "2036-01-12T00:00:00Z",
            "tiers": [
              {
                "id": "tier_1",
                "name": "GOLD",
                "level": 1,
                "benefits": [
                  {
                    "id": "BEN_001",
                    "title": "Early access",
                    "description": "24-hour early access to seasonal sales",
                  }
                ]
              },
              {
                "id": "tier_2",
                "name": "CREDIT_CARD",
                "level": 2,
                "benefits": [
                  {
                    "id": "BEN_001",
                    "title": "Membership credit card holder saves 5%",
                    "description": "Members receive $5 discount on eligible products",
                    "applies_on": "$.discounts.applied[0]"
                  },
                  {
                    "id": "BEN_002",
                    "title": "Early access",
                    "description": "24-hour early access to seasonal sales",
                  }
                ]
              }
            ],
            "activated_tiers": ["tier_1", "tier_2"],
            "provisional": false,
            "eligibility": "com.example.loyalty_gold"
          }
        ]
      }
```
To me level is just a numeric identifier of the position of the tier within the system, rather than a ranked ladder (maybe the naming is a bit misleading). Also, I felt we shouldn't view Circle (the free one), Circle 360 and CircleCard as 3 different memberships but rather 3 different tiers within the larger Target Circle membership system?

source/schemas/common/types/membership_reward.json
```diff
@@ -0,0 +1,26 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/common/types/membership_reward.json",
+  "title": "Membership Reward",
+  "description": "Quantifiable and collectable reward of loyalty membership of the customer.",
```
maximenajim, line 5: No support for non-fungible / one-time-use rewards.  Target's birthday reward is 5% off a single purchase (one-time use, percentage-based, 30-day validity). The reward model assumes fungible balances with integer amounts. There's no way to express:

- One-time-use rewards vs. balance-based
- Percentage-based rewards vs. point amounts
- Personal triggers (birthday, anniversary)

source/schemas/common/types/loyalty_membership.json
```diff
@@ -0,0 +1,66 @@
...
+      },
+      "description": "List of quantifiable rewards value the user holds. Each object encapsulates one type of reward the membership offers."
+    },
+    "provisional": {
+      "type": "boolean",
+      "default": false,
+      "description": "True if this membership requires additional verification.",
+      "ucp_request": "omit"
+    },
+    "eligibility": {
```
maximenajim, line 60: Should this be an array? Business has multiple concurrent verification-based discounts: military, student, teacher — each with different seasonal availability and discount mechanics. Can a single eligibility claim reference represent this?

libcairo.2.dylib
```diff
@@ -0,0 +1 @@
+/opt/homebrew/lib/libcairo.2.dylib
```
maximenajim: Was this committed accidentally?

ziwuzhou-google: Yup, already reverted in my latest commit.

docs/specification/loyalty.md
ziwuzhou-google: Yup added one dedicated session for how context can work together with this extension.

source/schemas/common/types/membership_balance.json
ziwuzhou-google: Ran some literature research and I do find some loyalty programs offering points bearing non-integer amount. Added a decimal_places field as suggested to balance_currency:

```json
    "decimal_places": {
      "type": "integer",
      "default": 0,
      "description": "The position of a digit to the right of a decimal point. Applies to all amount related fields for balance."
    }
```

source/schemas/common/types/loyalty_membership.json
ziwuzhou-google: It should still be a singular field. This eligibility holds the claim for the membership itself. If there're member-only benefits that requires further verification (student, military, etc), the business should publish another eligibility claim like `com.example.loyalty.student` that platform can put in the request context.eligibility (on top of `com.example.loyalty` claim for membership). This `com.example.loyalty.student` eligibility will then be used at the corresponding discount's eligibility field (i.e. `discount.applied.eligibility`).

source/schemas/common/types/membership_reward.json
ziwuzhou-google: Thanks for the insight. In this initial version of proposal, `rewards` really just hold the accumulable balances and not like a wallet that holds rewards/coupons with various constraints (time/further eligibility on top of membership). In the meantime, I think business can rely on the `benefits` array to communicate these rewards, run the required verification to decide when and if they should be applied?

source/schemas/common/types/loyalty_membership.json
maximenajim: To avoid the confusion maybe we need to make tiers optional? Or we add a membership_type or similar field to distinguish between hierarchical ladder programs and composable/flat membership structures?

source/schemas/common/types/membership_balance.json
maximenajim: Thanks but all balance fields remain type: integer and balance_currency.json has no decimal_places or conversion_factor. How should a loyalty currency that uses fractional values (e.g., 2.5 points per dollar, or cash-back balances like $4.73) be represented?I think we either switch to type: number or add a decimal_places field to balance_currency.json so implementers know how to interpret the integer (e.g., 1000 with decimal_places: 2 = 10.00).

source/schemas/common/types/membership_reward.json
maximenajim: Target's birthday reward (5% off a single purchase, 30-day validity) cannot be expressed in the current model because: (a) rewards are modeled as fungible integer balances, not one-time-use vouchers; (b) there's no way to express percentage-based vs. fixed-amount rewards; (c) there's no trigger/occasion concept (birthday, anniversary, milestone). Should we add a reward_type enum (e.g., "balance", "voucher", "coupon") and optional fields for usage_limit, discount_type ("fixed" / "percentage"), and valid_from/valid_until? Or is the intent that these one-time rewards be modeled through the discounts extension instead? If so, that mapping should be documented.

source/schemas/common/types/loyalty_membership.json
ACSchil: To me tier implies a hierarchy. My recommendation is we do not force hierarchical memberships in the spec. One could have two memberships that have _different_ benefits, but one is not a "higher" tier to the other.

source/schemas/common/types/membership_balance.json
ziwuzhou-google: Maybe I missed something but the latest commit should have it added in the balance_currency: https://github.com/Universal-Commerce-Protocol/ucp/pull/251/changes#diff-4bdb722c7b032db00e47a67dfdd1529f73904ed79774a37a6cf0b1d1ee4e8410R18

draspall:
Superseded by [340 feat: loyalty extension for catalog/cart/checkout capability](https://github.com/Universal-Commerce-Protocol/ucp/pull/340)

PR #254, opened 2026-03-11
feat!: update Order capability

richmolj:
# Description

Collection of updates to the Order capability to make the spec more flexible, robust, and clear.

- Soften "append-only" on adjustments and "immutable" on line items
- Remove "derived from events" quantity language
- Model Order:Checkout as 1:N via `checkouts` array (order edits create new sessions)
- Add signed quantities/amounts on adjustments and `quantity.original` on line items
- Rename adjustment `amount` to `totals` for consistency with Order and OrderLineItem

## Type of change

      or fields**)
- [x] Documentation update

---

- `checkout_id` (string) replaced with `checkouts` (array of `{ id, created_at }` objects) — orders can reference multiple checkout sessions via edits and exchanges
- `amount` on Adjustment replaced with `totals` (array of Total objects) — aligns with the pattern used by Order and OrderLineItem
- `minimum: 1` removed from adjustment line item quantities — signed values needed for returns and exchanges
- `minimum: 0` removed from `amount.json` — amounts can be negative for refunds, credits, and adjustment totals across all capabilities

---

## Checklist

### DIFF
--- a/source/schemas/shopping/types/adjustment.json
+++ b/source/schemas/shopping/types/adjustment.json
@@ -54,19 +54,7 @@
     "totals": {
       "type": "array",
       "items": {
-        "allOf": [
-          {
-            "$ref": "total.json"
-          },
-          {
-            "type": "object",
-            "properties": {
-              "amount": {
-                "$ref": "signed_amount.json"
-              }
-            }
-          }
-        ]
+        "$ref": "total.json"
       },
       "description": "Adjustment totals breakdown. Signed values - negative for money returned to buyer (refunds, credits), positive for additional charges (exchanges)."
     },


source/schemas/shopping/order.json
```diff
@@ -39,9 +39,25 @@
...
+    "checkouts": {
+      "type": "array",
+      "minItems": 1,
+      "items": {
+        "type": "object",
+        "required": ["id", "created_at"],
+        "properties": {
+          "id": {
+            "type": "string",
+            "description": "Checkout session identifier."
```
alexpark20:
```suggestion
            "description": ""Unique identifier of the checkout session."
```

To be consistent with checkout docs / schema desc

source/schemas/shopping/order.json
```diff
@@ -39,9 +39,25 @@
...
+          },
+          "created_at": {
+            "type": "string",
+            "format": "date-time",
+            "description": "RFC 3339 timestamp when this checkout session was created."
+          }
```
alexpark20: is the purpose of created_at to identify the sequential order of multiple checkouts? If so, do we need to include updated_at as well?

source/schemas/shopping/types/adjustment.json
```diff
@@ -45,16 +45,18 @@
...
       },
       "description": "Which line items and quantities are affected (optional)."
     },
-    "amount": {
-      "$ref": "amount.json",
-      "description": "Amount in ISO 4217 minor units for refunds, credits, or price adjustments."
+    "totals": {
+      "type": "array",
+      "items": {
+        "$ref": "total.json"
```
alexpark20, line 57: I don't think the current total and amount schema allows negative - should update them as part of this PR?

source/schemas/shopping/order.json
alexpark20: Or is the purpose here just to distinguish the original checkout and edit/exchange checkouts ?

docs/specification/order.md
```diff
@@ -31,17 +31,20 @@ Orders have three main components:
 **Line Items** — what was purchased at checkout:
 
 * Includes current quantity counts (total, fulfilled)
+* Can change post-order (e.g. order edits, exchanges)
```
jingyli: With order edits now in the picture, can we add a guideline around what `line_items` **SHOULD** include post these changes, is it:
1) A comprehensive list of all the items that once existed in the order, even if they were altered/removed via an edit?
OR 
2) Only contain remaining items that are present after the latest edit?

My hunch is that 1) is more comprehensive from a data/audit perspective given there may have been `adjustments` that reference back to these altered products?

docs/specification/order.md
```diff
@@ -50,7 +53,7 @@ Orders have three main components:
 Line items reflect what was purchased at checkout and their current state:
 
 * Item details (product, price, quantity ordered)
-* Quantity counts and status are derived
+* Quantity counts and fulfillment status
```
jingyli, line 59: Can you clarify what this change mean here given I think `status` of the line_item is still derived based on it's relationship with `quantity.total` and `quantity.fulfilled`?

docs/specification/order.md
```diff
@@ -83,15 +86,17 @@ Expectations can be split, merged, or adjusted post-order. For example:
...
 
 * Type is an open string field - businesses can use any values that make sense
   (typically money movements like `refund`, `return`, `credit`,
   `price_adjustment`, `dispute`, `cancellation`)
 * Can be any post-order change
 * Optionally link to line items (or order-level for things like shipping refunds)
-* Include amount when relevant
+* Quantities and amounts are signed—negative for reductions (returns, refunds),
+  positive for additions (exchanges)
+* Include totals breakdown when relevant
```
jingyli, line 102: Trying to visualize how `totals` can be used here:
- For a full refund, I can see this would be very similar to the `totals` construct in checkout.
- However, for things like partial refund or any credit/price adjustment (i.e. resulting in an arbitrary money movement), I fail to fully understand what should be returned in the array here - if we have `type: TOTAL` for these scenarios here, then it "violates"/"invalidates" our general rule of how `total` should be calculated that currently is present in `totals.json`.

Alternatively, we can consider adding another categorization into `totals.type` for `adjustment`?

source/schemas/shopping/types/order_line_item.json
```diff
@@ -23,18 +23,23 @@
       "type": "object",
       "required": ["total", "fulfilled"],
       "properties": {
+        "original": {
```
jingyli, line 26: It's not super clear about the use case we have in-mind for this field (especially given other `quantity` fields are used for deriving `status`).

Tied with a comment below, I think maybe one use case is to use it along with `quantity.total` to under if `status` should be in `cancelled` state?

docs/specification/order.md
```diff
@@ -103,15 +108,15 @@ fulfillment:
...
-Status and quantity counts should reflect the event logs.
 
 {{ schema_fields('order_line_item', 'order') }}
 
 **Quantity Structure:**
 
 ```json
 {
-  "total": 3,      // Current total quantity
+  "original": 3,   // Quantity at checkout
```
jingyli: nit: Maybe we should clarify a bit further that this is the checkout with the oldest `created_at` timestamp (and not just any checkout)?

source/schemas/shopping/types/order_line_item.json
jingyli: I think we will need some more status or at least add some more conditions to this described logic to cover the following 2 scenarios:
1) If order is cancelled prior to fulfillment, then by this logic today, item-level status will be stuck in `processing`. We should probably add some terminal state to catch `cancelled`.
2) If an item is fulfilled, then edited from an exchange, then I'd imagine the old item that got exchanged would have `totals = 0` but `fulfilled > 0`, in that case, what status should it be in (maybe we should have `fulfilled if quantity.fulfilled >= quantity.total` instead)?

source/schemas/shopping/order.json
richmolj: `created_at` is there so the sequence is clear. Checkouts are frozen after creation - they are never updated - so `updated_at` is not needed here.

source/schemas/shopping/types/adjustment.json
richmolj: Good catch. Removed `minimum: 0` from `amount.json` so negative values are allowed for signed adjustment totals.

docs/specification/order.md
richmolj: Option 1 - comprehensive list. Line items should include all items that ever existed on the order, even if altered/removed via an edit. This is why `quantity.original` exists on line items - it preserves what was originally ordered at checkout, even when `quantity.total` changes to 0 after an edit. Adjustments can then always reference back to these items.

docs/specification/order.md
richmolj: The change here is removing the claim that quantities are "derived from events" - merchants may not send events, so we cannot say that is where the values come from, even though that may be the case the majority of the time. Status is still derived from the quantity fields, and that logic is still documented in the Status Derivation section below.

docs/specification/order.md
richmolj: `totals` on an adjustment is scoped to that adjustment, not the order. A partial refund of $5 would just be `[{ "type": "total", "amount": -500 }]`. If the business wants to break it down further (e.g. the subtotal and tax portions of the refund separately), they can, but a single `total` entry works fine for simple cases. The calculation rules in `totals.json` apply to order-level and checkout-level totals, not adjustment-level.

source/schemas/shopping/types/order_line_item.json
richmolj: Primary use case is what we discussed above - line items are a comprehensive list that includes items removed via edits. `quantity.original` lets platforms see what was originally ordered without fetching the Checkout record. For example, an item edited down to 0 would have `original: 2, total: 0, fulfilled: 0` - the platform can see it was a real item that got removed, not a phantom entry.

docs/specification/order.md
richmolj: Good point. This refers to the original checkout (the one with the earliest `created_at`). Updated the description to clarify.

source/schemas/shopping/types/order_line_item.json
richmolj: Good observations. The status derivation documented here covers the common case - it is not meant to be exhaustive. The spec treats status as an open string, so businesses can use additional values like `cancelled` or `exchanged` for these scenarios. We will keep the derivation simple for now and can expand the documented examples in a follow-up if needed. Does that sound good?

source/schemas/shopping/types/amount.json
```diff
@@ -3,6 +3,5 @@
   "$id": "https://ucp.dev/schemas/shopping/types/amount.json",
   "title": "Amount",
   "description": "Monetary amount in the currency's minor unit as defined by ISO 4217. Refer to the currency's exponent to determine minor-to-major ratio (e.g., 2 for USD, 0 for JPY, 3 for KWD).",
-  "type": "integer",
-  "minimum": 0
```
igrigorik, line 7: @richmolj I think this is a footgun. I understand the motivation, but removing this constraint on shared type opens up a class of unexpected scenarios — e.g. negative prices on line items, etc. 

If we move forward with allowing negative values, I think we need to be more surgical. PTAL @ https://github.com/Universal-Commerce-Protocol/ucp/pull/261#commits-pushed-6fd2c53, we can use similar approach here.

igrigorik (changes requested):
See comment on removing `"minimum": 0` on amount.json.

source/schemas/shopping/types/amount.json
richmolj: Sure, I've updated the PR with `signed_amounts.json`.

That said, I disagree. The current field requires mental math to understand `type = discount` means negative, why not make this explicit? Negative line item prices, or negatives from other business-specific types, should be up to the business logic not the spec.

I don't feel strongly here, though, and it's a more significant breaking change. So went ahead and updated.

source/schemas/shopping/order.json
lemonmade: Should we just rely on order here instead? If we really want to add `created_at`, it feels like we should add it to `checkout` too, so there is no information you get about the checkout on order you could not get directly.

source/schemas/shopping/types/order_line_item.json
```diff
@@ -23,18 +23,23 @@
...
+          "type": "integer",
+          "minimum": 0,
```
lemonmade, line 28: Could all of the properties in here use the https://ucp.dev/schemas/shopping/types/amount.json shared type?

source/schemas/shopping/types/signed_amount.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/signed_amount.json",
+  "title": "Signed Amount",
+  "description": "Monetary amount in the currency's minor unit as defined by ISO 4217. Refer to the currency's exponent to determine minor-to-major ratio (e.g., 2 for USD, 0 for JPY, 3 for KWD). May be negative — the sign is intrinsic to the value (e.g., discounts are negative, charges are positive).",
+  "type": "integer"
+}
```
lemonmade, line 7: I think this is in `main` now, a rebase should remove the diff.

generated/schema-types.ts
lemonmade, line 1: It looks like we haven't run this in quite awhile, I am not even sure why it is still here. Is there a reason you had to update it for this PR?

docs/specification/order.md
jingyli: Thanks, makes sense, is it possible to actually add this on top of this bullet here?

i.e. Something along the lines of `; **MUST** include all line items that ever existed on the order regardless of edits or alterations`?

source/schemas/shopping/order.json
```diff
@@ -39,9 +39,25 @@
       "type": "string",
       "description": "Unique order identifier."
     },
-    "checkout_id": {
-      "type": "string",
-      "description": "Associated checkout ID for reconciliation."
+    "checkouts": {
```
jingyli: Looking at this a bit more, a naive QQ: What do we feel is the value by having this full array (especially as a required field)?

The reason I question this is 2-fold:
1) This seems to suggest an additional dependency between UCP's checkout & order capabilities. What does it mean for a platform (like a post-order management agent) that only implements/supports order capability in UCP? To them, the concept of `checkout_id` is fairly useless since they don't really need/require knowledge on the upstream process.
2) Per [this guideline](https://ucp.dev/latest/specification/checkout/#get-checkout), there is no guarantee that business will persist the checkout session after completion so I'm unsure what can the platform truly do with this full list of checkout_ids (having the initial one makes sense to me for the reconciliation use case). It also feels a bit odd conceptually since we don't generally go back from order to checkout.

source/schemas/shopping/types/order_line_item.json
jingyli: The explanation makes sense, the only nit-pick from me would be if we can represent this "open string is accepted" concept properly here? My understanding of JSON schema is that as soon as we have `enum` key, it restricts the value to a finite fixed value set where no further extension is accepted (which is the case here).

docs/specification/order.md
jingyli: This makes sense (especially post PR#261 where we have now removed the formula anchored on `type = total`)!

source/schemas/shopping/order.json
gsmith85: I agree with @lemonmade here on the checkouts array structure feeling askew, I'll attempt to summarize my concerns:

1. Denormalization: Including `created_at` in the array items feels leaky. If the checkout resource is the source of truth, we should rely on a timestamp within.
2. Ordering Contract: I understand the path to including `created_at` to imply an ordering but, even with it, there is no _actual_ schema enforcement.  If the client really cares about ordering they should double-check it, this is not expensive even if it's a 1:1000 relationship.
3. Naming: `created_at` is misleading here. I assume we should only be referencing checkouts that reached a terminal (success) state.  Should this be `finalized_at` or `executed_at` instead?
4. Dereferencing: I'm not persuaded that we want to include the full checkout object rather than just references.  An ordered array of checkout IDs provides the same lineage, and optionality to the client to fetch the full objects (or not).

gsmith85:
I have a significant concern regarding the domain mismatch introduced by using the checkouts array to handle post-purchase modifications, I suggest escalating it to the TC:

- Domain Mismatch & Incongruous Properties: A Checkout is a session-based negotiation primitive; an exchange is a persistent accounting entry. Shunting the latter into a Checkout forces a tool built for point-in-time discovery to perform ledger reconciliation logic it wasn’t designed for. This causes core properties to become semantically askew; for example, `total` must shift from representing gross intent (the absolute cost of items) to a delta (the balance of a swap). This can break standard validation logic, where a checkout engine expects a positive total and a payment method, whereas an exchange accounting session may result in a zero or negative balance.
- Functional Divergence: Even if a validation session is modeled with a Checkout, the use cases are fundamentally different. For an exchange, the Checkout is a point-check for cost and availability. Forcing the Order to act as a permanent folder for these IDs creates a dependency on transient reconciliation data that may not remain accessible in the same way the primary order record does.
- Complexity: The proposed change to `Orders` adds considerable complexity to the object and erodes desirable properties like the append-only log for modifications. It forces the client to re-calculate the current truth by replaying a list of session intents rather than following a clean log of finalized adjustments.

I would propose, we should lean on the Adjustments capability (which already supports signed quantities/amounts) to handle edits, returns, and exchanges. This keeps the Order as a clean, fulfillment-centric record. If we must link to a validation session for audit purposes, we should do so via a `source_checkout_id` reference on the Adjustment itself, rather than at the Order root.

source/schemas/shopping/order.json
richmolj: The main use case would have been post-purchase upsells, but [removed from this PR](https://github.com/Universal-Commerce-Protocol/ucp/pull/254#issuecomment-4120558459) and we can tackle that in a separate one.

source/schemas/shopping/types/order_line_item.json
richmolj: Sorry I think some terms may have gotten muddled here. `Adjustment.type` [is the open string](https://github.com/Universal-Commerce-Protocol/ucp/blob/434bb0cd6f711ceae2ad4ae489c5e4b4987f9226/source/schemas/shopping/types/adjustment.json#L20) that can have arbitrary values like `cancellation` or `exchange`. The breakdown of `line_item.quantity` I feel is a true enum (`original/total/fulfilled`). IOW the events that _cause_ quantity to change are freeform, but quantity itself is a defined enum.

Is there value in taking this enum and turning it into an open string?

source/schemas/shopping/types/order_line_item.json
richmolj: These are quantities, not monetary amounts, so I don't think `amount.json` would be the right type here.

source/schemas/shopping/types/signed_amount.json
richmolj: Done, rebased.

generated/schema-types.ts
richmolj: Dropped — it was deleted on main and the rebase removed it from this branch.

source/schemas/shopping/types/adjustment.json
```diff
@@ -45,16 +45,30 @@
...
+            "description": "Signed quantity affected by this adjustment. Negative values represent reductions (e.g. returns); positive values represent additions (e.g. exchanges)."
           }
         }
       },
       "description": "Which line items and quantities are affected (optional)."
     },
-    "amount": {
-      "$ref": "amount.json",
-      "description": "Amount in ISO 4217 minor units for refunds, credits, or price adjustments."
+    "totals": {
```
jingyli, line 54: Can we just use `totals.json` directly here instead of rebuilding the `allOf` relationship?

One benefit of reusing `totals.json` here is that if this is set, it also enforces the array needing exactly 1 `total` and 1 `subtotal` so it brings consistency.

source/schemas/shopping/types/order_line_item.json
jingyli: So there are 3 things:
1) `adjustment.type`: aligned -> open string enum value
2) `order_line_item.quantity`: aligned -> fixed enum value (`original/total/fulfilled`)
3) `order_line_item.status`: what this comment is meant for and an open discussion remains -> right now it's defined as a fixed enum value (`processing/partial/fulfilled`), of which my concern is that it creates odd semantics per the original comment (i.e. `status = processing` forever if order/item is cancelled prior to fulfillment)

I would propose either we update the description describing how `status` derivation should take place or we add more enum values to the set so it makes more sense (another extreme alternative we can consider is dropping this field entirely and have platform relies on `fulfillment.events` and `adjustments` purely).

jingyli (commented):
Thanks for the updates! Just left 2 more comments - the only concern I have on PR remains on the `order_line_item.status` representation, especially now that `order_line_item.quantity` is extended to support 3 types.

source/schemas/shopping/types/adjustment.json
richmolj: `totals.json` enforces exactly one `subtotal` and one `total` entry, which is too strict for adjustments - a simple refund might just be `[{ "type": "total", "amount": -500 }]` without a subtotal breakdown. The `allOf` here lets adjustments reuse the `total.json` item shape with signed amounts but without the array-level constraints.

source/schemas/shopping/types/order_line_item.json
richmolj: Added `removed` to cover the `quantity.total == 0` case (edits, exchanges, cancellations). Updated the derivation to check for it first. See latest push.

sinhanurag:
I need to chat with a couple of folks before approving these changes. Will comment and close out by Monday EOD.

docs/specification/order.md
```diff
@@ -31,17 +31,20 @@ Orders have three main components:
...
 * **Events** (append-only log) — what actually happened (e.g. 👕 was shipped)
 
-**Adjustments** (append-only log) — post-order events independent of fulfillment:
+**Adjustments** — post-order events independent of fulfillment:
 
 * Typically money movements (refunds, returns, credits, disputes, cancellations)
 * Can be any post-order change
 * Can happen before, during, or after fulfillment
+* Businesses may append new entries or update existing ones in place
+  (e.g. a single `return` adjustment can transition from `pending` to `completed`)
```
gsmith85, line 50: @richmolj, I feel like we're really close here, thanks for going deep on this one with me!

On re-review, I think we can maintain the append-only ledger while still achieving the flexibility you're pursuing. To do that, we could use a "last state wins" strategy for adjustments.

In the provided example, this would involve appending a net-new Adjustment for the `completed` return state that includes a `replaces_adjustment_id` referencing the `pending` one. Clients just read the newest entry to get the current status, but the protocol preserves the state transitions.

This avoids implementation inconsistencies across businesses where adjustment revisions might be handled differently. More importantly, it prevents data loss. If an adjustment is created on Monday as pending and finalized on Wednesday as completed, mutating the `status` and `occurred_at` in place effectively erases the historical timeline.

source/schemas/shopping/types/order_line_item.json
```diff
@@ -23,18 +23,23 @@
...
+          "description": "Quantity from the original checkout."
+        },
         "total": {
           "type": "integer",
           "minimum": 0,
-          "description": "Current total quantity."
+          "description": "Current total quantity. May differ from original due to order edits or exchanges."
```
gsmith85: Overall like the `Quantity` object modifications, but I think the descriptions here could benefit from some minor clarification to avoid conflating cart actions with adjustments.

For `quantity.total`, in a post-order context, an exchange usually means returning this line item (dropping its total to 0) and appending a new line item for the replacement SKU. It might be safer to describe this as:

> Current total active quantity. May differ from original due to post-order adjustments (e.g., returns or cancellations)."

source/schemas/shopping/types/order_line_item.json
```diff
@@ -48,9 +53,10 @@
       "enum": [
         "processing",
         "partial",
-        "fulfilled"
+        "fulfilled",
+        "removed"
```
gsmith85, line 57: For `status = removed`: Removed is more aligned with a cart operation. Since this is an `Order`, if `quantity.total == 0`, it means the item was either cancelled (before fulfillment) or returned (after fulfillment). Should we update the enum/description to reflect those more specific states?

richmolj: I see the intent here as "what is outstanding vs completed vs no longer relevant". Fulfillment events and adjustments can give the details on how that state is derived, but this enum is just the state itself.

docs/specification/order.md
richmolj: I think it's fine to encourage append-only, but businesses have existing systems that may not work this way and should not have to rewrite their systems in order to adopt UCP.

source/schemas/shopping/types/order_line_item.json
richmolj: Good call, I took this suggestion but used "modifications" instead of "adjustments" to avoid naming confusion with the `adjustments` field.

docs/specification/order.md
igrigorik: Agree with both positions. I think current text is right, we can't mandate append-only, but we _can_ make a firmer preference without enforcing it as hard constraint. E.g...

> * Businesses SHOULD append new entries rather than mutating existing ones; append-only ledger is preferred. Businesses that do not maintain adjustment history MAY perform in-place updates of existing entries.

source/schemas/shopping/types/order_line_item.json
igrigorik: This tripped me as well, but I think I agree with Lee..

This enum describes fulfillment status for this line: processing/partial/fulfilled answer "where is this in the fulfillment lifecycle"; `removed` says "nowhere — it's no longer relevant". The "why" (cancelled, returned, exchanged) lives in adjustments.

source/schemas/shopping/types/order_line_item.json
```diff
@@ -48,9 +53,10 @@
...
       ],
-      "description": "Derived status: fulfilled if quantity.fulfilled == quantity.total, partial if quantity.fulfilled > 0, otherwise processing."
+      "description": "Derived status: removed if quantity.total == 0, fulfilled if quantity.fulfilled == quantity.total, partial if quantity.fulfilled > 0, otherwise processing."
```
igrigorik: Subtle issue here: if total==0 then ordering of how you implement these matters. To make it bulletproof...

```suggestion
      "description": "Derived status: removed if quantity.total == 0, fulfilled if quantity.total > 0 and quantity.fulfilled == quantity.total, partial if quantity.total > 0 and quantity.fulfilled > 0, otherwise processing."
```

richmolj: Good call, updated 👍

PR #260, opened 2026-03-12
fix: promote reverse_domain_name to standalone type

igrigorik:
Fixes [current mkdocs build failures](https://github.com/Universal-Commerce-Protocol/ucp/actions/runs/23012845260/job/66827982052?pr=250#step:13:22).

Extract reverse_domain_name from ucp.json $defs into shopping/types/reverse_domain_name.json. Update all propertyNames refs in ucp.json to point to the standalone file directly.

The docs renderer generates page-local anchors for $defs entries (prefixed with the parent schema name), but standalone types get rendered on the reference page with a linkable section. The $defs indirection caused broken #ucp-reverse-domain-name anchors on checkout, cart, catalog, and reference pages.

## Type of change

### DIFF
--- a/source/schemas/shopping/types/reverse_domain_name.json
+++ b/source/schemas/shopping/types/reverse_domain_name.json
@@ -1,0 +1,8 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/reverse_domain_name.json",
+  "title": "Reverse Domain Name",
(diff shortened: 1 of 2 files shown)


source/schemas/shopping/types/reverse_domain_name.json
wry-ry, line 1: Since this will be used across all domains, would it make sense for this to live outside of shopping/, maybe directly under schemas/?

igrigorik: Agree, but right now it would be an outlier at that level.. let's defer this as a refactor (there are other types we'll want to move too) until we have more domains beyond shopping.

wry-ry: SGTM

Issue #219, opened 2026-02-26
Clarify itemized fee support in totals[]


maximenajim:
# Context
The UCP total.json schema already defines "fee" as a valid type enum value in the totals[] array, but the specification does not document:
- Whether multiple type: "fee" entries are permitted in the same totals[] array
- How platforms should render multiple fees (collapsed vs. itemized)
- How fees interact with the total calculation formula
- Whether fees can appear at different levels (checkout, line item, fulfillment option)
Businesses implementing UCP have no guidance on how to express itemized fees (e.g., service fees, handling fees, recycling fees, convenience fees).

# Proposal
We propose documenting the following semantics for type: "fee" entries. 

1. Multiple fee entries are allowed
A business MAY include multiple type: "fee" entries in the same totals[] array. Each entry represents a distinct, itemized charge:
```
{
  totals: [
    {type: subtotal, display_text: Subtotal, amount: 5000},
    {type: fee, display_text: Service Fee, amount: 200},
    {type: fee, display_text: Handling Fee, amount: 150},
    {type: fee, display_text: Gift Wrap Fee, amount: 350},
    {type: fulfillment, display_text: Standard Shipping, amount: 500},
    {type: tax, display_text: Tax, amount: 520},
    {type: total, display_text: Total, amount: 6720}
  ]
}
```
2. Total calculation formula includes all fees
`total = subtotal - discount + fulfillment + tax + fee₁ + fee₂ + ... + feeₙ`

3. Fees at multiple levels
Fees MAY appear at any level where totals[] exists:
- Checkout-level: `$.totals[]`
- Line-item-level: `$.line_items[].totals[]`
- Fulfillment-option-level: `$.fulfillment_options[].totals[]`

# Key Design Question: Should fees be a formal Extension like Discounts?
## Option A: Document within existing totals[] 
Fees remain a type value in the existing Total schema. No new schema files, no capability registration. This is the minimal approach — it clarifies what's already structurally possible.

Pros:
- No new schemas
- Fees are conceptually simpler than discounts (no codes, no stacking, no allocations)
- Already works today — just needs documentation

Cons:
- No structured metadata (no way to express fee reason, legal / regulatory disclosures, tax applicability, whether fee is waivable, etc.)
- No capability discovery — platforms can't know in advance if a business will include fees
- No allocation breakdown — can't express "this fee is because of line item X"
- Limited to display_text + amount — not machine-readable

## Option B: Create a Fee Extension
[...]

PR #261, opened 2026-03-12
feat: formalize totals contract

igrigorik:
Related context: #219 and #245

On first pass, [fees-as-extension](https://github.com/Universal-Commerce-Protocol/ucp/pull/245) seemed "right" to me, but reviewing discussion on the PR led me into a dead end. If fees are modeled as an extension, a business in a jurisdiction that mandates itemized fee disclosure (e.g., state mandated electronics recycling fees), would have to escalate via `continue_url` when the platform doesn't advertise support for the Fee Extension — the platform simply can't render what the business is obligated to show. The same problem applies to taxes: some countries/regions require display of separate tax lines at checkout. This is a repeating pattern and problem.

This is a non-starter. You can conduct commerce without negotiating discounts — absence of ability to negotiate a discount does not "block" checkout from proceeding. But displaying accurate, itemized totals to the buyer is not optional functionality. We should not force a dependency on an extension for something the business is obligated to present, and I think our existing contract is basically there... modulo a few edits and upgrades.

**Claim: fees and taxes must live on `totals[]`; `totals[]` must support repeating types.** (and it does, but underspecified)

This draft is formalizing the totals contract: the business dictates what must be rendered — content and order — because the correct presentation is subject to regional, product, and regulatory requirements that the business is obligated to satisfy. The platform renders what it's given. The contract is simple: iterate, display `display_text` + `amount` for each entry. The platform is a receipt printer, it does not and must not enforce own display logic -- this makes it universal, because business dictates the presentation and can meet its obligations.

## What this contract enables

* **Itemized fees**: multiple `type: "fee"` entries, or collapsed with `lines` breakdown
* **Multi-jurisdiction tax**: separate Federal Tax / State Tax entries
* **Custom types**: `account_credit`, `tip`, `surcharge` — open `type` string with explicit `effect` field
* **Sub-line breakdown**: `lines` array for progressive enhancement — business controls what MUST be rendered (top-level) vs what MAY be optionally surfaced (sub-lines)
* **Cardinality enforcement**: exactly one `subtotal` and one `total` (schema-enforced)
* **Self-describing sign convention**: `effect: "add"` / `"subtract"` — well-known types have implicit defaults, custom types must include it

## Examples

**Split tax, itemized at top-level:**

```json
"totals": [
  { "type": "subtotal",    "display_text": "Subtotal",      "amount": 5750 },
  { "type": "fulfillment", "display_text": "Shipping",      "amount": 899 },
  { "type": "tax",         "display_text": "Federal Tax",   "amount": 332 },    // <--
  { "type": "tax",         "display_text": "State Tax",     "amount": 465 },    // <--
  { "type": "total",       "display_text": "Total",         "amount": 7446 }
]
```

**Collapsed fees with optional breakdown:**
```json
"totals": [
  { "type": "subtotal", "display_text": "Subtotal", "amount": 4999 },
  {
    "type": "fee", "display_text": "Fees", "amount": 549,
    "lines": [              // <--
      { "display_text": "Service Fee", "amount": 399 },
      { "display_text": "Recycling Fee", "amount": 150 }
    ]
  },
  { "type": "tax",   "display_text": "Tax",   "amount": 444 },
  { "type": "total", "display_text": "Total", "amount": 5992 }
]
```

**Account credit — custom type with explicit `effect`:**
```json
"totals": [
  { "type": "subtotal",     "display_text": "Subtotal",     "amount": 10000 },
  { "type": "tax",          "display_text": "Tax",          "amount": 800 },
  { "type": "account_credit", "display_text": "Account Credit", "amount": 2500, "effect": "subtract" }, // <--
  { "type": "total",        "display_text": "Amount Due",   "amount": 8300 }
]
```

## Schema enforcement for cardinality
The array has two layers of constraints, each using allOf:

1. **Item shape** (items.allOf): composes the base `total_line.json` entry with checkout-level requirements — `display_text` becomes required, and the `effect` and `lines` fields are added. This tightening only applies at checkout/cart/order level.
2. **Array cardinality** (top-level allOf): uses `contains`/`minContains`/`maxContains` to enforce exactly one subtotal and exactly one total entry. All other types are unconstrained — they may appear zero or more times.

## Backward compatibility

* **Additive**: `effect` field (optional, well-known types have defaults), `lines` array (optional).
* **Tightening**: 
   - `display_text` required at checkout/cart/order level. 
   - Cardinality constraints on subtotal and total makes implicit contract explicit in schema.
* **Relaxed**: `type` enum opened to string — all existing values remain valid, schema additionally accepts custom types.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -156,8 +156,22 @@ jobs:
           # Extract the date (e.g., release/2026-01-11 -> 2026-01-11)
           VERSION_NAME=${GITHUB_REF#refs/heads/release/}
 
-          # Deploy this version, tag it as 'latest', and set it as the default site root
-          uv run mike deploy --push --update-aliases $VERSION_NAME latest
+          # Fetch all release branches to determine if this is the latest
+          git fetch origin "+refs/heads/release/*:refs/remotes/origin/release/*"
+          
+          # Find the highest version number among all release branches
+          LATEST_VERSION=$(git branch -r --list "origin/release/*" | sed 's|origin/release/||' | sort -V | tail -n 1 | tr -d ' ')
+
+          echo "Deploying version: $VERSION_NAME"
+          echo "Latest detected version: $LATEST_VERSION"
+
+          if [ "$VERSION_NAME" = "$LATEST_VERSION" ]; then
+            echo "This is the latest version. Updating 'latest' alias."
+            uv run mike deploy --push --update-aliases "$VERSION_NAME" latest
+          else
+            echo "This is NOT the latest version. Deploying without updating 'latest' alias."
+            uv run mike deploy --push "$VERSION_NAME"
+          fi
 
       - name: Create GitHub Release and Tag
         if: startsWith(github.ref, 'refs/heads/release/')


source/schemas/shopping/types/totals.json
```diff
@@ -0,0 +1,64 @@
...
+      {
+        "$ref": "total_line.json"
+      },
+      {
+        "type": "object",
+        "required": [
+          "display_text"
+        ],
+        "properties": {
+          "lines": {
```
lemonmade, line 18: Here, we are using a nested `lines` field to create parent/ child relationships. In contrast, [`line_items` uses a `parent_id` field](https://ucp.dev/latest/specification/checkout/#line-item_1) to create this structure. I guess it would be too awkward to apply the same approach in both spots, since totals do not have an `id`?

igrigorik: Right, this is intentional. With lines+parent_id the contract is that all lines are default visible but can be optionally grouped if platform follows parent_id. The contract here is that nested lines are optional / supplementary and MAY be rendered by a platform; the top-level is the MUST render.

source/schemas/shopping/types/total_line.json
```diff
@@ -1,7 +1,8 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://ucp.dev/schemas/shopping/types/total.json",
-  "title": "Total",
+  "$id": "https://ucp.dev/schemas/shopping/types/total_line.json",
```
maximenajim: Renaming total.json to total_line.json and changing the canonical $id is a breaking public-contract change.  Is that ok?  Should we keep a deprecated total.json shim with the old $id that delegates to the new schema, or version the capability/schema explicitly, so existing integrations do not hard-fail on rollout?

source/schemas/shopping/types/total_line.json
```diff
@@ -28,8 +20,7 @@
       "ucp_request": "omit"
     },
     "amount": {
-      "$ref": "amount.json",
-      "description": "If type == total, sums subtotal - discount + fulfillment + tax + fee. Should be >= 0. Amount in ISO 4217 minor units.",
+      "$ref": "signed_amount.json",
```
maximenajim: Changing the shared base type to signed_amount.json broadens every consumer of total_line—not just cart/checkout/order, but also line_item.totals, order_line_item.totals, and fulfillment_option.totals. The sign convention, however, is only documented/enforced in types/totals.json. That makes the meaning of the same type value depend on where it is nested. Please either scope signed amounts to the top-level totals.json contract (override amount there) or propagate the same sign rules and documentation to the other consumers in the same PR.

source/schemas/shopping/types/totals.json
```diff
@@ -0,0 +1,64 @@
...
+  "type": "array",
+  "items": {
+    "allOf": [
+      {
+        "$ref": "total_line.json"
+      },
+      {
+        "type": "object",
+        "required": [
+          "display_text"
```
maximenajim: Requiring display_text on every top-level entry is not just tightening an implicit contract; it invalidates multiple published examples in this repo (cart-rest.md, checkout-rest.md, order.md, ap2-mandates.md, playground.md) where top-level subtotal/tax lines omit it.

source/schemas/shopping/types/total_line.json
igrigorik: Great catch, reverted. I walked into this because I wanted to avoid total vs totals, but not worth it.

sinhanurag (approved):

docs/specification/checkout.md
```diff
@@ -610,11 +610,136 @@ field or omitting them.
...
+Platforms MUST NOT substitute their own computed totals for the business's
+values. Platforms MAY verify the provided totals:
+
+```python
+assert sum(e.amount for e in totals if e.type != "total") == total_entry.amount
+```
+
+If the computed sum does not match the `type: "total"` entry, the platform
+MUST NOT alter the rendered output — the business's total is authoritative.
+Platforms MAY prevent checkout completion or surface the discrepancy to the
```
maximenajim: The new verification guidance is too permissive for autonomous financial flows. MAY verify / MAY prevent checkout completion allows a business to provide an internally inconsistent breakdown and still have a platform auto-complete it. If sum(non-total entries) != total, that should have a deterministic failure path—at minimum, platforms should MUST block autonomous completion or require explicit human escalation, and the spec should define a standard error/message for that case.

docs/specification/checkout.md
```diff
@@ -610,11 +610,136 @@ field or omitting them.
...
+| `subtotal`        | +          | Sum of line item prices                        |
+| `discount`        | −          | Order or line-item level discount              |
+| `items_discount`  | −          | Rollup of line-item discounts                  |
+| `fulfillment`     | +          | Shipping, delivery, or pickup charges          |
+| `tax`             | +          | Tax charges                                    |
+| `fee`             | +          | Fees and surcharges                            |
+| `total`           | =          | Authoritative grand total (exactly one)        |
+
+The sign convention for well-known types is schema-enforced: subtractive
+types (discount, items_discount) MUST have negative amounts; additive types
```
maximenajim: This flips the sign convention for discount / items_discount, but the discount capability still defines discount amounts as positive absolute values and its examples show positive totals lines. Without updating that companion contract, integrators will either double-negate or fail validation depending on which section they implemented from. Please update discount.md / discount.json in the same PR and explicitly document the mapping between discounts.applied.amount and signed totals[].amount.

igrigorik: Hmm. This would be a breaking change. Let me make an argument for keeping it as is...

The totals array and the discount extension serve different layers. The extension describes WHAT was applied — "a $10 discount, code SAVE10, allocated across these items." That's the definition of the discount, and its magnitude is naturally positive: the value of the applied discount is $10. The totals entry describes the EFFECT on the receipt — "this line reduces the total by $10." That's signed: -1000.

This difference is not without precedent. For example, Stripe uses coupon.amount_off = 1000, invoice.line_items[].amount = -1000. In effect...

* The discount extension's applied[].amount = 1000 answers: "what discount was applied?". That's a definition and it refs amount.json (unsigned, minimum: 0).
* The totals entry amount = -1000 answers: "what's the line on the receipt?" — minus $10. That's an effect.
   It refs signed_amount.json.

maximenajim: The distinction between the discount extension's applied[].amount (positive, definitional) and the totals entry's amount (signed, presentational) makes sense and has good precedent (Stripe, etc.). I'd just suggest making this two-layer relationship explicit in the spec docs somewhere — a single sentence like "The discount extension's applied[].amount describes the magnitude of the applied discount; the totals entry's amount represents its signed effect on the receipt" would save integrators from having to infer it.

igrigorik: 👍  done: https://github.com/Universal-Commerce-Protocol/ucp/pull/261/changes/29d949e2f2e628dfe2fd30692064189bb6b458fe

PR #264, opened 2026-03-13
feat: attestation extension for eligibility claims

douglasborthwick-crypto:
# Description

Adds a capability extension that complements eligibility (#250) with
cryptographic attestation proofs for non-instrument claims (token holdings,
on-chain credentials, membership status).

Discussed and invited in #203 and #250:

- **#203**: @igrigorik established the `provider_jwks` / `kid` / `payload` /
  `sig` attestation pattern in the signals spec
- **#250**: @igrigorik confirmed attestation as ["a great candidate for a
  capability extension that complements eligibility. Additive and
  non-breaking."](https://github.com/Universal-Commerce-Protocol/ucp/pull/250#issuecomment-4048233895)
  and specified the [sibling map
  pattern](https://github.com/Universal-Commerce-Protocol/ucp/pull/250#issuecomment-4050845563)
  — attestations keyed by eligibility claim, on the cart/checkout object
  similar to the discount extension

### How it works

- Platforms relay signed attestations from third-party verifiers alongside
  `context.eligibility` claims
- Businesses verify offline: fetch JWKS from `provider_jwks`, select key by
  `kid`, verify `sig` over `payload`
- `payload` is an opaque passthrough of the provider's signed object — no
  transformation, so signature verification works end-to-end
- `expires_at` lives outside `payload` because it is not covered by the
  signature — the Business checks this independently

Follows the discount extension pattern (`allOf` on Cart and Checkout, sibling
map keyed by reverse-domain eligibility claims).

### Wire format example

```json
{
  "context": {
    "eligibility": ["com.example.token_holder"]
  },
  "attestations": {
    "com.example.token_holder": {
      "provider_jwks": "https://example.com/.well-known/jwks.json",
      "kid": "example-key-2026-01",
      "payload": {
        "id": "att-7c3e9f",
        "pass": true,
        "results": ["..."],
        "attestedAt": "2026-03-13T19:20:32.530Z"
      },
      "sig": "base64...",
      "expires_at": "2026-03-13T19:50:32.530Z"
    }
  }
}
```

### Verification flow

1. Business receives checkout with `context.eligibility` claim + matching
   `attestations` entry
2. Fetches JWKS from `provider_jwks`, selects key matching `kid`
3. Verifies `sig` over the payload (platforms must preserve the provider's
   original serialization)
4. Checks `expires_at` hasn't passed
5. Reads `payload.pass` to confirm the claim is met

No callback to the Platform or attestation provider required — fully offline.

### Relationship to #280

[...]

PR #265, opened 2026-03-15
feat!: redesign identity linking with mechanism registry and capability-driven scopes

amithanda:
## Overview

This PR redesigns the Identity Linking capability from a hardcoded OAuth 2.0
specification into an extensible, security-hardened system built on three
core principles: a **Mechanism Registry pattern** for future-proof
authentication negotiation, **capability-driven least-privilege scope
derivation**, and hardened OAuth 2.0 security requirements aligned with
current best practices.

---

### Mechanism Registry Pattern (`identity-linking.md`, `identity_linking.json`)

The capability no longer hardcodes OAuth 2.0 as the only authentication
mechanism. Instead, businesses declare an ordered `supported_mechanisms`
array in their capability config, and platforms select the first mechanism
type they support (business-preference ordering, mirroring TLS cipher suite
negotiation).

A formal **Mechanism Selection Algorithm** is defined:
- Iterate from index 0 (highest business preference)
- Select the first `type` the platform supports
- Abort if no supported type is found — no partial or fallback flows permitted

This enables future mechanism types (e.g., `verifiable_credential`) without
breaking changes to the protocol.

### Capability-Driven Least-Privilege Scope Negotiation (`overview.md`, `checkout.json`, `identity_linking.json`)

Authorization scopes are no longer hardcoded or pre-declared in the identity
linking config. Instead:

- Each capability schema declares its own required scopes via a top-level
  `identity_scopes` annotation (e.g., `checkout.json` declares
  `dev.ucp.shopping.scopes.checkout_session`)
- Scopes are derived **only** from the finalized capability intersection —
  capabilities excluded during negotiation contribute zero scopes
- The capability intersection algorithm gains two new steps:
  - **Step 3 — Scope Dependency Pruning**: capabilities declaring
    `identity_scopes` are removed if `dev.ucp.common.identity_linking` is not
    in the intersection
  - **Step 5 — Derive Scopes (Final Pass)**: the authorization scope set is
    computed from the stabilized intersection only

This makes over-permissioning impossible by construction, not convention.

### Scope Naming Convention

Scopes now use **reverse DNS dot notation** with a mandatory `.scopes.`
segment, consistent with UCP capability names:

- UCP-defined: `dev.ucp.<domain>.scopes.<capability>`
  (e.g., `dev.ucp.shopping.scopes.checkout_session`)
- Third-party: `<reverse-dns>.scopes.<capability>`
  (e.g., `com.example.loyalty.scopes.points_balance`)

A regex pattern enforcing this convention is defined in
`$defs/identity_scopes` in `identity_linking.json` as the canonical
reference.

### Hardened OAuth 2.0 Security (`identity-linking.md`)

The OAuth 2.0 mechanism requirements are significantly tightened:

| Requirement | Before | After |
|---|---|---|
| PKCE (RFC 7636) | Not required | **MUST** use `S256` |
| `iss` validation (RFC 9207) | Not mentioned | **MUST** validate to prevent Mix-Up Attacks |
| `redirect_uri` matching | Not mentioned | **MUST** enforce exact string match |
| Issuer normalization | Not mentioned | **MUST** strip trailing slashes before comparison |
| Discovery error handling | Unspecified | Only HTTP 404 triggers OIDC fallback; all other failures abort |

### Discovery Resolution Hierarchy (`identity-linking.md`)

A formal 3-tier hierarchy replaces the previous unspecified discovery:

1. **Explicit `discovery_endpoint`** (hard abort on any failure)
2. **RFC 8414** `/.well-known/oauth-authorization-server` (only `404`
   triggers fallback; `5xx`, timeouts abort)
3. **OIDC fallback** `/.well-known/openid-configuration` (abort on non-2xx)

### Schema (`identity_linking.json`)

New schema file with:
- `platform_schema`: passthrough only — platforms are consumers of
  mechanisms, not providers
- `business_schema`: requires `config` with `supported_mechanisms`
  (`minItems: 1`)
- `mechanism`: open base type (consistent with `payment_credential.json`)
  — `type` required, `additionalProperties: true`
- `oauth2`: named `$def` for explicit strict validation
- `identity_scopes`: canonical annotation definition with enforced regex
  pattern

### End-to-End Workflow Example (`identity-linking.md`)

A complete walkthrough added: AI Shopping Agent (platform) + Merchant
(business), covering discovery, scope derivation, mechanism selection,
OAuth authorization request (with PKCE and `iss` parameters), and token
exchange.

---

## Files Changed

| File | Change |
|---|---|
| `docs/specification/identity-linking.md` | Major rewrite |
| `docs/specification/overview.md` | Pruning algorithm + Step 5 scope derivation |
| `source/schemas/common/identity_linking.json` | New file |
| `source/schemas/shopping/checkout.json` | Adds `identity_scopes` annotation |
| `docs/index.md` | Scope value updated to reverse DNS notation |

## Type of change

- [x] Documentation update

---

files or fields:

```text
The change from hardcoded OAuth 2.0 to a registry and new scope naming conventions constitutes a breaking change for the protocol
```
---

## Checklist

---

### DIFF
--- a/docs/specification/identity-linking.md
+++ b/docs/specification/identity-linking.md
@@ -105,7 +105,7 @@ intersection of negotiated capabilities**.
    the finalized intersection. If a capability (e.g., `order`) is excluded from
    the active capability set, its respective scopes **MUST NOT** be requested by
    the platform. If the final derived scope list is completely empty, the platform
-   **SHOULD** abort the identity linking process, as there are no secured resources
+   **MUST** abort the identity linking process, as there are no secured resources
    to authorize.
 
 ### Scope Structure & Mapping
@@ -137,7 +137,7 @@ Example capability-to-scope mapping based on UCP schemas:
 
 | Resources       | Operation                                     | Scope Action                               |
 | :-------------- | :-------------------------------------------- | :----------------------------------------- |
-| CheckoutSession | Get, Create, Update, Delete, Cancel, Complete | `dev.ucp.shopping.scopes.checkout_session` |
+| CheckoutSession | Get, Create, Update, Cancel, Complete         | `dev.ucp.shopping.scopes.checkout_session` |
 
 ## Supported Mechanisms
 


douglasborthwick-crypto:
Great design. The mechanism registry pattern is exactly right — hardcoding OAuth as the only path was the main limitation of the previous spec.

This pairs well with #264 (attestation extension for eligibility). The two address different surfaces:

- **#265 (this PR)**: *Identity* — "who is this user?" Mechanism negotiation for linking accounts.
- **#264**: *Eligibility* — "what does this user hold?" Signed attestation proofs attached to cart/checkout.

But there's also a natural overlap: wallet attestation as an identity mechanism type. Some commerce flows don't need OAuth account linking at all — the wallet *is* the identity. A business selling token-gated merchandise needs to verify holdings, not link an account.

The registry already supports this cleanly:

```json
"supported_mechanisms": [
    {
        "type": "oauth2",
        "issuer": "https://auth.merchant.example.com"
    },
    {
        "type": "wallet_attestation",
        "provider_jwks": "https://verifier.example.com/.well-known/jwks.json",
        "attestation_endpoint": "https://verifier.example.com/v1/attest"
    }
]
```

The platform calls the attestation endpoint, receives a signed payload with `kid` + `sig`, and the business verifies offline via JWKS — the same pattern from #264 and the signals work in #203. No redirect flow, no token exchange, no account creation. The base mechanism schema's `additionalProperties: true` means this works today without schema changes.

For scope derivation: a `wallet_attestation` mechanism wouldn't need OAuth scopes — verification is stateless and self-contained. The pruning algorithm in Step 3 already handles this correctly: capabilities without `identity_scopes` contribute zero scopes.

One question on the spec: the Discovery Bridging section (RFC 8414 / OIDC fallback hierarchy) is tightly coupled to OAuth. Would it make sense to scope that hierarchy under the `oauth2` mechanism section rather than presenting it as general? Other mechanism types would define their own resolution — JWKS endpoint for attestation, DID resolution for verifiable credentials, etc.

docs/specification/identity-linking.md
```diff
@@ -16,146 +16,342 @@
...
+
+**Issuer Validation**: Regardless of the discovery method used above, the
+platform **MUST** validate that the `issuer` value returned in the metadata
+exactly matches the `issuer` string defined in the capability configuration.
+
+*Issuer Normalization*: To prevent brittle integrations, the platform **MUST**
+normalize both the locally configured `issuer` and the remotely returned `issuer`
+by stripping any trailing slashes (`/`) before performing the exact string match
+comparison (i.e., `https://auth.example.com` and `https://auth.example.com/`
+**MUST** be evaluated as identical).
```
cusell-google: Under [RFC 8414 3.3](https://datatracker.ietf.org/doc/html/rfc8414#section-3.3) it mentions that the string must be an identical match. 
I understand that the intent to reduce brittle integrations due to a simple slash, but this can have other consequences. In the case of a trailing slash, the exact string would be incldued in the the signed `iss` claim of the JWT access token, and standard JWT validation libraries will check for equality (not stripping). If UCP strips the slash during discovery and passes the normalized string, the equality will fail.

Can you drop the normalization step? And can we instead add a mandate that the `issuer` string in UCP must match exactly the metadata, and the `iss` parameter?

docs/specification/identity-linking.md
```diff
@@ -16,146 +16,342 @@
...
+- **UCP-defined scopes:** `dev.ucp.<domain>.scopes.<capability>` (e.g.,
+  `dev.ucp.shopping.scopes.checkout_session`)
+- **Third-party scopes:** `<reverse-dns>.scopes.<capability>` (e.g.,
+  `com.example.loyalty.scopes.points_balance`)
+
+Example capability-to-scope mapping based on UCP schemas:
+
+| Resources       | Operation                                     | Scope Action                               |
+| :-------------- | :-------------------------------------------- | :----------------------------------------- |
+| CheckoutSession | Get, Create, Update, Delete, Cancel, Complete | `dev.ucp.shopping.scopes.checkout_session` |
```
cusell-google: +1 for the change from colon-separation to reverse DNS. It aligns with what I've seen in [PR 250 for eligibility](https://github.com/Universal-Commerce-Protocol/ucp/pull/250) and other UCP changes.

From an OAuth perspective, [RFC 6749 3.3](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3) restricts scope tokens to `%x21 / %x23-5B / %x5D-7E`, which makes dots perfectly valid and widely supported by Authorization Servers.

(Optionally) add a short note mentioning this format strictly adheres to the scope token syntax defined in RFC 6749 Section 3.3.

docs/specification/identity-linking.md
```diff
@@ -16,146 +16,342 @@
...
+
+1. **Schema Declaration:** Each individual capability schema explicitly defines
+   its own required identity scopes (e.g., `dev.ucp.shopping.checkout` declares
+   `dev.ucp.shopping.scopes.checkout_session`).
+2. **Dynamic Derivation:** During UCP Discovery, when the platform computes the
+   intersection of supported capabilities between itself and the business, it
+   extracts the required scopes from **only** the successfully negotiated
+   capabilities.
+3. **Authorization:** The platform initiates the connection requesting **only**
+   the derived scopes. If a capability (e.g., `order`) is excluded from the
```
cusell-google: Are we intentionally pushing platforms toward incremental authorization here? If a platform negotiates both `checkout` and `order`, but only needs `checkout` for the current session, requesting a subset of scopes means they'll have to trigger a re-linking flow later for order management. This introduces UX friction compared to requesting the full negotiated union upfront.

To clarify the intent, I recommend replacing "only" with either "at most" (if we want to allow subsets for progressive consent) or "exactly" (if we expect platforms to request the full derived set to minimize re-prompting).

docs/specification/identity-linking.md
```diff
@@ -16,146 +16,342 @@
...
+      cross-account protection.
+
+#### For businesses
+
+- **MUST** implement OAuth 2.0
+  ([RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749))
+- **MUST** adhere to [RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414)
+  to declare the location of their OAuth 2.0 endpoints
+  (`/.well-known/oauth-authorization-server`)
+- **SHOULD** populate `scopes_supported` in their RFC 8414 metadata to allow
```
cusell-google: Should this be MUST? Given that the Platform will only request the derived scopes.

docs/specification/identity-linking.md
```diff
@@ -16,146 +16,342 @@
...
+   defined `issuer` string and fetch. If this fetch returns any non-2xx response
+   other than `404 Not Found` (e.g., `500 Internal Server Error`, `503 Service
+   Unavailable`), or if a connection timeout or network error occurs, the
+   platform **MUST** abort the discovery process and **MUST NOT** proceed to the
+   OIDC fallback.
+3. **OIDC Fallback (Lowest Priority)**: If and only if the RFC 8414 fetch
+   returns exactly `404 Not Found`, the platform **MUST** append
+   `/.well-known/openid-configuration` to the defined `issuer` string and fetch.
+   If this final fetch returns any non-2xx response or a network error, the
+   platform **MUST** abort the identity linking process.
```
cusell-google, line 173: no actionable comment, just wanted to mention that having this as a fallback is a great idea

docs/specification/identity-linking.md
```diff
@@ -16,146 +16,342 @@
...
+    - **MAY** support Client Metadata
+    - **MAY** support Dynamic Client Registration mechanisms to supersede static
+      credential exchange.
+- The platform must include the token in the HTTP Authorization header using the
+  Bearer schema (`Authorization: Bearer <access_token>`)
+- **MUST** implement the OAuth 2.0 Authorization Code flow
+  (<a href="https://datatracker.ietf.org/doc/html/rfc6749#section-4.1" target="_blank">RFC
+  6749 4.1</a>) as the primary linking mechanism.
+- **MUST** strictly implement Proof Key for Code Exchange (PKCE)
+  ([RFC 7636](https://datatracker.ietf.org/doc/html/rfc7636)) using the `S256`
```
cusell-google, line 229: This seems like an extension going towards what OAuth 2.1 is pushing for. OAuth 2.0 doesn't specifiy it, and OAuth 2.1 isn't, as of today, an official RFC. Do you want to already include PKCE as an extension of UCP or hold off and add it later? [Oauth 2.1 Draft](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-15.txt)

I suspect this may be significant overhead, especially as a MUST, for platforms and businesses while it's not-quite-yet industry standard.

cusell-google: Feel free to dismiss this one. PKCE S256 is supported more widely than I anticipated.

docs/specification/identity-linking.md
amithanda: yes, makes sense. Argument for should was - RFC 8414 defines scopes_supported as OPTIONAL. Upgrading to MUST makes UCP stricter than the underlying RFC, which could be a barrier for businesses retrofitting existing OAuth servers that don't currently populate it.  But I think we can go with MUST.

docs/specification/identity-linking.md
amithanda: clarifying the language.

docs/specification/identity-linking.md
amithanda: Clarified above.

docs/specification/identity-linking.md
```diff
@@ -16,146 +16,349 @@
...
+
+To maintain the **Principle of Least Privilege**, authorization scopes are
+**NOT** hardcoded within the identity linking capability.
+
+Instead, **authorization scopes are dynamically derived from the final
+intersection of negotiated capabilities**.
+
+1. **Schema Declaration:** Each individual capability schema explicitly defines
+   its own required identity scopes (e.g., `dev.ucp.shopping.checkout` declares
+   `dev.ucp.shopping.scopes.checkout_session`).
```
drewolson-google, line 98: Help me understand this a bit. Capabilities will work with a linked identity and without a linked identity -- checkout is a good example of this. Is this idea that _if an identity is linked_ then the link must contain all of the scopes from the negotiated set of capabilities?

If so, this feels a bit odd, in that the capabilities worked _before identity linking_, when there were no scopes at all, and could work post-identity linking even without the appropriate scopes by falling back to the behavior without a linked identity.

Thoughts on this?

amithanda: @drewolson-google 
No, This is to facilitate identity linking itself not the case where identity is already linked. 
Today, the discovery of identity linking url's and scopes is not very well defined via UCP. This PR allows us to have a UCP capability `dev.ucp.common.identity_linking` , where you can find out the auth scheme, config that both platform and business can support, algorithm for agents to facilitate identity_linking. The scopes are used during this initial identity_linking flow based on the capabilities.

Once the identities are linked, this configuration is not used as the tokens are used directly to facilitate authentication can call operations on capabilties. Let me know if that makes sense or if you suggest I add some more documentation to clarify this piece better somewhere.

drewolson-google: Got it, makes sense, thanks!

docs/specification/identity-linking.md
```diff
@@ -16,146 +16,349 @@
...
+  `com.example.loyalty.scopes.points_balance`)
+
+This format strictly adheres to the scope token syntax defined in
+[RFC 6749 Section 3.3](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3).
+
+Example capability-to-scope mapping based on UCP schemas:
+
+| Resources       | Operation                                     | Scope Action                               |
+| :-------------- | :-------------------------------------------- | :----------------------------------------- |
+| CheckoutSession | Get, Create, Update, Delete, Cancel, Complete | `dev.ucp.shopping.scopes.checkout_session` |
```
jingyli: nit: We currently don't have `Delete` operation for checkout capability.

docs/specification/identity-linking.md
```diff
@@ -16,146 +16,349 @@
...
+   defined `issuer` string and fetch. If this fetch returns any non-2xx response
+   other than `404 Not Found` (e.g., `500 Internal Server Error`, `503 Service
+   Unavailable`), or if a connection timeout or network error occurs, the
+   platform **MUST** abort the discovery process and **MUST NOT** proceed to the
+   OIDC fallback.
+3. **OIDC Fallback (Lowest Priority)**: If and only if the RFC 8414 fetch
+   returns exactly `404 Not Found`, the platform **MUST** append
+   `/.well-known/openid-configuration` to the defined `issuer` string and fetch.
+   If this final fetch returns any non-2xx response or a network error, the
+   platform **MUST** abort the identity linking process.
```
jingyli, line 173: nit: This should also be `abort the discovery process` for consistency with the steps above..?

docs/specification/identity-linking.md
```diff
@@ -16,146 +16,349 @@
...
+      (<a href="https://datatracker.ietf.org/doc/html/rfc7009" target="_blank">RFC
+      7009</a>) when a user initiates an unlink action on the platform side.
+    - **SHOULD** support
+      [OpenID RISC Profile 1.0](https://openid.net/specs/openid-risc-1_0-final.html)
+      to handle asynchronous account updates, unlinking events, and
+      cross-account protection.
+
+#### For businesses
+
+- **MUST** implement OAuth 2.0
```
jingyli, line 249: A super naive QQ: Here we are saying business **MUST** implement OAuth 2.0, a few lines above we also mention `**MUST** implement the OAuth 2.0 Authorization Code flow` for platforms.

Given these 2 bullets, wouldn't the negotiated mechanism type always be `oauth2` (which feels like it's going against the main design point around the mechanism registry)..?

docs/specification/identity-linking.md
```diff
@@ -16,146 +16,349 @@
...
+2. **Dynamic Derivation:** During UCP Discovery, when the platform computes the
+   intersection of supported capabilities between itself and the business, it
+   extracts the required scopes from **only** the successfully negotiated
+   capabilities.
+3. **Authorization:** The platform initiates the connection requesting **exactly**
+   the derived scopes — the union of `identity_scopes` from all capabilities in
+   the finalized intersection. If a capability (e.g., `order`) is excluded from
+   the active capability set, its respective scopes **MUST NOT** be requested by
+   the platform. If the final derived scope list is completely empty, the platform
+   **SHOULD** abort the identity linking process, as there are no secured resources
```
jingyli: Curious why this guideline is a **SHOULD** rather than a **MUST** if there are no secured resources to authorize as part of the linking process (is there ever a use case for showing an empty consent screen)?

docs/specification/overview.md
```diff
@@ -650,14 +672,25 @@ for a session:
...
-    - For single-parent extensions (`extends: "string"`): parent must be present
-    - For multi-parent extensions (`extends: ["a", "b"]`): at least one parent
-        must be present
+3. **Prune orphaned extensions & unauthorized capabilities**: Remove any capability that lacks its required structural or functional dependencies:
+    - **Structural Dependencies**: Remove any capability where `extends` is
+      set but **none** of its parent capabilities are in the intersection.
+        - For single-parent extensions (`extends: "string"`): parent must be present
+        - For multi-parent extensions (`extends: ["a", "b"]`): at least one parent
+          must be present
+    - **Scope Dependencies**: Remove any capability declaring `identity_scopes`
```
jingyli, line 681: Naive QQ on this: Given the UCP checkout capability will declare `identity_scopes` per the change on checkout.json schema below, does this mean it will always be removed if `dev.ucp.common.identity_linking` is not part of the intersection between business & platform?

docs/specification/identity-linking.md
amithanda: This is the last stage of discovery so if this fails, we want platform to abort the identity linking process. We could have said "abort the discovery process" but I think both are valid. Let me know if you feel strongly about changing this. I don't have a strong opinion and both seem ok to me.

docs/specification/identity-linking.md
amithanda: These are both mentioned under OAuth 2.0 (`"type"`: `"oauth2"`) heading, so they apply only when Supported Mechanism is `oauth2`. When we add new mechanisms types, we will have a separate section under `Supported Mechanisms` heading for that mechanism, at the same level as OAuth 2.0. Does that make sense?

docs/specification/identity-linking.md
amithanda: Good catch.
The only weak argument for keeping SHOULD could be forward-compatibility with future mechanism types that don't use scopes — but the spec already handles that through the mechanism registry: future non-scope-based mechanisms would define their own requirements in their own section and wouldn't be constrained by this sentence.
I will flip it to MUST.

docs/specification/overview.md
amithanda: Yes, I think that is how it should work. `identity_scopes` on checkout.json is checkout's declaration that it requires an authorized Bearer token to operate. Without identity_linking in the intersection, there's no negotiated mechanism to obtain that token — so keeping checkout active would leave the platform with a capability it can't securely call. Pruning it is the correct outcome.

In practice this means identity_linking acts as a soft prerequisite for any capability declaring identity_scopes. Both parties need to advertise it for checkout to survive the intersection.

Good callout though, let me know if we should add something more to the documentation to make it clearer.

docs/specification/identity-linking.md
amithanda: Good catch, removed.

amithanda:
> Added some nits/naive questions but overall looks good!
> 
> Non-blocking, but one side question I'm curious about is how do we think about extension application (if any) on this new identity linking capability moving forward given the schema is focused on the discovery & negotiation of authorization/linking between platforms & businesses?

Yes, that may not make sense for this capability, given the purpose is to define the mechanism/config for identity_linking. It will just extend via new mechanism types. If you were thinking about the loyalty use cases, I think the extensions there would probably apply to cart/checkout/order etc. For identity management, I think we will probably need a new capability (e.g Profile or Account) which will be a separate from this capability (identity_linking).

Let me know if that makes sense.

docs/specification/overview.md
lemonmade: Does this mean you can only use the checkout capability if you also use the identity linking capability?

PR #272, opened 2026-03-18
feat: Add `ec.totals.change` to Embedded Checkout Protocol

lemonmade:
# Description

This PR adds an `ec.totals.change` message to the Embedded Checkout Protocol. This message allows hosts to be informed of changes to the checkout state that do not have an accompanying `ec.{domain}.change` event today, like taxes and fees, which are entirely contained in `totals`. It also allows hosts to be informed of totals-only changes that may result from asynchronous updates after a domain event, such as potentially-slow shipping rate updates after a `ec.fulfillment.change` message has indicated the initial application of the change.

## Type of change

      or fields**)
- [ ] Documentation update

---

## Checklist


westeezy (approved):
I think this seems reasonable and appreciate the strict rule on ordering to keep it predictable.

docs/specification/embedded-checkout.md
```diff
@@ -822,6 +822,61 @@ informational notices about the checkout state.
...
+
+Checkout totals have been updated. This message covers all total line changes
+including taxes, fees, discounts, and fulfillment costs — many of which have no
+other domain-specific change message. Businesses **MUST** send this message
+whenever `checkout.totals` changes for any reason.
+
+When a change also triggers a domain-specific message (e.g.,
+`ec.line_items.change`, `ec.buyer.change`, or `ec.payment.change`), the business
+**MUST** send the domain-specific message first, then follow it with
+`ec.totals.change`.
```
jingyli, line 838: Naive QQ: I understand this is nice from a consistency perspective, but is there any additional benefits from mandating business to **MUST** send both notifications if the domain-specific message also has its own mandate to send the entire latest checkout state (which naturally encapsulates `ec.totals.change`)? It feels like we are going to be exchanging the same checkout object multiple times throughout the embedded checkout's lifecycle (especially if we are going to mirror this pattern for other EP bindings like Embedded Cart, where it's possible many `line_item` changes will occur).

Should we loosen the guideline to **MAY** (or **SHOULD**) or do the opposite where if a domain-specific message exist, then that must be the one message sent back to host with the updated totals?

docs/specification/embedded-checkout.md
```diff
@@ -822,6 +822,61 @@ informational notices about the checkout state.
...
+    "method": "ec.totals.change",
+    "params": {
+        "checkout": {
+            "id": "checkout_123",
+            // The entire checkout object is provided, including the updated totals
+            "totals": [
+                {
+                    "type": "subtotal",
+                    "display_text": "Subtotal",
+                    "amount": { "value": 4000, "currency": "USD" }
```
jingyli: `amount` is [integer only](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/shopping/types/amount.json) with no `currency`, these examples should be updated?

lemonmade: Good catch 👍

docs/specification/embedded-checkout.md
lemonmade: We discussed this internally too, as were also a little worried by the amount of data flying over the wire for these changes. In the end, I think this contract is still right, though — it is weird for only some changes to totals to be communicated with this message, but not others, and I felt uncomfortable with the implementation instructions for both host and embedded checkout when I tried writing docs for a version that deduplicated `ec.totals.change` with other messages.

PR #276, opened 2026-03-18
feat(order): Get Order operation w/ platform-auth

richmolj:
## Description

Add synchronous GET endpoint for order retrieval, complementing the
existing webhook push mechanism. Both return the same current-state
snapshot shape - webhooks deliver proactively to avoid polling, GET
provides on-demand access for reconciliation and conversational use.

Authorization model:
- **Who** - MUST authenticate (non-negotiable, any UCP mechanism)
- **What** - MAY scope access (platform credentials -> own orders, buyer authorization -> buyer's orders, or custom policies). MAY omit or redact fields based on context or business policy.
- **When** - MAY enforce data availability (retention, erasure)

## Type of change

## Checklist

### DIFF
--- a/docs/specification/cart-mcp.md
+++ b/docs/specification/cart-mcp.md
@@ -31,29 +31,32 @@ Businesses advertise MCP transport availability through their UCP profile at
   "ucp": {
     "version": "{{ ucp_version }}",
     "services": {
-      "dev.ucp.shopping": {
-        "version": "{{ ucp_version }}",
-        "spec": "https://ucp.dev/{{ ucp_version }}/specification/overview",
-        "mcp": {
+      "dev.ucp.shopping": [
+        {
+          "version": "{{ ucp_version }}",
+          "spec": "https://ucp.dev/{{ ucp_version }}/specification/overview",
+          "transport": "mcp",
           "schema": "https://ucp.dev/{{ ucp_version }}/services/shopping/mcp.openrpc.json",
           "endpoint": "https://business.example.com/ucp/mcp"
         }
-      }
+      ]
     },
-    "capabilities": [
-      {
-        "name": "dev.ucp.shopping.checkout",
-        "version": "{{ ucp_version }}",
-        "spec": "https://ucp.dev/{{ ucp_version }}/specification/checkout",
-        "schema": "https://ucp.dev/{{ ucp_version }}/schemas/shopping/checkout.json"
-      },
-      {
-        "name": "dev.ucp.shopping.cart",
-        "version": "{{ ucp_version }}",
-        "spec": "https://ucp.dev/{{ ucp_version }}/specification/cart",
-        "schema": "https://ucp.dev/{{ ucp_version }}/schemas/shopping/cart.json"
-      }
-    ]
+    "capabilities": {
+      "dev.ucp.shopping.checkout": [
+        {
+          "version": "{{ ucp_version }}",
+          "spec": "https://ucp.dev/{{ ucp_version }}/specification/checkout",
+          "schema": "https://ucp.dev/{{ ucp_version }}/schemas/shopping/checkout.json"
+        }
+      ],
+      "dev.ucp.shopping.cart": [
+        {
+          "version": "{{ ucp_version }}",
+          "spec": "https://ucp.dev/{{ ucp_version }}/specification/cart",
+          "schema": "https://ucp.dev/{{ ucp_version }}/schemas/shopping/cart.json"
+        }
+      ]
+    }
   }
 }
 ```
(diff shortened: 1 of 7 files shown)


docs/specification/order-mcp.md
```diff
@@ -0,0 +1,294 @@
...
+
+Businesses advertise MCP transport availability through their UCP profile at
+`/.well-known/ucp`.
+
+```json
+{
+  "ucp": {
+    "version": "2026-01",
+    "services": {
+      "dev.ucp.shopping": {
```
jingyli: This seems to be inconsistent with what our schema in https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/service.json is describing, should this instead be:
```
"services": {
    "dev.ucp.shopping": [
      {
          "version": "{{ ucp_version }}",
          "spec": "https://ucp.dev/{{ ucp_version }}/specification/overview",
          "transport": "mcp",
          "schema": "https://ucp.dev/{{ ucp_version }}/services/shopping/mcp.openrpc.json",
          "endpoint": "https://business.example.com/ucp/mcp"
        }
    ]
  }
```

Similar comment to other occurrences of `services`. Also `{{ ucp_version }}` should be added to all the `version` placeholder and `spec` & `schema` URLs.

docs/specification/order.md
```diff
@@ -263,14 +263,102 @@ Examples: `refund`, `return`, `credit`, `price_adjustment`, `dispute`,
...
+order it already has context for.
+
+**Buyer authorization** - the platform obtains buyer authorization via
+[Identity Linking](identity-linking.md) with the necessary scopes, or a
+similar mechanism. This grants access to the buyer's orders regardless of
+which platform originated them.
+
+Businesses **MAY** define additional access policies (e.g., trusted partner
+agreements), enforce data availability constraints (e.g., retention
+windows, regulatory erasure), and omit or redact fields from the response
```
jingyli: > and omit or redact fields from the response based on context, business policy, or other requirements - independently of authorization.

A little bit confused on what this sentence mean - if a data field is marked as `required` in the response, I don't feel business can just omit them due to this reason (else they would not be UCP compliant)?

docs/specification/order.md
```diff
@@ -263,14 +263,102 @@ Examples: `refund`, `return`, `credit`, `price_adjustment`, `dispute`,
...
+based on context, business policy, or other requirements - independently
+of authorization.
+
+#### Error Responses
+
+When the business cannot return an order, the response includes a `messages`
+array describing the outcome:
+
+**Order not found:**
```
jingyli, line 326: Should this instead be in the `error_response` shape for consistency with other resource not found scenarios?

source/schemas/shopping/order.json
```diff
@@ -98,6 +98,13 @@
     "totals": {
       "$ref": "types/totals.json",
       "description": "Different totals for the order."
+    },
+    "messages": {
```
jingyli, line 106: ditto, I think we can get away from this addition if we just reuse `error_response.json`..?

source/schemas/shopping/order.json
jingyli, line 1: Will we also will need the `ucp_request: omit` annotation given we are starting to introduce proper operations  now?

docs/specification/order.md
```diff
@@ -263,14 +263,102 @@ Examples: `refund`, `return`, `credit`, `price_adjustment`, `dispute`,
...
-| Event Mechanism                             | Method | Endpoint              | Description                                            |
+The `permalink_url` is the authoritative reference for the full order
+experience - timeline, post-purchase operations, returns. The API provides
+programmatic access to current state for conversational and operational use
+cases.
+
+| Operation                                   | Method | Endpoint              | Description                                            |
 | :------------------------------------------ | :----- | :-------------------- | :----------------------------------------------------- |
+| [Get Order](#get-order)                     | `GET`  | `/orders/{id}`        | Platform retrieves current order state.                |
 | [Order Event Webhook](#order-event-webhook) | `POST` | Platform-provided URL | Business sends order lifecycle events to the platform. |
```
jingyli: I see `Order Webhook` is now also defined as an operation (vs. event like we originally had), is this intended? If `Order Webhook` is an operation, then does it mean the request for webhooks is equivalent to the response of the other platform-initiated operations?

docs/specification/order-rest.md
```diff
@@ -0,0 +1,261 @@
...
+Request and response signatures follow the
+[Message Signatures](signatures.md) specification using RFC 9421 HTTP Message
+Signatures. See
+[REST Request Signing](signatures.md#rest-request-signing) and
+[REST Request Verification](signatures.md#rest-request-verification) for
+the complete algorithm.
+
+## Conformance
+
+Platforms implementing the REST binding:
```
jingyli, line 250: A couple of the bullets in this section is duplicated from the general platform guidelines mentioned in `order.md`, should we consider deleting them here to reduce duplication?

docs/specification/order.md
igrigorik: Yeah, good catch, fixed.

docs/specification/order-rest.md
```diff
@@ -0,0 +1,274 @@
...
+=== "Response"
+
+    ```json
+    HTTP/1.1 200 OK
+    Content-Type: application/json
+
+    {
+      "ucp": {
+        "version": "{{ ucp_version }}",
+        "capabilities": [
```
ihoosain: do we need to update these responses to use the capabilities format, similar to other examples?

```
"capabilities": {
      "dev.ucp.shopping.checkout": [
        {"version": "{{ ucp_version }}"}
      ]
},
```

igrigorik: Good catch, fixed!

PR #283, opened 2026-03-20
feat!: Update the Order schema to make currency a required field

deinck6:
# Description

This PR is making the currency field in Order required. This is a followup change from PR 210. When adding the currency field to the Order schema the intention was to make it required during the next version update / breaking change.

## Type of change

      or fields**)

---

files or fields:

```text

The total values within the Order capability are only represented by an amount and rely on platforms having context of the currency established during Checkout. Currency has sense been added to the Order schema but as an optional field to avoid a breaking change. The intention was to make this required during the next version update / breaking change. Currency should be a required field because capabilities are meant to function independently of one another and the monetary values contained within Order are incomplete without this context. 
```

### DIFF
--- a/source/schemas/shopping/order.json
+++ b/source/schemas/shopping/order.json
@@ -86,13 +86,7 @@
     "currency": {
       "type": "string",
       "description": "ISO 4217 currency code. MUST match the currency from the originating checkout session.",
-      "ucp_request": {
-        "transition": {
-          "from": "optional",
-          "to": "required",
-          "description": "Global requirement for currency in all Order requests."
-        }
-      }, 
+      "ucp_request": "omit", 
       "ucp_response": {
         "transition": {
           "from": "optional",


source/schemas/shopping/order.json
```diff
@@ -29,6 +29,7 @@
     "permalink_url",
     "line_items",
     "fulfillment",
+    "currency",
```
jingyli, line 32: For future reference: you can also use the `transition` schema attribute to avoid an immediate breaking change (example: https://github.com/Universal-Commerce-Protocol/ucp/blob/e8d048064fa45f192735916c11c4d42f857687e2/source/schemas/shopping/checkout.json#L32).

igrigorik: +1 please add a transition annotation.

deinck6: Is this needed if I am making the field required immediately for the new version? My understanding was this annotation signals a future change

deinck6: Added annotation

source/schemas/shopping/order.json
```diff
@@ -84,7 +85,21 @@
     },
     "currency": {
       "type": "string",
-      "description": "ISO 4217 currency code. MUST match the currency from the originating checkout session."
+      "description": "ISO 4217 currency code. MUST match the currency from the originating checkout session.",
+      "ucp_request": {
```
igrigorik: shouldn't it be request: omit? currency is a response only attribute, the caller does not provide currency.

jingyli: A fly-by comment: I think we need to revisit `Order` capability schema more holistically on top of the `currency` field - given it currently doesn't have any associated operations (besides webhook support - of which `currency` will be required in the "request" body or webhook payload that business sends to platform), none of the required attributes are actually marked as response-only..

igrigorik: @jingyli yes, that's https://github.com/Universal-Commerce-Protocol/ucp/pull/276 -- let's tackle that there. cc @richmolj

deinck6: updated to omit

PR #288, opened 2026-03-22
feat: add card verification and billing address constraints

jamesandersen:
# Enhancement Proposal: Card Payment Constraints

## Summary

Adds named constraints to `card_payment_instrument.json` that enable merchants to communicate per-checkout instrument requirements to platforms. This gives platforms a clear signal of what the merchant requires — no guesswork, no trial-and-error tokenization.

## Motivation

Many instrument and credential properties are optional at the protocol level but required by individual merchants and their PSPs. For example, CVC is optional in `card_credential.json` but most merchants require it for card-not-present transactions. A handler like `dev.shopify.card` serves thousands of merchants — some require CVC, some don't (it's a per-merchant fraud setting). Without a way to communicate these requirements, the platform must guess which optional fields the merchant actually needs.

## Design

Uses the existing `constraints` mechanism on `available_instruments`, consistent with how `brands` constraints already work. Constraints participate in the existing resolution flow and can vary per merchant without changing the handler schema.

### New Constraints

| Constraint                     | Type    | Description                                                                                                        |
|--------------------------------|---------|--------------------------------------------------------------------------------------------------------------------|
| `requires_card_verification`   | boolean | When `true`, the handler requires card verification data. For FPAN: CVC. For network tokens: cryptogram and ECI.   |
| `requires_billing_address`     | boolean | When `true`, the handler requires a billing address on the instrument.                                             |
| `requires_billing_postal_code` | boolean | When `true`, the handler requires a billing postal code for AVS verification. Ignored when `requires_billing_address` is `true`. |

### Example

```json
{
  "available_instruments": [
    {
      "type": "card",
      "constraints": {
        "brands": ["visa", "mastercard"],
        "requires_card_verification": true,
        "requires_billing_postal_code": true
      }
    }
  ]
}
```

### Per-`card_number_type` Semantics

The `requires_card_verification` constraint adapts to the credential mode:

| `card_number_type` | Verification fields required      |
|--------------------|-----------------------------------|
| `fpan`             | `cvc`                             |
| `network_token`    | `cryptogram`, `eci_value`         |

The spec documents these mappings. A future PR could make the schema fully self-documenting via a `verification` object with conditionals per `card_number_type` (as discussed in PR review), but named constraints provide a pragmatic step forward today.

## Code Changes

**Modified Files:**
- `source/schemas/shopping/types/card_payment_instrument.json` — Added `requires_card_verification`, `requires_billing_address`, and `requires_billing_postal_code` to the `available_card_payment_instrument` constraints
- `docs/specification/payment-handler-guide.md` — Added Card Constraints section with constraint table, example, and resolution flow documentation

## Test Plan

- [x] JSON Schema validation: new constraints accepted on `card_payment_instrument.json`
- [x] Spell check (cspell): 0 issues
- [x] Markdown lint (markdownlint): 0 issues
- [ ] Round-trip test: verify constraints flow through resolution in `available_instruments`

## References

- [PR #187](https://github.com/Universal-Commerce-Protocol/ucp/pull/187) — Original constraints mechanism
- [PR #49](https://github.com/Universal-Commerce-Protocol/ucp/pull/49) — Prior instrument schema discussion

---

## Type of change

- [x] Documentation update

## Checklist

### DIFF
--- a/docs/specification/payment-handler-guide.md
+++ b/docs/specification/payment-handler-guide.md
@@ -212,7 +212,7 @@ signal of what the merchant requires — no guesswork, no retry loops.
 |---------------------------------|---------|----------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `brands`                        | array   | -        | Limit to specific card brands (e.g. `["visa", "mastercard"]`).                                                                                                         |
 | `requires_card_verification`    | boolean | `false`  | When `true`, the handler requires card verification data appropriate to the credential type (e.g., CVC for FPAN).                                                      |
-| `requires_billing_address_data` | string  | `"none"` | Level of billing address data required. `full_address`: complete billing address. `postal_code`: postal code only (e.g. for AVS). `none`: no billing address required. |
+| `billing_address_granularity`   | string  | `"none"` | Level of billing address required. `full_address` — complete billing address; `postal_code` — postal code only (e.g., for AVS); `none` — not required. |
 
 **Example — merchant requires card verification and billing address:**
 
@@ -224,7 +224,7 @@ signal of what the merchant requires — no guesswork, no retry loops.
       "constraints": {
         "brands": ["visa", "mastercard"],
         "requires_card_verification": true,
-        "requires_billing_address_data": "full_address"
+        "billing_address_granularity": "full_address"
       }
     }
   ]
@@ -323,7 +323,7 @@ and typically includes different configuration:
       "constraints": {
         "brands": ["visa", "mastercard"],
         "requires_card_verification": true,
-        "requires_billing_address_data": "full_address"
+        "billing_address_granularity": "full_address"
       }
     }
   ],
@@ -575,7 +575,7 @@ defines `available_card_payment_instrument` with card-specific constraints.
 | Schema                                                                                                 | Constraints                                                                                           |
 | :----------------------------------------------------------------------------------------------------- | :---------------------------------------------------------------------------------------------------- |
 | [`available_payment_instrument.json`](site:schemas/shopping/types/available_payment_instrument.json)   | Base: type, constraints (open object)                                                                 |
-| `card_payment_instrument.json#/$defs/available_card_payment_instrument`                                | Extends base with `constraints.brands`, `requires_card_verification`, `requires_billing_address_data` |
+| `card_payment_instrument.json#/$defs/available_card_payment_instrument`                                | Extends base with `constraints.brands`, `requires_card_verification`, `billing_address_granularity` |
 
 Handlers reference these instrument-defined schemas when declaring
 `available_instruments`. The **instrument schema authors** define what
--- a/docs/specification/payment-handler-guide.md
+++ b/docs/specification/payment-handler-guide.md
@@ -197,6 +197,54 @@ instrument types defined by its handler schema. When present, it narrows the
 advertised types and/or applies additional constraints (e.g., limiting card
 brands to `["visa", "mastercard"]`).
 
+#### Card Constraints
+
+For card instruments, the `constraints` object supports additional properties
+beyond `brands` that communicate per-merchant requirements to the platform.
+
+Many instrument and credential properties are optional at the protocol level
+but required by individual merchants and their PSPs. For example, CVC is
+optional in `card_credential.json` but most merchants require it for
+card-not-present transactions. Named constraints give the platform a clear
+signal of what the merchant requires — no guesswork, no retry loops.
+
+| Constraint                     | Type    | Description                                                                                                                      |
+|--------------------------------|---------|----------------------------------------------------------------------------------------------------------------------------------|
+| `brands`                       | array   | Limit to specific card brands (e.g., `["visa", "mastercard"]`)                                                                   |
+| `requires_card_verification`   | boolean | When `true`, the handler requires card verification data. For FPAN: CVC. For network tokens: cryptogram and ECI.                 |
+| `requires_billing_address`     | boolean | When `true`, the handler requires a billing address on the instrument.                                                           |
+| `requires_billing_postal_code` | boolean | When `true`, the handler requires a billing postal code for AVS verification. Ignored when `requires_billing_address` is `true`. |
+
+**Example — merchant requires card verification and billing address:**
+
+```json
+{
+  "available_instruments": [
+    {
+      "type": "card",
+      "constraints": {
+        "brands": ["visa", "mastercard"],
+        "requires_card_verification": true,
+        "requires_billing_address": true
+      }
+    }
+  ]
+}
+```
+
+The platform reads `constraints` from the resolved `available_instruments`
+in the response and knows exactly what the merchant requires before
+constructing the instrument.
+
+Because constraints live on `available_instruments`, they participate in the
+existing resolution flow and can vary per merchant — a handler like
+`dev.shopify.card` serves thousands of merchants, some requiring CVC and
+some not, and each checkout resolves the correct constraints for that
+merchant.
+
+When a constraint is absent or `false`, the handler places no additional
+requirement for that field. Platforms fall back to existing behavior.
+
 ---
 
 #### Handler Declaration Variants
@@ -267,7 +315,9 @@ and typically includes different configuration:
     {
       "type": "card",
       "constraints": {
-        "brands": ["visa", "mastercard"]
+        "brands": ["visa", "mastercard"],
+        "requires_card_verification": true,
+        "requires_billing_address": true
       }
     }
   ],
@@ -313,6 +363,11 @@ authoritative value returned in the `response_schema`.
 
 In this example, the business's PSP is not configured for Discover, so Discover
 is excluded from the response even though the platform supports it.
+
+Constraints participate in this resolution naturally: the business
+includes constraint values in the resolved `available_instruments` returned
+in the response. The platform reads these requirements at the same time it
+discovers which instruments are available — no additional round-trip needed.
 
 ---
 
@@ -509,12 +564,12 @@ multiple instrument types for different payment flows.
 Each instrument schema defines its own `available_*` variant in `$defs` that
 specifies what constraints are valid for that instrument type. For example,
 [`card_payment_instrument.json`](site:schemas/shopping/types/card_payment_instrument.json)
-defines `available_card_payment_instrument` with a `brands` constraint.
+defines `available_card_payment_instrument` with card-specific constraints.
 
-| Schema                                                                                                 | Constraints                                                     |
-| :----------------------------------------------------------------------------------------------------- | :-------------------------------------------------------------- |
-| [`available_payment_instrument.json`](site:schemas/shopping/types/available_payment_instrument.json)   | Base: type, constraints (open object)                           |
-| `card_payment_instrument.json#/$defs/available_card_payment_instrument`                                | Extends base with `constraints.brands` for card networks        |
+| Schema                                                                                                 | Constraints                                                                                                                      |
+| :----------------------------------------------------------------------------------------------------- | :------------------------------------------------------------------------------------------------------------------------------- |
+| [`available_payment_instrument.json`](site:schemas/shopping/types/available_payment_instrument.json)   | Base: type, constraints (open object)                                                                                            |
+| `card_payment_instrument.json#/$defs/available_card_payment_instrument`                                | Extends base with `constraints.brands`, `requires_card_verification`, `requires_billing_address`, `requires_billing_postal_code` |
 
 Handlers reference these instrument-defined schemas when declaring
 `available_instruments`. The **instrument schema authors** define what
--- a/docs/specification/payment-handler-guide.md
+++ b/docs/specification/payment-handler-guide.md
@@ -208,12 +208,11 @@ optional in `card_credential.json` but most merchants require it for
 card-not-present transactions. Named constraints give the platform a clear
 signal of what the merchant requires — no guesswork, no retry loops.
 
-| Constraint                     | Type    | Default | Description                                                                                                                      |
-|--------------------------------|---------|---------|----------------------------------------------------------------------------------------------------------------------------------|
-| `brands`                       | array   | —       | Limit to specific card brands (e.g., `["visa", "mastercard"]`)                                                                   |
-| `requires_card_verification`   | boolean | `false` | When `true`, the handler requires card verification data. For FPAN: CVC. For network tokens: cryptogram and ECI.                 |
-| `requires_billing_address`     | boolean | `false` | When `true`, the handler requires a billing address on the instrument.                                                           |
-| `requires_billing_postal_code` | boolean | `false` | When `true`, the handler requires a billing postal code for AVS verification. Ignored when `requires_billing_address` is `true`. |
+| Constraint                      | Type    | Default  | Description                                                                                                                                                            |
+|---------------------------------|---------|----------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `brands`                        | array   | -        | Limit to specific card brands (e.g. `["visa", "mastercard"]`).                                                                                                         |
+| `requires_card_verification`    | boolean | `false`  | When `true`, the handler requires card verification data appropriate to the credential type (e.g., CVC for FPAN).                                                      |
+| `requires_billing_address_data` | string  | `"none"` | Level of billing address data required. `full_address`: complete billing address. `postal_code`: postal code only (e.g. for AVS). `none`: no billing address required. |
 
 **Example — merchant requires card verification and billing address:**
 
@@ -225,7 +224,7 @@ signal of what the merchant requires — no guesswork, no retry loops.
       "constraints": {
         "brands": ["visa", "mastercard"],
         "requires_card_verification": true,
-        "requires_billing_address": true
+        "requires_billing_address_data": "full_address"
       }
     }
   ]
@@ -242,8 +241,9 @@ existing resolution flow and can vary per merchant — a handler like
 some not, and each checkout resolves the correct constraints for that
 merchant.
 
-When a constraint is absent or `false`, the handler places no additional
-requirement for that field. Platforms fall back to existing behavior.
+When a constraint is absent or set to its default value, the handler places
+no additional requirement for that field. Platforms fall back to existing
+behavior.
 
 Constraints are additive to schema requirements. If the instrument or
 credential schema marks a field as required, the corresponding constraint
@@ -323,7 +323,7 @@ and typically includes different configuration:
       "constraints": {
         "brands": ["visa", "mastercard"],
         "requires_card_verification": true,
-        "requires_billing_address": true
+        "requires_billing_address_data": "full_address"
       }
     }
   ],
@@ -572,10 +572,10 @@ specifies what constraints are valid for that instrument type. For example,
 [`card_payment_instrument.json`](site:schemas/shopping/types/card_payment_instrument.json)
 defines `available_card_payment_instrument` with card-specific constraints.
 
-| Schema                                                                                                 | Constraints                                                                                                                      |
-| :----------------------------------------------------------------------------------------------------- | :------------------------------------------------------------------------------------------------------------------------------- |
-| [`available_payment_instrument.json`](site:schemas/shopping/types/available_payment_instrument.json)   | Base: type, constraints (open object)                                                                                            |
-| `card_payment_instrument.json#/$defs/available_card_payment_instrument`                                | Extends base with `constraints.brands`, `requires_card_verification`, `requires_billing_address`, `requires_billing_postal_code` |
+| Schema                                                                                                 | Constraints                                                                                           |
+| :----------------------------------------------------------------------------------------------------- | :---------------------------------------------------------------------------------------------------- |
+| [`available_payment_instrument.json`](site:schemas/shopping/types/available_payment_instrument.json)   | Base: type, constraints (open object)                                                                 |
+| `card_payment_instrument.json#/$defs/available_card_payment_instrument`                                | Extends base with `constraints.brands`, `requires_card_verification`, `requires_billing_address_data` |
 
 Handlers reference these instrument-defined schemas when declaring
 `available_instruments`. The **instrument schema authors** define what
--- a/docs/specification/payment-handler-guide.md
+++ b/docs/specification/payment-handler-guide.md
@@ -208,12 +208,12 @@ optional in `card_credential.json` but most merchants require it for
 card-not-present transactions. Named constraints give the platform a clear
 signal of what the merchant requires — no guesswork, no retry loops.
 
-| Constraint                     | Type    | Description                                                                                                                      |
-|--------------------------------|---------|----------------------------------------------------------------------------------------------------------------------------------|
-| `brands`                       | array   | Limit to specific card brands (e.g., `["visa", "mastercard"]`)                                                                   |
-| `requires_card_verification`   | boolean | When `true`, the handler requires card verification data. For FPAN: CVC. For network tokens: cryptogram and ECI.                 |
-| `requires_billing_address`     | boolean | When `true`, the handler requires a billing address on the instrument.                                                           |
-| `requires_billing_postal_code` | boolean | When `true`, the handler requires a billing postal code for AVS verification. Ignored when `requires_billing_address` is `true`. |
+| Constraint                     | Type    | Default | Description                                                                                                                      |
+|--------------------------------|---------|---------|----------------------------------------------------------------------------------------------------------------------------------|
+| `brands`                       | array   | —       | Limit to specific card brands (e.g., `["visa", "mastercard"]`)                                                                   |
+| `requires_card_verification`   | boolean | `false` | When `true`, the handler requires card verification data. For FPAN: CVC. For network tokens: cryptogram and ECI.                 |
+| `requires_billing_address`     | boolean | `false` | When `true`, the handler requires a billing address on the instrument.                                                           |
+| `requires_billing_postal_code` | boolean | `false` | When `true`, the handler requires a billing postal code for AVS verification. Ignored when `requires_billing_address` is `true`. |
 
 **Example — merchant requires card verification and billing address:**
 
@@ -244,6 +244,12 @@ merchant.
 
 When a constraint is absent or `false`, the handler places no additional
 requirement for that field. Platforms fall back to existing behavior.
+
+Constraints are additive to schema requirements. If the instrument or
+credential schema marks a field as required, the corresponding constraint
+is redundant and **MUST NOT** be interpreted as overriding the schema. A
+`requires_*` constraint of `false` or absent does not make a
+schema-required field optional.
 
 ---
 
--- a/source/schemas/shopping/types/card_payment_instrument.json
+++ b/source/schemas/shopping/types/card_payment_instrument.json
@@ -24,6 +24,18 @@
                   "minItems": 1,
                   "uniqueItems": true,
                   "description": "Limit to specific card brands (e.g., ['visa', 'mastercard', 'amex'])."
+                },
+                "requires_card_verification": {
+                  "type": "boolean",
+                  "description": "When true, the handler requires card verification data. For FPAN: CVC. For network tokens: cryptogram and ECI."
+                },
+                "requires_billing_address": {
+                  "type": "boolean",
+                  "description": "When true, the handler requires a billing address on the instrument."
+                },
+                "requires_billing_postal_code": {
+                  "type": "boolean",
+                  "description": "When true, the handler requires a billing postal code for AVS verification. Ignored when requires_billing_address is true."
                 }
               }
             }
--- a/source/schemas/shopping/types/card_payment_instrument.json
+++ b/source/schemas/shopping/types/card_payment_instrument.json
@@ -30,11 +30,11 @@
                   "default": false,
                   "description": "When true, the handler requires card verification data appropriate to the credential type (e.g., CVC for FPAN)."
                 },
-                "requires_billing_address_data": {
+                "billing_address_granularity": {
                   "type": "string",
                   "enum": ["full_address", "postal_code", "none"],
                   "default": "none",
-                  "description": "Specifies the level of billing address data the handler requires. full_address: complete billing address. postal_code: billing postal code only (e.g., for AVS). none: no billing address required."
+                  "description": "Level of billing address required. `full_address` — complete billing address; `postal_code` — postal code only (e.g., for AVS); `none` — not required."
                 }
               }
             }
--- a/source/schemas/shopping/types/card_payment_instrument.json
+++ b/source/schemas/shopping/types/card_payment_instrument.json
@@ -27,14 +27,17 @@
                 },
                 "requires_card_verification": {
                   "type": "boolean",
+                  "default": false,
                   "description": "When true, the handler requires card verification data. For FPAN: CVC. For network tokens: cryptogram and ECI."
                 },
                 "requires_billing_address": {
                   "type": "boolean",
+                  "default": false,
                   "description": "When true, the handler requires a billing address on the instrument."
                 },
                 "requires_billing_postal_code": {
                   "type": "boolean",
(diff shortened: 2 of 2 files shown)


alexpark20 (commented):
I understand the motivation here, but I am concern that this solution is not scalable, especially for any new schema type to be introduced. I think the right solution is to leverage the `.schema` field in entity (`ucp.json#/$defs/entity`).

Before PR #49 , payment handlers had separate `config_schema` (URI) and `instrument_schemas` (array of URIs). PR #49 consolidated these into a single schema URI — but this was a simplification of the reference structure, not a reduction in capability. A single JSON Schema document can contain $defs covering all instrument types and their requirements. That said, I think the protocol's existing schema field is already capable of expressing everything `required_fields` is trying to express, and more. 

Although, one can argue that the current `schema` in the entity (`ucp.json#/$defs/entity`) cannot really express its intention. I imagine people often assume that schema is only for the payment handler config. As an alternative to introducing `required_fields`, I think re introducing `instrument_schema` (just URI, not array though) could solve the problem. What do you think?

alexpark20:
@jamesandersen , thanks for the detailed response and the examples! It really helps me see the full motivation. The per-merchant variance point makes sense, and I think it actually clarifies why we need both required_fields and the static schema rather than one or the other.

I agree that `required_fields` inline in available_instruments is the right place for per-merchant variance to provide that information at run time. 

The one gap I see in `required_fields` alone is that the dot-notation strings are unvalidated (i.e typo such as credential.cvv). Nothing ties "credential.cvc" back to an actual property in the composed instrument and credential schema. I think having the static instrument schema (either via existing schema or new instrument_schema, but I lean into the latter) that platform can fetch / resolve during the discovery closes that gap. In addition, the spec doc can address it explicitly by stating something like

> the business MUST only include paths in `required_fields` that correspond to valid properties defined in ucp schema and `instrument_schema`, and the platform SHOULD validate accordingly

Note that the platform already fetches and evaluates the handler during the discovery, so there's no additional operational overhead. This gives the platform a static, structural definition of what the handler's instrument looks like.

---

Here is how I picture both plays its role:

Similar to how buyer.json (or any other type in the schema) work today, fields are declared as optional unless they are strongly required by the handler itself (e.g. a handler that fundamentally cannot process without a billing address would mark it required). Everything else stays optional in the schema, and per-merchant requirements are communicated at checkout time via `required_fields`. The schema defines the shape and the outer bounds. `required_fields` fills in the runtime specifics. 

---

As suggestion for this PR, would you open to add more language in the doc? I am okay either deferring the decision of introducing a separate schema field in another PR or combining it as part of this PR.

jamesandersen:
@alexpark20 thanks for the thoughtful review and the context on PR #49! I've updated this PR to incorporate your suggestion — the schema now includes both `required_fields` and `instrument_schema`:

- **`instrument_schema`** (new) — a URI to a JSON Schema defining the full instrument structure for the handler. Gives platforms a static, structural definition to validate `required_fields` paths against. No new operational overhead since platforms already fetch/evaluate handler schemas during discovery.
- **`required_fields`** (existing) — runtime per-merchant requirements, resolved per checkout via the existing `available_instruments` resolution flow.

I've also added normative language per your suggestion:

> The business **MUST** only include paths in `required_fields` that correspond to valid properties defined in the applicable UCP instrument and credential schemas or the `instrument_schema`. The platform **SHOULD** validate accordingly.

The PR description and docs have been updated to reflect both properties and how they complement each other. Let me know what you think!

docs/specification/payment-handler-guide.md
```diff
@@ -197,6 +197,86 @@ instrument types defined by its handler schema. When present, it narrows the
...
+Each entry in `available_instruments` **MAY** include a `required_fields`
+array that declares which optional fields from the instrument or credential
+schemas this handler requires for payment processing.
+
+Many instrument and credential properties are optional at the protocol level
+but required by individual merchants and their PSPs. For example, `cvc` is
+optional in `card_credential.json` but most merchants require it for
+card-not-present transactions. Without `required_fields`, the platform must
+guess which optional fields the merchant actually needs — leading to
+trial-and-error tokenization or over-collection of buyer data.
```
raginpirate: I don't think this statement is true. The handler config and specs can carry the way to describe this context, take a look at a real example here with "billingAddressRequired": https://developers.google.com/merchant/ucp/guides/google-pay-payment-handler

jamesandersen:
@raginpirate thanks for the thoughtful pushback — the tokenization example is a great illustration of why this is tricky. The core tension is: merchants need to express per-checkout (not per schema) requirements, but instruments and credentials are modeled as separate structures in UCP, so some requirements are naturally instrument-level (billing address) while others span the credential boundary (card verification).

I considered narrowing `required_fields` to instrument-level paths only (e.g., `billing_address.postal_code`) and using named constraints for cross-credential concerns like card verification. But that still requires constraints for the CVC case — which was the original motivation for this PR — and leaves you with two mechanisms to understand and implement.

I think the cleaner path is to **standardize on `constraints` only**: e.g. add new constraints fields like 
* `requires_card_verification`,
* `requires_billing_address`, etc. 

This gives us a single mechanism, consistent with #187. The tradeoff is that constraint semantics become implicit per credential scheme — `requires_card_verification` means "provide CVC" for raw cards but "provide cryptogram" for network tokens — so the spec needs to document those mappings. But one mechanism with clear documentation feels preferable to two mechanisms.

If this direction works for you and @alexpark20, I'll update the PR accordingly.

jamesandersen:
@kmcduffie good question — you're right that `payment_instrument` references `credential`, so the hierarchy exists. The challenge is that `card_credential` models **both** raw cards and network tokens in a single type via the `card_number_type` discriminator (`fpan` | `network_token` | `dpan`). The fields that matter differ depending on which mode is in play:

- `card_number_type: "fpan"` → merchant may require `cvc`
- `card_number_type: "network_token"` → `cryptogram` and `eci_value` are relevant, but `cvc` is not

So the hierarchy tells the platform *that* a credential is attached, but since it's the same `card_credential` type either way, it can't distinguish which fields are actually required for a given checkout. A flat `required_fields` list like `["credential.cvc"]` would incorrectly demand CVC even when the buyer is paying with a network token.

Full disclosure — this unified modeling initially threw me as well. In PR #296 I proposed a separate `card_network_token_credential` type, which would have naturally sidestepped this problem (distinct types = distinct requirements). But as @raginpirate pointed out there, UCP already models both modes in the unified `card_credential` — consistent with how many processor APIs handle it. Given that direction, the requirements-per-mode problem needs to be solved elsewhere, which is what led me toward named `constraints` like `requires_card_verification: true`. The spec would define what that means per `card_number_type` — "provide `cvc`" for `fpan`, "provide `cryptogram`/`eci_value`" for `network_token` — without the merchant needing to enumerate credential-level paths.

Quick illustration:

```json
{
  "available_instruments": [{
    "type": "card",
    "constraints": {
      "brands": ["visa", "mastercard"],
      "requires_card_verification": true
    }
  }]
}
```

The platform sees `requires_card_verification` and knows from the spec to collect `cvc` for an FPAN or `cryptogram`/`eci_value` for a network token — one mechanism, consistent with #187.

alexpark20:
Ah, I missed the purpose of constraints here and this is a good learning! thank you @raginpirate ! 

After reading comments and new context, here some thoughts:

**On required_fields vs. constraints**

I can see how `constraints` already has the power to express what `required_fields` is trying to do, and I agree we should use a single mechanism rather than introducing a parallel one 👍 

That said, I think the implicit _per-credential-scheme_ semantics (e.g., requires_card_verification meaning CVC for FPAN but cryptogram for network tokens) exposes a few gaps in the current modeling.

1. Instrument-credential binding is loose

From my understanding, the current type-specific mapping (card instrument → card credential) is convention, not schema enforced. When we start putting credential level requirements on instrument level constraints (like `requires_card_verification`), this loose binding becomes a real gap. There is nothing in the schema that formally ties which credential type applies to which instrument type.

As you mentioned already @raginpirate , should we consider introducing `available_credentials` to formalize this binding? Now that we have a concrete use case (credential-level requirements expressed at the instrument level) ? Or is there other nuances I might missed?

2. card_credential.json

Right now cvc, cryptogram, and eci_value are flat siblings on `card_credential`. The meaning of "card verification" is entirely implicit per `card_number_type`. I think we could make this more explicit by grouping them into a verification object (naming could be different) with conditionals that enforce which fields are required per mode:

```json
"verification": {
    "type": "object",
    "description": "Card verification data. Required fields depend on card_number_type."
    "properties": {
      "cvc": {
        "type": "string",
        "maxLength": 4,
        "description": "Card verification code. Applicable to fpan."
      },
      "cryptogram": {
        "type": "string",
        "description": "Cryptogram provided with network tokens."
      },
      "eci_value": {
        "type": "string",
        "description": "Electronic Commerce Indicator provided with network tokens."
      }
    }
}
  
With

  {
    "if": { "properties": { "card_number_type": { "const": "fpan" } } },
    "then": { "properties": { "verification": { "required": ["cvc"] } } }
  },
  {
    "if": { "properties": { "card_number_type": { "const": "network_token" } } },
    "then": { "properties": { "verification": { "required": ["cryptogram", "eci_value"] } } }
  }
```

This way, `requires_card_verification: true` as a named constraint would have a clear and enforceable meaning. The verification object is required, and the schema handles which subfields are needed per mode. We can avoid relying on spec documentation to map implicit semantics.

----

**On instrument_schema**

@raginpirate , I get your point, and that's fair, and this isn't a hill I'll die on. 

But I do want to make sure we're making this call with good reasoning. The current schema description on entity is "URL to JSON Schema defining this entity's structure and payloads", which doesn't clearly indicate it will include input schema for payment instruments when using this particular handler. In practice, the platform and business fetch the handler schema to build/verify the _payment_handlers_ response in both well-known and checkout calls. An `instrument_schema` would serve a different purpose, which is building/verifying the _instrument_ payload submitted. These are related but technically separated entities. Is the benefit of keeping both in the same schema simplicity? Convenience? Just want to understand the reasoning.

jamesandersen:
Thanks @alexpark20 — the `verification` object proposal is a clean way to make the schema self-documenting per `card_number_type`. I think that's a valuable direction, though getting the modeling right there will take some more involved alignment across the group.

Would folks be open to landing this PR as a pragmatic step forward? Concretely, that would mean adding fields like `requires_card_verification` and `requires_billing_address` to `payment_instrument.constraints`, with spec language documenting the per-`card_number_type` semantics (CVC for FPAN, cryptogram/ECI for network tokens). This gives platforms a way to understand per-merchant requirements today, with fully self-documenting schema (the `verification` object, `available_credentials` formalization) as a potential follow-up PR.

I'd update this PR to drop `required_fields` and `instrument_schema`, and add the named constraints instead. @raginpirate @alexpark20 does that work?

raginpirate:
Sorry for the very long circle back on this one @jamesandersen @alexpark20.

@jamesandersen I'm aligned with your proposal to add a few extra constraints for quick product wins 🚀 

@alexpark20 responding to your thoughts, I love seeing you dive into this domain 🤿

> An instrument_schema would serve a different purpose, which is building/verifying the instrument payload submitted. These are related but technically separated entities. Is the benefit of keeping both in the same schema simplicity? Convenience? Just want to understand the reasoning.

I think I get what you are saying, but how would we think about structuring that? 1 handler has N instrument schemas, so I view this just like how a capability defines N payload schemes under it; the problem seems to be the same there. I will not act as a blocker over the delivery of this if integrators agree this is important to optimize how they type-check and type-share across UCP, but I think this is not a problem isolated to payments... maybe we can chat about this problem deeper if you view one.

> As you mentioned already @raginpirate , should we consider introducing available_credentials to formalize this binding? Now that we have a concrete use case (credential-level requirements expressed at the instrument level) ? Or is there other nuances I might missed?

Yeah this one is fair and makes instruments and credentials more expressive 😄 Its not a prioritized investment because most handlers we've looked at really just use a handful of instruments with single credentials right now, but happy to see that opened independently with a clear set of examples we're empowering and/or a real adopter!

> This way, requires_card_verification: true as a named constraint would have a clear and enforceable meaning. The verification object is required, and the schema handles which subfields are needed per mode. We can avoid relying on spec documentation to map implicit semantics.

I think this schema proposal is fair but it doesn't match the shape industry payments APIs; folks keep those values flat on the card object, and I think you can still write a constraint regardless of how it's nested. I would advocate to not take on this refactor.

jamesandersen:
PR updated — dropped `required_fields` and `instrument_schema`, replaced with three named constraints on `card_payment_instrument.json`:

- `requires_card_verification`
- `requires_billing_address`
- `requires_billing_postal_code`

Rebased on latest main, docs updated. Awaiting any further review or, if aligned, approval to merge. @raginpirate @alexpark20

alexpark20:
Thanks for the updates @jamesandersen, and @raginpirate for the thoughts on my earlier point 👏 

I am also aligned with landing this as named constraints `on card_payment_instrument.json`, which is consistent with #187. We can keep this PR tight.

---

Additional notes: 

@raginpirate — fair point on **instrument_schema**, it is not isolated to payments. We can take that to a separate conversation if it turns out to be a real problem worth solving across the protocol 👍 

 On the **verification object**,  I am not fully agree with the idea that constraints work regardless of how fields are nested. With the current flat structure, `requires_card_verification` still carries implicit per type semantics (CVC for FPAN, cryptogram for network tokens) that the spec has to document and implementors have to know. Grouping under a verification object with conditionals would make the schema self-enforcing rather than relying on that documentation. 

Although, your point about matching industry payment API shapes is fair, and I don't think it needs to be solved here as part of this PR.

docs/specification/payment-handler-guide.md
```diff
@@ -197,6 +197,54 @@ instrument types defined by its handler schema. When present, it narrows the
...
+signal of what the merchant requires — no guesswork, no retry loops.
+
+| Constraint                     | Type    | Description                                                                                                                      |
+|--------------------------------|---------|----------------------------------------------------------------------------------------------------------------------------------|
+| `brands`                       | array   | Limit to specific card brands (e.g., `["visa", "mastercard"]`)                                                                   |
+| `requires_card_verification`   | boolean | When `true`, the handler requires card verification data. For FPAN: CVC. For network tokens: cryptogram and ECI.                 |
+| `requires_billing_address`     | boolean | When `true`, the handler requires a billing address on the instrument.                                                           |
+| `requires_billing_postal_code` | boolean | When `true`, the handler requires a billing postal code for AVS verification. Ignored when `requires_billing_address` is `true`. |
```
alexpark20: worth considering the specify whats the default (false I assume) here?

docs/specification/payment-handler-guide.md
```diff
@@ -197,6 +197,54 @@ instrument types defined by its handler schema. When present, it narrows the
...
+
+Because constraints live on `available_instruments`, they participate in the
+existing resolution flow and can vary per merchant — a handler like
+`dev.shopify.card` serves thousands of merchants, some requiring CVC and
+some not, and each checkout resolves the correct constraints for that
+merchant.
+
+When a constraint is absent or `false`, the handler places no additional
+requirement for that field. Platforms fall back to existing behavior.
+
```
alexpark20, line 247: Somewhere in the doc, if not already, can we mention that constraints are additive to schema requirements? For future proofed?

For example, something like, 

"If the instrument or credential schema marks a field as required, the corresponding constraint is redundant and MUST NOT be interpreted as overriding the schema. A requires_* constraint of false or absent does not make a schema-required field optional."

docs/specification/payment-handler-guide.md
jamesandersen: Good call — lines 245-246 covered this but it was easy to miss below the table. Added `default: false` to each constraint in the schema and a Default column to the table so the behavior is clear inline.

docs/specification/payment-handler-guide.md
jamesandersen: Agreed — added normative language in the Card Constraints section:

> Constraints are additive to schema requirements. If the instrument or credential schema marks a field as required, the corresponding constraint is redundant and **MUST NOT** be interpreted as overriding the schema. A `requires_*` constraint of `false` or absent does not make a schema-required field optional.

alexpark20 (approved):
Thank you for iterating and addressing all the feedbacks 👏

source/schemas/shopping/types/card_payment_instrument.json
```diff
@@ -24,6 +24,21 @@
                   "minItems": 1,
                   "uniqueItems": true,
                   "description": "Limit to specific card brands (e.g., ['visa', 'mastercard', 'amex'])."
+                },
+                "requires_card_verification": {
+                  "type": "boolean",
+                  "default": false,
+                  "description": "When true, the handler requires card verification data. For FPAN: CVC. For network tokens: cryptogram and ECI."
```
raginpirate: > For network tokens: cryptogram and ECI

This constraint description is a bit odd to me, because network tokens **can** be valid without either of these; a network token can use cvv. Might be worth focusing this only on the requirement of cvv for fpans in specific, and leaving network tokens as open to the platform to make the right decision on.

source/schemas/shopping/types/card_payment_instrument.json
```diff
@@ -24,6 +24,21 @@
...
+                "requires_billing_address": {
+                  "type": "boolean",
+                  "default": false,
+                  "description": "When true, the handler requires a billing address on the instrument."
+                },
+                "requires_billing_postal_code": {
+                  "type": "boolean",
+                  "default": false,
+                  "description": "When true, the handler requires a billing postal code for AVS verification. Ignored when requires_billing_address is true."
                 }
```
raginpirate: Is this better to represent as an address constraint with three example values (full_address, postal_code, nil)?

raginpirate (commented):
great work, just sharing two notes about the constraints which feel odd.

jamesandersen:
@raginpirate thanks for the review and the approval from @alexpark20 — two good catches here.

**1. `requires_card_verification` and network tokens**

Agreed the current description is too prescriptive — happy to soften it. Before I update, I want to make sure I understand the scenario you're pointing to. My understanding has been that for CIT transactions with network tokens, the cryptogram and ECI are the expected verification mechanism — the cryptogram proves the token is being used by the authorized requestor for this specific transaction, and the ECI indicates the authentication level. Using CVV as an alternative to the cryptogram in a CIT flow is new to me. Is there a specific PSP or card network flow you're thinking of where that applies? Just want to make sure I get the description right and learn something in the process.

Either way, I think the fix is to, as you suggested, simplify the description to something like: *"When true, the handler requires card verification data (e.g., CVC for FPAN credentials)"* — and leave it to the platform to provide the appropriate verification for their credential mode, rather than the constraint trying to enumerate fields per `card_number_type`. The spec docs can provide guidance on common verification patterns without the schema itself being prescriptive.

**2. Billing address — single enum vs. two booleans**

I like this. The two booleans create an awkward "ignored when" relationship and a 4th combination (street-only without postal code) that doesn't map to how AVS is actually used by issuers. A single constraint with enumerated values aligns better with the practical AVS modes merchants configure:

```json
"billing_address_requirement": {
  "type": "string",
  "enum": ["full", "postal_code"],
  "description": "When present, specifies the billing address data the handler requires. 'full' requires a complete billing address. 'postal_code' requires a billing postal code for AVS verification. When absent, no billing address data is required."
}
```

I'll hold off pushing changes until I hear back on #1 so I can address both together.

docs/specification/payment-handler-guide.md
```diff
@@ -197,6 +197,60 @@ instrument types defined by its handler schema. When present, it narrows the
...
+signal of what the merchant requires — no guesswork, no retry loops.
+
+| Constraint                     | Type    | Default | Description                                                                                                                      |
+|--------------------------------|---------|---------|----------------------------------------------------------------------------------------------------------------------------------|
+| `brands`                       | array   | —       | Limit to specific card brands (e.g., `["visa", "mastercard"]`)                                                                   |
+| `requires_card_verification`   | boolean | `false` | When `true`, the handler requires card verification data. For FPAN: CVC. For network tokens: cryptogram and ECI.                 |
+| `requires_billing_address`     | boolean | `false` | When `true`, the handler requires a billing address on the instrument.                                                           |
+| `requires_billing_postal_code` | boolean | `false` | When `true`, the handler requires a billing postal code for AVS verification. Ignored when `requires_billing_address` is `true`. |
```
kmcduffie: What do you think of merging this with `required_billing_address`?

If there was an enum for `requires_billing_address_data`: {FULL_ADDRESS, POSTAL_CODE, NONE}

It would avoid this situation where this property is ignored based on the value of `requires_billing_address`

raginpirate:
@jamesandersen Check out VGS's documentation on it :) https://docs.verygoodsecurity.com/agentic-commerce/provision-network-tokens-for-payments#types-of-cryptograms

> and leave it to the platform to provide the appropriate verification for their credential mode

👍 totally agree!

docs/specification/payment-handler-guide.md
jamesandersen: Yep - similar to the comment above - this is a good suggestion; I just updated the PR accordingly.

jamesandersen:
@raginpirate  ahh ... I see you're talking about the short-form cryptogram that might be suitable for browser "autofill" network tokens - gotcha.

@raginpirate / @kmcduffie - I've swapped to the enum based approach you both suggested.   LMK if anything seems needed here  (🤞 )

source/schemas/shopping/types/card_payment_instrument.json
```diff
@@ -24,6 +24,17 @@
                   "minItems": 1,
                   "uniqueItems": true,
                   "description": "Limit to specific card brands (e.g., ['visa', 'mastercard', 'amex'])."
+                },
+                "requires_card_verification": {
+                  "type": "boolean",
+                  "default": false,
+                  "description": "When true, the handler requires card verification data appropriate to the credential type (e.g., CVC for FPAN)."
+                },
+                "requires_billing_address_data": {
```
raginpirate: Nit: Not a fan of the requires_*_data name...
Should we just simplify to
```suggestion
                "billing_address": {
```

or maybe `billing_address_granularity`?

source/schemas/shopping/types/card_payment_instrument.json
```diff
@@ -24,6 +24,17 @@
...
+                  "type": "string",
+                  "enum": ["full_address", "postal_code", "none"],
+                  "default": "none",
+                  "description": "Specifies the level of billing address data the handler requires. full_address: complete billing address. postal_code: billing postal code only (e.g., for AVS). none: no billing address required."
```
raginpirate: nit on phrasing / punctuation here:

```suggestion
                  "description": "Level of billing address required. `full_address` — complete billing address; `postal_code` — postal code only (e.g., for AVS); `none` — not required."
```

raginpirate (commented):
Sorry for a last round of nits here; expect significantly higher response time from me so we can close this out!

source/schemas/shopping/types/card_payment_instrument.json
jamesandersen: @raginpirate I like that granularity option - just pushed an update for this (and the description nit below)

source/schemas/shopping/types/card_payment_instrument.json
jamesandersen: updated!

docs/specification/payment-handler-guide.md
```diff
@@ -197,6 +197,60 @@ instrument types defined by its handler schema. When present, it narrows the
...
+
+```json
+{
+  "available_instruments": [
+    {
+      "type": "card",
+      "constraints": {
+        "brands": ["visa", "mastercard"],
+        "requires_card_verification": true,
+        "billing_address_granularity": "full_address"
```
gsmith85, line 227: The `billing_address_granularity` enum is doing a lot of work for a concept that
UCP already expresses precisely. `postal_address.json` defines nine named, all-optional fields:

`street_address`, `extended_address`, `address_locality`, `address_region`, `address_country`, `postal_code`, `first_name`, `last_name`, `phone_number`.

The three-level enum collapses all of that into an abstraction that doesn't map cleanly to what PSPs actually require, particularly internationally:
- US AVS needs `postal_code` (full AVS adds `street_address`)
- UK AVS needs `street_address` + `postal_code`
- Billing name matching needs `first_name` + `last_name`
- `full_address` is ambiguous on whether name fields are included

Rather than a granularity level, consider a field-set constraint that uses the vocabulary `postal_address.json` already defines:

```json
"required_billing_address_fields": {
  "type": "array",
  "items": { "type": "string" },
  "description": "Billing address fields the handler requires. Values correspond to field names in postal_address.json (e.g., 'postal_code', 'address_country', 'first_name')."
}

File: docs/specification/payment-handler-guide.md [hunk 10]
```diff
@@ -197,6 +197,60 @@ instrument types defined by its handler schema. When present, it narrows the
...
+but required by individual merchants and their PSPs. For example, CVC is
+optional in `card_credential.json` but most merchants require it for
+card-not-present transactions. Named constraints give the platform a clear
+signal of what the merchant requires — no guesswork, no retry loops.
+
+| Constraint                      | Type    | Default  | Description                                                                                                                                                            |
+|---------------------------------|---------|----------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `brands`                        | array   | -        | Limit to specific card brands (e.g. `["visa", "mastercard"]`).                                                                                                         |
+| `requires_card_verification`    | boolean | `false`  | When `true`, the handler requires card verification data appropriate to the credential type (e.g., CVC for FPAN).                                                      |
```
gsmith85, line 214: Covered by others (@raginpirate, @alexpark20), but I agree that `requires_card_verification` feels off. It's a boolean with no direct linkage to the field it's actually requiring — the platform is left to "know" that this means cvc for FPAN. For network tokens and DPANs, `cryptogram` and `eci_value` are structurally required regardless, so the boolean doesn't meaningfully apply.

Are there other examples you're aiming to solve for with `requires_card_verification`? Otherwise I think we can lean towards directly addressing the CVC case. See [below](https://github.com/Universal-Commerce-Protocol/ucp/pull/288#discussion_r3185742825) for an idea on possible structuring.

source/schemas/shopping/types/card_payment_instrument.json
```diff
@@ -24,6 +24,17 @@
                   "minItems": 1,
                   "uniqueItems": true,
                   "description": "Limit to specific card brands (e.g., ['visa', 'mastercard', 'amex'])."
+                },
+                "requires_card_verification": {
+                  "type": "boolean",
+                  "default": false,
+                  "description": "When true, the handler requires card verification data appropriate to the credential type (e.g., CVC for FPAN)."
+                },
+                "billing_address_granularity": {
```
gsmith85, line 33: Bringing it all together: The two new fields point toward a pattern worth making explicit. Rather than top-level boolean/enum fields on constraints, what if constraints used sub-objects that mirror the instrument's own property hierarchy?

Instead of:

```
"constraints": {
  "requires_card_verification": true,
  "billing_address_granularity": "postal_code"
}
```

Something like:

```
"constraints": {
  "credential": {
    "card_credential": {
      "required": ["cvc"]
    },
    "bank_account_credential": {
      "required": ["account_type"]
    }
  },
  "billing_address": {
    "required": ["postal_code"]
  }
}
```

The credential key mirrors the property on the instrument; the `card_credential` sub-key handles the polymorphic case, constraints are scoped to a specific subtype and can't bleed across incompatible schemas.

The required arrays use field names from the schemas they govern (`card_credential.json`, `postal_address.json`). No new vocabulary to maintain, no translation layer between granularity enum values and actual address fields.

gsmith85: @raginpirate curious if this fits the constraints model you had in mind in #187.

jamesandersen:
@gsmith85 thanks for the detailed review! I want to make sure you have the full context on the discussion arc — I did force push and update the PR content mid-way through the discussion here, which makes it harder to follow the evolution from the diff alone. The comment thread tells the fuller story.

This PR originally proposed `required_fields` (an array of dot-notation paths like `credential.cvc`, `billing_address.postal_code`) — which is conceptually similar to the structured `required` arrays you're proposing. After discussion, @raginpirate [pointed out](https://github.com/Universal-Commerce-Protocol/ucp/pull/288#issuecomment-3035024284) that `payment_instrument.constraints` already has this power (see [#187](https://github.com/Universal-Commerce-Protocol/ucp/pull/187)) and suggested using named constraints as a single mechanism consistent with that existing pattern. @alexpark20 [aligned](https://github.com/Universal-Commerce-Protocol/ucp/pull/288#issuecomment-3056131720) with landing named constraints on `card_payment_instrument.json` as consistent with #187.

So the current approach — flat named constraints — is the result of iterating *away* from the field-reference pattern toward something the maintainers felt was more consistent with the existing constraints model.

On your specific points:

**`requires_card_verification`** — The evolution here was: @raginpirate [noted](https://github.com/Universal-Commerce-Protocol/ucp/pull/288#discussion_r2118979485) that the description was too prescriptive about network tokens (they can use CVV too), so I simplified it to just signal that card verification is required and leave it to the platform to provide the appropriate data for their credential mode. That's where it stands now — intentionally abstract rather than field-specific.

**`billing_address_granularity`** — Your point about international AVS variation is fair. The enum doesn't cover every combination. The tradeoff the group landed on was keeping this simple and extensible — `full_address` and `postal_code` cover the dominant patterns, and additional values can be added to the enum as real adopter needs emerge. That said, if @raginpirate thinks a field-set array like you're proposing would better fit the constraints model, I'm open to it — though it does circle back toward the `required_fields` approach we moved away from.

**Structured sub-object constraints** — This is the most interesting suggestion, but it's also the closest to what `required_fields` was originally doing (field-level references into the schema hierarchy). @raginpirate would have the best perspective on whether this fits the constraints model from #187 — I see you've already tagged him there.

Happy to iterate further if @raginpirate or @alexpark20 feel any of these proposals better serve the constraints direction. The goal has been to keep this PR tight and land incremental value, with room for richer constraint modeling in follow-ups.

raginpirate:
Love the thinking here @gsmith85!
I'll augment @jamesandersen summary above to give my perspective on the feedback.

> billing_address_granularity

Great push on this one; originally I also thought about just having a full-resolution answer of using billing_address and then specifying required fields under it, but I swayed away from it for a more payments-focused abstraction that matches how providers generally solve this problem; full, postal-only, and not required; keeps it simple. I do think this _**works**_ for most if not all merchants.

But seeing this push again echo'd by you @gsmith85 makes me wonder more about constraints generally across UCP. I think it is right for us to build this as constraints for addresses, and _**to share this constraint primitive with other domains**_. Looking around, it seems like fulfillment actually has a gap where it does not expose any concept of address constraints, and we could be building a reasonable foundation here! With this we also future-proof ourselves; we don't have to maintain some payments constraint abstraction for addresses.

TLDR; agreed with this push, and we have an action to take back to the UCP TC to discuss how we share this into the fulfillment extension as an example 😄

> requires_card_verification

For this one, agreed and we've discussed available_credentials needing to be advertised as well in UCP at some point. The nuance here is that although technically we want to describe this requirement about the raw credential here, you can have "credential inception" by which a provider token is what you support ingesting (no cvv field on it for example), but yet we're trying to communicate that we need a cvv on the raw card behind that credential. As well, you would have to redefine the same constraint N times for each credential type.

I see two options here:
* proceed with a proposal to push a simple, abstract helper to cover the concept of required verification across any used credential. This is simple, and generally fits what any merchant needs.
* dive deep here and solve this abstract concern, which I think bringing in a subtype of credentials I'd like to call the funding sources would solve this issue. Communicate available credential constraints as well as available funding source constraints; the two sets of types do not need to overlap.

I did some pairing with Claude and I opened this: https://github.com/Universal-Commerce-Protocol/ucp/pull/424 <-- PTAL @gsmith85 and @jamesandersen! Maybe we can just shift to a better state of the world with the abstract fundamentals here.

jamesandersen:
Just left some comments on #424 - overall that looks promising so opted to bump discussion there ;-)

kmcduffie:
@jamesandersen if you want to continue with this, do you mind resolving the conflicts?

jamesandersen:
While the original problem statement for this PR is still open, it feels like we're closer to resolving it with #424 at this point so closing this one to tidy up.

Issue #185, opened 2026-02-16
EP: content citation and attribution telemetry


jalexspringer:
## Summary

`org.openattribution.telemetry` is a vendor extension that tracks which content influenced AI agent outcomes in commerce conversations. It embeds an `attribution` object in UCP checkout sessions, recording which content was retrieved and cited during the conversation that led to a purchase.

This starts as a vendor extension under the `org.openattribution.*` namespace, per UCP's governance model for new use cases.

## Motivation

Content creators — reviewers, guide authors, comparison sites — invest significantly in producing the expert content that AI shopping agents rely on to generate recommendations. This content is the raw material of AI-assisted commerce: without reviews, there are no credible recommendations; without guides, there is no informed comparison.

Yet creators currently have **zero visibility** into whether their work influences purchases. This is a market failure:

1. **No measurement exists.** There is no standardised way to track which content was retrieved and cited in the conversation that led to a purchase.
2. **Attribution is impossible.** Commerce outcomes (purchases, cart additions) cannot be linked to specific content pieces, so creators cannot demonstrate ROI to partners or advertisers.
3. **Privacy is unaddressed.** Without a purpose-built signal, ad-hoc tracking solutions will emerge that compromise user privacy rather than preserving it.
4. **Multi-session journeys are invisible.** Research, comparison, and purchase often happen across separate conversations over days. Without temporal signals on content retrieval, these journeys are invisible to attribution.

Without measurement, the rational response from creators is to restrict AI access to their content — threatening the content supply chain that the entire AI commerce ecosystem depends on. Attribution telemetry is the feedback loop that makes the ecosystem sustainable.

### Value to UCP

- **First-mover advantage.** UCP becomes the first commerce protocol with native content attribution, differentiating it from competitors.
- **Merchant insight.** Merchants gain content partnership ROI data they cannot get from any other source, enabling data-driven content investment.
- **Zero-risk adoption.** The extension is additive and non-blocking — it cannot interfere with existing checkout flows. Merchants and agents can adopt incrementally.

### Goals

* Define an optional attribution extension that embeds content citation data in UCP checkout sessions
* Support privacy-preserving data sharing with lightweight conversation context (`turn_count`, `topics`)
* Enable multi-conversation attribution by accumulating content from prior conversations into the final checkout's `content_retrieved` array — timestamps on entries provide the temporal signal
* Support negative attribution via `contradiction` citation type (content retrieved but disagreed with)
[...]

Issue #180, opened 2026-02-14
Enhancement Proposal: Attribution Signals Extension


sanderslee2:
## Summary

Today, merchants receive the cart, buyer info, and payment data but no context about the journey that led to the purchase. This extension solves this with a `discovery_attribution` object that describes how a product was discovered and what factors drove the recommendation.

## Motivation

Traditional e-commerce generates rich attribution data through UTM parameters, referral headers, and click paths. UCP identifies which agent initiated a transaction via the agent profile URI but merchants have no visibility into how a product was discovered, why it was selected, or what product attributes drove the agent's recommendation. Without this, merchants cannot measure agent channel ROI, optimize product feeds for agent selection, or understand their competitive positioning in agent recommendations.

### Goals

* Define an optional `discovery_attribution` extension on checkout requests carrying discovery context and match factors
* Use closed enumerations to prevent raw user query leakage
* Enable merchant analytics: agent channel measurement, product feed optimization, and competitive positioning

### Non-Goals

* Affiliate crediting or referral tracking (separate concern)
* Merchant-to-agent attribution feedback
* Extracting how agents compute match factors or rank products

### Capability

```json
{
  "capabilities": [{
    "name": "dev.ucp.shopping.checkout",
    "version": "2026-01-11",
    "extensions": [{
      "name": "dev.ucp.shopping.discovery_attribution",
      "version": "2026-02-18",
      "spec": "https://ucp.dev/extensions/discovery_attribution/spec.json",
      "schema": "https://ucp.dev/extensions/discovery_attribution/schema.json"
    }]
  }]
}
```

### Schema

```json
{
  "discovery_attribution": {
    "schema_version": "2026-02-18",
    "discovery_context": {
      "method": "direct_search | recommendation | comparison | browsing | reorder | wishlist_match | promotion | external_referral",
      "session_type": "active_shopping | passive_suggestion | automated_reorder | gift_recommendation",
      "query_category": "string (max 100 chars)",
      "alternatives_considered": "integer",
      "rank_position": "integer (1 = top)"
    },
    "match_factors": [
      {
        "type": "price | relevance | availability | rating | brand_preference | location | past_purchase | specification_match | promotion | sustainability | delivery_speed",
        "weight": "number [0.0–1.0]",
        "description": "string (max 200 chars, optional)"
      }
    ]
  }
}
```

### Full Example

```json
{
  "line_items": [{ "product_id": "prod_trail_runner_42", "quantity": 1 }],
[...]

PR #10, opened 2026-01-11
feat: add affiliate attribution extension

0xtotaylor:
# Enhancement Proposal: Affiliate Attribution Extension

  | Status | **Provisional** |
  |--------|-----------------|
  | Author | @0xtotaylor |
  | RFC | `rfcs/affiliate-attribution.md` |

  ## Summary

  This proposal extends `POST /checkout-sessions/{id}/complete` to accept an optional `affiliate_attribution` object, enabling agents to transmit fraud-resistant attribution claims for crediting third-party publishers without relying on cookies, redirects, or client-side tracking.

  ## Type of change

  - [x] Enhancement Proposal (new capability/extension per GOVERNANCE.md)

  ## Proposal Contents

  Per the Enhancement Proposal process, the RFC (`rfcs/affiliate-attribution.md`) includes:

  - Summary
  - Motivation
  - Detailed Design
  - Risks
  - Test Plan
  - Graduation Criteria

  ## Implementation

  This PR also includes a complete implementation ready to merge once the proposal reaches **Implementable** status:

  - `source/schemas/shopping/affiliate_attribution.json` — Extension schema
  - `docs/specification/affiliate-attribution.md` — Specification documentation
  - `mkdocs.yml` — Navigation updates

  ## Checklist

  - [x] RFC document follows Enhancement Proposal template
  - [x] Implementation aligns with RFC detailed design
  - [x] Documentation complete
  - [x] Schema validation passes
  - [x] `mkdocs build --strict` passes

  ## Requested Action

  Requesting TC review to advance from **Provisional** → **Implementable**.

PR #295, opened 2026-03-23
feat: add event context extension for cart and checkout sessions

jamesandersen:
# Enhancement Proposal: Event Context Extension

## Summary

An optional `event_context` extension for UCP cart and checkout sessions that enables referring platforms to pass referral context and deduplication keys through agentic checkout flows — preserving the merchant's analytics and ROI measurement capabilities that are otherwise lost when the checkout bypasses the merchant's website.

## Motivation

When a user discovers a product through any external channel — a paid ad, an organic recommendation, an influencer link, or an AI agent's suggestion — and completes a purchase via an agentic commerce protocol, the merchant loses the referral context they normally receive. In the traditional web flow, this data arrives via URL parameters on the merchant's landing page (e.g., `utm_source`, `utm_campaign`, platform-specific click IDs like [gclid](https://support.google.com/google-ads/answer/9744275?hl=en), [fbclid](https://developers.facebook.com/docs/marketing-api/conversions-api/parameters/fbp-and-fbc/), [ttclid](https://ads.tiktok.com/help/article/tiktok-click-id?lang=en)). With agentic checkout, the user never visits the landing page, so these parameters are never set.

This creates two critical problems for merchants:

1. **Merchants lose channel attribution and ROI visibility** — Without campaign and channel identifiers flowing through to the conversion event, merchants cannot attribute sales to the channels that drove them. This breaks the analytics pipelines that merchants rely on to measure channel performance, optimize budgets, and evaluate return on spend. Tools like Google Analytics, Northbeam, and Triple Whale all depend on UTM parameters and click IDs arriving with the transaction to build attribution models. When these signals are missing, the merchant's analytics show agentic purchases as "direct" or "unattributed" traffic.

2. **Conversion reporting double-counts transactions** — When a purchase completes through an agentic flow, both the referring platform and the merchant's existing server-side integration (e.g., Meta Conversions API, Google Enhanced Conversions, TikTok Events API) may report the same transaction independently. Without a shared deduplication key, the same conversion is counted twice — inflating metrics and corrupting automated bidding data.

This affects every platform that refers users to merchant products through agentic checkout — including Google, Meta, TikTok, Snap, Pinterest, OpenAI/ChatGPT, and others.

## Goals

- Enable referring platforms to pass structured referral context (UTM parameters, platform-specific identifiers) through agentic cart and checkout flows to the merchant
- Provide deduplication keys so merchants can reconcile agentic transactions with their existing server-side reporting
- Preserve compatibility with existing merchant analytics tools (Google Analytics, Northbeam, Triple Whale, etc.)
- Support any referring platform — not limited to a specific platform or channel type
- Supply context that downstream analytics systems consume to perform attribution — without UCP itself performing attribution

## Non-Goals

- Multi-touch attribution modeling — this is the merchant's domain, handled by their analytics tools downstream
- Prescribing how merchants should process or store event context data
- Defining platform-specific deduplication protocols — each platform has its own model (Google uses `gclid` + `order_id`, Meta/TikTok use `event_id`, etc.)
- Content creator attribution (covered by [Issue #185](https://github.com/Universal-Commerce-Protocol/ucp/issues/185))
- Agent discovery telemetry (covered by [Issue #180](https://github.com/Universal-Commerce-Protocol/ucp/issues/180))

## Related Work

| Proposal | Focus | Relationship |
|----------|-------|-------------|
| [Issue #180](https://github.com/Universal-Commerce-Protocol/ucp/issues/180) — Attribution Signals | Agent discovery context (how/why product recommended) | Complementary — #180 is about the agent's reasoning; this is about the referral channel data |
| [Issue #185](https://github.com/Universal-Commerce-Protocol/ucp/issues/185) — Content Citation | Content creator attribution (reviews/guides influence) | Complementary — focuses on content influence tracking |
| [PR #10](https://github.com/Universal-Commerce-Protocol/ucp/pull/10) — Affiliate Attribution | Affiliate tracking (stalled) | Overlapping — this proposal is broader, covering all referral platforms not just affiliates |

Neither #180 nor #185 addresses the use case where the platform that referred a user to a product needs to pass referral context through the checkout flow to the merchant.

### Extension Schema

The extension follows the same composition pattern as `fulfillment.json` and `discount.json`: defines the extended cart and checkout in `$defs["dev.ucp.shopping.cart"]` and `$defs["dev.ucp.shopping.checkout"]` using `allOf` to compose onto `cart.json` and `checkout.json`, with `ucp_request` annotations specifying per-operation behavior.

**Event Context Payload:**

| Field | Type | Required | Description |
|-------|------|----------|-------------|
| `platform` | string | Yes | Referring platform identifier (reverse domain naming). MUST correspond to the domain of the platform's `UCP-Agent` profile URL. |
| `dedup_keys` | object | No | Deduplication keys for reconciling with the merchant's server-side reporting |
| `utm` | object | No | Standard UTM parameters for web analytics compatibility |
| `custom` | object | No | Platform-specific key-value pairs routed by the merchant based on the `platform` field. Values MUST be string, number, or boolean. Maximum 20 properties. |

**Dedup Keys:**

| Field | Type | Description |
|-------|------|-------------|
| `event_id` | string | Shared dedup key for reconciling with the merchant's server-side reporting |
| `session_id` | string | Platform session identifier for cross-event correlation (e.g., `fbp`, `ga_session_id`) |

**UTM Parameters:** `utm_source`, `utm_medium`, `utm_campaign`, `utm_content`, `utm_term`, `utm_id`, `utm_source_platform`

**Request Behavior (Cart):**

| Operation | Behavior |
|-----------|----------|
| `create` | `optional` |
| `update` | `optional` |

**Request Behavior (Checkout):**

| Operation | Behavior |
|-----------|----------|
| `create` | `optional` |
| `update` | `optional` |
| `complete` | `optional` |

### Usage Examples

**Google (Google Ads / Gemini)** — A user discovers a product through a Google Shopping ad and completes the purchase via agentic checkout:

```json
{
  "event_context": {
    "platform": "com.google",
    "dedup_keys": {
      "session_id": "GA1.2.1234567890.1710300000"
    },
    "utm": {
      "utm_source": "google",
      "utm_medium": "cpc",
      "utm_campaign": "spring_collection_2026",
      "utm_content": "60123456789",
      "utm_id": "18234567890",
      "utm_source_platform": "Google"
    },
    "custom": {
      "click_id": "EAIaIQobChMI8bXe7...",
      "ad_group_id": "142345678901",
      "placement": "Google_Shopping",
      "gbraid": "WVLA4QjBkaJkZW..."
    }
  }
}
```

Google Ads deduplicates conversions using `gclid` + `order_id` + `conversion_date_time`. The `order_id` is available via UCP's Order capability, so no separate `event_id` is required for Google's dedup model.

**Meta** — A user discovers a product through Meta's platform and their AI agent completes the purchase:

```json
{
  "event_context": {
    "platform": "com.meta",
    "dedup_keys": {
      "event_id": "evt_abc123def456",
      "session_id": "fb.1.1710300000000.1234567890"
    },
    "utm": {
      "utm_source": "meta",
      "utm_medium": "paid_social",
      "utm_campaign": "spring_sale_2026",
      "utm_content": "6861203971771",
      "utm_id": "6861196821371",
      "utm_source_platform": "Meta"
    },
    "custom": {
      "click_id": "IwY2xjawOR56Fle...",
      "placement": "Meta_AI"
    }
  }
}
```

The merchant passes `event_id` and `fbp` (from `session_id`) to their Conversions API integration. The shared `event_id` prevents duplicate event counting between the platform's first-party event and the merchant's server-side event.

**TikTok** — A user discovers a product via TikTok and completes the purchase through an agentic flow:

```json
{
  "event_context": {
    "platform": "com.tiktok",
    "dedup_keys": {
      "event_id": "evt_tt_xyz789abc012"
    },
    "utm": {
      "utm_source": "tiktok",
      "utm_medium": "paid_social",
      "utm_campaign": "spring_launch_2026",
      "utm_source_platform": "TikTok"
    },
    "custom": {
      "click_id": "E.C.P.abcdef123456..."
    }
  }
}
```

**OpenAI (ChatGPT Shopping)** — Even without a traditional campaign, the platform provides event context:

```json
{
  "event_context": {
    "platform": "com.openai",
    "dedup_keys": {
      "event_id": "evt_oai_def456ghi789"
    },
    "utm": {
      "utm_source": "chatgpt",
      "utm_medium": "agentic",
      "utm_campaign": "shopping_recommendations"
    }
  }
}
```

### Design Rationale

1. **Follows UCP extension pattern** — uses `allOf` composition onto `cart.json` and `checkout.json` with `ucp_request` annotations, matching `fulfillment.json` and `discount.json`. Capability declarations (`platform_schema`, `business_schema`) follow the same pattern as `fulfillment.json`.
2. **Separate from `signals`** — UCP's `signals` property carries environment data for authorization and abuse prevention. Event context serves a different purpose: referral and marketing analytics context. Mixing them would conflate security/risk data with marketing data.
3. **Single event context object** — Each platform provides one event context payload per cart or checkout. Multi-touch attribution across platforms is the merchant's domain.
4. **UTM as the universal analytics layer** — UTM parameters are the universal language of web analytics that merchants already know how to process.
5. **Dedup keys as first-class concept** — Different platforms use different dedup models. The `dedup_keys` object accommodates all models with optional fields, and different analytics platforms (GA4, Adobe, Northbeam, Meta CAPI, etc.) handle dedup differently.
6. **Platform-namespaced** — Reverse domain naming provides a clean discriminator. The `platform` value MUST correspond to the domain of the platform's `UCP-Agent` profile URL, linking identity to protocol identity without coupling to UCP-specific constructs.
7. **Custom field for platform-specific data** — Platform-specific identifiers go in `custom`, keeping the core schema minimal and universal. Values constrained to string, number, or boolean with a maximum of 20 properties.
8. **Event context, not attribution** — UCP supplies the context that downstream analytics systems consume to perform attribution. UCP itself does not perform attribution or prescribe attribution models.

## Risks and Mitigations

| Risk | Mitigation |
|------|-----------|
| Privacy concerns about passing tracking data | Event context data flows from the referring platform (which already has it) to the merchant (who would have received it via URL params in the web flow). No new data is created. Privacy note included in specification. |
| Schema bloat from platform-specific fields | Core schema covers only universal fields (UTM, dedup keys). All platform-specific identifiers go in `custom`. |
| Inconsistent dedup models across platforms | `dedup_keys` fields are all optional. Each platform populates only the fields relevant to its dedup model. |
| Overlap with existing proposals (#180, #185, PR #10) | Complementary to #180 and #185 (different data, different use case). PR #10 is stalled and narrower in scope. |

## Test Plan

- JSON Schema validation: validate all usage examples against the schema
- Round-trip test: create checkout with `event_context`, verify it persists through update and complete operations
- Missing fields: verify checkout succeeds when `event_context` is omitted entirely (optional extension)
- Platform routing: verify merchants can discriminate by `platform` field and route to correct analytics integration

## Graduation Criteria

- Schema merged into UCP source
- Specification documentation complete
- 2+ independent implementations (per UCP Enhancement Proposal process)
- TC majority vote for Candidate status

## Code Changes

**New Files:**
- `source/schemas/shopping/event_context.json` — Extension schema with `$defs` containing `event_context_payload`, `dev.ucp.shopping.cart`, and `dev.ucp.shopping.checkout` composition
- `docs/specification/event-context.md` — Extension documentation

**Modified Files:**
- `mkdocs.yml` — Added event context extension to nav and llmstxt sections
- `.cspell/custom-words.txt` — Added related terms

**Note:** No modifications to `checkout.json` or `cart.json` are required. UCP extensions compose onto cart and checkout via `allOf` in the extension schema (same pattern as `fulfillment.json`, `discount.json`).

## References

- [Google Analytics UTM Parameters](https://support.google.com/analytics/answer/1033863)
- [Google Click Identifier (GCLID)](https://support.google.com/google-ads/answer/9744275?hl=en)
- [Google Ads Offline Conversion Deduplication](https://support.google.com/google-ads/answer/7014069)
- [Google Enhanced Conversions](https://developers.google.com/google-ads/api/docs/conversions/enhanced-conversions/web)
- [Meta Conversions API](https://developers.facebook.com/docs/marketing-api/conversions-api)
- [Meta Conversions API Deduplication](https://developers.facebook.com/docs/marketing-api/conversions-api/deduplicate-pixel-and-server-events)
- [TikTok Events API](https://business-api.tiktok.com/portal/docs?id=1741601162187777)
- [TikTok Click ID](https://ads.tiktok.com/help/article/tiktok-click-id?lang=en)
- [Northbeam](https://www.northbeam.io/) — Multi-touch attribution analytics
- [Triple Whale](https://www.triplewhale.com/) — Ecommerce analytics and attribution
- UCP [Payment Handler Guide](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/docs/specification/payment-handler-guide.md)
- UCP `checkout.json` — Base checkout schema (extended via `allOf`)
- UCP `cart.json` — Base cart schema (extended via `allOf`)

---

## Type of change

- [x] Documentation update

## Checklist

### DIFF
--- a/docs/specification/event-context.md
+++ b/docs/specification/event-context.md
@@ -1,0 +1,216 @@
+<!--
+   Copyright 2026 UCP Authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+
+# Event Context Extension
+
+## Overview
+
+The Event Context extension enables referring platforms to pass referral context
+and deduplication keys through agentic checkout flows. When a user discovers a
+product through an external channel (a paid ad, an organic recommendation, an
+influencer link, or an AI agent's suggestion) and completes a purchase via
+agentic checkout, the merchant loses the referral context they normally receive
+via URL parameters on their landing page. This extension preserves that data.
+
+**Key features:**
+
+- Pass structured UTM parameters for compatibility with existing analytics tools
+- Provide deduplication keys so merchants can reconcile agentic transactions
+  with their server-side reporting
+- Support any referring platform via reverse domain naming
+- Pass platform-specific identifiers through an opaque `custom` field
+
+**Dependencies:**
+
+- Cart Capability or Checkout Capability
+
+## Discovery
+
+Businesses advertise event context support in their profile:
+
+```json
+{
+  "ucp": {
+    "version": "{{ ucp_version }}",
+    "capabilities": {
+      "dev.ucp.shopping.event_context": [
+        {
+          "version": "{{ ucp_version }}",
+          "extends": "dev.ucp.shopping.cart",
+          "spec": "https://ucp.dev/{{ ucp_version }}/specification/event-context",
+          "schema": "https://ucp.dev/{{ ucp_version }}/schemas/shopping/event_context.json"
+        },
+        {
+          "version": "{{ ucp_version }}",
+          "extends": "dev.ucp.shopping.checkout",
+          "spec": "https://ucp.dev/{{ ucp_version }}/specification/event-context",
+          "schema": "https://ucp.dev/{{ ucp_version }}/schemas/shopping/event_context.json"
+        }
+      ]
+    }
+  }
+}
+```
+
+## Schema Composition
+
+The event context extension extends both **cart** and **checkout** objects:
+
+- **Base schemas extended**: `cart.json`, `checkout.json`
+- **Path**: `cart.event_context`, `checkout.event_context`
+- **Schema reference**: `event_context.json`
+
+The `event_context` property is composed onto each object via `allOf`,
+following the same extension pattern as `fulfillment.json` and `discount.json`.
+
+## Schema Definition
+
+### Event Context Payload
+
+The `event_context` object contains referral context from the platform that
+drove the cart or checkout:
+
+| Field         | Type   | Required | Description                                                                                                                                               |
+|---------------|--------|----------|-----------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `platform`    | string | Yes      | Referring platform identifier using reverse domain naming (e.g., `com.google`, `com.meta`)                                                                |
+| `dedup_keys`  | object | No       | Deduplication keys for reconciling with the merchant's server-side reporting                                                                              |
+| `utm`         | object | No       | Standard UTM parameters for web analytics compatibility                                                                                                   |
+| `custom`      | object | No       | Platform-specific key-value pairs routed by the merchant based on the `platform` field. Values MUST be string, number, or boolean. Maximum 20 properties. |
+
+### Deduplication Keys
+
+The `dedup_keys` object prevents double-counting when both the referring
+platform and the merchant report the same transaction independently. If
+present, at least one key MUST be provided:
+
+| Field        | Type   | Description                                                                            |
+|--------------|--------|----------------------------------------------------------------------------------------|
+| `event_id`   | string | Shared dedup key for reconciling with the merchant's server-side reporting             |
+| `session_id` | string | Platform session identifier for cross-event correlation (e.g., `fbp`, `ga_session_id`) |
+
+### UTM Parameters
+
+Standard UTM parameters for compatibility with web analytics tools (Google
+Analytics, Northbeam, Triple Whale, etc.):
+
+| Field                  | Type   | Description                                                          |
+|------------------------|--------|----------------------------------------------------------------------|
+| `utm_source`           | string | Traffic source                                                       |
+| `utm_medium`           | string | Marketing medium                                                     |
+| `utm_campaign`         | string | Campaign name                                                        |
+| `utm_content`          | string | Content identifier                                                   |
+| `utm_term`             | string | Search term                                                          |
+| `utm_id`               | string | Campaign ID                                                          |
+| `utm_source_platform`  | string | Advertising platform that served the ad (e.g., Google, Meta, TikTok) |
+
+### Request Behavior
+
+The `event_context` property supports the following operations:
+
+| Operation  | Behavior   | Description                                     |
+|------------|------------|-------------------------------------------------|
+| `create`   | `optional` | Platform MAY include event context on create    |
+| `update`   | `optional` | Platform MAY update event context               |
+| `complete` | `optional` | Platform MAY include event context on complete  |
+
+## Privacy Note
+
+This extension replaces referral context and conversion signals that
+traditionally flow from the referring platform to the business via the
+merchant's website. When a purchase occurs through an agentic flow, the website
+is bypassed and these signals would otherwise be lost. While the fields
+primarily carry marketing and campaign metadata, some values — such as click
+identifiers and session identifiers in `dedup_keys` or `custom` — may be
+considered advertising data subject to user consent under applicable privacy
+regulations.
+
+Implementations **MUST**:
+
+- Only use event context data for the purposes of conversion measurement,
+  reporting, attribution, and analytics
+- **NOT** transmit personally identifiable information (e.g., email addresses,
+  phone numbers, names) through any field in this extension
+- Comply with applicable consent requirements for advertising and marketing
+  data in the jurisdictions where they operate
+
+## Usage Examples
+
+### Google (Google Ads / Gemini)
+
+A user discovers a product through a Google Shopping ad and completes the
+purchase via Gemini's agentic checkout:
+
+```json
+{
+  "event_context": {
+    "platform": "com.google",
+    "dedup_keys": {
+      "event_id": "evt_abc123def456"
+    },
+    "utm": {
+      "utm_source": "google",
+      "utm_medium": "cpc",
+      "utm_campaign": "spring_collection_2026",
+      "utm_content": "60123456789",
+      "utm_id": "18234567890",
+      "utm_source_platform": "Google"
+    },
+    "custom": {
+      "click_id": "EAIaIQobChMI8bXe7...",
+      "ad_group_id": "142345678901",
+      "placement": "Google_Shopping",
+      "gbraid": "WVLA4QjBkaJkZW..."
+    }
+  }
+}
+```
+
+The merchant includes this `event_id` when reporting the conversion via their
+server-side integration to prevent double-counting.
+
+### Meta
+
+A user discovers a product through Meta's platform and their AI agent completes
+the purchase:
+
+```json
+{
+  "event_context": {
+    "platform": "com.meta",
+    "dedup_keys": {
+      "event_id": "evt_abc123def456",
+      "session_id": "fb.1.1710300000000.1234567890"
+    },
+    "utm": {
+      "utm_source": "meta",
+      "utm_medium": "paid_social",
+      "utm_campaign": "spring_sale_2026",
+      "utm_content": "6861203971771",
+      "utm_id": "6861196821371",
+      "utm_source_platform": "Meta"
+    },
+    "custom": {
+      "click_id": "IwY2xjawOR56Fle...",
+      "placement": "Meta_AI"
+    }
+  }
+}
+```
+
+The merchant passes `event_id` and `fbp` (from `session_id`) to their
+Conversions API integration. The shared `event_id` prevents duplicate event
+counting between the platform's first-party event and the merchant's server-side
+event.
--- a/source/schemas/shopping/event_context.json
+++ b/source/schemas/shopping/event_context.json
@@ -1,0 +1,115 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/event_context.json",
+  "name": "dev.ucp.shopping.event_context",
+  "title": "Event Context Extension",
+  "description": "Extends Cart and Checkout with event context, enabling referring platforms to pass referral context and deduplication keys through agentic checkout flows.",
+  "$defs": {
+    "event_context_payload": {
+      "type": "object",
+      "required": ["platform"],
+      "properties": {
+        "platform": {
+          "$ref": "types/reverse_domain_name.json",
+          "description": "Referring platform identifier (reverse domain naming). MUST correspond to the domain of the platform's UCP-Agent profile URL.",
+          "examples": ["com.google", "com.meta", "com.tiktok", "com.openai"]
+        },
+        "dedup_keys": {
+          "type": "object",
+          "description": "Deduplication keys for reconciling this event with the merchant's own server-side reporting. Without these, the same transaction may be counted twice in analytics and reporting platforms.",
+          "minProperties": 1,
+          "properties": {
+            "event_id": { "type": "string", "description": "Shared dedup key. The merchant includes this same event_id when reporting the event via their server-side integration (e.g., CAPI, Enhanced Conversions, Events API) to prevent double-counting." },
+            "session_id": { "type": "string", "description": "Platform session identifier for cross-event correlation (e.g., fbp, ga_session_id)." }
+          }
+        },
+        "utm": {
+          "type": "object",
+          "description": "Standard UTM parameters for compatibility with web analytics tools (Google Analytics, Northbeam, Triple Whale, etc.).",
+          "properties": {
+            "utm_source": { "type": "string" },
+            "utm_medium": { "type": "string" },
+            "utm_campaign": { "type": "string" },
+            "utm_content": { "type": "string" },
+            "utm_term": { "type": "string" },
+            "utm_id": { "type": "string" },
+            "utm_source_platform": { "type": "string" }
+          }
+        },
+        "custom": {
+          "type": "object",
+          "description": "Platform-specific key-value pairs not covered by the structured fields above. Merchants pass these through to their analytics integrations based on the platform field.",
+          "additionalProperties": {
+            "type": ["string", "number", "boolean"]
+          },
+          "maxProperties": 20
+        }
+      }
+    },
+    "dev.ucp.shopping.cart": {
+      "title": "Cart with Event Context",
+      "description": "Cart extended with referral event context.",
+      "allOf": [
+        {
+          "$ref": "cart.json"
+        },
+        {
+          "type": "object",
+          "properties": {
+            "event_context": {
+              "$ref": "#/$defs/event_context_payload",
+              "description": "Referral event context from the platform that drove the cart.",
+              "ucp_request": {
+                "create": "optional",
+                "update": "optional"
+              }
+            }
+          }
+        }
+      ]
+    },
+    "dev.ucp.shopping.checkout": {
+      "title": "Checkout with Event Context",
+      "description": "Checkout extended with referral event context.",
+      "allOf": [
+        {
+          "$ref": "checkout.json"
+        },
+        {
+          "type": "object",
+          "properties": {
+            "event_context": {
+              "$ref": "#/$defs/event_context_payload",
+              "description": "Referral event context from the platform that drove the checkout.",
+              "ucp_request": {
+                "create": "optional",
+                "update": "optional",
+                "complete": "optional"
+              }
+            }
+          }
+        }
+      ]
+    },
+    "dev.ucp.shopping.event_context": {
+      "platform_schema": {
+        "title": "Event Context Capability (Platform)",
+        "description": "Platform-level event context capability. Indicates the platform can provide referral event context on cart and checkout requests.",
+        "allOf": [
+          {
+            "$ref": "../capability.json#/$defs/platform_schema"
+          }
+        ]
+      },
+      "business_schema": {
+        "title": "Event Context Capability (Business)",
+        "description": "Business-level event context capability. Indicates the business accepts and processes referral event context.",
+        "allOf": [
+          {
+            "$ref": "../capability.json#/$defs/business_schema"
+          }
+        ]
+      }
+    }
+  }
+}
--- a/.cspell/custom-words.txt
+++ b/.cspell/custom-words.txt
@@ -107,6 +107,16 @@ zapatillas
 yml
 jwks
 keyid
+CAPI
+dedup
+Dedup
+gbraid
+gclid
+Northbeam
 reauth
 reprepare
 sandboxing
+sess
+tiktok
+xjaw
+WVLA
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -47,6 +47,7 @@ nav:
           - AP2 Mandates Extension: specification/ap2-mandates.md
           - Buyer Consent Extension: specification/buyer-consent.md
           - Discounts Extension: specification/discount.md
+          - Event Context Extension: specification/event-context.md
           - Fulfillment Extension: specification/fulfillment.md
       - Cart Capability:
           - Overview: specification/cart.md
@@ -230,6 +231,7 @@ plugins:
           - specification/ap2-mandates.md
           - specification/buyer-consent.md
           - specification/discount.md
+          - specification/event-context.md
           - specification/fulfillment.md
         Cart Capability:
           - specification/cart.md


source/schemas/shopping/event_context.json
```diff
@@ -0,0 +1,88 @@
...
+          }
+        },
+        "custom": {
+          "type": "object",
+          "description": "Platform-specific key-value pairs not covered by the structured fields above. Merchants pass these through to their analytics integrations based on the platform field.",
+          "additionalProperties": true
+        }
+      }
+    },
+    "dev.ucp.shopping.checkout": {
```
amithanda, line 71: Should attribution extension be applied to Cart as well, in addition to Checkout? Please see discounts extension as an example where it can extend both cart and checkout. In future, it could extend other events as well, but for now maybe starting with Cart and Checkout is a good idea. What do you think?

source/schemas/shopping/event_context.json
```diff
@@ -0,0 +1,88 @@
...
+      "properties": {
+        "platform": {
+          "$ref": "types/reverse_domain_name.json",
+          "description": "Referring platform identifier (reverse domain naming). SHOULD correspond to the domain of the platform's UCP-Agent profile URL.",
+          "examples": ["com.google", "com.meta", "com.tiktok", "com.openai"]
+        },
+        "dedup_keys": {
+          "type": "object",
+          "description": "Deduplication keys for reconciling this event with the merchant's own server-side reporting. Without these, the same transaction may be counted twice in analytics and reporting platforms.",
+          "properties": {
```
amithanda, line 21: Both `event_id` and `session_id` are optional, making an empty dedup_keys: {} schema-valid and semantically meaningless. Should we add "minProperties": 1 to prevent this? 

If `dedup_keys` is there, we should have alteast 1 additional property.

source/schemas/shopping/attribution.json
```diff
@@ -0,0 +1,88 @@
...
+            "utm_campaign": { "type": "string" },
+            "utm_content": { "type": "string" },
+            "utm_term": { "type": "string" },
+            "utm_id": { "type": "string" }
+          }
+        },
+        "custom": {
+          "type": "object",
+          "description": "Platform-specific key-value pairs not covered by the structured fields above. Merchants pass these through to their analytics integrations based on the platform field.",
+          "additionalProperties": true
```
amithanda: I think you flagged this as a discussion point, I think it is a good idea to add some constraints e.g
```
"additionalProperties": {
    "type": ["string", "number", "boolean"]
  },
 "maxProperties": 50
```

docs/specification/event-context.md
```diff
@@ -0,0 +1,187 @@
...
+      "placement": "Meta_AI"
+    }
+  }
+}
+```
+
+The merchant passes `event_id` and `fbp` (from `session_id`) to their
+Conversions API integration. The shared `event_id` prevents duplicate event
+counting between the platform's first-party event and the merchant's server-side
+event.
```
amithanda, line 216: Should we add a callout for privacy:
e.g something like - 
### Privacy Note: 
Platforms and Businesses implementing this extension **SHOULD** ensure that the transmission of attribution data complies with applicable privacy laws and the user's consent preferences. Data passed via the `utm`, `dedup_keys,` or `custom` fields should only be used for the purposes of attribution and conversion reporting as authorized by the user.

source/schemas/shopping/attribution.json
```diff
@@ -0,0 +1,88 @@
...
+  "title": "Attribution Extension",
+  "description": "Extends Checkout with attribution support, enabling referring platforms to pass attribution data and deduplication keys through agentic checkout flows.",
+  "$defs": {
+    "attribution_payload": {
+      "type": "object",
+      "required": ["platform"],
+      "properties": {
+        "platform": {
+          "$ref": "types/reverse_domain_name.json",
+          "description": "Referring platform identifier (reverse domain naming). SHOULD correspond to the domain of the platform's UCP-Agent profile URL.",
```
amithanda: What do you think about using **MUST** for better security?  With **SHOULD** it is possible that a platform could pass com.google attribution while being a different platform entirely.

docs/specification/event-context.md
jamesandersen: Good callout ... let me think about this a bit - the data flowing through here is ideally unrelated to the user (e.g. not derived from the user or user identifiable) but rather a flow between the agent and the business to replace data that normally flows via the website.

source/schemas/shopping/event_context.json
jamesandersen: Makes sense - updated

source/schemas/shopping/attribution.json
jamesandersen: Even 50 feels pretty generous so just added this but set `maxProperties` to 20 as a start

docs/specification/event-context.md
jamesandersen: Alright added some language to reinforce the privacy posture of this extension ... definitely open to more tweaks on the language but was trying to find something striking a reasonable balance of firmness without being anchored or implying any specific jurisdictions/ regulations / privacy policy / terms of service etc.

source/schemas/shopping/attribution.json
jamesandersen: Again, good call; updated.

jamesandersen:
@amithanda just checking in ... is that round of questions available?  As I'm in calls with some early candidates for UCP adoption - lack of visibility to the attribution source of the transaction is recurring theme so I'd love to find an aligned path forward so this can become available to merchants.

docs/specification/attribution.md
```diff
@@ -0,0 +1,187 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+
+# Attribution Extension
+
+## Overview
+
+The Attribution extension enables referring platforms to pass attribution data
```
ashutosh-goog: Is it expected for UCP to  support attribution ? 
May be UCP can provide the event_context and let  the analytics / conversion tracking system perform the right attribution ?

docs/specification/event-context.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Overview
+
+The Attribution extension enables referring platforms to pass attribution data
+and deduplication keys through agentic checkout flows. When a user discovers a
+product through an external channel (a paid ad, an organic recommendation, an
+influencer link, or an AI agent's suggestion) and completes a purchase via
+agentic checkout, the merchant loses the attribution data they normally receive
+via URL parameters on their landing page. This extension preserves that data.
+
```
ashutosh-goog, line 27: Completely agree on the premise hence the suggestion to supply this information as part of event context rather than as attribution information.

docs/specification/attribution.md
```diff
@@ -0,0 +1,187 @@
...
+## Discovery
+
+Businesses advertise attribution support in their profile:
+
+```json
+{
+  "ucp": {
+    "version": "{{ ucp_version }}",
+    "capabilities": {
+      "dev.ucp.shopping.attribution": [
```
ashutosh-goog: Can we rename this event_context. 

Consider user journey as follows. 
The end user clicks an ad link  which opens UCP experience in agent, End user then views the item, adds it to cart, and make purchase. 

Advertiser is interested in all the events mentioned above for understanding the user-funnel, building the audience list for remarketing, as well as ads bidding including both lower funnel and upper funnel bidding. 

These events can happen in short duration to one another or can be hours apart. In both the cases if UCP starts sending attribution information on all events, then it actually turns into marketing attribution solution. 

Another option can be that UCP send event_context information which contains the clicks/utms/referral information and signals what causes this informationt to start. 

In the example above only the view_item event would have the click data and rest would not have any additional information. The marketing analytics solution would perform the right attribution to help Advertiser understand what is driving there conversions.

docs/specification/attribution.md
```diff
@@ -0,0 +1,187 @@
...
+## Schema Definition
+
+### Attribution Payload
+
+The `attribution` object contains referral context from the platform that drove
+the checkout:
+
+| Field         | Type   | Required | Description                                                                                          |
+|---------------|--------|----------|------------------------------------------------------------------------------------------------------|
+| `platform`    | string | Yes      | Referring platform identifier using reverse domain naming (e.g., `com.google`, `com.meta`)           |
```
ashutosh-goog: UTM supports the utm_source_platform.
Do you think we could potentially use that and move it to UTM object?

docs/specification/event-context.md
```diff
@@ -0,0 +1,213 @@
...
+**Key features:**
+
+- Pass structured UTM parameters for compatibility with existing analytics tools
+- Provide deduplication keys so merchants can reconcile agentic transactions
```
ashutosh-goog, line 31: Can txn-id for purchase and event-id for non-purchase be used for dedupes ?

docs/specification/attribution.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Attribution Payload
+
+The `attribution` object contains referral context from the platform that drove
+the cart or checkout:
+
+| Field         | Type   | Required | Description                                                                                          |
+|---------------|--------|----------|------------------------------------------------------------------------------------------------------|
+| `platform`    | string | Yes      | Referring platform identifier using reverse domain naming (e.g., `com.google`, `com.meta`)           |
+| `dedup_keys`  | object | No       | Deduplication keys for reconciling with the merchant's server-side reporting                         |
```
ashutosh-goog: Is this for event dedupe then does it needs to be part of attribution Jason or may be outside of it ?

docs/specification/attribution.md
jamesandersen: Thanks for taking a look @ashutosh-goog. I agree with the framing: UCP should supply context that downstream analytics and attribution systems consume, not perform attribution itself. Renaming this to event_context better reflects that role.
                                                 
I'll update the PR with the rename from attribution → event_context along with the other changes discussed below.

docs/specification/attribution.md
jamesandersen: Yep, aligned with the reframing as per the earlier comment.

One nuance for agentic flows: unlike a traditional web funnel where view_item → add_to_cart → purchase are discrete events hitting the merchant's site, an agent integrated _only for checkout_ has no UCP-level view_item or add_to_cart events at all — the Create Checkout call is the first and potentially only meaningful touchpoint before Complete Checkout. The event context needs to be available on whichever operations the integration actually uses, which is why keeping it optional across cart and checkout makes sense.

The merchant and their analytics stack can then decide how to attribute — UCP just ensures the context isn't lost when the web flow is bypassed. Does that work for you?

docs/specification/attribution.md
jamesandersen: I think your suggested reframing/renaming to `event_context` helps here 👍 . 
The dedup keys are useful event context ... but I think fit better when not framed as attribution data.

docs/specification/attribution.md
jamesandersen: I'll add `utm_source_platform` to the UTM object so it's available for analytics tools that consume it.

I'd propose keeping the top-level platform field as well, since it serves a slightly different purpose: platform is the verified source identifier (reverse domain naming, MUST match the agent's UCP profile) that the merchant uses to route custom field values to the right integration. utm_source_platform is an analytics parameter consumed by tools like [GA4](https://support.google.com/analytics/answer/10917952?hl=en). They'll often align, but separating them keeps the protocol-level identity distinct from the analytics payload.

Happy to discuss further if you think that's redundant — but my thinking is that the cost of having both is low, and it avoids overloading a marketing parameter with a protocol-identity role.

source/schemas/shopping/event_context.json
jamesandersen: Forgot to note it here earlier but did update the PR for this already.

docs/specification/event-context.md
jamesandersen: Certainly possible.  Among the major conversion funnel reporting tools I  was looking at there seemed to be some repeated use of event_id (uniquely identifying discrete events along a purchase journey - add to cart, start checkout, complete checkout) and some concept of sessions.   

The transaction id (`order.id` or possibly even `order.name`) is a good dedupe candidate - but as you noted kind of a special case that's possible when a purchase is complete.  In this case the flow works in the opposite direction e.g. business -> platform (instead of platform -> business for `event_id`).  In the interest of simplicity I didn't include it here but those fields are available via UCP and, if platform and business align, they could also be used.

jamesandersen:
@amithanda  / @ashutosh-goog  - Just updated the PR - again thanks for your comments; I'll be on the lookout for any more follow-ups.

docs/specification/event-context.md
```diff
@@ -0,0 +1,217 @@
...
+
+A user discovers a product through a Google Shopping ad and completes the
+purchase via Gemini's agentic checkout:
+
+```json
+{
+  "event_context": {
+    "platform": "com.google",
+    "dedup_keys": {
+      "session_id": "GA1.2.1234567890.1710300000"
```
ashutosh-goog: sorry for being pedantic. Can we make this event_id for Google traffic as well.

Generally session is the  Analytics concept (with configurable options) so UCP events will just have event-id.

jamesandersen: Not pedantic at all — good catch. Updated the Google example to use `event_id` and simplified the explanatory text.

source/schemas/shopping/event_context.json
mnaga: Would/should this be sent with every request one the Agent identifies the checkout session was initiated by an Event?
If so, what would this mean with multi-item cart/checkout where multiple items were added

igrigorik (commented):
**Fresh-eyes: can `signals` offer a simpler and more robust solution for this problem?**

As a thought experiment, consider...

> Platforms SHOULD emit, under a reverse-domain key in `signals`, the same key-value event and attribution pairs they would encode as URL query parameters on the equivalent browser-based landing flow, enabling businesses to use them with existing analytics integrations.

**Example:**

```json
"signals": {
  "dev.ucp.buyer_ip": "203.0.113.1",

  "com.google.event": {
    "utm_source": "google", "utm_medium": "cpc", "utm_campaign": "spring_2026",
    "utm_id": "18234567890", "gclid": "EAIaIQobChMI...", "gbraid": "WVLA4QjBkaJk...",
    "ad_group_id": "142345"
  },
  "com.meta.event": {
    "utm_source": "meta", "utm_medium": "paid_social", "utm_campaign": "spring_sale",
    "fbclid": "IwY2xjaw...", "fbc": "fb.1.1710300000000.IwY2xjaw...",
    "fbp": "fb.1.1710300000000.1234567890", "event_id": "evt_abc123"
  }, 
 ...
}
```

Existing attribution systems own the vocabulary, UCP is the pipe. This gives each platform full control over how they model attribution, same as they do on the web today, and most importantly there is 1:1 parity between web (parsed `window.location.search`) vs. agent-provided attribution. The merchant takes the signals and forwards them verbatim to their existing infrastructure, no remapping, etc.

#### Why `signals` is appropriate

- **Provenance match** — platform-asserted, write-only, observational.
- **Covers the full journey** — catalog → cart → checkout → order. 
- **Namespace safety** — `propertyNames` prevents `com.google.event` / `com.meta.event` collisions.

Only adjustment would be to update the description from "authorization and abuse prevention" to include marketing/referral provenance. Beyond that, we can document a best-practice and pattern for how platforms should emit this signal, but we don't need to define a unified schema -- each platform maps its existing attribution signals.

This path also eliminates the need for merchant to advertise support: this is a platform signal, its presence/absence does not change anything about the response or API shape offered by the merchant -- same contract as on the web, these signals are provided by the referrer and consumed at-will by the business.

source/schemas/shopping/event_context.json
jamesandersen: @mnaga good questions.

  **Per-request behavior**: The extension uses `ucp_request: "optional"` on create and update operations, so the platform can include it on every request but isn't required to. In practice, I'd expect agents to send it when they want to model a new event — consider how current JS SDKs integrated with websites model events e.g.: 
  * [Google gtag](https://developers.google.com/tag-platform/gtagjs/reference/events) (`add_to_cart`, `begin_checkout`)
  * [Meta Pixel](https://developers.facebook.com/docs/meta-pixel/reference#standard-events) (`AddToCart`, `InitiateCheckout`)

This actually raises an interesting consideration: a single UCP API call could map to what the website would model as *multiple* analytics events — e.g., a cart update that adds one item and removes another would map to both an `add_to_cart` and `remove_from_cart` events from the Google spec above. @ashutosh-goog / @mnaga I sense this might be getting to an edge case outside the scope of UCP but just to test it out - if the UCP platform was going to make server-side calls back to the event reporting APIs of the agent ... would a single event_id be sufficient here (used with multiple event names).  My sense is that modeling multiple event isn't necessary but welcome any pressure testing of this idea.

  **Multi-item carts**: I'm imagining that much of the event context e.g. the important UTM params won't vary across a UCP *session*.  The dedupe data e.g. `event_id` likely *would* change with each UCP API call though allowing platforms to report distinct funnel events where needed (e.g. back to the agent alongside website traffic with appropriate dedupe keys and/or to 3P reporting tools like Triple Whale, NorthBeam, Fivetran, Criteo, etc)

jamesandersen:
@igrigorik an interesting alternative! — I think there are a few tradeoffs either way. I put together a quick comparison to help me think it through:

  | Dimension | Event Context Extension | Add Event Context to Signals |
  |---|---|---|
  | **Adoption cost** | New extension + capability declarations. Platforms learn new extension; merchants declare support. | No new schema. Platforms already send signals. Merchants still need processing logic but don't declare support upfront. |
  | **Cross-platform interop** | Standardized UTM + dedup_keys — one parser for all platforms. | Per-platform opaque payloads — per-platform extraction logic. |
  | **Extensibility** | Platform-specific data in `custom` dict. New shared fields need schema updates. | Platforms add keys freely under their namespace. New shared conventions need only docs. |
  | **Data routing & integrity** | Marketing data structurally separate from fraud/risk signals — different pipelines. Schema enforces types. | Marketing mixed into fraud/risk channel. Merchants must distinguish. No structural validation. |
  
Overall I think using `signals` could work.  The main counter point I'd raise is around *data routing*.  Pragmatically, it feels like `signals` is intended for risk/integrity systems whereas `event_context` would be routed (possibly bumped off to an async tier) to marketing system integrations (e.g. reporting back to agent event APIs, internal marketing systems and/or 3P analytics tools).  Using `signals` could work for both but I'm pondering if a dedicated extension keeps those concerns more separate / clean.
* Some of the other data in `signals` may be more sensitive (as discussed in #203) and put a little more burden on platform implementations to split up before sending along to these disparate systems. 
* having schematized UTM params (which are a de facto analytics standard) could be advantageous for downstream systems

WDYT?

amithanda:
Thanks for the suggestion, @igrigorik! I think there is probably a good case for keeping `event_context` as a separate extension rather than folding it into signals.

My PoV is that they serve fundamentally different use cases:

- **Necessity vs. Optionality:** signals are typically mandatory operational data for risk, fraud, and identity verification. event_context is purely optional marketing and attribution data.

- **Capability Discovery:** by making this an extension, merchants can explicitly declare support for this functionality and the platforms they integrate with. If we put it in signals, it becomes an opaque payload, and we lose that structured negotiation layer where platforms can understand if the merchant actually supports it.

- **Data Routing & Consent:** These payloads go to different places. Security signals are processed in-line for transaction integrity, while marketing context is usually routed asynchronously to analytics platforms. They also have different privacy consent requirements (GDPR/CCPA), and separating them makes compliance cleaner for merchants.

I am leaning more towards keeping this as an extension. WDYT?

igrigorik:
@amithanda @jamesandersen fair points on `signals`: it fits the "platform-asserted" contract, but event/attribution context would require broadening its scope beyond original intent.

That said, I still think capability negotiation is a mismatch for the problem at hand. Event/attribution data is informational: it doesn't affect the negotiation state or the response from the business, and it's not a feature merchants gate integration on. Just as a merchant accepts traffic from any referrer on the web today, it negotiates cart and checkout regardless of source. It _can_ gate based on UCP-Agent's profile, if it only wants to deal with approved agents. It's on the platform to provide (optional) attribution if it wants to claim credit or pass through relevant metadata.

Lastly, I don't think we need to be in the business of inventing how this gets modeled. The simplest and most effective contract is to take exactly what a platform would have sent via query string and serialize it into an open object. Businesses can then feed that object directly into their event/analytics pipeline. Least friction, full symmetry.

To make it concrete, a quick sketch: https://github.com/Universal-Commerce-Protocol/ucp/pull/391

jamesandersen:
@igrigorik just left a comment on#391, but TL;DR I think that approach fits the goals that motivated this PR *and* has simpler schema, no capability negotiation overhead. As far as I'm concerned we can continue the discussion on #391 :-)

igrigorik:
Closing, resolved via #391.

PR #299, opened 2026-03-24
fix!: update total to use signed_amount.json

wry-ry:
# Description

`totals.json` composed `total.json` (which referenced `amount.json` with
`minimum: 0`) via `allOf` and attempted to override the amount with
`signed_amount.json`. But `allOf` is conjunctive in JSON Schema — all
subschemas must pass simultaneously. The override doesn't relax the base
constraint; both apply. This made negative discount amounts unsatisfiable:
the value must be `>= 0` (from `amount.json`) AND `< 0` (from the per-type
conditional).

## Changes

**Schema:**

- `total.json`: use `signed_amount.json` (no minimum) and add per-type sign
  conditionals directly (`discount/items_discount < 0`,
  `subtotal/tax/fee >= 0`)
- `totals.json`: remove the broken `signed_amount` override and the per-type
  conditionals (now inherited from `total.json`)
- `line_item.json`, `order_line_item.json`, `fulfillment_option.json`:
  unchanged — sign rules inherited via `$ref` to `total.json`

**Docs (`discount.md`):**

- Fix references to phantom `line_items[].discount` field that never existed
  in the schema (day-one doc/schema mismatch from initial commit). The actual
  mechanism is `line_items[].totals[type=items_discount]`
- Correct stale "Amount convention" prose that described discount amounts as
  positive — written for the pre-#261 effect field model. Clarify the two
  amounts: `discounts.applied[].amount` is positive (the value),
  `totals[type=discount].amount` is negative (the effect on the receipt)

## Verification

```bash
ucp-schema resolve source/schemas/shopping/types/totals.json \
  --response --op read --bundle --pretty | \
  jq '.items.allOf[0].properties.amount'

# Before: { "type": "integer", "minimum": 0, ... }
# After:  { "type": "integer", ... }  (no minimum — sign-agnostic)
```

## Type of change

      functionality to not work as expected, **including removal of schema
      files or fields**)
- [x] Documentation update

## Type of change

      or fields**)
- [ ] Documentation update

### DIFF
--- a/source/schemas/shopping/types/total.json
+++ b/source/schemas/shopping/types/total.json
@@ -23,5 +23,15 @@
       "$ref": "signed_amount.json",
       "ucp_request": "omit"
     }
-  }
+  },
+  "allOf": [
+    {
+      "if": { "properties": { "type": { "enum": ["discount", "items_discount"] } }, "required": ["type"] },
+      "then": { "properties": { "amount": { "exclusiveMaximum": 0 } } }
+    },
+    {
+      "if": { "properties": { "type": { "enum": ["subtotal", "fulfillment", "tax", "fee"] } }, "required": ["type"] },
+      "then": { "properties": { "amount": { "minimum": 0 } } }
+    }
+  ]
 }
--- a/docs/specification/discount.md
+++ b/docs/specification/discount.md
@@ -138,7 +138,7 @@ operations. The same semantics apply to both resources.
 
 - `discounts.applied` contains all active discounts (code-based + automatic)
 - Rejected codes communicated via `messages[]` (see below)
-- Discount amounts reflected in `totals[]` and `line_items[].discount`
+- Discount amounts reflected in `totals[]` and `line_items[].totals[]`
 
 **Cart-to-checkout continuity:** When a cart is converted to a checkout via the
 cart capability's `cart_id` field, businesses MUST carry forward any discount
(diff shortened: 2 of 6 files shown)


source/schemas/shopping/types/total.json
```diff
@@ -20,7 +20,7 @@
       "ucp_request": "omit"
     },
     "amount": {
-      "$ref": "amount.json",
+      "$ref": "signed_amount.json",
```
alexpark20, line 23: totals.json, which is used in the example already overrides amount to be signed_amount [here](https://github.com/Universal-Commerce-Protocol/ucp/blob/02120f52b3182120f86d3f49b81e1722fe077106/source/schemas/shopping/types/totals.json#L15-L17).

alexpark20 (commented):
What this PR is trying to fix I believe is already possible via the current `totals.json`. 

To understand its intention better, is the change here to update all other places (i.e order_line_item, fulfillment_option) to use signed_amount?

source/schemas/shopping/types/total.json
wry-ry: The problem is that totals.json has allOf referencing total.json, but total.json itself references amount.json and not signed_amount.json which prevents totals from being negative. Should we instead delete { "$ref": "total.json" } from totals.json?

igrigorik:
@alexpark20 good catch, pulling on that thread some more.. also caught a longstanding bug in our prose on discount. Fixing both in same commit.

PR #315, opened 2026-03-27
feat: add com.razorpay.upi UPI Intent payment handler

paraskathuria-dev:
## Summary

Introduces a complete UCP payment handler specification for **UPI Intent payments via Razorpay** (`com.razorpay.upi`), bringing Indian UPI support into the UCP ecosystem.

UPI (Unified Payments Interface) is India's real-time bank-to-bank payment system managed by NPCI. This handler uses the **UPI Intent flow** — the platform deep-links the buyer into their preferred UPI app (Google Pay, PhonePe, Paytm, BHIM, etc.) which authorizes the payment. The resulting cryptographic proof from the Razorpay SDK is submitted at checkout.

### What's included

**Handler spec** (`docs/specification/razorpay-upi-payment-handler.md`):
- Full participant table (Business, Platform, Razorpay as PSP)
- Step-by-step payment protocol with ASCII flow diagram
- Business integration: Razorpay Order creation, HMAC signature verification, payment capture
- Platform integration: SDK initialization, UPI Intent flow, checkout completion
- Security considerations table
- End-to-end test checklist + signature verification code sample

**JSON Schemas** (`source/handlers/razorpay-upi/`):
- `schema.json` — root handler schema for `com.razorpay.upi`
- `types/business_config.json` — `key_id`, `environment`, `merchant_name`
- `types/platform_config.json` — `upi_apps` (optional app constraint)
- `types/response_config.json` — runtime config including `razorpay_order_id`
- `types/upi_intent_instrument.json` — `upi_intent` instrument with `available_upi_intent_instrument` declaration
- `types/razorpay_payment_credential.json` — `razorpay_payment_proof` credential (payment_id + order_id + HMAC signature)

**Navigation** (`mkdocs.yml`): handler added to nav and llmstxt plugin.

### Key design decisions

| Decision | Rationale |
|---|---|
| **No tokenization endpoint** | UPI doesn't transfer raw payment credentials. The Razorpay HMAC-SHA256 proof is verified locally by the business using their `key_secret` — no token storage or detokenize API needed. |
| **`razorpay_order_id` as binding** | The business creates a Razorpay Order before responding to checkout and embeds the `order_id` in `response_config`. The credential must reference this exact ID, tying the payment proof to the checkout and preventing replay attacks. |
| **Manual capture (`payment_capture: 0`)** | Gives the business the opportunity to verify the HMAC signature before funds are committed. Capture only proceeds after successful verification. |
| **UPI Intent only** | UPI Collect is being discontinued. This handler covers Intent exclusively. |

### How it differs from the processor-tokenizer pattern

The `com.razorpay.upi` handler is structurally closer to Shop Pay (`dev.shopify.shop_pay`) than to the processor-tokenizer example — it uses a single fixed instrument type (`upi_intent`) with no card-brand-style constraints, and the credential is a payment proof rather than a stored token. The key addition is `razorpay_order_id` in `response_config`, which serves as the UCP checkout binding and is created by the business backend before returning the checkout response.

## Test plan

- [ ] Validate JSON schemas against JSON Schema draft 2020-12
- [ ] Verify `/.well-known/ucp` business profile declaration parses correctly
- [ ] Create a Razorpay test order (`rzp_test_*` keys) and confirm `order_id` format matches schema pattern `^order_[A-Za-z0-9]+$`
- [ ] Initialize Razorpay SDK with `key_id` + `razorpay_order_id` from `response_config`; confirm intent flow opens
- [ ] Confirm SDK callback delivers `{razorpay_payment_id, razorpay_order_id, razorpay_signature}`
- [ ] Verify HMAC-SHA256 signature using the reference Python snippet in the spec
- [ ] Confirm binding check rejects a credential with a mismatched `razorpay_order_id`
- [ ] Confirm checkout complete payload matches `upi_intent` instrument schema
- [ ] Confirm mkdocs builds without errors (`mkdocs build`)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


source/schemas/shopping/types/upi_intent_credential.json
```diff
@@ -0,0 +1,49 @@
...
+        "type": {
+          "const": "upi_intent",
+          "description": "Discriminator for UPI Intent credential."
+        },
+        "intent_uri": {
+          "type": "string",
+          "pattern": "^upi://pay\\?.*pa=[^&]+.*$",
+          "minLength": 20,
+          "maxLength": 2048,
+          "description": "NPCI-compliant UPI Intent URI. Must contain the 'pa' (payee VPA) parameter per NPCI UPI Linking Specification. Example: upi://pay?pa=merchant@rzp&pn=MerchantName&tr=TXN123&am=500.00&cu=INR"
```
shiv9091, line 23: Given this only includes merchant VPA and amount, how could user confirm that they are indeed paying for the right items?

source/schemas/shopping/types/upi_intent_credential.json
```diff
@@ -0,0 +1,49 @@
...
+    },
+    {
+      "type": "object",
+      "required": ["type", "intent_uri", "transaction_reference"],
+      "properties": {
+        "type": {
+          "const": "upi_intent",
+          "description": "Discriminator for UPI Intent credential."
+        },
+        "intent_uri": {
```
shiv9091, line 18: Since the intent_uri is passed through the Agent/Platform, what prevents a malicious app or a compromised Agent from tampering with the Payee VPA (pa) or Merchant Code (mc) before the user reaches the bank app? Does the protocol provide a way for the Merchant to cryptographically sign the URI itself?

source/schemas/shopping/types/upi_intent_credential.json
```diff
@@ -0,0 +1,49 @@
...
+  "allOf": [
+    {
+      "$ref": "payment_credential.json"
+    },
+    {
+      "type": "object",
+      "required": ["type", "intent_uri", "transaction_reference"],
+      "properties": {
+        "type": {
+          "const": "upi_intent",
```
shiv9091, line 15: Does the handler mandate a system chooser for the upi:// intent? If a malicious app registers itself as the default handler for UPI intents, it could 'mimic' a bank app UI to steal the user's PIN. How does the protocol ensure the user has landed on a genuine, trusted Bank/TPAP surface?

source/schemas/shopping/types/upi_intent_credential.json
paraskathuria-dev: @shiv9091 - This is a fair UX concern, but it's actually a UPI protocol level limitation. UPI intent URIs support core transaction/order references and payment fields, but not complex structured like line-items.

DOC for reference - https://developers.google.com/pay/india/api/android/in-app-payments

source/schemas/shopping/types/upi_intent_credential.json
paraskathuria-dev: In practice, the UPI intent URI is generated server-side by Razorpay and delivered over TLS, which protects it against network-level interception or tampering in transit.

UPI intent flows do not currently use a universal digital signature in the URI that all UPI apps verify. Instead, trust is ensured through merchant onboarding checks, VPA validation, PSP/bank-side validations, user approval within the UPI app, and post-transaction reconciliation.

If a malicious actor were to alter the `pa` (payee VPA):

1. **If changed to a personal VPA**
   The transaction would typically no longer align to a merchant (P2M) payment flow. Depending on the payer app / PSP controls, it may be blocked, reclassified, or flagged due to metadata inconsistencies.

2. **If changed to another merchant VPA**
   Even if a bank app were to process the transfer, it would no longer match the expected Razorpay merchant beneficiary / order mapping. As a result, webhook confirmation, settlement attribution, and reconciliation against the original order would fail, and the merchant payment flow would not complete successfully.

So while generic intent URIs are not the sole trust anchor, the ecosystem uses multiple downstream controls to prevent unauthorized modification from resulting in a valid merchant payment outcome.

source/schemas/shopping/types/upi_intent_credential.json
paraskathuria-dev: The **platform handler** config supports a upi_apps allowlist (e.g., ["gpay", "phonepe", "paytm", "bhim"]). Native platforms should use this to target a specific trusted package, bypassing the chooser entirely.

source/schemas/shopping/types/upi_intent_credential.json
shiv9091: This is little different than regular in-app payments as regular payments only involve humans, here we have agents in the middle which can hallucinate and show different items.

source/schemas/shopping/types/upi_intent_credential.json
shiv9091: How would TPAPs verify payee & order mapping, TPAPs at max can only verify that the payee is of type merchant and has a valid VPA, so if a compromised agent changes the vpa of the merchant to another one, TPAP would allow the payment? Even on the downstream when UPI checks with the other merchant acquirer, it can say yes to incoming transaction and allow payment to happen? User would loose money in this case. As a payment mechanism, there should be a way to verify certain details on the TPAP/PSP side to validate that the URI coming to the TPAP is valid and not tempered with.

paraskathuria-dev: You’re correct, today TPAPs don’t have a cryptographic way to verify that the `pa`, `mc`, or other fields in a UPI intent URI haven’t been tampered with upstream.

In the current design, the intent URI is not a signed payload. TPAPs can validate that the VPA is valid and whether it’s a merchant, but they cannot verify that it matches the original merchant or order generated by the PG. So if a compromised agent changes the VPA to another valid merchant VPA, the TPAP/PSP would still process it, and the transaction could go through if the user approves it.

The main protections today are layered outside the URI itself:

- The UPI app resolves and shows the payee name, which the user is expected to verify before approving.
- PSP/TPAP risk checks and app integrity controls reduce chances of large-scale abuse.
- On the merchant side, order mapping and reconciliation will fail if the payment goes to a different VPA, so the transaction won’t complete as a valid merchant payment from the PG’s perspective.

That said, these are defense-in-depth measures, not a cryptographic guarantee. There isn’t currently a standard mechanism (e.g., signed intent payloads) enforced across the ecosystem by National Payments Corporation of India that allows TPAPs to independently verify URI integrity.

You can read more about UPI Intent as an instrument from this PDF: [https://www.labnol.org/files/linking.pdf](https://www.labnol.org/files/linking.pdf)

source/schemas/shopping/types/upi_intent_credential.json
paraskathuria-dev: Yes, this is how UPI architecture works currently. This is similar to how card transactions work, we don't get to see line items/order details on the 3DS page while completing the transactions.

damaz91:
Closing for staleness.  @paraskathuria-dev please feel free to re-open if you still want to work on this. Thanks!

PR #325, opened 2026-03-31
feat!: align embedded protocol errors with UCP error conventions

westeezy:
Aligns ECP error and success responses with UCP's transport-agnostic conventions. What started as moving delegation errors from JSON-RPC `error` to `result` expanded into broader envelope, error model, and schema alignment.

### UCP envelope on all ECP results

Every ECP request result now uses `oneOf [success, error_response]` with `result.ucp.status` as the discriminator — same pattern as REST and MCP. Applies to all checkout delegation methods (`ec.ready`, `ec.auth`, `ec.payment.*`, `ec.fulfillment.*`, `ec.window.*`) and cart methods (`ep.cart.ready`, `ep.cart.auth`). The JSON-RPC `error` field is reserved for transport-level failures only.

### Decouple `ucp.status` from severity

Previously the spec coupled `ucp.status: "error"` to `severity: "unrecoverable"`. This was accurate for REST/MCP (error_response replaces a missing resource) but breaks for ECP where delegation errors occur within a live session. `ucp.status` is now a pure shape discriminator; `severity` independently prescribes the action.

### Schema refactoring

- **`ucp.json`**: Added `$defs/success` and `$defs/error` — shared building blocks replacing inline `allOf + status const` patterns
- **`embedded.openrpc.json`**: Extracted `components/schemas` for `ucp_success`, `payment_instruments`, `payment_instrument_selection`
- **Removed** `embedded_error_code.json`, `embedded_message_error.json`, `embedded_error_response.json` — three files duplicating the error chain for different `examples` on a freeform string. ECP error codes are documented in spec prose.

### Other fixes

- **ec.ready checkout**: Replaced `$ref` to full `checkout.json` with inline partial type matching what ec.ready actually promises (display-only host hints)
- **ec.ready failure lifecycle**: Host MUST tear down on handshake error
- **Notification cleanup**: Reverted `oneOf [resource, error_response]` on `messages.change` notifications back to plain `$ref` — session errors go through `ec.error` / `ep.cart.error`

---

## Type of change

- [X] **Breaking change** (fix or feature that would cause existing
      or fields**)
- [ ] Documentation update

---

files or fields:

- [X] **I have added `!` to my PR title** (e.g., `feat!: remove field`).

```text

JSON-RPC 2.0 reserves the top-level error field for transport-level failures (parse errors, method not found, invalid params). Delegation outcomes like "user cancelled" or "payment method not supported" are application-level results — the RPC call succeeded, but the outcome was negative.

The previous design conflated these two failure planes, causing two problems:

  1. Middleware can't distinguish transport failures from domain outcomes.
  2. Inconsistency with other UCP transports — REST and MCP return domain errors inside the response body, not via transport error mechanisms.

Moving delegation errors to result.error restores correct JSON-RPC layering: error means the message itself failed; result.error means the delegation produced a negative outcome.
```

---

## Checklist

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules


source/services/shopping/embedded.openrpc.json
```diff
@@ -272,11 +353,20 @@
...
+            {
+              "$ref": "../../schemas/shopping/types/error_response.json"
+            }
+          ]
         }
       }
     }
   ],
-  "x-delegations": ["payment.instruments_change", "payment.credential", "window.open"]
+  "x-delegations": ["payment.instruments_change", "payment.credential", "fulfillment.address_change", "window.open"]
```
westeezy, line 600: Noticed this missing while making the updates for errors.

source/services/shopping/embedded.openrpc.json
```diff
@@ -227,30 +242,96 @@
...
+            },
+            {
+              "$ref": "../../schemas/shopping/types/error_response.json"
             }
+          ]
+        }
+      }
+    },
+
+    {
```
westeezy, line 326: Noticed the address events missing from specification files missing while making the updates for errors.

source/schemas/shopping/types/error_code.json
```diff
@@ -9,6 +9,12 @@
     "item_unavailable",
     "address_undeliverable",
     "payment_failed",
-    "eligibility_invalid"
+    "eligibility_invalid",
```
igrigorik: Since these are EP specific, should we define embedded_error_code that provides this as superset? Otherwise these values will leak as examples to other transports and schemas as well.

igrigorik (changes requested):
_Thinking & writing out loud..._

UCP's promise is to be transport-agnostic. REST and MCP deliver on this and ECP (unfortunately) is the outlier where this promise is broken. As a north star, we _should_ work towards fixing this; we should prefer steps that get us closer to transport-agnostic promise instead of away or preserving status quo.

This PR is a step in the right direction. Moving delegation errors into `result` using `error_response` is correct, but it only fixes the error side, which creates the highlighted asymmetric discriminator: `ucp.status` exists on errors but not on success. And so, why not take the next logical step and complete it: add`ucp` to the success side too. 

```json
// success
{ "result": {
    "ucp": { "version": "2026-01-15", "status": "success" },
    "checkout": { ... } }
}}

// error
{ "result": {
    "ucp": { "version": "2026-01-15", "status": "error" },
    "messages": [{ ... }]
}}
```

With that, `ucp.status` is a consistent discriminator on both branches, matching the same shape we have with REST and MCP. This applies to all delegation results:
- **`ec.ready`**: `{ ucp, upgrade?, checkout? }` — the handshake ack now explicitly confirms the negotiated version
- **`ec.payment.*`**: `{ ucp, checkout }` — partial update with envelope
- **`ec.fulfillment.*`**: same
- **`ec.window.open_request`**: `{ ucp }` — the empty ack becomes a protocol-level confirmation instead of a structureless `{}`

I see all of the above as strict improvements. The version echo is the host explicitly confirming the negotiated `ec_version` -- similarly, explicit confirmation is a feature.

source/schemas/shopping/types/error_code.json
westeezy: Yes I think we should!

docs/specification/embedded-checkout.md
```diff
@@ -426,20 +429,62 @@ All ECP messages **MUST** use JSON-RPC 2.0 format
...
+reserved for transport-level failures (parse errors, method not found, invalid
+params). Implementations **MUST NOT** use the JSON-RPC `error` field for
+delegation-level error codes.
 
 **Success Response:**
 
+All delegation success results **MUST** include the `ucp` envelope with
+`status: "success"`. The `version` **MUST** echo the `ec_version` negotiated
+during session initialization and confirmed by the host in the `ec.ready`
+response. The version is session-bound: once established during the handshake,
```
jingyli, line 439: Curious: Why does the host need to confirm the version again in `ec.ready` response if they are also the ones appending the initial `ec_version` parameter to the `continue_url` to initiate the embedded session?Doesn't this act as a confirmation that both platform & business have successfully negotiated ECP per the version returned in the checkout response?

docs/specification/embedded-checkout.md
```diff
@@ -426,20 +429,62 @@ All ECP messages **MUST** use JSON-RPC 2.0 format
...
+    "checkout": { ... }
+  }
+}
 ```
 
 **Error Response:**
 
+Delegation errors **MUST** be returned as a `result` using the standard UCP
+`error_response` shape: `ucp.status: "error"` with a `messages` array. This is
+the same error shape used by REST and MCP transports — ECP does not define its
```
jingyli: In REST & MCP transports, an `error_response` indicates that the requested resource no longer exists and **MUST** use `severity: "unrecoverable"` (https://ucp.dev/draft/specification/checkout/#error-handling) - platforms are required to recreate a resource once receiving this response.

Seems like EP is still allowing for some flexibility & exception where some errors returned via `error_response` can still be `recoverable`? Is that interpretation correct? While writing this comment, it also made me think back to some of the examples I have added in https://github.com/Universal-Commerce-Protocol/ucp/pull/244 😅

source/schemas/shopping/types/embedded_message_error.json
```diff
@@ -0,0 +1,50 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/embedded_message_error.json",
+  "title": "Embedded Protocol Message Error",
+  "description": "Error message for Embedded Protocol delegation responses. Same as message_error but uses EP-specific error codes.",
```
jingyli: This feels a little bit odd to me... Is there any reason why we can't reuse `message_error` but use an `oneOf` on `code` to point to either `error_code.json` or `embedded_error_code.json`?

Same feedback to `embedded_error_response.json`, is it possible to just compose using existing files if the only thing that differs is the example error codes?

source/services/shopping/embedded.openrpc.json
```diff
@@ -179,33 +197,50 @@
...
+                  "allOf": [
+                    { "$ref": "../../schemas/ucp.json#/$defs/base" },
+                    {
+                      "properties": { "status": { "const": "success" } },
+                      "required": ["status"]
+                    }
+                  ],
+                  "description": "UCP protocol metadata. Status MUST be 'success' for successful delegation."
+                },
+                "checkout": {
```
jingyli, line 264: `checkout.json` also holds `ucp` metadata field - does this structurally mean we will have 2 `ucp` in the response here?

source/services/shopping/embedded.openrpc.json
jingyli: They are currently defined in https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/shopping/fulfillment.json#L83

I recall the motivation was the [composition nature of extensions](https://ucp.dev/latest/specification/overview/#schema-composition) - i.e. transport references core capability and the extension schemas will outline how they augment these base schemas (including any additional embedded delegation methods it may add).

If we are looking to consolidate everything within `embedded.openrpc.json`, should we delete those methods from `fulfillment.json` to eliminate the duplicate?

docs/specification/embedded-checkout.md
```diff
@@ -651,6 +705,29 @@ information:**
 }
 ```
 
+**Example Error Response:**
+
+If the host cannot complete the handshake (e.g., origin validation failure or
+protocol state violation), it **MUST** respond with an `error_response` result:
```
jingyli, line 714: What's the expectation once Embedded Checkout receives this `error_response`? Should we follow a similar pattern as https://github.com/Universal-Commerce-Protocol/ucp/pull/244/changes/073a04b3af27992b275be68e15cefb22026e3d8f where Embedded Checkout must echo back this error + `continue_url` in a way for host to kickoff the iframe tear down & buyer handoff?

docs/specification/embedded-checkout.md
igrigorik: `ucp.version` in the `ec.ready` response is the host's explicit confirmation that it's operating at this version. For subsequent delegation results: mildly redundant given the session contract, but `version` is a required field in `ucp.json#/$defs/base`, and ECP reusing the same UCP envelope as REST/MCP is an explicit goal.

source/schemas/shopping/types/embedded_message_error.json
igrigorik: Fair, simplified.

source/services/shopping/embedded.openrpc.json
igrigorik: Fixed, the problem here was that we were ref'ing checkout json instead of targeted definition.

source/services/shopping/embedded.openrpc.json
igrigorik: Good catch, fixed.

docs/specification/embedded-checkout.md
igrigorik: Fixed

docs/specification/checkout.md
```diff
@@ -142,7 +141,8 @@ placement (e.g., high-value order approval, first-purchase policy).
 When the business cannot create a new resource or the requested resource
 no longer exists, the response contains `ucp.status: "error"` with
 `messages` describing the failure — no resource is included in the
-response body. Error responses MUST use `severity: "unrecoverable"`.
+response body. When no resource exists to act on, messages SHOULD use
```
jingyli, line 144: Naive QQ: Should this be `**MUST**` instead? Is it `**SHOULD**` only because we anticipate `severity: "requires_escalation"` to be a legitimate alternative..?

docs/specification/embedded-cart.md
```diff
@@ -182,20 +182,42 @@ All ECaP messages **MUST** use JSON-RPC 2.0 format
 
 ### Response Handling
 
-For requests (messages with `id`), receivers **MUST** respond with either:
+For requests (messages with `id`), receivers **MUST** respond with a `result`.
+Both success and error outcomes **MUST** be returned via the `result` field,
+consistent with UCP's two-layer error model. The JSON-RPC `error` field is
+reserved for transport-level failures (parse errors, method not found, invalid
+params). Implementations **MUST NOT** use the JSON-RPC `error` field for
+delegation-level error codes.
```
jingyli, line 190: nit: Should we generalize this to `session-level` or `application-level` error codes since with `ep.cart.error`, we are no longer dealing with delegation errors only?

docs/specification/embedded-cart.md
```diff
@@ -350,6 +377,13 @@ over the upgraded communication channel, and wait for a new response. All
 subsequent messages **MUST** be sent only over the upgraded communication
 channel.
 
+If the host cannot complete the handshake (e.g., origin validation failure or
+protocol state violation), it **MUST** respond with an `error_response` result.
+When the host responds with an error, the session cannot proceed. The host
```
jingyli, line 382: > The host **MUST** tear down the embedded context and **MAY** redirect the buyer to
`continue_url` if present. The Embedded Cart **MUST NOT** send further
messages after receiving a handshake error.

I wonder if this is missing a step in-between? In my mind, the following makes slightly more sense:
1) Host responds with `error_response`
2) Embedded Cart echoes with `ep.cart.error` containing a possible `continue_url`
3) Upon receiving 2), host proceed with tear down & redirect
4) Embedded Cart **MUST NOT** send further messages

Without step 2), I don't think there is any opportunity for host to ever receive the `continue_url` from business for handoff.

docs/specification/embedded-checkout.md
```diff
@@ -192,6 +192,9 @@ parameters from business-specific query parameters:
 
 - `ec_version` (string, **REQUIRED**): The UCP version for this session
     (format: `YYYY-MM-DD`). Must match the version from the checkout response.
+    The version is negotiated at session initialization and **MUST** remain
```
jingyli, line 195: nit: Should this same blurb be added to ECaP's documentation as well until we consider stripping out the common pieces post ECP namespace update?

docs/specification/embedded-checkout.md
```diff
@@ -430,16 +433,41 @@ For requests (messages with `id`), receivers **MUST** respond with either:
 
 **Success Response:**
 
+All delegation success results **MUST** include the `ucp` envelope with
```
jingyli, line 436: This content also slightly differ from what is in `embedded-cart.md`, is it possible to keep them consistent?

source/schemas/shopping/fulfillment.json
```diff
@@ -84,59 +84,65 @@
...
                 "properties": {
-                  "fulfillment": {
-                    "$ref": "../../schemas/shopping/types/fulfillment.json",
-                    "description": "Fulfillment with updated destinations and selected_destination_id."
+                  "ucp": { "$ref": "../ucp.json#/$defs/success" },
+                  "checkout": {
+                    "type": "object",
+                    "description": "Partial checkout update with fulfillment address selection.",
+                    "properties": {
+                      "fulfillment": {
```
jingyli, line 131: To confirm on the expectation here: Does this mean in these delegation responses, we would only get back the delta data (i.e. updated `fulfillment`) and won't actually get the entire checkout object back again?

This feels like it differs from other transport and also differs from what the example in https://ucp.dev/draft/specification/embedded-checkout/#ecfulfillmentchange mentions..?

source/services/shopping/embedded.openrpc.json
```diff
@@ -141,11 +166,8 @@
           "name": "checkout",
           "required": true,
           "schema": {
-            "oneOf": [
-              { "$ref": "../../schemas/shopping/checkout.json" },
-              { "$ref": "../../schemas/shopping/types/error_response.json" }
-            ],
-            "description": "Current checkout state with updated messages or an error response."
+            "$ref": "../../schemas/shopping/checkout.json",
```
jingyli, line 169: Referencing back to https://github.com/Universal-Commerce-Protocol/ucp/pull/244#discussion_r3025631266, by removing the `error_response` option here - how would business actually notify hosts that an unrecoverable error has occurred (i.e. OOS) with the resource no longer exist on the session?

jingyli (commented):
Thanks @westeezy @igrigorik, the main constructs LGTM! Left some style-related comments + 2 structural ones related to:
1) Should delegation response echo back the entire checkout object? If not, should we clarify why (is it because host wouldn't hold/be responsible for the other business-only fields) and clean up the comment in the examples?
2) By removing `error_response` from `xx.messages.change`, how will business notify host on actually unrecoverable errors where the resource is no longer being maintained on their end..? Are we expecting them to send that via `xx.error` as well?

Can approve from my end once the 2 points above are resolved!

Also a fly-by comment to @mmohades's note above, specifically on this:
> Highlight the unique direction of Embedded errors: Adding a quick note that outcome errors in ECP/ECaP are returned by the Host (versus the Business in REST/MCP) will make the spec much easier to digest contextually.

I think there is actually an exception to this point, namely `xx.error` messages like https://ucp.dev/draft/specification/embedded-cart/#epcarterror where outcome errors are also being sent in the direction of `Embedded Cart/Checkout -> host`.

source/services/shopping/embedded.openrpc.json
igrigorik: For unrecoverable case, the business emits ec.error with error_response + continue_url, host tears down and redirects. If the session is valid but has errors (e.g. items out of stock), ec.messages.change carries the checkout resource: its status reflects the state (incomplete, requires_escalation) and the messages array contains the errors. The resource still exists; it just needs attention.

docs/specification/embedded-cart.md
igrigorik: The host initiated the embedded flow, it already has the cart/checkout response from the previous step, including continue_url. A handshake failure should be terminal: host sends error response, tears down. To recover, the host can retry the embedded session or fall back to continue_url from the original cart/checkout response. No echo needed — the host has full context from before it loaded the iframe

source/schemas/shopping/fulfillment.json
igrigorik: This is consistent with other delegation responses in embedded, wihch return only the delta. 

The spec prose already describes this: "The Embedded Checkout MUST treat this update as a PUT-style change by entirely replacing the existing state for the provided fields." The example in the spec matches; it shows only checkout.fulfillment, not the full checkout object.

docs/specification/embedded-checkout.md
igrigorik: These are all addressed in #339 which extracts the shared content (Response Handling, version semantics, error code language) into embedded-protocol.md.

docs/specification/embedded-cart.md
igrigorik: Good catch — addressed in #339 where the shared embedded-protocol.md uses "application-level error codes."

docs/specification/embedded-checkout.md
igrigorik: Addressed in #339 — shared embedded-protocol.md covers version semantics once for both capabilities.

docs/specification/checkout.md
igrigorik: `requires_escalation` is a checkout status, not a severity — the severity values are `recoverable`, `requires_buyer_input`, `requires_buyer_review`, `unrecoverable`. When no resource exists, `unrecoverable` is correct, because there's nothing to recover against.

We're using SHOULD instead of MUST because `error_response` is also used in ECP delegation results where the session is still alive and `recoverable` is valid (e.g., user cancelled a payment sheet). The constraint is accurate for REST/MCP but too strict as a universal rule. The error processing algorithm is now updated to route on severity instead of short-circuiting on `ucp.status` — so the behavior is the same for REST/MCP (`unrecoverable` -> retry with new resource), but ECP delegation errors can flow through severity-based routing correctly.

docs/specification/embedded-cart.md
westeezy: I agree with @igrigorik in this regards. The `continue_url` should already be available and in terminal errors I wouldn't expect step 2.

source/schemas/shopping/fulfillment.json
westeezy: @jingyli I think this also may answer your question in another comment of should the delegations echo back the entire checkout object and my response would be no as this precedent exists.

jingyli (approved):
Thanks @igrigorik! My outstanding questions are resolved between https://github.com/Universal-Commerce-Protocol/ucp/pull/339/ and inline responses here.

igrigorik (approved):

PR #330, opened 2026-04-02
 feat!: delegated IdP, identity chaining, and capability-driven scope model

igrigorik:
Redesign identity linking for multi-merchant agentic commerce. Businesses can offer delegated identity providers, allowing buyers to authenticate through a trusted IdP (e.g., Google, Shop) instead of creating a per-merchant account. Platforms can support these IdPs to streamline identity linking — authenticate once, reuse across businesses that trusts the same provider.

This PR carries builds on / supersedes #265, reverted in #329.

## Delegated Identity Providers

The current identity linking spec assumes 1:1 — each business operates its own OAuth authorization server, requiring a full OAuth dance per merchant. In agentic commerce, where platforms shop across many businesses on a buyer's behalf, this means N merchants = N handoffs.

Businesses can now declare trusted identity providers in `config.providers`, keyed by reverse-domain identifier:

```json
"dev.ucp.common.identity_linking": [{
  "config": {
    "providers": {
      "com.google": {
      },
      "com.example.merchant": {
        "auth_url": "https://merchant.example.com/"
      }
    }
  }
}]
```

Here the business trusts Google as an external IdP and also lists itself (`com.example.merchant`) for business-hosted OAuth — both use the same discovery mechanism. Each provider's `auth_url` is resolved via RFC 8414, with OIDC `/.well-known/openid-configuration` fallback on 404.

Two distinct flows:

- **Account Linking**: One-time OAuth flow between platform and IdP. Standard Authorization Code + PKCE, with support for additional grant types (device authorization for headless agents via RFC 8628).
- **Accelerated IdP Flow**: Platform already holds a valid IdP token (e.g., from Google), chains the buyer's identity to a new business without a browser redirect — per [draft-ietf-oauth-identity-chaining-08](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-identity-chaining-08).

### Identity chaining

The accelerated flow implements a two-phase pattern:
1. Platform requests a JWT authorization grant from the IdP via token exchange (RFC 8693), using the `resource` parameter to identify the target business
2. Platform presents the JWT grant to the business's token endpoint via JWT bearer assertion (RFC 7523)
3. Business validates the grant, resolves buyer identity, issues its own access token

The IdP controls consent — it MUST NOT issue grants for businesses the buyer has not authorized. Businesses MAY auto-provision accounts or return a `continue_url` for buyer onboarding (terms acceptance, etc.). Two independent token lifecycles with independent revocation. Refresh tokens SHOULD NOT be issued per the chaining draft.

## Capability-driven scope model

Scopes define the permissions an identified buyer grants to a platform within a capability. Unlike #265's approach of annotating individual capability schemas with `identity_scopes`, scope declarations live centrally on the identity linking config. This is because whether a capability requires buyer auth is a business decision — a B2B wholesaler gates catalog access, a B2C retailer serves it publicly.

### Three access levels

| Level | Authentication | Example |
| :---- | :------------- | :------ |
| **Public** | None | Browse a public catalog |
| **Agent-authenticated** | Platform credentials | Guest checkout, read agent's own orders |
| **Buyer-authenticated** | Platform + buyer identity | Saved addresses, full order history, personalized pricing |

Identity linking upgrades capabilities to buyer-authenticated access. Capabilities are **never** pruned from the intersection based on identity linking; there is no capability pruning algorithm.

### Config shape

Building on the provider example, businesses declare which capabilities offer buyer-scoped features alongside their trusted IdPs:

```json
"dev.ucp.common.identity_linking": [{
  "config": {
    "providers": {
      "com.google": {
      }
    },
    "capabilities": {
      "dev.ucp.shopping.checkout": {},
      "dev.ucp.shopping.order": {
        "required": true,
        "scopes": ["read", "manage"]
      }
    }
  }
}]
```

- **Absent from capabilities map**: no buyer-scoped features (fully public)
- **Present, no `required`**: accessible without buyer auth, identity upgrades the experience
- **`required: true`**: business returns `identity_required` error without buyer identity
- **`scopes`**: sub-scope operation groups defined by each capability's spec, joined with colon on the wire (e.g., `dev.ucp.shopping.order:read`)
- Scope tokens use capability names directly — no separate scope namespace

## Carried forward from #265

- PKCE MUST (platforms and businesses)
- `iss` validation MUST (RFC 9207, both sides)
- Exact `redirect_uri` matching MUST (businesses)
- `scopes_supported` MUST in RFC 8414 metadata
- 2-step discovery hierarchy: RFC 8414 primary, OIDC fallback on 404 only, strict abort on other errors
- JWT authorization grants: 60s lifetime, `jti` single-use enforcement

---

## Type of change
      or fields**)

## Checklist


aglazer:
Thanks @igrigorik. Agree with the changes. Great addition to UCP!

aglazer (approved):

docs/specification/identity-linking.md
```diff
@@ -16,349 +16,685 @@
...
+* **SHOULD** include a `state` parameter per
+    [OAuth 2.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-14){ target="_blank" }.
+* When `config.providers` is present in the business's identity linking
+    capability, the platform **SHOULD** select a provider it supports from
+    the business's advertised list (see
+    [Identity Providers](#identity-providers)).
+* Before initiating identity chaining with a business, the platform
+    **SHOULD** offer the buyer a choice of available identity providers and
+    indicate that the selected provider's identity will be shared with the
+    business.
```
gsmith85, line 81: Is SHOULD the right keyword here if there is only one in-common provider, which would effectively be a no-op?

Also worth considering that a platform may have a preferred IdP (or operate its own), which isn't just a UX convenience, it has potential security advantages around revocation, since a platform controlling its IdP can enforce consistent token revocation across all trusting businesses. That said, the intent to preserve buyer choice and visibility over which identity is shared is an important counterbalance.

docs/specification/identity-linking.md
```diff
@@ -16,349 +16,685 @@
...
+* Revocation and security events
+    * **SHOULD** revoke tokens at both layers when a buyer initiates an
+        unlink action (see [Token Lifecycle](#token-lifecycle)).
```
gsmith85, line 84: Consider splitting the platform's revocation obligation into two levels:
- Revoking the business-issued token should be a MUST as that's the direct relationship the buyer explicitly asked to sever.
- Revoking the IdP token is a broader action that would cascade across all business relationships established through that IdP, so MAY feels more appropriate, leaving the platform to determine the right blast radius.

docs/specification/identity-linking.md
```diff
@@ -16,349 +16,685 @@
...
+    resource.
+* Additional permissions **MAY** be granted beyond those explicitly requested,
+    provided that the requested scopes are, at minimum, included.
+* The platform and business **MAY** define additional custom scopes beyond the
+    minimum scope requirements.
+* Revocation and security events
+    * **MUST** implement standard Token Revocation as defined in
+        [RFC 7009](https://datatracker.ietf.org/doc/html/rfc7009){ target="_blank" }.
+    * **MUST** revoke the specified token and **SHOULD** recursively revoke
+        all associated tokens (e.g., revoking a `refresh_token` **MUST** also
```
gsmith85, line 139: The example here promotes the `SHOULD` guidance into a `MUST`.  Should revocation really be optional?

docs/specification/identity-linking.md
```diff
@@ -16,349 +16,685 @@
...
+grants for businesses the buyer has not authorized.
+
+### Token Lifecycle
+
+Identity chaining produces two independent token lifecycles. Each layer is
+managed through its own OAuth endpoints, and revocation at one layer does not
+automatically propagate to the other.
+
+**Business-issued tokens.** Access tokens issued by the business follow the
+business's token lifecycle. Businesses **SHOULD NOT** issue refresh tokens
```
gsmith85, line 616: Architecturally, discouraging refresh tokens for the JWT bearer grant flow makes a lot of sense as it keeps the IdP as the ongoing trust anchor. But that makes the access token lifetime the effective trust re-verification interval, should we be explicit that it should be minutes not hours?

docs/specification/identity-linking.md
```diff
@@ -16,349 +16,685 @@
...
+When account linking fails at the IdP (standard OAuth flow), the platform
+receives standard OAuth errors. If the buyer declines consent, the IdP
+returns `access_denied`. The platform **SHOULD** offer alternative providers
+or fall back to direct business-hosted OAuth.
+
+### Buyer Awareness
+
+When multiple identity providers are available, the platform **SHOULD** offer
+the buyer a choice of providers. Before initiating identity chaining, the
+platform **SHOULD** inform the buyer which business will receive their
```
gsmith85, line 603: I'm left wondering if strengthening this obligation would be a buyer-centric privacy improvement.  The IdP's consent gate covers who and where: it knows the buyer and the target business, and actively gates that relationship. But the fine-grained scope negotiation happens between the platform and the business, outside the IdP's visibility. That makes the platform the only participant that sees both sides: which business receives the identity and what scopes that identity unlocks. The buyer should have visibility into both, not just where their identity is going, but what it will be used for.

source/schemas/common/identity_linking.json
```diff
@@ -1,86 +1,84 @@
...
+          "items": {
+            "type": "string",
+            "pattern": "^[a-z][a-z0-9_]*$"
+          },
+          "uniqueItems": true,
+          "minItems": 1
         }
       },
-      "additionalProperties": true
+      "additionalProperties": false
```
gsmith85, line 30: Is there a reason we're closing off extensibility here and leaving it open on provider?

source/schemas/common/identity_linking.json
```diff
@@ -1,86 +1,84 @@
...
+    "dev.ucp.common.identity_linking": {
+      "platform_schema": {
+        "title": "Identity Linking (Platform)",
+        "description": "Platform-level identity linking capability declaration.",
+        "allOf": [
+          { "$ref": "../capability.json#/$defs/platform_schema" }
+        ]
       },
-      "additionalProperties": true
+      "business_schema": {
```
gsmith85, line 54: Nit: Should `config` be required here as it was previously? The old schema required it, and declaring identity linking with no config is effectively a no-op: no providers, no capability scope requirements. It feels like a state that's more likely to be a misconfiguration than an intentional choice. That said, there's an argument for allowing it as an incremental rollout path, where a business signals infrastructure support before capabilities opt in.

docs/index.md
```diff
@@ -287,7 +287,7 @@ image: assets/banner.png
     "token_endpoint": "https://example.com/oauth2/token",
     "revocation_endpoint": "https://example.com/oauth2/revoke",
     "scopes_supported": [
-      "dev.ucp.shopping.scopes.checkout_session"
+      "ucp:scopes:checkout_session",
```
gsmith85, line 290: Nit: This scope format doesn't match the new scope naming convention defined in the Scope Naming section.

docs/specification/identity-linking.md
```diff
@@ -20,67 +20,110 @@
...
+    ([RFC 9207](https://datatracker.ietf.org/doc/html/rfc9207){ target="_blank" }).
+* When the business lists external identity providers in `config.providers`,
+    the business **MUST** support the JWT bearer assertion grant type
+    ([RFC 7523](https://datatracker.ietf.org/doc/html/rfc7523){ target="_blank" })
+    at its token endpoint to accept JWT authorization grants from trusted IdPs.
+    The business **MUST** include
+    `urn:ietf:params:oauth:grant-type:jwt-bearer` in
+    `grant_types_supported` in its
+    [RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414){ target="_blank" } metadata.
+* **MUST** provide an account creation flow if the buyer does not already have
```
richmolj, line 126: Thoughts on moving this to `MAY`? I'm thinking of the scenario: "_as a Business, I want to automatically sign in existing users, but don't want to interrupt the flow and force users to sign up_."

proshoumma:
Really like the direction here, the delegated IdP model and identity chaining solve a real pain point for multi-merchant agentic commerce. We've been working through this exact problem on our end and this is a big step forward.

One thing we keep coming back to is **per-business buyer consent in the Accelerated IdP Flow**.

With the traditional Account Linking flow, the buyer sees the business's consent screen and explicitly approves. The Accelerated IdP Flow removes that touchpoint - identity chains to new businesses server-to-server, which is great for UX but means the buyer never explicitly consents to (or may not even be aware of) each individual business receiving their identity.

The spec has the right instincts here - the MUST NOT on IdPs issuing grants for unauthorized businesses, and the SHOULD on platforms informing the buyer, but neither defines a concrete mechanism. And the underlying `draft-ietf-oauth-identity-chaining-08` doesn't help either. It's entirely silent on consent, which makes sense for its enterprise federation roots, but consumer commerce is a different world. Under GDPR especially, per-purpose consent and data subject awareness aren't optional.

A few things we're thinking about:
- Should the spec provide guidance on what "buyer has not authorized" actually means in practice? Right now it's a strong requirement with no defined path to satisfy it.
- Would it make sense to strengthen buyer awareness from SHOULD to MUST for consumer-facing contexts?
- Happy to help think through what a workable consent model could look like here.

igrigorik:
Closing, superseded by #423.

PR #339, opened 2026-04-07
docs: extract shared EP core, group transports in nav

igrigorik:
Extracts shared Embedded Protocol behavior from `embedded-checkout.md` and `embedded-cart.md` into a new `embedded-protocol.md`. Both files duplicate many parts (e.g. auth, security, ...) of the embedded protocol.

This PR depends on / stacked on #325 - we need to land that first. There are **no wire-level changes: method names, OpenRPC schemas, and URL parameters are unchanged.** 

---

**Shared infrastructure moved to `embedded-protocol.md`:**
- Transport & messaging (JSON-RPC 2.0 format, two-layer error model)
- Communication channels (postMessage, MessageChannel, native host globals)
- Authentication pattern (reauth flow, error escalation)
- Session error pattern (teardown contract, continue_url handoff)
- Lifecycle and state change contracts
- Error codes (shared W3C DOMException set)
- Security (CSP, iframe sandbox, credentialless iframes, origin validation)

**What remains in each capability spec:**
- Discovery and URL parameters (`ec_*` / `ep_*`)
- Delegation contracts and negotiation (checkout-specific)
- All method definitions with Direction / Type / Payload metadata
- All JSON wire-format examples
- Capability-specific error codes (`not_allowed_error`, `window_open_rejected_error`)
- Extensions (payment, fulfillment, window)
- Schema definitions

---
## Nav restructure

We never define "binding" and "EP Binding" is doubling confusing as a nav item under Cart & Checkout. Instead, we can cleanly group these under "Transports" section in each capability and replace EP with "Embedded". 

image

- Dropped "Binding" and "EP" from transport labels; group transports by name under a "Transports" grouping
- Shared Embedded Protocol page is linked from each capability's Embedded transport page

```
Before:                        After:
Checkout Capability            Checkout Capability
  Overview                       Overview
  HTTP/REST Binding              Transports
  MCP Binding                      REST
  A2A Binding                      MCP
  EP Binding                       A2A
  AP2 Mandates Extension           Embedded
  ...                            AP2 Mandates Extension
                                 ...
```

---

## Checklist
- [x] Documentation update


docs/specification/embedded-protocol.md
```diff
@@ -0,0 +1,514 @@
...
+```json
+{
+    "jsonrpc": "2.0",
+    "id": "auth_1",
+    "result": {
+        "ucp": { "version": "{{ ucp_version }}", "status": "error" },
+        "messages": [
+            {
+                "type": "error",
+                "code": "timeout_error",
```
westeezy, line 277: I don't believe we have timeout in the example table so minor nit but we can either add it there or show an example using one of the existing errors.

docs/specification/embedded-protocol.md
```diff
@@ -0,0 +1,514 @@
...
+- `messages` (array, **REQUIRED**): One or more messages describing the failure.
+- `continue_url` (string, **OPTIONAL**): URL for buyer handoff or session
+    recovery.
+
+**Example:**
+
+```json
+{
+    "jsonrpc": "2.0",
+    "method": "ep.cart.error",
```
westeezy: One more small cosmetic bit: we seem to mix examples between `ep.cart.error` and `ec.error`. Is it worth picking one for consistency in  the example payloads?

westeezy (approved):
LGTM outside a few nitpicks

ihoosain (approved):

docs/specification/embedded-protocol.md
igrigorik: Good catch. This was lifted from existing examples, which never defined it in the tables. Fixed.

docs/specification/embedded-protocol.md
```diff
@@ -0,0 +1,536 @@
...
+2. Host responds with:
+    - `ucp` envelope confirming the negotiated protocol version
+    - Optional `upgrade` with a `MessagePort` for channel upgrade
+    - Optional `credential` with requested authorization data
+    - Optional capability-specific state (e.g., checkout delegation state)
+3. If the host includes an `upgrade`, the embedded context **MUST** discard all
+   other fields in the response, switch to the upgraded channel, and send a new
+   `ready` message over that channel
+4. If the host responds with an `error_response` result, the session cannot
+   proceed — the host **MUST** tear down the embedded context and **MAY**
```
jingyli, line 217: Similar comment to https://github.com/Universal-Commerce-Protocol/ucp/pull/325#discussion_r3048343766, I'm wondering if there is a middle step missing here where the embedded context should echo back an `error_response` to the host (else I'm unsure when host will ever be able to get the `continue_url` for buyer handoff)?

docs/specification/embedded-protocol.md
```diff
@@ -0,0 +1,536 @@
...
+the embedded context is unable to process a host-provided credential (e.g.,
+the credential is corrupted). The session error **SHOULD** include a
+`continue_url` to allow buyer handoff.
+
+**Example — auth failure escalated to session error:**
+
+```json
+{
+    "jsonrpc": "2.0",
+    "method": "ec.error",
```
jingyli, line 301: nit: I wonder if we should replace this with `"method": "..."` so the example json remains EP namespace agnostic (alternatively, we can consider using `"method": "ep.{capability}.error"`)?

Similar comment to the other examples below for `xx.error`.

docs/specification/embedded-protocol.md
```diff
@@ -0,0 +1,536 @@
...
+
+**Example — messages changed (cart):**
+
+```json
+{
+    "jsonrpc": "2.0",
+    "method": "ep.cart.messages.change",
+    "params": {
+        "cart": {
+            "id": "cart_123",
```
jingyli, line 452: Should we add a comment alluding to the presence of other cart fields? At the very least, it should include `"line_items": [...]"` given `messages[0]` is referencing back to it..?

Similar comment to the other examples above where we do want the full checkout object to be returned back (i.e. `ec.complete` probably should hold more values than just `id` and `order`..?).

jingyli (approved):
Thanks @igrigorik for helping with the EP doc consolidation! Left some minor style comments and one question on expectation of host behaviour for handshake failures.

docs/specification/embedded-protocol.md
westeezy: I had similar thoughts initially when reading this but I came to a conclusion that it was helpful to show "real" examples so to speak. If we were to tweak this I would prefer `ep.{capability}.error`, but I do rather like the docs having these examples and folks looking to the schema for additional details and full payloads.

igrigorik: Agree, `ep.{capability}...` is right long-term shape, but we need to do the consolidation first (e.g. ec and ep prefixes collapsing to shared ep.error) before that can apply cleanly. 

This PR is just shuffling existing content to minimize duplication. We need to and *will* do the consolidation in followup PRs. My rec is to leave as is.

docs/specification/embedded-protocol.md
```diff
@@ -0,0 +1,540 @@
...
+The Embedded Protocol defines a shared set of error codes. Capabilities
+**MAY** define additional error codes for capability-specific scenarios.
+
+| Code                  | Severity        | Description                                                          |
+| :-------------------- | :-------------- | :------------------------------------------------------------------- |
+| `abort_error`         | `recoverable`   | The user cancelled the interaction (e.g., closed the sheet).         |
+| `security_error`      | `unrecoverable` | The host origin validation failed.                                   |
+| `invalid_state_error` | `unrecoverable` | Handshake was attempted out of order.                                |
+| `not_supported_error` | `unrecoverable` | The requested operation or authorization type is not supported.      |
+| `timeout_error`       | `recoverable`   | An internal service timed out. The operation may be retried.         |
```
mmohades, line 480: nit: do we want to have the error suffix for these?

mmohades (approved):

docs/specification/embedded-protocol.md
igrigorik: You mean `_error`? We followed / borrowed DOMException naming, which we snake_cased (SecurityError -> security_error). Let's keep as this, this PR is shuffling existing names, not changing any behaviors.

mmohades: Sounds good, thanks Ilya!

PR #342, opened 2026-04-08
fix!: Pull out dangling webhook fields into headers so transport reference only base schemas

jingyli:
# Description

There is a [requirement](https://ucp.dev/draft/specification/overview/#requirements) in UCP transport definition:
> Transport definitions (OpenAPI/OpenRPC) MUST reference base schemas only. They MUST NOT enumerate fields or define payload shapes inline.

This requirement is complied for all operations with the exception of `Order Webhook` today (https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/services/shopping/rest.openapi.json#L749), where the payload actually attempts to build additional fields (`event_id`, `created_time`) on top of the core Order capability.

To make sure we are fully compliant with the requirement, proposing to pull out the 2 webhook-specific fields into required headers.

## Type of change

      or fields**)
- [ ] Documentation update

---

files or fields:

```text

Bug fix on schema to make sure we are complying with UCP transport requirement.
```

---

## Checklist

---

### DIFF
--- a/docs/specification/order.md
+++ b/docs/specification/order.md
@@ -402,7 +402,7 @@ during partner onboarding. The URL format is platform-specific.
 | Header               | Description                                 |
 | :------------------- | :------------------------------------------ |
 | `Webhook-Timestamp`  | Event creation timestamp (RFC 3339)         |
-| `Request-Id`         | Unique event identifier                     |
+| `Webhook-Event-Id`   | Unique event identifier                     |
 
 {{ method_fields('order_event_webhook', 'rest.openapi.json', 'order') }}
 
--- a/docs/specification/order.md
+++ b/docs/specification/order.md
@@ -397,12 +397,16 @@ payload is the same **current-state snapshot** described in
 Businesses POST order events to a webhook URL provided by the platform
 during partner onboarding. The URL format is platform-specific.
 
+Headers follow **[Standard Webhooks](https://www.standardwebhooks.com/){ target="_blank" }**;
+except for request signing, which follows [RFC 9421](https://www.rfc-editor.org/rfc/rfc9421).
+See [Message Signatures](signatures.md) for more details.
+
 **Required Headers:**
 
 | Header               | Description                                 |
 | :------------------- | :------------------------------------------ |
-| `Webhook-Timestamp`  | Event creation timestamp (RFC 3339)         |
-| `Webhook-Event-Id`   | Unique event identifier                     |
+| `Webhook-Timestamp`  | Event occurrence timestamp (unix)           |
+| `Webhook-Id`         | Unique event identifier                     |
 
 {{ method_fields('order_event_webhook', 'rest.openapi.json', 'order') }}
 
--- a/docs/specification/order.md
+++ b/docs/specification/order.md
@@ -397,6 +397,13 @@ payload is the same **current-state snapshot** described in
 Businesses POST order events to a webhook URL provided by the platform
 during partner onboarding. The URL format is platform-specific.
 
+**Required Headers:**
+
+| Header               | Description                                 |
+| :------------------- | :------------------------------------------ |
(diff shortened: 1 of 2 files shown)


docs/specification/order.md
```diff
@@ -397,6 +397,13 @@ payload is the same **current-state snapshot** described in
 Businesses POST order events to a webhook URL provided by the platform
 during partner onboarding. The URL format is platform-specific.
 
+**Required Headers:**
+
+| Header               | Description                                 |
+| :------------------- | :------------------------------------------ |
+| `Webhook-Timestamp`  | Event creation timestamp (RFC 3339)         |
+| `Request-Id`         | Unique event identifier                     |
```
richmolj: Should this have the `Webhook-` prefix as well? Otherwise the name reads too generic.

jingyli: `Request-Id` is an existing header that is used within OpenAPI. I was thinking to just reuse this header instead of introducing a new one specific to webhooks.

But giving it some more thought just now, maybe reusing is not necessarily ideal - given we want this ID to be stable at an event level (regardless of how many potential redelivery the business may attempt with the platform). We could also reuse `idempotency-key` here - it is closer to what we are looking for, but I feel it's not explicitly in what it is representing.

tl;dr - introduced a custom header `Webhook-Event-Id` for our use case here.

igrigorik (changes requested):
Overall, 👍  on this change.

This is a breaking change and since we're shifting into headers, I think there is an opportunity to align with the [Standard Webhooks spec](https://github.com/standard-webhooks/standard-webhooks/blob/main/spec/standard-webhooks.md#webhook-headers-sending-metadata-to-consumers) rather than inventing custom names:

- `Webhook-Event-Id` -> `webhook-id`
  - I would drop the `format: "uuid"` constraint on the event ID -- the original `event_id` didn't have it, and Standard Webhooks doesn't require it. 
- `Webhook-Timestamp` -> is fine but with `type: "integer"`
  - Note: `webhook-timestamp` would be moving from RFC 3339 to unix seconds, which is a translation / inconsistency vs other places in UCP, but I'm OK with that at this layer because it's conforming to existing standard / expected behavior of webhooks.

This gives us interop with anyone who's implemented support for standard webhooks.

Standard Webhooks defines its own HMAC-SHA256 signature scheme using `webhook-signature`. UCP already has a stronger signing story via RFC 9421 (`Signature` + `Signature-Input` + `Content-Digest`), which are already wired up as parameters on this operation. I think this one we can leave as is and lean on existing infra.  A one-liner in the docs noting "metadata headers follow [Standard Webhooks](https://github.com/standard-webhooks/standard-webhooks); request signing follows RFC 9421" (link to our signing doc) would help implementers understand the split.

jingyli:
Thanks @igrigorik for the feedback, aligning with [Standard Webhooks](https://github.com/standard-webhooks/standard-webhooks/blob/main/spec/standard-webhooks.md) instead of inventing custom headers make sense!

The thing I did keep as-is is having the `"format": "uuid"` annotation on `Webhook-Id` header. Reasons being:
1) The description on the original `event_id` field mentioned it should be a "unique event identifier" - so introducing the constraint shouldn't be odd for implementers.
2) In my mind, `Webhook-Id` behaves very similarly to `Idempotency-Key` (in fact the Standard Webhooks spec literally talks about `Use the webhook-id header as an idempotency key` in their [Verifying signatures](https://github.com/standard-webhooks/standard-webhooks/blob/main/spec/standard-webhooks.md#verifying-signatures) section). The latter has the uuid annotation already in UCP (https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/services/shopping/rest.openapi.json#L870) and I feel it's better to keep them consistent.

igrigorik (approved):
👍

richmolj (approved):

PR #352, opened 2026-04-10
fix: render Total and Totals fields on schema reference page

ynachiket:
The Total and Totals sections on the specification reference page rendered empty tables because `_render_table_from_schema` entered the `allOf` branch for schemas that have both `properties` and `allOf` validation constraints (if/then, contains). The allOf items contained no renderable content, so no table rows were emitted.

Two fixes in `_render_table_from_schema`:

1. Only enter the `allOf` rendering branch when `properties` is empty (`and not properties`). This lets Total's three fields (type, display_text, amount) render through the normal property iteration path.

2. Treat array-typed schemas (like Totals) the same as scalar types in the description fallback, since their top-level `allOf` only carries `contains` constraints, not composition. Totals now renders its description text, consistent with Amount and Signed Amount.

Before:

image

After:
image

image

### DIFF
--- a/main.py
+++ b/main.py
@@ -600,9 +600,9 @@ def define_env(env):
 
     if (
       not properties
-      and "allOf" not in schema_data
       and "oneOf" not in schema_data
       and "$ref" not in schema_data
+      and ("allOf" not in schema_data or schema_data.get("type") == "array")
     ):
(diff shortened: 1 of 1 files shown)


aksbro-gpu (approved):

wry-ry (approved):
Thanks for your contribution!

PR #354, opened 2026-04-12
feat!: identity linking OAuth 2.0 foundation with capability-driven scopes

amithanda:
## Overview

This PR establishes the OAuth 2.0 foundation for the Identity Linking capability
(`dev.ucp.common.identity_linking`). It is an intentionally scoped first step —
business-hosted OAuth 2.0 only — designed so that delegated identity providers
and non-OAuth auth mechanisms can be added in future PRs as non-breaking
extensions.

**Context:** #265 introduced a mechanism registry and capability-driven scopes
but shipped a critical bug where the intersection algorithm's scope-dependency
pruning rule gated checkout behind identity linking, breaking guest checkout.
#329 reverted it. #330 proposed a full redesign including delegated IdP and
identity chaining. This PR carries forward the improvements from both without
the delegated IdP complexity, and lays explicit extensibility groundwork for
that work to land cleanly.

---

### New: `source/schemas/common/identity_linking.json`

Schema for the identity linking capability. Two context-specific views, nested
under `"dev.ucp.common.identity_linking"` in `$defs` per the project's
capability schema convention:

- **`platform_schema`**: passthrough — platforms advertise support; no
  auth-specific config needed.
- **`business_schema`**: requires `config.capabilities` — a map declaring
  which capabilities offer buyer-scoped features and whether buyer identity
  is required for each.

The `config` object uses `additionalProperties: true` with a `$comment`
naming one reserved extension point:
- `providers` — map of trusted identity providers keyed by reverse-domain,
  with a `type` discriminator defaulting to `oauth2`. This single extension
  point covers delegated IdP, identity chaining, and future non-OAuth
  mechanisms such as wallet attestation (future PR).

### New: `identity_required` error code (`error_code.json`)

Standard protocol signal for the case where a capability is configured with
`auth_required: true` and a request arrives without a buyer identity token.
Businesses MAY include a `continue_url` in the error body for buyer
onboarding flows.

### Rewrite: `docs/specification/identity-linking.md`

Major rewrite of the spec. Key sections:

**Access levels.** Capabilities operate at three access levels — public,
agent-authenticated, buyer-authenticated. Identity linking upgrades capabilities
to buyer-authenticated access. Capabilities are **never** excluded from
capability negotiation based on identity linking. This directly addresses the
root cause of the #265 bug.

**Security hardening** (carried forward from #265 and #330):
- PKCE `S256`: MUST for all authorization code flows; `plain` MUST NOT be used
- `iss` validation: MUST (RFC 9207) — prevents Mix-Up Attacks; validation is
  unconditional, not gated on presence of the `iss` parameter
- Exact `redirect_uri` matching: MUST — no partial/prefix matching
- `issuer` byte-for-byte match: MUST — no normalization (trailing slash stripping
  is a known `iss` validation bypass)
- `scopes_supported`: MUST in RFC 8414 metadata — enables early scope mismatch
  detection before consent screen

**Strict discovery hierarchy** (carried forward from #265 and #330):
1. RFC 8414 (`/.well-known/oauth-authorization-server`) — `2xx`: use it; `404`:
   proceed to step 2; any other non-2xx response, network error, or timeout:
   **MUST** abort, **MUST NOT** fall through to step 2
2. OIDC fallback (`/.well-known/openid-configuration`) — `2xx`: use it; any
   non-2xx response, network error, or timeout: **MUST** abort

**Capability-driven scope model** (redesigned from #265, aligned with #330):
Scope declarations live in `config.capabilities` on the identity linking config,
not as `identity_scopes` annotations on individual capability schemas. This is
architecturally correct because whether a capability requires buyer auth is a
business decision — a B2B wholesaler gates catalog access, a B2C retailer
doesn't. Per-schema annotations can't express this variance.

Scope naming uses capability names directly (`dev.ucp.shopping.checkout`,
`dev.ucp.shopping.order:read`) — reusing UCP's existing reverse-DNS governance
and eliminating a separate scope namespace. Replaces the old
`ucp:scopes:checkout_session` format.

**Future Extensibility section** — explicit normative spec for how future
versions extend this capability without breaking v1 implementations (see below).

### Updated: `docs/specification/overview.md`

Identity linking added back to both business and platform profile examples
using the new config shape, including the `auth_required` field naming.

### Fixed: `docs/index.md`

Scope naming in the RFC 8414 metadata example updated from
`ucp:scopes:checkout_session` to `dev.ucp.shopping.checkout`.

---

## Forward Compatibility Design

The schema and spec are explicitly designed for non-breaking extension via a
single `config.providers` extension point.

### `config.providers` — Delegated Identity Providers and Mechanism Extensibility

A future PR will add a `config.providers` map (keyed by reverse-domain
identifier) allowing businesses to declare trusted external identity providers
alongside their own hosted OAuth server. Each provider entry carries a `type`
discriminator defaulting to `oauth2`, making the map extensible to non-OAuth
mechanisms (wallet attestation, verifiable credentials) without introducing a
separate `mechanisms` array.

This covers two use cases:

- **Delegated IdP**: platforms with an existing IdP session can chain identity
  to new businesses without a browser redirect (per
  draft-ietf-oauth-identity-chaining-08), solving the N-merchant = N-OAuth-dances
  problem for agentic commerce.
- **Non-OAuth mechanisms**: future entries with a non-`oauth2` `type` value
  enable wallet attestation and similar schemes. Platforms select the first
  entry whose `type` they support — analogous to TLS cipher suite negotiation,
  preserving business-preference ordering.

This addition is non-breaking because:
- `config` uses `additionalProperties: true` — the schema will not reject the
  new field
- The spec defines a normative forward-compat rule: platforms that do not
  recognize `config.providers` MUST ignore it and fall back to RFC 8414
  discovery on the business domain

---

## Type of Change

- [x] New feature (non-breaking addition of new capability schema)
- [x] Breaking change (new scope naming convention replaces `ucp:scopes:*` format)
- [x] Documentation update

### Breaking Changes Justification

The scope naming convention change (`ucp:scopes:checkout_session` →
`dev.ucp.shopping.checkout`) is breaking for any implementation that hardcoded
the old format. The old format was defined in the pre-#265 spec and was never
part of a stable release — the capability was at Working Draft status throughout.
The new format is consistent with UCP's reverse-DNS naming governance and
eliminates the need for a parallel scope namespace.

---

## Checklist


douglasborthwick-crypto:
Thanks for laying the extensibility groundwork explicitly. The reserved `config.mechanisms` slot and the RFC 8414 fallback rule together make it straightforward for non-OAuth mechanisms to land as non-breaking additions — exactly the shape we were looking for when we opened #264.

Once this lands, we can follow up with a concrete `config.mechanisms` entry for wallet attestation — JWKS-discovered public key, ES256-signed boolean verdict per evaluated condition, deterministic offline verification. The `type` discriminator pattern (chosen inside the array, TLS-cipher-suite-style) works cleanly for wallet auth since the business declares the supported chains and condition types per entry.

Happy to wait for this PR to merge before opening the follow-up so we're not creating extension-point churn.

igrigorik (changes requested):
@amithanda nice work on this, lots of good improvements!

We need to land support for delegated IdPs. This is a key and painfully missing feature, which we proposed the shape for in #330. Comparing the two branches, we need to reconcile and merge, and I think it's easier to build on what you've scoped here. 

I took a run at bringing over additional logic from #330 in https://github.com/Universal-Commerce-Protocol/ucp/commit/603f4c832081cb5aebb8377ac357509b805b38d0. Because your PR is against your own fork I can't open a stacked PR. That said, if the commit looks good, I can push it into your branch directly and we can continue iterating against your current branch and close out #330. PTAL, any objections?

--- 

With above in place, I think we can collapse providers vs mechanisms into single primitive. They are both attempting to answer the same question: **"who does the business trust to vouch for buyers, and what proof protocol do they speak?"**

An OAuth IdP and a wallet attestor both have:
- A reverse-domain trust anchor (com.google, com.example.attestor)
- A discovery mechanism (auth_url → RFC 8414, provider_jwks → JWKS endpoint)
- A proof format (OAuth tokens, ES256 signed payloads)
- An explicit trust relationship (business lists them)

Keeping them as separate keys creates ambiguity: are mechanisms *alternatives* to OAuth (platform picks one) or *complements* (identity via OAuth + wallet proof for eligibility)? The TLS cipher-suite negotiation analogy on `mechanisms` implies the former, but your actual use case in #264/#280 looks more like the latter. 

## Proposed reconciliation

Unify under `providers` with a `type` discriminator. Default type is `oauth2` when absent — zero wire-format change for existing entries:

```json
"config": {
  "providers": {
    "com.google": {
    },
    "com.example.merchant": {
      "auth_url": "https://merchant.example.com/"
    },
    "com.example.attestor": {
      "type": "wallet_attestation",
      "provider_jwks": "https://attestor.example.com/.well-known/jwks.json",
      ...
    }
  },
  "capabilities": { ... }
}
```

**What this gets us:**
- **Single trust surface.** Business declares all trusted identity/attestation sources in one map.
- **Clean forward-compat.** Platforms skip provider entries whose `type` they don't recognize. No separate "ignore unknown config fields" rule for a sibling key.

**What we drop:** The `config.mechanisms` reserved extension point. It's replaced by new `type` values on provider entries, which is a strictly simpler extension model.

The relationship between providers then becomes clear from context: a platform that finds both `type: "oauth2"` and `type: "wallet_attestation"` entries knows these are different proof protocols. Whether they're alternatives or complements depends on how `config.capabilities` references them — that's the follow-up design question for when wallet attestation lands, not something we need to over-specify now.

igrigorik:
One gotcha that's bothering me with scopes as flat array of strings...

```json
"dev.ucp.shopping.order": {
  "auth_required": true,
  "scopes": ["read", "manage"]
}
```

With above contract every scope inherits uniform requirements from a single `auth_required` flag at the capability level. Fine for simple cases, but I think we need more flexibility. Consider the following example...

#### B2C retailer
- `search` does not require buyer auth — public catalog access
- `vip` requires buyer authorization — unlocks premium/vip products and inventory

#### B2B wholesaler
- `search` requires buyer auth
- `wholesale_pricing` requires elevated auth context (e.g. hardware-backed auth)

Same capability (`dev.ucp.shopping.catalog`), overlapping but different scope vocabulary, and different access policies for different scopes. Array of strings can't capture this. Capability-level `auth_required` over-triggers for B2C (flags `search` as requiring auth) or under-triggers for B2B (can't signal scope step-up).

## Rec: scope level policies + explicit scopes (no bare capability scope)

Each listed capability declares its scopes explicitly. Each scope is an object. Wire format is always `capability:scope` — there is no "bare capability name as scope". The resulting shape is:

```json
"dev.ucp.common.identity_linking": [{
  "config": {
    "providers": { ... },
    "capabilities": {
      "dev.ucp.shopping.catalog": {
        "scopes": {
          "search": { "auth_required": false },
          "vip":    { "auth_required": true }
        }
      },
      "dev.ucp.shopping.order": {
        "scopes": {
          "read":   { "auth_required": true },
          "manage": { "auth_required": true, "min_acr": "urn:tl:TL3" }
        },
        "some_future_foo": "bar"
      }
    }
  }
}]
```

The contract is:
 - Each listed capability MUST declare a `scopes` map with at least one entry.
 - Scopes name always follow `{capability}:{scope}` convention.

What this unlocks
- **Per-scope configs** — e.g, `auth_required` can be defined with scope granularity
- **Extensible scope configs** — e.g., `min_acr` for step-up auth, `max_token_age` for freshness; ...
- **Convention alignment** — providers, capabilities, payment_handlers are all keyed maps

## Alt: flat `scopes` map keyed by wire-format

```json
"config": {
  "providers": { ... },
  "scopes": {
    "dev.ucp.shopping.catalog:search": { "auth_required": false },
    "dev.ucp.shopping.catalog:vip":    { "auth_required": true },
    "dev.ucp.shopping.order:read":     { "auth_required": true },
    "dev.ucp.shopping.order:manage":   { "auth_required": true, "min_acr": "..." }
  }
}
```

The benefit of the above pattern is that key is the exact scope that you can copy-pase into Oauth `scope=`. The tradeoff is that we lose grouping and future extensibility / ability to provide capability-wide config values.

amithanda:
How about an alternate option where we can extend at the capability entry level, 
keep `scopes` as array as is and add scope_config in a future PR?

The per-scope flexibility concern is valid — capability-level `auth_required`
can't express cases where some scopes require buyer auth and others don't, or
where specific scopes need step-up auth (`min_acr`). However, rather than
changing `scopes` from an array to a map, can we get the same expressiveness by
opening the capability entry itself for future extension.

### The change

`capability_identity_config` currently has `"additionalProperties": false`.
Changing that to `true` and naming a reserved extension point in `$comment`
is all that's needed. The current format stays valid and unchanged:

```json
"dev.ucp.shopping.order": {
  "auth_required": true,
  "scopes": ["read", "manage"]
}
```

A future PR adds `scope_config` — a map keyed by scope name for per-scope
overrides — without any schema version bump or wire-format change:

```json
"dev.ucp.shopping.order": {
  "auth_required": true,
  "scopes": ["read", "manage"],
  "scope_config": {
    "manage": { "min_acr": "urn:tl:TL3" }
  }
}
```

The B2C case (public catalog search, buyer-gated VIP access):

```json
"dev.ucp.shopping.catalog": {
  "auth_required": false,
  "scopes": ["search", "vip"],
  "scope_config": {
    "vip": { "auth_required": true }
  }
}
```

`auth_required` at the capability level acts as the default for all scopes.
`scope_config` provides per-scope overrides only where needed — step-up auth,
scope-level auth exceptions, or future properties we haven't defined yet.

#### Why this over changing `scopes` to a map

The current `scopes` array covers the common case cleanly — most capabilities
have uniform auth requirements across all their scopes, and the simple array
format is easy to implement and read. Rather than redesigning the shape now to
handle the advanced cases, this approach lets us ship the current format as-is
and add `scope_config` as a follow-on PR once we have real implementation
experience with where per-scope policies are actually needed.

The extension point costs nothing to reserve now (one schema flag change) and
keeps the door open for `min_acr`, scope-level `auth_required` overrides, or
any other per-scope property we identify later — without touching anything that
has already shipped.

Mental model for identity linking discovery is - does this capability require identity linking, but adding
`auth_required` at the scope level adds complications as platform will have to 
understand the semantic meaning of the scope and it becomes a scope level discovery
problem. Rather, we can do a step up if someone wants to get additional scope. If there is 
any scope that doesn't require identity linking, the capability level `auth_required` if `false`
but we can do a step up if the agent requires operations with scopes that need buyer to be
authenticated.

#### Schema change (minimal)

In `capability_identity_config`:

```diff
- "additionalProperties": false
+ "additionalProperties": true,
+ "$comment": "Reserved extension point: 'scope_config' (map keyed by scope
+   name for per-scope overrides of auth_required and future properties such
+   as min_acr for step-up auth). Platforms MUST ignore unrecognized fields
+   and apply the capability-level auth_required as the default."
```

This is the same two-line pattern used on `config` itself for `providers`.
No spec text changes needed in v1 beyond naming the extension point in
the schema comment and the Future Extensibility section.

mnaga:
@igrigorik , while I agree with you about the use-cases such as B2C or B2B, I feel some of those need to be modeled at a higher level of granularity (hierarchical or flat structure). with B2C or B2B, it not just access to capabilities but there could be variations to data returned as part of existing operations. For example, the price returned could be different, and with the below suggestion, if the agent chooses to **shop** as B2B, then it would enforce authorization.

```
"dev.ucp.common.identity_linking": [{
  "config": {
    "providers": { ... },
    "capabilities": {
      "dev.ucp.shopping" : {
        "scopes": {
          "b2c": { "auth_required": false, default: true },
          "b2b":    { "auth_required": true }
        }
      }
      "dev.ucp.shopping.catalog": {
        "scopes": {
        }
      },
      "dev.ucp.shopping.order": {
        "scopes": {
          "read":   { "auth_required": true },
          "manage": { "auth_required": true, "min_acr": "urn:tl:TL3" }
        },
      }
    }
  }
}]
```
BTW, I am split on whether "dev.ucp.shopping" treated as a capability or something else, but this should give an idea of how i see individual capability authorization and uber authorization are kinda orthogonal.

amithanda:
@igrigorik @mnaga, I’ve been thinking about the discussion around B2B/B2C use cases and scope granularity. I’d like to propose a path forward that I think simplifies the data model significantly by stepping back to first principles of authentication and authorization.

**1. Separation of Concerns: AuthN/AuthZ vs. Roles**
*   **Authentication** (who the user is) and **Authorization** (what scopes they have) are protocol concerns that UCP should define.
*   **Roles** (how a business groups scopes for a user, e.g., "Wholesaler" vs. "Retailer") are highly specific to internal business logic.
*   If we leave role mapping to the merchant's side, UCP doesn't need to model complex role structures or scope inheritance. The client simply requests the scopes it needs, and the merchant grants them based on internal rules. If a user lacks a scope required for an operation, the merchant can trigger a "scope bump" (re-auth). This keeps the protocol thin and flexible.

**2. Principle: Scopes Imply Authentication**
*   If an operation is public (e.g., basic catalog search), it doesn't require authentication, and thus no scope is checked or required.
*   Therefore, the mere listing of a scope in the configuration should imply that authentication is required to access that functionality. Putting `auth_required: true` on every scope is redundant.

**Proposal:**
Based on this, I lean heavily towards Ilya's alternative of a **flat `scopes` map keyed by wire-format**, but with the simplification that we remove the `auth_required` field entirely. The presence of the scope key is the signal that auth is required.

Using a map (instead of a simple array) still gives us the extensibility Ilya mentioned for advanced cases (like step-up auth), but keeps the common case extremely clean.

```json
"config": {
  "providers": { ... },
  "scopes": {
    "dev.ucp.shopping.catalog:vip": {}, 
    "dev.ucp.shopping.order:read": {},
    "dev.ucp.shopping.order:manage": { "min_acr": "urn:tl:TL3" }
  }
}
```

# Examples scenarios with a representative Order capability 

 Here are a few examples for `order` capability (`dev.ucp.shopping.order`) under the simplified model where **scopes imply authentication**.

The core principle is: **If an action requires authentication, its scope is listed in the map. If it does not require authentication (it is public), it is not listed.**

---

## Scenario 1: Mixed Operations (Some require auth, some don't)

Imagine the `order` capability allows public guest tracking via an Order ID, but viewing full order history requires authentication.
*   **Public**: `dev.ucp.shopping.order:track` (Not listed)
*   **Requires Auth**: `dev.ucp.shopping.order:read` (Listed)

```json
{
  "config": {
    "providers": {
      "com.example.idp": {
        "auth_url": "https://idp.example.com/"
      }
    },
    "scopes": {
      "dev.ucp.shopping.order:read": {}
    }
  }
}
```
---

## Scenario 2: Fully Gated Operations (All operations require auth)

In this case, the merchant decided that *all* interactions with orders (both tracking, reading, and managing) require the user to be authenticated.

```json
{
  "config": {
    "providers": {
      "com.example.idp": {
        "auth_url": "https://idp.example.com/"
      }
    },
    "scopes": {
      "dev.ucp.shopping.order:track": {},
      "dev.ucp.shopping.order:read": {},
      "dev.ucp.shopping.order:manage": {}
    }
  }
}
```

---

## Scenario 3: Fully Public Operations (Hypothetical)

If a merchant allowed complete public access to all order functions (highly unlikely in practice, but useful for illustration), the `scopes` map would be empty for the order capability.

```json
{
  "config": {
    "providers": {
      "com.example.idp": {
        "auth_url": "https://idp.example.com/"
      }
    },
    "scopes": {}
  }
}
```

---

## Scenario 4: Advanced Case (Step-up auth per scope)

Here, reading orders requires standard authentication, but managing/canceling orders requires a stronger authentication context (step-up auth).

```json
{
  "config": {
    "providers": {
      "com.example.idp": {
        "auth_url": "https://idp.example.com/"
      }
    },
    "scopes": {
      "dev.ucp.shopping.order:track": {},
      "dev.ucp.shopping.order:read": {},
      "dev.ucp.shopping.order:manage": {
        "min_acr": "urn:tl:TL3" 
      }
    }
  }
}

File: docs/specification/identity-linking.md [hunk 1]
```diff
@@ -17,145 +17,508 @@
...
 * **Capability Name:** `dev.ucp.common.identity_linking`
+* **Schema:** `https://ucp.dev/schemas/common/identity_linking.json`
 
 ## Overview
 
-The Identity Linking capability enables a **platform** (e.g., Google, an agentic
-service) to obtain authorization to perform actions on behalf of a user on a
-**business**'s site.
+The Identity Linking capability enables a **platform** to obtain authorization
+to perform actions on behalf of a **buyer** on a **business**'s (relying
```
cusell-google: nit: "buyer" may be overly specific. The user doesn't necessarily need to seek a purchase. Same on several mentions of "buyer" below.

docs/specification/identity-linking.md
```diff
@@ -17,145 +17,508 @@
...
+    ([RFC 6749 §4.1](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1){ target="_blank" })
+    as the account linking mechanism.
+* **MUST** use PKCE
+    ([RFC 7636](https://datatracker.ietf.org/doc/html/rfc7636){ target="_blank" })
+    with `code_challenge_method=S256` for all authorization code exchanges.
+* **MUST** validate the `iss` parameter in the authorization response
+    ([RFC 9207](https://datatracker.ietf.org/doc/html/rfc9207){ target="_blank" })
+    to prevent Mix-Up Attacks. The platform **MUST** verify that the `iss`
+    value matches the authorization server's issuer URI (as declared in its
+    RFC 8414 metadata). If the values do not match, the platform **MUST**
```
cusell-google: Should link to the corresponding RFCs in each mention? https://datatracker.ietf.org/doc/html/rfc8414? Same on line 104.

docs/specification/identity-linking.md
```diff
@@ -17,145 +17,508 @@
...
+* Revocation and security events:
+    * **MUST** call the business's token revocation endpoint
+        ([RFC 7009](https://datatracker.ietf.org/doc/html/rfc7009){ target="_blank" })
+        when a buyer initiates an unlink action on the platform side.
     * **SHOULD** support
-        [OpenID RISC Profile 1.0](https://openid.net/specs/openid-risc-1_0-final.html)
-        to handle asynchronous account updates, unlinking events, and
-        cross-account protection.
+        [OpenID RISC Profile 1.0](https://openid.net/specs/openid-risc-1_0-final.html){ target="_blank" }
+        to handle asynchronous account updates and cross-account protection
```
cusell-google, line 154: "unlinking events" were removed, is that intentional and meant to be implicitly covered under "asynchronous account updates"?

docs/specification/identity-linking.md
```diff
@@ -17,145 +17,508 @@
...
+```
+
+## Security Considerations
 
-### Structure
+* **PKCE.** PKCE (`S256`) is REQUIRED for all authorization code flows.
+  Plain PKCE (`plain`) **MUST NOT** be used. Businesses **MUST** reject
+  authorization code exchanges without a valid `code_verifier`.
+* **Mix-Up Attack prevention.** Platforms **MUST** validate the `iss`
+  parameter in the authorization response (RFC 9207). Businesses **MUST**
```
cusell-google: link to rfc https://datatracker.ietf.org/doc/html/rfc9207. Also on line 405

docs/specification/identity-linking.md
```diff
@@ -17,145 +17,508 @@
...
-    "refresh_token"
-  ],
-  "token_endpoint_auth_methods_supported": [
-    "client_secret_basic"
+    "dev.ucp.shopping.checkout",
+    "dev.ucp.shopping.order:read",
+    "dev.ucp.shopping.order:manage"
   ],
+  "response_types_supported": ["code"],
+  "grant_types_supported": ["authorization_code", "refresh_token"],
```
cusell-google, line 676: (not actionable): just want to mention that explicitly adding `grant_types_supported` as per RFC 8414 (that I initially left out) is great because it allows extending with `client_credentials` in the future!

amithanda:
## Sample Flow

### Step 1: Discovery
1.  The Agent fetches the Business Profile at `/.well-known/ucp`.
2.  The Agent reads the configuration to understand which scopes require authentication (e.g., `dev.ucp.shopping.order:read`).
3.  The Agent also fetches the OAuth metadata at `/.well-known/oauth-authorization-server` to know where the auth endpoints are. or use provider URL for delegated identity.

### Step 2: Capability Negotiation
1.  The Agent and Business negotiate which capabilities they both support.
2.  **Authentication requirements do not block negotiation.** The capability is enabled even if the user is not yet authenticated.

### Step 3: Execution (Gated Operation)
1.  The Agent attempts to call an operation that requires a scope (e.g., reading order history).
2.  If the Agent does not have a token, the Business rejects the request with a specific error code `identity_required` and provides a `continue_url`.

### Step 4: Authentication (Scope Bump)
1.  The Agent redirects the user to the `continue_url` to authenticate and authorize the requested scope.
2.  Upon success, the Agent receives an Access Token.

### Step 5: Retry Operation
1.  The Agent retries the operation, this time including the token in the `Authorization: Bearer <token>` header.
2.  The Business validates the token and returns the requested data.

cusell-google (changes requested):
A few thoughts on the authorization flow and scope management:

1. +1 to @amithanda 's suggestion above to drop `auth_required` and just include the required scopes. It eliminates a redundant state and simplifies the schema.

2. Even if the merchant returns a narrow subset of scopes for a specific operation, there should be a mechanism to (optionally) request a broader set of scopes when performing the initial account linking. The user can still be scope-bumped in the future via incremental auth, but capturing a wider set of scopes upfront will reduce UX friction downstream, without the need to ask them to link accounts every time they perform a new operation.
3. When linking again for a scope-bump, the returned scopes should include all the previous scopes that the user already has access to. Otherwise, the previous authorization will be overridden, resulting in an accidental "scope replacement" instead of a true scope bump. Should this be explicitly called out in the spec, requiring businesses to issue a superset of previously granted scopes for a given user to prevent access regressions?

mnaga:
Amit,
In general, I agree with just defining scope and not requiring auth _required since it's implicit. 
Like I said before, there are two broad use cases 
1. Access control which can be addressed by scope as you suggested.
2. Api behavior: the catalog or search api is the same and might require read access.  but the price or delivery options change based on the type of user. Example: returns default non member price when unauthenticated versus returns a member price if authenticated and  user has an active member subscription.
IMO, we would need a construct like role, in addition to scopes thats at the uber level. In the example you have, I would say vip is more of a role than scope if it's impacting the behavior of that api and not limiting access to that api. I'd say we need something like this

```
"config": {
  "providers": { ... },
  "roles": [
     "vip": {auth_required:true},
     "b2b":{auth_required:true}
  ],
  "scopes": {
    "dev.ucp.shopping.order:read": {},
    "dev.ucp.shopping.order:manage": { "min_acr": "urn:tl:TL3" }
  }
}
``` 
So that the agent is forced to authenticate if it wants to interact as b2b or premium member.

It could be called Memberships as well and be part of Memberships, but here is where I feel auth and Memberships intersect.

amithanda:
@mnaga - I feel role is a concept that is internal to business, it is not required to be exposed to the public or to platforms. We already have a well defined way for business to provide a continue_url in case they think the identity being linked doesn't have the right scopes based on the roles configured internally in their systems. Let me know what you think and why you think having a role exposed to platform and public is critical?

raginpirate:
👋 thanks for a nudge to review this in TC, sorry for the late loop!

I think the simplified proposal looks great, avoiding forced coupling with capabilities while also introducing the flexible map for per-scope auth requirements. I think this is an elegant edge of the problem to isolate. We can model the complexities of what a resource (capabilities, payment handlers, services, transports, specific identities, etc.) needs from the authorized identity in its own follow-up proposal. Hopefully one that allows shared-scopes between resources and describes whats required vs optional (perhaps that should actually live on the given UCP entities themselves?)

> IMO, we would need a construct like role

@mnaga I think this is what you are pointing to: there _**needs to exist**_ this intersection between what scopes to request and what your identity model represents, what your capability represents, etc. And if I'm interpreting the spirit of this proposal, we'll push this out and model that better external to this scope and authorization primitive.

docs/specification/identity-linking.md
```diff
@@ -17,145 +17,508 @@
...
-    through HTTP Basic Authentication
-    ([RFC 7617](https://datatracker.ietf.org/doc/html/rfc7617){target="_blank"})
-    when exchanging codes for tokens.
-    * **MAY** support Client Metadata
-    * **MAY** support Dynamic Client Registration mechanisms to supersede
-        static credential exchange.
-* The platform must include the token in the HTTP Authorization header using
-    the Bearer schema (`Authorization: Bearer <access_token>`)
+Identity linking upgrades capabilities to buyer-authenticated access.
+Capabilities are **never** excluded from negotiation based on the presence or
```
raginpirate: Out of curiosity, why are we saying never? Why can't a business make a decision to turn off their x capability because they saw an identity, or vice-versa?

docs/specification/identity-linking.md
amithanda: I was thinking of this being implicitly covered under "asynchronous account updates". Let me know if it there is a reason it will be good to call it more explicitly?

cusell-google: Leaving it implicit is fine, as RISC doesn't include an explicit unlinking event anyway. Marking as resolved.

drewolson-google:
I agree with many of the points above -- I'm in favor of the flatten structure and the simplified approach, only listing the scopes that are required and dropping `auth_required`. Nice work!

igrigorik:
Ditto, I think we converged on a good (simplified) shape -- ty all!

@amithanda @mnaga we're working through scopes vs. claims. I agree with Amit that claims are merchant-internal: agent initiates auth, business looks up associated roles and can encode them into the granted token, which can then be used on subsequent requests to affect access and response (e.g. vip-only vs. vip-pricing). We _could_ get into modeling what claims are available for agents to request, but I think this is optional and can be safely deferred. If and when we do, we can model it as a scope property, which maps to OIDC conventions:

```json
"scopes": {
  "dev.ucp.shopping.catalog:search": {
    "claims": ["vip", "b2b", "member_tier"]
  }
}
```

docs/specification/identity-linking.md
amithanda: Updated `buyer` to `user` across various files.

docs/specification/identity-linking.md
amithanda: updated.

docs/specification/identity-linking.md
amithanda: updated.

cusell-google (approved):

igrigorik (commented):
@amithanda pushed an update to match what the group is converging on -- ptal.

amithanda:
Thanks @igrigorik , Overall looks good. I added documentation in the checkout.md file around the scope that it exposes. Without this, we won't know the supported scope for checkout, since we may backport this and implementors may keep using the old scope format (which we want to deprecate).
Let me know if that looks good.

amithanda:
@maximenajim - Can you please take a look at this as well?

igrigorik:
@amithanda yep, makes sense. We should define scope groups for other core capabilities as well? Catalog, Cart and Order.

docs/specification/identity-linking.md
```diff
@@ -17,145 +17,565 @@
...
+## General Guidelines
+
+### For Platforms
+
+* **MUST** authenticate using `client_id` and `client_secret`
+    ([RFC 6749 §2.3.1](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1){ target="_blank" })
+    via HTTP Basic Authentication
+    ([RFC 7617](https://datatracker.ietf.org/doc/html/rfc7617){ target="_blank" })
+    when exchanging authorization codes for tokens.
+* **MUST** include user identity tokens in the HTTP `Authorization` header
```
maximenajim, line 121: The access-level table says user-authenticated access is platform credentials plus a user identity token, but this requirement consumes the HTTP `Authorization` header with the user bearer token. Existing REST guidance already allows `Authorization: Bearer` for platform OAuth, and many businesses will still need a platform-authenticated caller for tenant isolation, rate limits, fraud controls, and audit even when a user token is present. Without an explicit binding rule, a stolen or replayed user token can become detached from the platform client it was issued to.

I would add normative resource-server validation language here: user identity access tokens MUST be issued to the authenticated platform OAuth client, and businesses MUST validate issuer, audience, expiry, scopes, and the platform/client identity (`client_id`, `azp`, or equivalent) before accepting a user-authenticated request. If UCP intends platform authentication to be carried separately on the same request, name the mechanism as well, for example HTTP Message Signatures for the platform request plus `Authorization: Bearer <user_token>` for the user token, or a dedicated user-token header if `Authorization` remains available for platform credentials.

docs/specification/identity-linking.md
```diff
@@ -17,145 +17,565 @@
...
+scopes it supports. Consent screens **SHOULD** group related scopes
+intelligibly rather than listing individual operations: for example, "Allow
+[platform] to view your order history" rather than "grant
+`dev.ucp.shopping.order:read`."
+
+## Error Handling
+
+### `identity_required`
+
+When an operation is gated by a scope listed in `config.scopes` and the
```
maximenajim, line 414: This handles the missing-token case, but the more common interoperability failure will be a valid token with the wrong scope or a token that fails a recognized per-scope policy such as `min_acr` or `max_token_age`. If a platform presents `order:read` and calls an operation that needs `order:manage`, should the business return `identity_required`, `unauthorized`, a transport 403, or a re-consent URL? Different answers here create retry loops and bad user handoffs.

I would split the cases normatively: absent, expired, invalid, or unverifiable user token returns `identity_required`; valid token lacking a required scope or policy returns a standard `insufficient_scope` error with the required scope names in the message object, and MAY include `continue_url` for re-consent. That probably means adding `insufficient_scope` next to `identity_required` in `source/schemas/shopping/types/error_code.json`.

docs/specification/checkout.md
```diff
@@ -493,6 +493,16 @@ platform can prefill checkout state when initiating a buy-now flow.
 > [REST transport binding](checkout-rest.md). Accessing a permalink returns a
 > redirect to the checkout UI or renders the checkout page directly.
 
+## Scopes
+
+When user authentication is required to interact with checkout sessions, the Checkout capability defines the following scope:
+
+| Scope | Description |
+| :--- | :--- |
+| `dev.ucp.shopping.checkout:manage` | Provides access to all checkout operations on behalf of the user. |
```
shopdave: Suggest we add all permissions we want to support here. Would suggest one per operation: `:create`, `:get`, `:update`, `:complete`. Same for other capabilities.

docs/specification/identity-linking.md
```diff
@@ -17,145 +17,565 @@
...
+* **MUST** return the `iss` parameter in the authorization response
+    ([RFC 9207](https://datatracker.ietf.org/doc/html/rfc9207){ target="_blank" }).
+* **MUST** enforce PKCE
+    ([RFC 7636](https://datatracker.ietf.org/doc/html/rfc7636){ target="_blank" })
+    validation at the token endpoint for all authorization code exchanges.
+    Requests without a valid `code_verifier` **MUST** be rejected.
+* **MUST** enforce exact string matching for the `redirect_uri` parameter
+    during authorization requests to prevent open redirects and token theft.
+    The `redirect_uri` in the token request **MUST** be identical to the one
+    in the authorization request.
```
shopdave: Would it make sense to bringing in the loopback/port exception from OAuth 2.1 for desktop applications?

docs/specification/identity-linking.md
shopdave: Agree with @maximenajim's point of view here.

docs/specification/checkout.md
igrigorik: Stress testing: is `get` + `complete` a valid combination? As in, when or why would you allow `create` but not `update`, or gate `get` but allow create/update/complete without auth?

docs/specification/identity-linking.md
igrigorik: 👍

docs/specification/checkout.md
amithanda: @shopdave - I am not sure about the use case here. I understand the intention to make it more granular but semantically it may not make much sense to split. It also opens the door to incoherent implementations e.g business provides create scope but not update or complete. WDYT?

docs/specification/identity-linking.md
amithanda: Good point. Updated the language. PTAL and see if this looks ok.

docs/specification/identity-linking.md
amithanda: Added `insufficient_scope`. PTAL.

docs/specification/identity-linking.md
amithanda: Makes sense, Added. AI agents can run as local processes (desktop agents, CLI tools) that use loopback redirects (127.0.0.1 or [::1]) with OS-assigned ephemeral ports. Without the exception, those implementations would need to pre-register a specific port, which is impractical. RFC 8252 §7.3 (OAuth 2.0 for Native Apps) already defines this exception, and OAuth 2.1 is standardizing it across the board.

amithanda:
> We should define scope groups for other core capabilities as well? Catalog, Cart and Order.

Added scopes for other capabilities. PTAL

docs/specification/identity-linking.md
```diff
@@ -328,6 +342,42 @@ account linking.
...
 
+### `insufficient_scope`
+
+When a request arrives with a valid user identity token but the token lacks a
+scope required by the operation (or fails a scope-level policy such as
+`min_acr` or `max_token_age`), the business **MUST** return a UCP error
+response containing a message with `code: "insufficient_scope"`. The message
+`content` **MUST** identify the missing scope(s) by name so the platform can
+construct a targeted incremental authorization request. When performing a
+scope-bump, businesses **MUST** issue the new token with a scope set that is a
```
igrigorik: I don't think we should dictate that it's a scope bump. When an insufficient scope is hit, the user needs to renegotiate scopes and there is no reason why they shouldn't be able to revoke a scope they no longer need or don't want the platform to have access to. 

It still holds that platform needs to / can construct a request with added scope, but we can't enforce that user accepts that strict set, they may use the UI to uncheck some of them.

igrigorik (changes requested):
Hmm, UCP-Identity-Token doesn't feel right. Let's step back...

In standard OAuth, client authentication happens at the `/token` endpoint (via `client_secret_basic`, `private_key_jwt`, mTLS, etc), not via a Bearer header at API time. **Authorization** at API time uses the access token that carries `client_id` claim. The user-authorized token IS the platform's presented credential to the resource server -- resource servers validate `client_id`/`azp` against expected values.

All that to say, we don't need to stack a "platform Bearer" and a "user Bearer." The collision Maxime flagged only appears because we listed OAuth Bearer as one of the **platform** auth options at API time -- which, in retrospect, I think is a mistake. OAuth Bearer at API time *is* the user-authorized token; it's not a separate platform credential.

I think we can unwind and fix this in this PR...

1. **Drop OAuth Bearer from the *platform* auth options** in `cart-rest.md` (line 540) and `checkout-rest.md` (line 1408). Platform identity at API time = HTTP Message Signatures (RFC 9421), mTLS, or API key.
2. **`Authorization: Bearer <user_token>`** unambiguously carries the user-authorized access token. Add a forward reference from those auth sections to `identity-linking.md` for scope vocabulary and validation requirements.
3. **Drop `UCP-Identity-Token`**

@maximenajim please sanity check!

docs/specification/identity-linking.md
maximenajim: @igrigorik sanity-checked, this lands cleanly. Agreed on all three.

amithanda:
Thanks for the feedback. @igrigorik - I agree that we should avoid stacking a 'platform Bearer' and a 'user Bearer' and eliminate the need for a custom `UCP-Identity-Token` header.

However, rather than removing OAuth 2.0 completely from the list of options (which might imply UCP REST bindings don't support standard OAuth at all), I have updated the language in both `cart-rest.md` and `checkout-rest.md` to be more targeted based on OAuth 2.0 fundamentals:

> **OAuth 2.0**: Via `Authorization: Bearer {token}` header. Identifies the platform for agent-authenticated access, or both platform and user for user-authenticated access (see [Identity Linking](identity-linking.md)).

**Why this modified language is better than removing the option:**

1. **Aligns with Standard OAuth2**: Per RFC 9068, a user-authorized access token already carries claims (like `client_id` or `azp`) identifying the client application it was issued to. Therefore, the user token *is* the platform's credential at API time.
2. **Clarifies Both Access Levels**: It explains that the same standard `Authorization: Bearer` header is used for both Agent-authenticated access (client credentials) and User-authenticated access (auth code flow).
3. **Prevents Collision**: By clarifying that the token represents *both* parties in the user flow, it makes it clear that a separate platform credential is not sent at API time, directly solving the collision problem without falsely implying that OAuth is unsupported.

Let me know what you think?

docs/specification/identity-linking.md
amithanda: I agree users should be able to revoke scopes, I think the language I used was making it confusing. What I meant was to clarify that business should send back complete set of scope(s) to avoid asking for scopes one after the other.

I've updated `identity-linking.md` to:
*   **Remove the strict superset requirement** on the business when issuing tokens, respecting the user's choice to manage permissions.
*   **Require businesses to list all scopes needed for the operation** (not just missing ones) in the error response. This prompts the platform to request the full set, preventing existing scopes from being dropped by strict authorization servers and also avoiding users to get multiple scope requests one after the other.

Let me know if this sounds good.

docs/specification/checkout.md
shopdave: That's fair feedback. I was looking for 1:1 between scope and operation as a base rule, but see how this can cause inconsistencies.

igrigorik:
Per the TC sync, we converged on a split of responsibilities between UCP and OAuth (RFC8414) for expressing auth requirements vs availability vs runtime hints. As a think-out-loud sanity-check exercise, this means...

1. UCP `config.scopes` declare:
   - **hard gates**: scopes that *require* user auth for the operations they cover
   - optional `description` string on the scope that provides human-readable explanation of the required scope.
1. OAuth `scopes_supported` metadata (RFC 8414) declare the **accepted scope vocabulary**: 
   - i.e. list every scope the AS will honor if requested
1. OAuth `scopes_supported` - UCP `config.scopes` is the _optional layer_: scopes offered but not gating operation.
1. UCP `messages[]` can be used as runtime contextual hints: per-request advisories ("sign in for member pricing") with a well-known info code and description (e.g. "sign-in to unlock member-pricing")

## Example walkthrough

### Step 1: discover capabilities

Agent fetches the UCP profile:

```
GET https://merchant.example.com/.well-known/ucp
```
```json
{
  "ucp": {
    "capabilities": {
      "dev.ucp.shopping.catalog.search":   [{ ... }],
      "dev.ucp.shopping.checkout":         [{ ... }],
      "dev.ucp.common.identity_linking":   [{
        "config": {
          "scopes": {
            "dev.ucp.shopping.checkout:manage": {}
          }
        }
      }]
    }
  }
}
```

Agent learns:
- Catalog search and checkout are offered.
- Checkout `:manage` is a **required gate** — guest cannot complete checkout.
- Catalog has **no required scope** — guest search works.
- Identity linking is supported.

### Step 2: discover auth machinery

The agent fetches the OAuth Authorization Server metadata:

```
GET https://merchant.example.com/.well-known/oauth-authorization-server
```
```json
{
  "issuer": "https://merchant.example.com",
  "authorization_endpoint": "...",
  "token_endpoint": "...",
  "jwks_uri": "...",
  "scopes_supported": [
    "dev.ucp.shopping.checkout:manage",
    "dev.ucp.shopping.catalog.search:read"
  ],
  "code_challenge_methods_supported": ["S256"],
  "token_endpoint_auth_methods_supported": ["private_key_jwt", "none"],
  ...
}
```

Agent learns:
- Endpoints, JWKS, flow parameters, PKCE/client-auth requirements.
- The **full vocabulary of scopes the AS will honor**: `:manage` *and* `.search:read`.

### Step 3: reason about the auth landscape

```
required = config.scopes              = { checkout:manage }
accepted = scopes_supported           = { checkout:manage, catalog.search:read }
optional = accepted - required        = { catalog.search:read }
```

Three sets, distinct semantics:
- **Required**: auth needed to access these operations.
- **Accepted**: the AS honors these scopes if the agent decides to request them.
- **Optional** (the diff): merchant offers but doesn't gate. The exact semantics of why and what benefits might be unlocked if the user signs in are dependent on the merchant and are left to the merchant to express -- e.g. by emitting a message on the response indicating benefits of sign-in.

### Step N: runtime contextual hints

Static metadata answers "is auth available?" but not "is auth valuable in this request?" That's contextual to the merchant and their implementation (persisted carts and wishlists; loyalty tiers; saved address and prefs). The merchant  signals this via an informational message in the operation response, using a well-known code:

```json
{
  ...
  "messages": [
    {
      "type": "info",
      "code": "identity_optional",
      "content": "Sign in for member pricing and personalized results."
    }
  ]
}
```

The agent can present this as an opportunity to the user, and if accepted, it uses the defined machinery and dance of identity linking to negotiate appropriate scopes by introspecting the two discovery documents and relevant descriptions. As a reminder, info messages are non-blocking for related operations (can be ignored by agent).

_p.s. `identity_optional` as counterpart to `identity_required` for gated operations._

igrigorik:
@amithanda separate from above, pushed two updates, please sanity check: 
- https://github.com/Universal-Commerce-Protocol/ucp/pull/354/changes/4f20c5829c8ac4042826f8838850cb303555c9f6: support public clients via PKCE
- https://github.com/Universal-Commerce-Protocol/ucp/pull/354/changes/a1ef95e8bf39d0818ca80d22fb422c9621e95d5a: normalize capability scope sections

amithanda:
@igrigorik - thank you - both your updates look good. I added the change to add optional description for required scopes. Please feel free to approve, if everything looks good.

igrigorik (approved):
FYI, stacked a few more changes to document `identity_optional` and adopt `WWW-Authenticate`. Please sanity check, but otherwise I believe this is now good to land.

shopdave (approved):

docs/specification/identity-linking.md
```diff
@@ -17,145 +17,824 @@
...
-A scope covering a capability must grant access to all operations associated to
-the capability. For example, ucp:scopes:checkout\_session must grant all of:
-Get, Create, Update, Delete, Cancel, Complete.
+Businesses **MAY** include this info-severity code in successful
+responses when authentication is available and would meaningfully
+unlock additional capabilities in the current context.
+
+The message functions as a contextual signal; per-scope semantics (what
+each scope unlocks, the consent text to present) are conveyed
+independently via the `description` field on scope declarations. These
```
jamesandersen: @igrigorik / @amithanda - for clarity, a `description` field on scope declarations can only exist on `config.scopes` which are the *required* scopes for a given capability, correct?   This section is talking about Optional Authentication so these scopes will only be listed in OAuth `scopes_supported` (RFC 8414) where there's no way to set a `description` field.

At least for me, presenting these two mechanisms here in the Optional Authentication section introduces a little confusion that somehow Per-scope `description` *could* be used to describe the scope.

What would you think about...
1. Changing line 525 to "Businesses **SHOULD** include this info-severity code..." - as I think this is really the only way platforms will have access to the business's value prompt for optional authentication
2. Removing mention of the two distinct mechanisms here or just further clarifying that only `identity_optional` messages apply for Optional Authentication use cases

amithanda: @jamesandersen  - Thanks for catching that. Added some changes to fix. PTAL.
@igrigorik - Added couple of other minor documentation updates. Let me know if all looks good and I think this should be good to land.

jamesandersen (approved):

docs/specification/identity-linking.md
igrigorik: Great catch, ty for addressing it.

igrigorik (approved):
👍

nino-triandafilidis:
@codex review

PR #239, opened 2026-03-04
fix: Remove 'checkout' wrapper from doc examples

richmolj:
# Description

https://github.com/Universal-Commerce-Protocol/ucp/pull/141 incorrectly added a `checkout` wrapper key in the documentation examples. This does not reflect the schema or what we believe to be correct, so fixing those examples.

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] **Breaking change** (fix or feature that would cause existing
      functionality to not work as expected, **including removal of schema files
      or fields**)
- [ ] Documentation update

---

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #360, opened 2026-04-16
docs: remove entity wrapper from MCP response examples

jamesandersen:
## Summary

MCP response examples incorrectly nest the UCP payload under a `"checkout"`, `"cart"`, or `"order"` key inside `structuredContent`. Per the [OpenRPC specification](https://spec.open-rpc.org/), the `result.name` field is descriptive metadata for documentation and tooling — it does **not** create a wrapper key in the JSON-RPC wire format. The `name` field only has wire-format significance on `params` (by-name parameter passing), not on `result`.

This creates an inconsistency in the current docs: **error responses** already place the UCP envelope directly in `structuredContent` (no wrapper), while **success responses** incorrectly wrap it.

### Before (success response)

```json
"structuredContent": {
  "checkout": {
    "ucp": { ... },
    "id": "checkout_abc123",
    ...
  }
}
```

### After (success response — consistent with error responses)

```json
"structuredContent": {
  "ucp": { ... },
  "id": "checkout_abc123",
  ...
}
```

### Files changed

- `docs/specification/checkout-mcp.md` — 4 response examples
- `docs/specification/cart-mcp.md` — 6 response examples
- `docs/specification/order-mcp.md` — 3 response examples
- `docs/specification/overview.md` — 1 MCP transport example

### Context

Addresses the same issue as #239, which was closed because #216 was expected to include this fix. #216 corrected error response examples but left success responses wrapped.

## Checklist

- [x] Docs only — no schema changes
- [x] Self-reviewed
- [x] markdownlint passes
- [x] cspell passes
- [x] No new warnings introduced

### DIFF
--- a/docs/specification/cart-mcp.md
+++ b/docs/specification/cart-mcp.md
@@ -168,49 +168,47 @@ Maps to the [Create Cart](cart.md#create-cart) operation.
       "id": 1,
       "result": {
         "structuredContent": {
-          "cart": {
-            "ucp": {
-              "version": "{{ ucp_version }}",
-              "capabilities": {
-                "dev.ucp.shopping.checkout": [{"version": "{{ ucp_version }}"}],
-                "dev.ucp.shopping.cart": [{"version": "{{ ucp_version }}"}]
-              }
-            },
-            "id": "cart_abc123",
-            "line_items": [
-              {
-                "id": "li_1",
-                "item": {
-                  "id": "item_123",
-                  "title": "Red T-Shirt",
-                  "price": 2500
-                },
-                "quantity": 2,
-                "totals": [
-                  {"type": "subtotal", "amount": 5000},
-                  {"type": "total", "amount": 5000}
-                ]
-              }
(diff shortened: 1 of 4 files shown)


alexpark20 (approved):
Good catch 👏 

I think this should be merged into 04-08 release branch as well

igrigorik (approved):
Thanks for catching this @jamesandersen. Tracing lineage...

1. **#141** introduced the wrapper when adding `tools/call` envelope to MCP examples -- input wrapper keys (`"checkout"`, `"cart"`) were copy-pasted into output `structuredContent` — doh.
2. **#239** opened the correct fix, was closed as "resolved via #216" But 216 rewrote error response examples flat (correct) but didn't touch success responses - miss, doh again.

Once we land https://github.com/Universal-Commerce-Protocol/ucp/pull/359, this kind of bugs should -- hopefully -- become a thing of the past.

drewolson-google (approved):

PR #296, opened 2026-03-23
feat: add card network token credential and instrument types

jamesandersen:
# Enhancement Proposal: Card Network Token Credential Type

## Summary

A new `card_network_token` credential type for UCP that enables platforms to present card network tokens (Visa VTS, Mastercard MDES, Amex Token Service) directly as payment credentials in checkout sessions — without requiring re-tokenization through a handler or falling back to raw PANs.

## Motivation

UCP currently supports payment tokenization through handler-specific tokens and a delegated payment model. However, there is no support for card network tokens — tokenized card credentials issued by card networks that replace the primary account number (PAN) with a network-level token.

Card network tokens are an industry-standard payment security mechanism with broad PSP acceptance. Every major processor supports directly processing network tokens with the same core fields (token/DPAN, cryptogram, ECI):

| PSP | Always Required | Conditional / Optional | Documentation |
|-----|----------------|----------------------|---------------|
| Adyen | DPAN, expiry, cardholder name, cryptogram, ECI | Brand (co-badged cards), CVC | [Network Tokenization](https://docs.adyen.com/online-payments/network-tokenization#existing-network-tokens) |
| Chase | Token, expiry | Cryptogram + ECI (may be omitted for MIT/recurring) | [Online Payments](https://developer.payments.jpmorgan.com/docs/commerce/online-payments/capabilities/online-payments/payment-enhancements/tokenization) |
| Braintree | DPAN, expiry, cryptogram (for CIT) | ECI, token requestor ID, transaction source (for MIT) | [Bring Your Own Token](https://developer.paypal.com/braintree/docs/guides/network-tokens/bring-your-own-token/java/) |
| Stripe | Card data, cryptogram + ECI | Network, 3DS version | PCI required for API docs |
| Cybersource | Token, cryptogram, ECI, expiry | — | [Auth with NT](https://developer.cybersource.com/docs/barclays/en-us/payments/developer/all/rest/payments/CreatingOnlineAuth/CreatingAuthReqPNT.html) |

UCP's own specification explicitly identifies network tokens as a mechanism for platforms to avoid PCI-DSS scope:

> *"Platforms can avoid PCI-DSS scope by using handlers providing opaque credentials, never accessing raw payment data."*
> — [UCP Specification Overview: Credential Flow & PCI Scope](https://ucp.dev/latest/specification/overview/#credential-flow-pci-scope)

Yet despite this design intent, UCP currently has no dedicated credential type for network tokens. Platforms that already hold card network tokens cannot use them directly.

[...]

PR #367, opened 2026-04-17
docs: clarify network token usage and PCI scope in card credentials

jamesandersen:
## Summary

- Add a **Pre-Provisioned Network Tokens** section to the tokenization guide, documenting the BYOT path where platforms present card network tokens directly (`card_number_type: "network_token"`) without a tokenize/detokenize round-trip
- Clarify **PCI scope differences** between FPAN and network token credential modes in the overview and card credential schema description
- Add a network token credential example with `cryptogram` and `eci_value`

Split from #296 into a focused docs-only PR per reviewer feedback. TRID intentionally omitted per discussion in #296.

## Type of change

- [x] Documentation update

## Checklist

- [x] cspell passes (added `MDES` to custom dictionary)
- [x] markdownlint passes


docs/specification/overview.md
```diff
@@ -1464,6 +1464,10 @@ Most platform implementations can **avoid PCI-DSS scope** by:
 - Forwarding credentials without the ability to use them directly
 - Using PSP tokenization payment handlers where raw credentials never pass
     through the platform
+- Presenting pre-provisioned card network tokens
+    (`card_number_type: "network_token"`) — DPANs with cryptograms are not
```
kmcduffie: What do you think of having this cover both DPANs and CPANs?

source/schemas/shopping/types/card_credential.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "$id": "https://ucp.dev/schemas/shopping/types/card_credential.json",
   "title": "Card Credential",
-  "description": "A card credential containing sensitive payment card details including raw Primary Account Numbers (PANs). This credential type MUST NOT be used for checkout, only with payment handlers that tokenize or encrypt credentials. CRITICAL: Both parties handling CardCredential (sender and receiver) MUST be PCI DSS compliant. Transmission MUST use HTTPS/TLS with strong cipher suites.",
+  "description": "A card credential containing payment card details. When card_number_type is 'fpan', this contains a raw Primary Account Number (PAN) — both parties MUST be PCI DSS compliant. When card_number_type is 'network_token', this contains a Device PAN (DPAN) with cryptogram and ECI, which operates under reduced PCI scope as DPANs are not considered cardholder data by most acquirers. FPAN credentials MUST NOT be used directly for checkout — use payment handlers that tokenize or encrypt credentials. Transmission MUST use HTTPS/TLS with strong cipher suites.",
```
kmcduffie: There is a situation where the card_number_type is fpan, but can be exchanged in a secure flow outside of UCP, where what is passed from the payment handler is effectively a foreign key to that value that was previously exchanged, or a reference to something that can be retrieved.

docs/specification/tokenization-guide.md
```diff
@@ -68,6 +68,42 @@ Tokenization handlers transform credentials between source and checkout forms:
 Tokenization handlers accept source credentials (e.g., card with FPAN) and
 produce checkout credentials (e.g., tokens).
 
+### Pre-Provisioned Network Tokens
+
+Not all credential flows require a tokenize/detokenize round-trip. When a
+platform already holds a **card network token** (provisioned via Visa VTS,
```
kmcduffie: Given the new APIs networks are standing up, what do you think of dropping the API names and having the documentation simply reference Visa / MC?

docs/specification/tokenization-guide.md
```diff
@@ -68,6 +68,42 @@ Tokenization handlers transform credentials between source and checkout forms:
...
+Mastercard MDES, or similar), it can present the token directly as a
+`card_credential` with `card_number_type: "network_token"`. This bypasses
+the tokenization handler lifecycle entirely — the credential is ready for
+checkout as-is.
+
+```json
+{
+  "type": "card",
```
kmcduffie, line 88: There is also a flow where this obj, even when DPAN / CPAN, could still be wrapped in a payment_handler wrapper. It wouldn't require a de-tokenization call to an out of band API, but would make the object opaque to any intermediate parties. 

Can you update the language to be open to that setup as well? I would expect the object definition of the card credential would specify what the object contains.

kmcduffie:
@jamesandersen At a high level I think this makes sense. At the same time, I would like to avoid setting the UCP documentation in a way that it needs to be updated if PCI guidance changes. Would you be open to adjusting the language to reference the PCI guidance and do the right thing?

jamesandersen:
@kmcduffie Good call — I was trying to find phrasing that didn't imply network tokens were *not* in scope of PCI DSS, but referring to the spec itself with guidance to consult it is a better path. Updated all three places where I made scope assertions (overview, tokenization guide, card_credential.json description) to point to current PCI DSS guidance instead. Pushing now.

docs/specification/overview.md
```diff
@@ -1464,6 +1464,11 @@ Most platform implementations can **avoid PCI-DSS scope** by:
 - Forwarding credentials without the ability to use them directly
 - Using PSP tokenization payment handlers where raw credentials never pass
     through the platform
+- Presenting pre-provisioned card network tokens
+    (`card_number_type: "network_token"`) — DPANs with cryptograms may
```
kmcduffie: With CPANs, not all cryptograms will be device bound. What do you think of updating the comments to be more broadly around cryptograms?

kmcduffie (commented):
Sorry for some of the outdated comments, I couldn't figure out how to delete the no longer relevant ones.

docs/specification/tokenization-guide.md
raginpirate: nit: I would drop any concern about who provisioned it here to stay bias-free in the protocol documentation 😄

docs/specification/tokenization-guide.md
```diff
@@ -68,6 +68,42 @@ Tokenization handlers transform credentials between source and checkout forms:
...
 produce checkout credentials (e.g., tokens).
 
+### Pre-Provisioned Network Tokens
+
+Not all credential flows require a tokenize/detokenize round-trip. When a
+platform already holds a **card network token** (provisioned via Visa VTS,
+Mastercard MDES, or similar), it can present the token directly as a
+`card_credential` with `card_number_type: "network_token"`. This bypasses
+the tokenization handler lifecycle entirely — the credential is ready for
+checkout as-is.
```
raginpirate: Hmm thats interesting: to make this work mechanically, we would require a base dev.ucp.network_token payment handler to allow proper negotiation of direct network token support at checkout + advertising constraints.

Until we decide to deliver something like that (and have a volunteer business & platform adopters), this statement is not necessarily true.

docs/specification/tokenization-guide.md
```diff
@@ -68,6 +68,42 @@ Tokenization handlers transform credentials between source and checkout forms:
 Tokenization handlers accept source credentials (e.g., card with FPAN) and
 produce checkout credentials (e.g., tokens).
 
+### Pre-Provisioned Network Tokens
+
+Not all credential flows require a tokenize/detokenize round-trip. When a
+platform already holds a **card network token** (provisioned via Visa VTS,
+Mastercard MDES, or similar), it can present the token directly as a
+`card_credential` with `card_number_type: "network_token"`. This bypasses
```
raginpirate: Hmmmm... From a business perspective nothing happened involving them, but from a platform perspective they still need to generate a cryptogram, which to me is a form of tokenization process that a handler can/should describe in the ecosystem 😄

docs/specification/tokenization-guide.md
raginpirate: Aligned and truthfully I think thats what would fit the current state of UCP!

docs/specification/tokenization-guide.md
```diff
@@ -68,6 +68,42 @@ Tokenization handlers transform credentials between source and checkout forms:
 Tokenization handlers accept source credentials (e.g., card with FPAN) and
 produce checkout credentials (e.g., tokens).
 
+### Pre-Provisioned Network Tokens
```
raginpirate: Just a note that this document, although it uses card as an example, does not actually get into the specifics of how one should handle card tokenization. This context is a bit jarring living here, it might be worth pivoting this to describe secure tokens and using network tokens as an example of reducing compliance risk with a secure tokenization protocol.

raginpirate (commented):
Really appreciate you working on this @jamesandersen! I'm definitely enjoying the discourse as we find the right place to fit network tokens into UCP 😄

docs/specification/overview.md
jamesandersen: Good call — updated the language to say "network tokens with cryptograms" rather than "DPANs with cryptograms," so it covers both DPANs and CPANs without assuming device-binding.

docs/specification/tokenization-guide.md
jamesandersen: Agreed — dropped the VTS/MDES references.

docs/specification/tokenization-guide.md
jamesandersen: Alright - I've tried to reframe the section as "Pre-Provisioned Credentials," with a general pattern of credentials acquired in advance that don't require a per-transaction tokenize/detokenize round-trip. Network tokens are a concrete example of this pattern (though details of provisioning / submitting for payment are outside the scope of this doc) rather than the sole focus.  WDYT?

docs/specification/tokenization-guide.md
jamesandersen: Yeah, I'm with you on the need for a handler through which the platform and business can "agree" on use of a network token.

docs/specification/tokenization-guide.md
jamesandersen: You're right — I've updated the language to cover a payment handler requirement for capability negotiation. The specific handler definition (while perhaps the spirit of what I was after with #296 - but didn't quite nail) is something we tackle in the future.  Hopefully you'll find the docs here framing that up correctly now though.

docs/specification/tokenization-guide.md
jamesandersen: Yep - language is updated to note a payment handler is needed for negotiation.  This should leave the door open for handlers that also that wrap credentials for intermediary opacity. 

I'm interested to discuss/learn more about pros/cons of wrapping vs. directly passing network tokens (which the schema seems to allow on a complete checkout call) BUT also interested in getting this PR over the finish line before diving into that further ;-)

jamesandersen (commented):
Thanks @kmcduffie  and @raginpirate for sticking with this PR - good feedback throughout and hopefully we're getting close with this latest round of updates ;-)

source/schemas/shopping/types/card_credential.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "$id": "https://ucp.dev/schemas/shopping/types/card_credential.json",
   "title": "Card Credential",
-  "description": "A card credential containing sensitive payment card details including raw Primary Account Numbers (PANs). This credential type MUST NOT be used for checkout, only with payment handlers that tokenize or encrypt credentials. CRITICAL: Both parties handling CardCredential (sender and receiver) MUST be PCI DSS compliant. Transmission MUST use HTTPS/TLS with strong cipher suites.",
+  "description": "A card credential containing payment card details. When card_number_type is 'fpan', this contains a raw Primary Account Number (PAN) — both parties MUST be PCI DSS compliant. When card_number_type is 'network_token', this contains a Device PAN (DPAN) with cryptogram and ECI, which may qualify for reduced PCI scope — consult current PCI DSS guidance for applicable requirements. FPAN credentials MUST NOT be used directly for checkout — use payment handlers that tokenize or encrypt credentials. Transmission MUST use HTTPS/TLS with strong cipher suites.",
```
kmcduffie: Can you update this to be DPAN / CPAN agnostic?

kmcduffie (changes requested):
Thanks @jamesandersen for sticking with this! I think the documentation updates will be very helpful.

One small nit around the wording in card_credential.json. Otherwise, I'm supportive of this PR.

source/schemas/shopping/types/card_credential.json
jamesandersen: Good catch - just updated this

kmcduffie:
If you want to resolve the conflicts, I'm supportive of this change.

damaz91:
@kmcduffie I resolved the merge conflicts - @jamesandersen let us know if you want to move forward with this!

Thanks!

PR #371, opened 2026-04-21
fix: Fix inconsistent schema on documentation examples & discounts extension schema

jingyli:
# Description

This PR aims to address some minor schema inconsistencies that are present in the current UCP spec:
1) In `discount.md`: Update all the examples to follow the correct UCP schema - `item.id` is the only field populated in requests, `quantity` pulled out to the `line_item` level, etc.
2) In `checkout-rest.md` and `checkout-mcp.md`: Remove unsupported `available_quantity` field and add in other required fields like `item` in business error example.
3) Fix incorrect usage of `readOnly` annotation in `discount.json`.

## Type of change

- [x] Documentation update

---

## Checklist

### DIFF
--- a/docs/specification/checkout-mcp.md
+++ b/docs/specification/checkout-mcp.md
@@ -656,10 +656,16 @@ as JSON-RPC `result` with `structuredContent` containing the UCP envelope and
       "line_items": [
         {
           "id": "li_1",
-          "quantity": 100,
-          "available_quantity": 12
+           "item": {
+              "id": "item_123",
+              "title": "Blue Jeans",
+              "price": 5000
+            },
+          "quantity": 12,
+          "totals": [...]
         }
       ],
+      "totals": [...],
       "messages": [
         {
           "type": "warning",
--- a/docs/specification/checkout-rest.md
+++ b/docs/specification/checkout-rest.md
@@ -1291,11 +1291,17 @@ with HTTP 200 and the UCP envelope containing `messages`:
   "status": "incomplete",
   "line_items": [
     {
-      "id": "item_456",
-      "quantity": 100,
-      "available_quantity": 12
+      "id": "li_1",
+        "item": {
+          "id": "item_123",
+          "title": "Blue Jeans",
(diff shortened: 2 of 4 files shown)


cusell-google (approved):
Some automated doc checks were skipped, could you make sure they run before merging?

docs/specification/checkout-mcp.md
```diff
@@ -656,10 +656,16 @@ as JSON-RPC `result` with `structuredContent` containing the UCP envelope and
       "line_items": [
         {
           "id": "li_1",
-          "quantity": 100,
-          "available_quantity": 12
+           "item": {
+              "id": "item_123",
+              "title": "Blue Jeans",
+              "price": 5000
```
kmcduffie, line 662: @jingyli Is it worth modeling this as the price.json obj that is both amount and currency?
https://github.com/Universal-Commerce-Protocol/ucp/blob/6b61c3f7dab1a7825dc6c8a7c8c433d426db21d5/source/schemas/shopping/types/price.json

jingyli: My thinking is that the current model is more explicit: we already have `currency` at the `checkout` object level - https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/shopping/checkout.json#L74 - and our expectation is that all amounts in the checkout object should follow that `currency` field.

jingyli:
@cusell-google Thanks for the review! The skipped checks are expected - those checks should only be executed as part of the release workflow.

raginpirate (approved):
Technically this is a breaking change, because the readOnly was not our expected control over the output schemas and validation here. But we should be fine accepting this given our belief that consumers of the spec will not be impacted and would have interpreted this the same way.

mmohades (approved):
Thanks for catching this inconsistency in the docs!

igrigorik (approved):
👍

westeezy (approved):
lgtm ty.

yanheChen (approved):

PR #383, opened 2026-04-24
feat: Add seamless checkout transition functionality to ECaP 

jingyli:
# Description

Follow-up to [feedback](https://github.com/Universal-Commerce-Protocol/ucp/pull/244#discussion_r3022718182) received on https://github.com/Universal-Commerce-Protocol/ucp/pull/244. We need some mechanism to allow host to facilitate seamless transition from Embedded Cart to Embedded Checkout without needing to renegotiate via `create_checkout` if they would like.

## Category (Required)

## Related Issues

Not an issue, but related to https://github.com/Universal-Commerce-Protocol/ucp/pull/244.

## Checklist


docs/specification/embedded-cart.md
```diff
@@ -381,6 +383,56 @@ proceed to initiate a checkout session based on the completed cart by issuing a
...
+        "cart": { ...cart fields... },
+        "transition": {
+            "checkout": {
+                "ucp": {
+                    "version": "{{ ucp_version }}",
+                    "status": "success",
+                    "capabilities": {
+                        "dev.ucp.shopping.checkout": [ { "version": "{{ ucp_version }}" } ]
+                    },
+                    "services": {
```
gsmith85, line 420: Nit: Malformed JSONSchema. You're missing the capability name for the nested object.

source/services/shopping/embedded.openrpc.json
```diff
@@ -496,6 +496,28 @@
...
+        {
+          "name": "transition",
+          "schema": {
+            "type": "object",
+            "description": "Optional advertisement from business for seamless transition from cart to checkout.",
+            "required": ["checkout"],
+            "properties": {
+              "checkout": {
+                "type": "object",
+                "description": "Custom object containing data needed by host to facilitate the transition.",
```
gsmith85: Nit: The description for the transition object is a bit of a tautology. Consider updating the description to something descriptive, like:

> The UCP capability metadata and destination URL required by the host application to bootstrap and transition into a new embedded checkout session.

source/services/shopping/embedded.openrpc.json
```diff
@@ -496,6 +496,28 @@
...
+                "required": ["ucp", "url"],
+                "properties": {
+                  "ucp": { "$ref": "#/components/schemas/ucp_success" },
+                  "url": {
+                    "type": "string",
```
gsmith85, line 514: Include `"format": "uri",` for basic string URL structure validation.

source/services/shopping/embedded.openrpc.json
```diff
@@ -496,6 +496,28 @@
...
+            "description": "Optional advertisement from business for seamless transition from cart to checkout.",
+            "required": ["checkout"],
+            "properties": {
+              "checkout": {
+                "type": "object",
+                "description": "Custom object containing data needed by host to facilitate the transition.",
+                "required": ["ucp", "url"],
+                "properties": {
+                  "ucp": { "$ref": "#/components/schemas/ucp_success" },
+                  "url": {
```
gsmith85, line 513: While a generic uri format permits http://, allowing an unencrypted handoff at the cart-to-checkout inflection point is something we should avoid. It leaves session parameters and any delegated authority contexts wide open to interception.

One solution would be to introduce a reusable `secure_url` type as a schema components. We can enforce ^https:// to guarantee production integrity, while explicitly whitelisting localhost so local dev remains frictionless.

Any thoughts?  Doesn't need to be part of this PR, I can create a standalone proposal.

docs/specification/embedded-cart.md
```diff
@@ -381,6 +383,56 @@ proceed to initiate a checkout session based on the completed cart by issuing a
...
 
+Business **MAY** also choose to specify a `transition` field to advertise a mechanism
+for hosts to seamlessly transition into Embedded Checkout from the completed
+Embedded Cart. This is represented by `transition` containing a minimal `checkout` object:
+
+- `ucp` (object, **REQUIRED**): Metadata to fully qualify the advertisement.
+    Business **MUST** include a checkout capability and an embedded service binding
+    with `config.delegate`.
+- `url` (string, **REQUIRED**): URL representing shift to checkout. Business **MAY** choose
+    to specify a redirect URL without having the actual checkout session created.
```
gsmith85: How I interpret the "Business **MAY** choose" case is that the business would advertise a URL like:

`https://merchant.com/.../checkout/start?cart_id=cart_999`

This effectively creates a lazy initiation mechanism, and allows the host to transition into checkout without a preprovisioned session.

However, the ECP spec (Section 3.1.2 and 3.2) is built on the assumption that the host has already received a JSON 'Checkout Response' containing the id *before* the iframe is loaded. By allowing a direct transition in embedded-cart.md without this response, we are potentially creating a gap.

Since the Binding Contract (Section 3.3.3) starts at the handshake, the host needs that checkout id immediately to anchor its orchestration, logging, and validation. Relying on the subsequent `ec.start` notification is too late, as it leaves a race condition where delegated actions can be triggered before the host has a session reference.

Recommendation: Require that the id (`checkout_id`) MUST be included in the `ec.ready` request params.  This shifts the id exchange earlier, and should allow for accommodation of this use case.

docs/specification/embedded-cart.md
```diff
@@ -381,6 +383,56 @@ proceed to initiate a checkout session based on the completed cart by issuing a
...
+                        {
+                            "version": "{{ ucp_version }}",
+                            "transport": "embedded",
+                            "config": {
+                                "delegate": [...]
+                            }
+                        }
+                    }
+                },
+                "url": "https://merchant.example.com/cart-to-checkout/checkout123",
```
gsmith85, line 432: Recommendation: We should not rely on the URL for (checkout) identity disambiguation.

While a merchant may bake an ID into the URL slug, that identity is opaque to the host's programmatic orchestration. To support agentic workflows and reliable logging, let's include the checkout  id as an explicit field in the transition.checkout object (when known).

This makes it slightly more explicit that there is a JIT variant too where the id isn't provided until `ec.ready`.

docs/specification/embedded-cart.md
jingyli: Thanks for catching it, updated.

source/services/shopping/embedded.openrpc.json
jingyli: Updated the description here, I'm a little bit hesitant to explicitly call out the sub-fields (i.e. UCP metadata, URL) in the object in case it changes in the future.

source/services/shopping/embedded.openrpc.json
jingyli: There are a couple of places in UCP today that emphasize the use of `https`:
- https://ucp.dev/latest/specification/overview/#transport-security
- https://ucp.dev/latest/specification/checkout/#format

I tightened the language of `url` in the spec documentation to also call out that it **MUST** be a HTTPS URL. I think the idea of enforcing it via schema directly makes sense, though agree with you that it can be treated as a standalone proposal/refactoring and should be consistently applied over all UCP URL fields.

docs/specification/embedded-cart.md
jingyli: > However, the ECP spec (Section 3.1.2 and 3.2) is built on the assumption that the host has already received a JSON 'Checkout Response' containing the id before the iframe is loaded. By allowing a direct transition in embedded-cart.md without this response, we are potentially creating a gap.

_(Preface: this was also the part that took me some time to wrap my head around)_ 
I don't think we are necessarily creating a gap here. To initiate ECP today, we need to append certain parameters onto the `continue_url` we have received via the Checkout response ([ref](https://ucp.dev/latest/specification/embedded-checkout/#loading-an-embedded-checkout-url)). However, there are [2 flavours of `continue_url` that are prescribed by UCP](https://ucp.dev/latest/specification/checkout/#format):
- Server-side stateful URL (the one where checkout session is proactively created by the business)
- Stateless permalinks (no actual session is created until the redirect lands)

There is nothing wrong reusing the pattern here and offer both URL formats for full flexibility. What it would mean is that the `checkout` resource would only be returned to the host for the first during `ec.start` in the case of a stateless permalink transition URL.

> Relying on the subsequent ec.start notification is too late, as it leaves a race condition where delegated actions can be triggered before the host has a session reference.

I don't think this race condition is ever possible given `ec.start` is a core lifecycle message that signals the readiness of the checkout UI for interaction. Without it, buyer cannot use Embedded Checkout to trigger a delegated action.

docs/specification/embedded-cart.md
jingyli: > To support agentic workflows and reliable logging, let's include the checkout id as an explicit field in the transition.checkout object (when known).

Would you be able to help me understand this argument a bit more? I actually feel like having an explicit `id` here is a little bit weirder given:
- This would be the first time where the identifier of a core capability can be optional in UCP schema.
- If this `id` is set, does it mean that in theory (since embedded context can be concluded after receiving `ep.cart.complete` and host can move to a different transport implementation to continue with checkout): platforms have the ability to directly call `update_checkout` against this ID with a request payload and bypass `create_checkout` entirely?

jingyli:
Thanks @gsmith85 for the feedback! Comments are addressed and requesting another pass (especially on the 2 open threads around presence of `checkout.id`).

PR #391, opened 2026-04-27
feat: attribution field for platform referral context

igrigorik:
Alternative proposal to [#295](https://github.com/Universal-Commerce-Protocol/ucp/pull/295). cc @jamesandersen 

Adds a top-level `attribution` field to UCP carrying platform-emitted referral and conversion-event context. It carries the same parameters platforms communicate via URL query parameters in browser-based flows, in the same flat key-value form.

```json
{
  "attribution": {
    "campaign_id": "18234567890",
    "campaign_source": "google",
    "campaign_medium": "cpc",
    "campaign_name": "spring_2026",
    "gclid": "EAIaIQobChMI..."
  }
}
```

Available on `cart`, `checkout`, and catalog requests as platform-provided input. Appears on `order` as a business-emitted snapshot of the originating checkout's attribution.

## Design decisions

- **Core field, not extension or capability.** Attribution is informational; its presence or absence does not affect protocol behavior. Capability negotiation earns nothing: there's no merchant decision to gate. The field lives in core schema alongside `signals` and `context` (the three platform/buyer-asserted session-metadata fields).
- **Open string-keyed map.** UCP does NOT prescribe attribution models, windows, or assignment logic. Platforms use their existing conventions (GA4 campaign parameters, click identifiers like `gclid` / `fbclid` / `ttclid`); businesses receive and process per their own analytics needs. No UCP-imposed vocabulary, existing event/analytics pipelines work transparently by piping contents of attribution into their existing event processing logic; attribution is equivalent to URL-parsed fields on browser-based flow.
- **Separate from `signals`.** `signals` carries authorization and abuse-prevention data with a specific trust posture. `attribution` carries marketing/referral context with different consent and routing expectations. Distinct fields keep these tiers clean and let each evolve independently.
- **Spans the journey.** Lives on every operation where referral context might matter — catalog discovery, cart assembly, checkout, and order — rather than scoped to a single capability.

---
## Checklist

### DIFF
--- a/docs/specification/cart.md
+++ b/docs/specification/cart.md
@@ -220,6 +220,15 @@ requirements.
 
 {{ schema_fields('types/signals', 'checkout') }}
 
+### Attribution
+
+Platform-provided referral and conversion-event context — campaign IDs,
+click identifiers, and source/medium markers communicated by the platform.
+See [Attribution](overview.md#attribution) for details and consent
+requirements.
+
+{{ schema_fields('types/attribution', 'checkout') }}
+
 ### Total
 
 The same totals contract applies to cart and checkout. See
--- a/docs/specification/overview.md
+++ b/docs/specification/overview.md
@@ -1722,6 +1722,48 @@ signal is provided; an `info` is advisory and non-blocking.
 }
 ```
 
+### Attribution
+
+Platforms refer users to businesses through many channels — paid ads,
+organic recommendations, influencer links, AI agents. In a browser-based
+flow, the referral context (campaigns, click identifiers, source/medium
+markers) flows through URL query parameters. The `attribution` field
+enables platforms to communicate the same parameters to businesses.
+
+UCP does **NOT** prescribe attribution models, windows, or assignment
+logic. Platforms use their existing conventions (GA4 campaign parameters,
+click identifiers like `gclid` / `fbclid` / `ttclid`, etc.); businesses
+receive and process them according to their own analytics needs.
+
+```json
+{
+  "attribution": {
+    "campaign_id": "18234567890",
+    "campaign_source": "google",
+    "campaign_medium": "cpc",
+    "campaign_name": "spring_2026",
+    "gclid": "EAIaIQobChMI..."
+  }
+}
+```
+
+Attribution is informational and optionally provided by the platform.
+Businesses do not negotiate or advertise support; the field's presence or
+absence MUST NOT affect the response or negotiation.
+
+The data can carry pseudonymous identifiers (click IDs, session keys)
+treated as personal data under applicable data protection laws. Platforms
+and businesses are each responsible for compliance in their respective
+jurisdictions: platforms determine what to emit and disclose; businesses
+apply their own data handling, retention, and consent policies. The
+`buyer_consent` extension provides a structured channel for buyers to
+communicate consent state. Direct identifiers (email, phone,
+name, postal address) belong on `buyer`, not `attribution`.
+
+Attribution appears on cart, checkout, and catalog requests as
+platform-provided attribution context; on order it appears as a
+business-emitted snapshot of the originating checkout's attribution.
+
 ### Transaction Integrity and Non-Repudiation
 
 For scenarios requiring cryptographic proof of authorization (e.g., autonomous
--- a/docs/specification/overview.md
+++ b/docs/specification/overview.md
@@ -1757,8 +1757,7 @@ and businesses are each responsible for compliance in their respective
 jurisdictions: platforms determine what to emit and disclose; businesses
 apply their own data handling, retention, and consent policies. The
 `buyer_consent` extension provides a structured channel for buyers to
-communicate consent state. Direct identifiers (email, phone,
-name, postal address) belong on `buyer`, not `attribution`.
+communicate consent state.
 
 Attribution appears on cart, checkout, and catalog requests as
 platform-provided attribution context; on order it appears as a
--- a/source/schemas/shopping/types/attribution.json
+++ b/source/schemas/shopping/types/attribution.json
@@ -1,0 +1,11 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/attribution.json",
+  "title": "Attribution",
+  "description": "Platform-emitted referral and conversion-event context — campaign identifiers, click IDs, source/medium markers, etc. The same parameters platforms communicate via URL query parameters in browser-based flows.",
+  "type": "object",
+  "additionalProperties": {
+    "type": "string",
+    "description": "URL-style parameter value, encoded as a string. Numeric or boolean values MUST be string-encoded as they would be in a URL query string."
+  }
+}
--- a/.cspell/custom-words.txt
+++ b/.cspell/custom-words.txt
@@ -110,3 +110,6 @@ keyid
 reauth
 reprepare
 sandboxing
+fbclid
+gclid
+ttclid
--- a/docs/documentation/schema-authoring.md
+++ b/docs/documentation/schema-authoring.md
@@ -367,6 +367,19 @@ typos in core metadata like the `ucp` block).
 }
 ```
 
+### Property-Count Constraints (`minProperties` / `maxProperties`)
+
+By default, UCP schemas do not set `minProperties` or `maxProperties` on
+object fields:
+
+- **`maxProperties`** — Limits are deferred to implementers. The protocol
+  does not define caps because any specific limit requires judgment calls
+  that inevitably run into exceptions. Implementers are encouraged to
+  impose their own constraints and surface clear error feedback to support
+  debugging and good behavior.
+- **`minProperties`** — Empty objects (`{}`) are well-formed and harmless.
+  Implementers should accept and process them as a no-op.
+
 ## Complete Example: Capability Schema
 
 A capability schema defines both payload structure and declaration variants:
--- a/docs/specification/catalog/index.md
+++ b/docs/specification/catalog/index.md
@@ -89,6 +89,15 @@ and abuse prevention. Signal values MUST NOT be buyer-asserted claims. See
 
 {{ schema_fields('types/signals', 'catalog') }}
 
+### Attribution
+
+Platform-provided referral and conversion-event context — campaign IDs,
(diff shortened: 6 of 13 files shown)


jamesandersen:
Thanks @igrigorik !  After the discussion on #295 and the concerns raised there, I'm ready to align with this approach:
                                                                                                                        
-  _Preserves the core goal_. Platforms can still pass campaign IDs, UTM params, and click identifiers (gclid, fbclid, ttclid) - the same data #295 was designed to carry — so merchants don't lose referral context in agentic flows.
- _Aligns with @ashutosh-goog's framing_. UCP provides the data; it doesn't perform attribution. The open map design makes this explicit — platforms write what they'd normally encode as URL query params, merchants interpret per their own models.
- _Capability negotiation isn't needed here._ As @igrigorik noted, this data is informational and doesn't affect protocol behavior or response negotiation. Merchants don't gate web traffic on whether UTM params are present — same logic applies here.
- _Addresses the data routing / consent separation_. Giving attribution its own top-level field (separate from signals) preserves the pipeline separation both @amithanda and I raised — fraud/risk signals and marketing context have different trust postures and compliance requirements.
- _Covers the full journey_. Spanning catalog → cart → checkout → order naturally resolves the per-request question @mnaga raised on #295.
- _Lower adoption cost_. Platforms emit existing conventions as string key-value pairs rather than conforming to a new structured schema.
                                                            
One suggestion: it would be worth adding minProperties: 1 (to prevent empty attribution objects) and a maxProperties  cap - maybe 20? - to the schema — @amithanda raised similar constraints on #295 and they'd be a good fit here too.

docs/specification/overview.md
```diff
@@ -1722,6 +1722,48 @@ signal is provided; an `info` is advisory and non-blocking.
...
+absence MUST NOT affect the response or negotiation.
+
+The data can carry pseudonymous identifiers (click IDs, session keys)
+treated as personal data under applicable data protection laws. Platforms
+and businesses are each responsible for compliance in their respective
+jurisdictions: platforms determine what to emit and disclose; businesses
+apply their own data handling, retention, and consent policies. The
+`buyer_consent` extension provides a structured channel for buyers to
+communicate consent state. Direct identifiers (email, phone,
+name, postal address) belong on `buyer`, not `attribution`.
```
westeezy: Nit: Should this be worded more strongly and along the lines of 

> **Platforms **MUST NOT** include direct identifiers, such as email, phone, name, postal address in,  attribution**. These fields belong on `buyer`.

source/schemas/shopping/types/attribution.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/attribution.json",
+  "title": "Attribution",
+  "description": "Platform-emitted referral and conversion-event context — campaign identifiers, click IDs, source/medium markers, etc. The same parameters platforms communicate via URL query parameters in browser-based flows.",
```
westeezy, line 5: I am assuming this is last-write-wins if there is duplicated keys based on how most parsers and `URLSearchParams.set` operate. Does it need to be explicitly specified?

westeezy (approved):
I like the flexibility in this approach. I left a few questions and agree with the suggestion on `maxProperties`.  Other than that this LGTM!

igrigorik:
@jamesandersen @westeezy I'm not opposed to the belt-and-suspenders semantics of `{min,max}Properties`, but practically we've not enforced max in any UCP schemas, and min in only two places. This discussion has come up before, and we punted: discussions on max inevitably run into arbitrary judgment calls -- e.g. is 20 attributes really enough? I am sure I've seen URLs with more than that, and I'd rather leave this as an implementer decision than encode it as a protocol-governance constraint.

I think this is a good topic to revisit at TC. If we want to start enforcing min/max, that's a breaking change for existing fields, and I'd prefer to do a sweep and be consistent.

source/schemas/shopping/types/attribution.json
igrigorik: Non-problem in this case since we're not passing the literal string but the parsed object representation, which means we can't have duplicate keys.

docs/specification/overview.md
igrigorik: Hmm, good prompt. After some thinking though, went in opposite direction: https://github.com/Universal-Commerce-Protocol/ucp/pull/391/changes/9044fde2548d5d4bf817d59278a7c6b7acf7d939

westeezy: 👍 i think this path is equally as good :)

westeezy:
> I am sure I've seen URLs with more than that, and I'd rather leave this as an implementer decision than encode it as a protocol-governance constraint.

Yeah after thinking on this a bit more, I agree with this. My original motivation around max was that we should have some rails on it due to URL length limits. However I feel that I agree with you it almost feels arbitrary at some point. I think I would rather leave off min/max so bit of a 180 from me. Leaving as an implementer decision feels like the right fidelity to me long term.

amithanda:
Thought about this a more -  I am aligned with the approach in this PR of using Flat Optional Key-Value Pairs for handling attribution data. Here is a summary:

*   **Separation of Concerns**: It keeps marketing/attribution data separate from operational `signals`, respecting their different trust postures and data routing needs (e.g., security vs. marketing tiers).
*   **No Negotiation Overhead**: Attribution is purely informational and does not affect the state of the transaction. It should not require capability negotiation or gate interactions.
*   **Low Friction & Flexibility**: A flat map allows platforms to pass existing conventions (like UTMs and click IDs) directly, without mapping to a rigid schema that would require constant updates as new advertising platforms emerge.

I also thought it might be a good idea to use this example to document the principles, we are using to distinguish between **extensions** (which are also optional) and **Optional flat key-value pairs** - helpful for similar discussions in future:

| Criterion | Use a **UCP Extension** | Use **Optional Flat Key-Value Pairs** |
| :--- | :--- | :--- |
| **Impact on Behavior** | The data **changes the state or execution** of the operation (e.g., applying a discount). | The data is **purely informational** and does not affect the protocol's execution or response logic. |
| **Data Stability** | The data structure is **stable, standardized**, and widely agreed upon across the ecosystem. | The data is **volatile, platform-specific**, or rapidly evolving (e.g., marketing tags, click IDs). |
| **Capability Negotiation** | Requires mutual agreement in profiles **AND** an active parent capability (otherwise pruned). | The data is provided on a best-effort basis and is **consumed at-will** by the receiver (no gating). |
| **Schema Validation** | **Strict validation** is critical to ensure transaction integrity and avoid errors. | **Flexibility** is preferred to accommodate various formats; validation happens downstream. |
| **Multi-Platform Scale** | High value for **data normalization** and automated **privacy compliance** across diverse platforms. | **Low friction** for platforms to adopt; shifts the normalization burden to the receiver. |
| **Typical Examples** | `discount`, `fulfillment` | `attribution` (UTMs), referral tracking, session tags. |

I was weighing against the structured extension (PR #295) which may potentially offer advantages as the ecosystem scales to support multiple diverse platforms e.g It would enforce cross-platform data normalization (saving merchants from custom parsing), ensure reliable deduplication by strictly requiring specific keys, and facilitate automated privacy compliance. However, for data that is purely informational and does not drive protocol state or response negotiation, I believe the simplicity and low adoption barrier of the flat map is the right tradeoff to keep the core protocol light.

Let's move forward with this approach.

amithanda:
> > I am sure I've seen URLs with more than that, and I'd rather leave this as an implementer decision than encode it as a protocol-governance constraint.
> 
> Yeah after thinking on this a bit more, I agree with this. My original motivation around max was that we should have some rails on it due to URL length limits. However I feel that I agree with you it almost feels arbitrary at some point. I think I would rather leave off min/max so bit of a 180 from me. Leaving as an implementer decision feels like the right fidelity to me long term.

How about this middle ground:

1.  **Support `minProperties: 1` in the schema**: If a platform includes the `attribution` field in a request, it signals the intent to pass referral context. Allowing an empty object `{}` adds noise to the payload without providing any value. Enforcing `minProperties: 1` is a simple structural guardrail that ensures intent matches content—if the field is provided, it must carry data.

2.  **Keep `maxProperties` as guidance rather than a schema constraint**: I agree with @igrigorik and @westeezy that hardcoding a limit (like 20) in the schema is arbitrary and could break valid use cases where platforms send extensive tracking parameters. Instead of a rigid schema constraint, we should add a note in the documentation providing **guidance** (e.g., recommending that platforms aim to keep properties below a threshold for optimal performance), leaving any actual enforcement to the receiver's discretion.

This approach gives us a basic structural guardrail with `minProperties` while preserving the flexibility of the open map design by keeping `maxProperties` as guidance.

What do you think?

docs/specification/cart.md
```diff
@@ -220,6 +220,15 @@ requirements.
...
 {{ schema_fields('types/signals', 'checkout') }}
 
+### Attribution
+
+Platform-provided referral and conversion-event context — campaign IDs,
+click identifiers, and source/medium markers communicated by the platform.
+See [Attribution](overview.md#attribution) for details and consent
+requirements.
+
+{{ schema_fields('types/attribution', 'checkout') }}
```
amithanda, line 230: nit: I noticed this uses 'checkout' instead of 'cart', following the existing pattern for other `schema_fields` in this file. Ideally, this should be 'cart' to prevent issues if the message contracts diverge in the future or pointing to permalinks in checkout rather than cart. If we change it, then we should do it consistently across the file. Mostly a clean up task, feel free to clean it up now or I can send a clean up in a followup PR.

igrigorik: Good flag. We should address this, but it's a separate PR. Opened: https://github.com/Universal-Commerce-Protocol/ucp/issues/412

igrigorik:
Per TC discussion, added an authoring note for avoiding `minProperties: 1` in https://github.com/Universal-Commerce-Protocol/ucp/pull/391/changes/deb50665b48fc056d5eb071788d00e01ae3daf3a.

@jamesandersen @amithanda with that, is there anything else outstanding? Please do final pass and we can land this.

jamesandersen (approved):
Thanks @igrigorik for finding a better home for attribution data than the original #295 proposal.

amithanda (approved):
Great work! The addition of the attribution field is a great improvement for maintaining platform context without overcomplicating the core protocol.

PR #380, opened 2026-04-24
fix: misleading JSONPath example and inconsistent message.path descriptions

sakinaroufid:
## Summary

The `path` property on `message_error`, `message_warning`, and `message_info` is an optional RFC 9535 JSONPath pointer used to anchor a message to a specific component of the response payload. Its JSON Schema `description` is currently inconsistent across the three sibling schemas, and the example in `message_error.json` references a field that does not exist in the protocol.

## Findings

- **`message_info.json`** — description omits an example, giving schema consumers no pattern to follow.
- **`message_warning.json`** — phrases the field as "JSONPath (RFC 9535) to related field", diverging from the sibling wording.
- **`message_error.json`** — canonical phrasing, but the example points at `$.items[1]`. No top-level `items` collection is defined anywhere in `checkout.json`, `cart.json`, or `order.json`; the actual array is `line_items` (see also the JSONPath examples used in `discount.json` and `message_warning.json`). The example is therefore unresolvable against any valid response and is actively misleading.

## Change

Unify the three `description` strings on a single canonical form — `"RFC 9535 JSONPath to the component the message refers to (e.g., \$.line_items[0])."` — matching the example pattern already used elsewhere in the schema set.

## Test plan

- [x] JSON parses cleanly on all three files.
- [x] \`cspell\` pass on changed files (no new lexemes introduced).
- [x] \`@biomejs/biome lint\` pass.
- [x] Example JSONPath (\`\$.line_items[0]\`) resolves against the \`line_items\` array declared in \`checkout.json\`, \`cart.json\`, and \`order.json\`.

PR #397, opened 2026-04-30
docs: add message.path field conventions to Error Handling

sbeashwar:
The `path` field on error/warning/info messages is used in examples
throughout the spec but has no prose explaining how to construct it.

Changes to `checkout.md`:

- Introduces `path` in the Error Handling opening paragraph so readers
  know it exists before hitting the severity table, with an anchor link
  to the new subsection.
- Adds `#### The path Field` subsection covering RFC 9535 JSONPath,
  snake_case convention, filter expressions, specificity rule, and
  create-vs-update keying differences.
- Adds `path` to the `invalid_phone` entry in the error processing
  example (`$.buyer.contact.phone`).
- Updates the error processing algorithm to show how to use `path` to
  identify the specific field to fix, with fallback to `error.code`
  when `path` is absent.

Complements #380 (which fixes the schema descriptions on the
message_* types) with implementation guidance in the spec prose.

### DIFF
--- a/docs/specification/checkout.md
+++ b/docs/specification/checkout.md
@@ -119,8 +119,10 @@ platform receives messages indicating what's needed to progress.
 The `messages` array contains errors, warnings, and informational messages
 about the checkout state. `ucp.status` is the shape discriminator —
 `"success"` means the response carries the expected payload, `"error"`
-means it carries error information instead. The `severity` field on each
-error message prescribes the recommended action:
+means it carries error information instead. Each message carries a `type`,
+`code`, `severity`, `content`, and an optional `path` that identifies the
+specific field or line item the message refers to (see [The `path` Field](#the-path-field) below).
+The `severity` field prescribes the recommended platform action:
 
 | Severity                | Meaning                                          | Platform Action                                                   |
 | :---------------------- | :----------------------------------------------- | :---------------------------------------------------------------- |
@@ -182,6 +184,7 @@ Businesses **SHOULD** surface such messages as early as possible, and platforms
       "type": "error",
       "code": "invalid_phone",
       "severity": "recoverable",
+      "path": "$.buyer.phone_number",
       "content": "Phone number format is invalid"
     },
     {
@@ -219,7 +222,11 @@ IF unrecoverable is not empty
 
 IF recoverable is not empty
   FOR EACH error IN recoverable
-    ATTEMPT to fix error (e.g., reformat phone number)
+    IF error.path is present
+      IDENTIFY the field at error.path in the request payload
+      ATTEMPT to fix that field (e.g., reformat phone at $.buyer.phone_number)
+    ELSE
+      ATTEMPT generic fix based on error.code
   CALL Update Checkout
   RETURN and re-evaluate response
 
@@ -249,6 +256,54 @@ messages or deferring to checkout completion.
 
 Example: `out_of_stock` requires specific upfront UX, whereas
 `payment_required` can be handled generically at submission.
+
+#### The `path` Field
+
+The optional `path` field on a message anchors the error to a specific
+component of the response payload. Platforms use it to associate error
+messages with the input field or line item that caused them - for example,
+highlighting a specific buyer field in a form or flagging a specific
+cart line.
+
+`path` **MUST** be an [RFC 9535](https://www.rfc-editor.org/rfc/rfc9535)
+JSONPath expression relative to the root of the UCP response object.
+Property names **MUST** use snake_case matching the request schema.
+When `path` is omitted, the message applies to the response as a whole.
+
+**Simple field reference:**
+
+```json
+{ "path": "$.buyer.email" }
+```
+
+**Indexed array element:**
+
+```json
+{ "path": "$.line_items[0].quantity" }
+```
+
+**Filter expression (optional, when referencing a specific item by ID):**
+
+```json
+{ "path": "$.line_items[?(@.id=='line-item-uuid')].quantity" }
+```
+
+Filter expressions are valid RFC 9535 syntax and **MAY** be used when
+referencing a specific line item by `id` is clearer than its index.
+Index-based paths are equally valid; the business returns indices that
+are unambiguous within the response.
+
+**Specificity rule:** A path to a specific field (e.g.,
+`$.line_items[0].quantity`) takes precedence over a path to its parent
+(e.g., `$.line_items[0]`). When multiple errors apply to the same field,
+each message **SHOULD** carry the most specific path applicable.
+
+**Create vs. Update keying:** On `create_checkout`, line item filter
+expressions key on `item.id` (the catalog item identifier). On
+`update_checkout`, they key on the line item's own `id` (the
+session-scoped line item identifier assigned in the create response).
+
+When `path` is omitted, the message applies to the response as a whole.
 
 #### Eligibility Verification at Completion
 
--- a/docs/specification/checkout.md
+++ b/docs/specification/checkout.md
@@ -119,7 +119,7 @@ platform receives messages indicating what's needed to progress.
 The `messages` array contains errors, warnings, and informational messages
 about the checkout state. `ucp.status` is the shape discriminator —
 `"success"` means the response carries the expected payload, `"error"`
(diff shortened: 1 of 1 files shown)


sbeashwar:
@yanheChen @alexpark20 @sakinaroufid friendly bump on this - docs addition for the `message.path` field. CI green, and sakinaroufid (author of related #380) reacted positively. Could you take a look when you get a chance?

sakinaroufid:
> @yanheChen @alexpark20 @sakinaroufid friendly bump on this - docs addition for the `message.path` field. CI green, and sakinaroufid (author of related #380) reacted positively. Could you take a look when you get a chance?

+1

docs/specification/checkout.md
```diff
@@ -182,6 +184,7 @@ Businesses **SHOULD** surface such messages as early as possible, and platforms
       "type": "error",
       "code": "invalid_phone",
       "severity": "recoverable",
+      "path": "$.buyer.contact.phone",
```
jingyli: This is an invalid field in the schema, we should have `$.buyer.phone_number` instead. Similar comments to other references to `buyer.contact`.

docs/specification/checkout.md
```diff
@@ -250,6 +257,51 @@ messages or deferring to checkout completion.
...
+{ "path": "$.line_items[0].quantity" }
+```
+
+**Filter expression (preferred when referencing a specific item by ID):**
+
+```json
+{ "path": "$.line_items[?(@.id=='line-item-uuid')].quantity" }
+```
+
+Filter expressions are preferred over index-based paths when the platform
```
jingyli: Trying to understand the motivation here a bit more: `messages[]` are set by business in the response based on a given platform request. Even if array items are reordered by platforms between different requests, the indexed approach should achieve equal clarity since its scope is limited to that given request.

Acknowledge if we want to recommend a preference on filter expression, but this rationale is a bit unclear to me personally..

docs/specification/checkout.md
```diff
@@ -250,6 +257,51 @@ messages or deferring to checkout completion.
...
+may reorder line items between requests.
+
+**Specificity rule:** A path to a specific field (e.g.,
+`$.buyer.contact.email`) takes precedence over a path to its parent
+(e.g., `$.buyer.contact`). When multiple errors apply to the same field,
+each message **SHOULD** carry the most specific path applicable.
+
+**Create vs. Update keying:** On `create_checkout`, line item filter
```
jingyli: This sub-section is very specific to when `path` is using `filter expression` strategy from above instead of a general guidance for `path`. I would recommend moving it closer to that option above and make it as a sub-section if possible to make the connection more explicit.

jingyli (commented):
Thanks @sbeashwar for the doc enhancement! Added some comments, the biggest one is the usage of a non-existent field in UCP.

docs/specification/checkout.md
sbeashwar: Good catch, that was a mistake on my end - I wrote `buyer.contact.email/phone` from a generic mental model of contact-info structures rather than reading `buyer.json`. Updated all four occurrences to the actual schema (`$.buyer.email`, `$.buyer.phone_number`, and the specificity-rule example to `$.line_items[0]` parent vs `$.line_items[0].quantity` child).

docs/specification/checkout.md
sbeashwar: Agree the "preferred" framing was overreach on my part. The filter syntax is valid RFC 9535 but the spec doesn't require it, and as you point out the business returns indices that are unambiguous within a given request/response. Updated to present filter expressions as optional (MAY), with index-based paths called out as equally valid.

docs/specification/checkout.md
sbeashwar: Agreed - moved "When `path` is omitted, the message applies to the response as a whole" up next to the JSONPath/snake_case rule so it reads as general guidance rather than a trailing note.

sbeashwar:
@jingyli ping on this - I addressed your three comments (schema fix to `buyer.email`/`buyer.phone_number`, dropped the "preferred" framing for filter expressions, restructured so the omission rule sits with the general guidance). Could you re-review when you have a moment?

@yanheChen @alexpark20 still open if either of you wants to weigh in.

docs/specification/checkout.md
```diff
@@ -119,8 +119,10 @@ platform receives messages indicating what's needed to progress.
 The `messages` array contains errors, warnings, and informational messages
 about the checkout state. `ucp.status` is the shape discriminator —
 `"success"` means the response carries the expected payload, `"error"`
-means it carries error information instead. The `severity` field on each
-error message prescribes the recommended action:
+means it carries error information instead. Each message carries a `type`,
```
jingyli: > Each message carries a `type`,
`code`, `severity`, `content`, and an optional `path` that identifies the
specific field or line item the message refers to

nit: I would suggest adding back the `error` discriminator to the beginning of this sentence (i.e. "Each **error** message carries...") since `severity` does not exist for warning & informational messages.

docs/specification/checkout.md
```diff
@@ -250,6 +257,54 @@ messages or deferring to checkout completion.
...
+`$.line_items[0].quantity`) takes precedence over a path to its parent
+(e.g., `$.line_items[0]`). When multiple errors apply to the same field,
+each message **SHOULD** carry the most specific path applicable.
+
+**Create vs. Update keying:** On `create_checkout`, line item filter
+expressions key on `item.id` (the catalog item identifier). On
+`update_checkout`, they key on the line item's own `id` (the
+session-scoped line item identifier assigned in the create response).
+
+When `path` is omitted, the message applies to the response as a whole.
```
jingyli: This sentence can be removed as it is verbatim to line 271 above.

docs/specification/checkout.md
```diff
@@ -250,6 +257,54 @@ messages or deferring to checkout completion.
...
+Index-based paths are equally valid; the business returns indices that
+are unambiguous within the response.
+
+**Specificity rule:** A path to a specific field (e.g.,
+`$.line_items[0].quantity`) takes precedence over a path to its parent
+(e.g., `$.line_items[0]`). When multiple errors apply to the same field,
+each message **SHOULD** carry the most specific path applicable.
+
+**Create vs. Update keying:** On `create_checkout`, line item filter
+expressions key on `item.id` (the catalog item identifier). On
```
jingyli: Rereading this again, I'm curious why is this needed given `messages` (and indirectly `path`) is a response-only field - by the time business responds to `create_checkout`, `line_item.id` must be present so we should be able to reuse it for keying purpose right?

It almost feels like this section can be removed since there won't be a difference between `create` and `update`..

jingyli (changes requested):
@sbeashwar Apologies for the delay - this is very close now with 3 more smaller comments from me. We should be good once they are addressed and the file conflict is addressed. Thanks!

docs/specification/checkout.md
sbeashwar: Good catch - severity is error-only (required on message_error, absent from warning/info). Changed to qualify the sentence as 'Each error message carries...'.

docs/specification/checkout.md
sbeashwar: Removed - that was a leftover duplicate from moving the sentence up in the earlier round.

docs/specification/checkout.md
sbeashwar: Agreed. Since messages/path is response-only and line_item.id is always present in the response, keying always uses the line item id regardless of create vs update - the distinction never surfaces. Removed the section.

damaz91 (approved):

jingyli:
Thanks! One last nit before approving: can we add some fixes to ensure our workflow can succeed (i.e. add `<!-- ucp:example skip reason="path schema example" -->` right above where we are using `json` annotation)?

sbeashwar:
@jingyli done - added `<!-- ucp:example skip reason="path schema example" -->` above each of the three `path` snippet blocks so the example-validation workflow skips them (they're illustrative fragments, not full response objects). Also rebased on latest main. lint is green now. Thanks for the reviews!

jingyli (approved):
Thanks! LGTM.

PR #398, opened 2026-04-30
feat: add return_url for post-handoff buyer return

sioked:
Ref #486 

## Summary

- Adds `return_url` as an optional request field (`create`/`update`) on checkout sessions
- Provides a platform-specified destination for the buyer after completing or exiting a hosted checkout UI reached via `continue_url`
- The redirecting party appends `?status=completed` or `?status=canceled` as a routing hint; platforms should verify actual session state independently via the checkout session API
- Includes open redirect security guidance recommending businesses validate `return_url` against known platform origins before redirecting

## Motivation

`continue_url` handles the outbound handoff — business sends buyer to hosted UI. `return_url` closes the loop — business sends buyer back to the platform after the hosted experience ends. Without it, there is no standard mechanism for a platform to receive the buyer back after a hosted checkout or payment handler redirect.

## Category

## Checklist

### DIFF
--- a/docs/specification/checkout-mcp.md
+++ b/docs/specification/checkout-mcp.md
@@ -181,6 +181,7 @@ Maps to the [Create Checkout](checkout.md#create-checkout) operation.
               }
             ],
             "currency": "USD",
+            "return_url": "https://platform.example.com/checkout/return",
             "fulfillment": {
               "methods": [
                 {
@@ -587,6 +588,24 @@ Maps to the [Update Checkout](checkout.md#update-checkout) operation.
       }
     }
     ```
+
+#### Handoff and Return Flow
+
+When a business responds with `continue_url`, the platform directs the buyer to
+that URL to complete the hosted checkout UI. If the platform provided a
+`return_url` in the `create_checkout` request, the business **SHOULD** redirect
+the buyer back to the platform when the hosted UI exits:
+
+```text
+# Buyer completes hosted checkout
+GET https://platform.example.com/checkout/return?status=completed
+
+# Buyer cancels or checkout cannot complete
+GET https://platform.example.com/checkout/return?status=canceled
+```
+
+Platforms **SHOULD** verify the actual checkout session state via `get_checkout`
+after receiving the buyer back — `status` is a routing hint only.
 
 ### `complete_checkout`
 
--- a/docs/specification/checkout-rest.md
+++ b/docs/specification/checkout-rest.md
@@ -70,7 +70,8 @@ All REST endpoints **MUST** be served over HTTPS with minimum TLS version
           },
           "quantity": 2
         }
-      ]
+      ],
+      "return_url": "https://platform.example.com/checkout/return"
     }
     ```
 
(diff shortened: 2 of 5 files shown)


jingyli:
Thanks @sioked for raising the PR! Are there any linked issues/discussion threads that describe this issue a bit more? Personally I'm having some hard time gauging whether this concept of `return_url` is truly needed due to:

1) **From business/user experience PoV**: Through handoff via `continue_url`, the buyer is able to continue their actions on business site to either a success (e.g., placing the transaction, finalizing cart, etc.) or failure (e.g., cart/checkout abandonment, rejected transaction upon placement). This would mean they should've already seen some terminal screen, like an order confirmation screen, on business side. What would be the expectation for platform to render when `return_url` is loaded (loading a duplicated error/confirmation screen on platform side feels odd)? What would be the incentive for businesses to redirect this "concluded" session back to the platform?
2) **From platform/data PoV**: Buyer handoff via `continue_url` can occur at various points in the purchase journey - in the most extreme case, it can occur at `create_checkout` when an [error_response](https://ucp.dev/latest/specification/checkout/#error-handling) is received (meaning there are no checkout resource attached). From platform perspective, there is no way to learn about the complete data expectation from business when `return_url` is initiated. Calling `get_checkout` may be one way to get this data snapshot, but it would be infeasible for the extreme case mentioned earlier. Tying back to 1), I'm unsure what will be the expectation for platforms to display as part of the UI that renders via `return_url`.

Also a general callout - `continue_url` is a very generic concept that is used by error handling, cart & checkout capabilities today whereas `return_url` is very checkout specific. Is this intended? Is not, how would it work with the other scenarios?

damaz91:
@sioked do you mind taking a look at @jingyli comment? 

If we don't hear from you in the next few days, we'll close this out. Thanks!

damaz91:
Closing this out for inactivity. @sioked please feel free to reopen!

sioked:
Thanks - was on a sabbatical for a little while and getting back to this now. I think it's ok to keep this closed for right now but I would like to keep the separate issue (#486) open so we can continue discussion - as @jingyli points out the continue_url is a very general concept that is used for a variety of different usecases, but my focus has specifically been around the payments aspect of checkout. In that use case a platform may need to redirect to a payment provider surface for approval - ie 3DS or a Wallet payment - but ultimately with the expectation that the user would come back to the platform for completing the checkout session. That provides a general separation between the overall checkout and the individual components (payments).

Regarding the points above:
> 1.  This would mean they should've already seen some terminal screen, like an order confirmation screen, on business side

This assumes that the terminal screen does not happen on the business side- instead the redirect goes to a separate 3rd party (3DS/Wallet) and the completion occurs directly on the platform. The incentive to return to the platform is on the wallet provider or 3ds provider and not on the merchant directly.

2. I think there is additional complexity for other usecases, but in general the goal would be to support continuation & completion directly on the platform instead of completing the payment directly on the business site

I'll monitor the issue for additional feedback but will be trying to implement these capabilities either via payment handlers or an extension so it doesn't need to be part of the core checkout capability.

Issue #400, opened 2026-04-30
[Proposal]: Structured product attributes via schema.org PropertyValue


sakinaroufid:
# Summary

Add an optional `attributes` array on `Product` and `Variant` for measurable facts like battery life, display size, weight, and storage. Each entry has a required `name` and `value`, plus optional `numeric_value` and `unit`. Same inline shape as the existing `Variant.barcodes`. About 30 lines of schema across two files. Additive: no new endpoints, transports, or request parameters.

The shape matches schema.org `PropertyValue`, which is what most catalog teams already publish via JSON-LD on their product pages. For them, adoption is mostly piping existing data into UCP catalog responses.

# Motivation

An agent gets the query "phone with at least 20 hours of battery, under $800." The catalog response usually contains the answer but it's stuck in prose: "all-day battery", "lightweight at 1.2 kg".

`tags` doesn't carry units. `metadata` accepts anything, but every merchant picks a different shape, which is the same fragmentation problem as parsing prose.

`attributes` gives a structured place for measurable facts. `numeric_value` lets agents filter without parsing display strings.

This isn't a vendor-specific use case. Every catalog has the same gap, and a vendor-namespaced version would be the same field with a different prefix, which doesn't help the cross-merchant comparison problem.

# Goals

* Get measurable attributes out of prose into structured form on `Product` and `Variant`.
* Expose a typed numeric field so agents can filter without LLM extraction.
* Stay additive. No breaking changes, no new endpoints, no new transports.

# Non-Goals

* Defining a UCP-canonical vocabulary of attribute names. Names follow merchant convention and schema.org practice. Vertical extensions can standardize identifiers later.
* Server-side filtering. Filtering happens agent-side over the catalog response.
* Localization of attribute names. Display strings come from the rendering layer.
* Range expressions. v1 is single values. Range siblings can land later.
* Categorical attributes (color, fit, material). Those go in `tags` or product `options`.

# Detailed Design

## Schema

```json
"attributes": {
  "type": "array",
  "description": "Structured product attributes for measurable facts (battery life, display size, weight). Use `tags` for categorical labels and `metadata` for business-specific data. Same shape as schema.org `PropertyValue`.",
  "items": {
    "type": "object",
    "required": ["name", "value"],
    "properties": {
      "name": {
        "type": "string",
        "description": "Human-readable attribute name (e.g. `Battery life`, `Display`, `Weight`). Same as schema.org `PropertyValue.name`."
      },
      "value": {
[...]

PR #401, opened 2026-05-01
feat: add structured product attributes via schema.org PropertyValue

sakinaroufid:
# Description

Adds an optional `attributes` array on `Product` and `Variant` for measurable facts like battery life, display size, weight, and storage. Shared type at `source/schemas/shopping/types/attribute.json`, referenced from both schemas. Same shape as schema.org PropertyValue.

Each entry has required `name` and `value` plus optional `key`, `numeric_value`, and `unit`. `numeric_value` lets agents filter without parsing display strings. `unit` lets them compare across merchants. `key` gives verticals a stable slot for standardized identifiers without UCP shipping a canonical vocabulary on day one.

Catalog spec doc gains an Attribute section: placement decision tree (`attributes` vs `options` vs `tags` vs `metadata`), Product/Variant per-name override semantics, SHOULD-list of well-known units, worked example.

Additive. No new endpoints, transports, request shapes, or capability negotiation. Existing producers and consumers are unaffected.

## Why

Agents handling queries like "phone with at least 20 hours of battery, under $800" need measurable facts in structured form. The catalog response usually contains the answer but it's stuck in prose: "all-day battery", "lightweight at 1.2 kg".

`tags` doesn't carry units. `metadata` accepts anything, so every merchant picks a different shape, which is the same fragmentation problem as parsing prose. `options` is for variant-defining axes (Color, Size), not measurable facts. None of them is a structured place for "Battery life: 22 hours."

The shape matches schema.org PropertyValue, which is what most catalog teams already publish via JSON-LD on their product pages. For them, adoption is mostly piping existing data into UCP responses.

## Design notes

**Shared type, not inlined.** `attributes` lives on both `Product` and `Variant`, which matches how `category.json`, `media.json`, and `rating.json` are referenced from both. Avoids drift between two parallel definitions. `Variant.barcodes` stays inlined because barcodes are variant-only.

**`key` in addition to `name`.** Schema.org PropertyValue has both `name` (human-readable) and `propertyID` (stable identifier). Using only `name` makes cross-merchant filtering brittle: "Battery life" vs "battery_life" vs "Battery". The optional `key` field is the `propertyID` analog. Producers can populate it now, verticals can standardize values later, consumers that only care about `name` are unaffected.

**Per-name override, not whole-array replace.** When the same attribute appears on both `Product` and `Variant`, Platforms MUST match by `key` (if present) or case-insensitive `name`, override the Product entry with the Variant entry, and inherit any Product entries the Variant doesn't redeclare.

**Open vocabulary on `unit`, with a SHOULD-list.** Per the extensibility guidance: no enum. The schema description carries a SHOULD-list of well-known units (SI/UCUM where they exist: `kg`, `mm`, `Hz`; common catalog units otherwise: `GB`, `inches`, `hours`) plus an `examples` array for codegen affinity. Same pattern as `barcodes.type`.

## Out of scope for v1

* A UCP-canonical vocabulary of `name` or `key` values. Vertical extensions can standardize identifiers later.
* Server-side filtering on attributes. Filtering happens agent-side over the catalog response.
* Range expressions. v1 is single values.
* Localization of attribute names.

## Category (Required)

Schema additions to `Product` and `Variant` qualify as Core Schema Modifications under the Contributing Guide, so this PR follows the Enhancement Proposal track. The proposal lives at #400.

## Related Issues

For #400. Currently in Proposal stage; this PR is the implementation reference and should not merge until the EP reaches Provisional via TC vote.

## Checklist

- [ ] I have regenerated Python Pydantic models by running `generate_models.sh` under `python_sdk`. The Pydantic regeneration lives in the `python_sdk` repo and will land there as a follow-up once these schemas merge.

Local checks against the configs in `.github/`:

* `ucp-schema lint source/`: 80 files checked, all passed.
* `DOCS_MODE=spec uv run mkdocs build --strict`: built in 9.66s, no warnings or errors on the new section.
* markdownlint with `.github/linters/.markdownlint.json` on the changed doc: 0 issues.
* cspell with `.cspell.json` on changed files: 4 files, 0 issues.

## Test plan

* [ ] Schema validation, positive: required `name`/`value` enforced; each combination of optional fields validates; existing fixtures without `attributes` validate unchanged.
* [ ] Schema validation, negative: missing `name` rejected; missing `value` rejected; `numeric_value` as string rejected.
* [ ] Conformance fixture: worked example (phone with product- and variant-level attributes).
* [ ] Conformance fixture: display-only case (`numeric_value` absent, only `value` populated).
* [ ] Conformance fixture: override case (same `key` on Product and Variant, assert per-name merge).


BuyWhere:
This is a welcome addition to the UCP spec. Structured product attributes via schema.org PropertyValue make it much easier for AI agents to do meaningful product comparison across merchants — an agent can filter on “processor generation” or “screen size” directly rather than parsing free-text descriptions.

We are doing something similar on the BuyWhere side — normalizing product data from Shopee, Lazada, and other marketplaces into a structured schema with typed attributes before surfacing it through our MCP server. Having UCP standardize this at the protocol level would save a lot of per-source mapping work.

One question: does this cover nested/complex attributes? For example, a laptop has CPU (with its own properties like cores, clock speed), RAM (type, size), and display (resolution, panel type).

BuyWhere:
Great to see progress on structured product attributes. At BuyWhere (buywhere.ai), we have been working on normalized product catalogs across multiple merchants and regions. We use schema.org as a base but extend with commerce-specific properties (pricing tiers, availability windows, multi-currency). Happy to share our category taxonomy if it helps inform the UCP spec.

sakinaroufid:
Hi @ptiper - would you mind pls tagging the right reviewers?

dylanfarrell:
This looks great. `attributes` feels like the right primitive for measurable product facts, and aligning with `PropertyValue` should make adoption easier for merchants already publishing JSON-LD.

A couple of thoughts:

- `PropertyValue` uses fields like `propertyID`, `unitCode`, and `unitText`, while this PR uses `key`, `unit`, `value`, and `numeric_value`. The simplified naming here is reasonable, but it may require JSON-LD producers to introduce a mapping layer. Would it be worth aligning the field names more closely with schema.org conventions (`property_id`, `unit_code`, `unit_text`)? Or if the current names are preferred, would it help to include a short mapping table in the spec?
- If a Product or Variant contains multiple attributes with the same `key`, how should consumers interpret that? Should it be considered invalid, treated as a multi-valued attribute, or resolved as last-one-wins?

PR #409, opened 2026-05-02
feat: Introduce the split payments extension

raginpirate:
# Description

Adds the `dev.ucp.shopping.split_payments` extension, enabling buyers to pay with multiple payment instruments in a single checkout. Businesses declare which instrument combinations they support via `allowed_combinations` (a constraint language built on instrument groups with type/min/max). Instruments are submitted in allocation priority order, with each contribution either platform-specified (`amount` set) or business-determined (`amount` omitted, e.g. gift card balance).

This unlocks the development of redeemable payment instruments, which are functionally incomplete without the support of split payments to cover their remaining balance.

## Category (Required)

## Checklist

^ WIP for the python_sdk commit.


docs/specification/split-payments.md
```diff
@@ -0,0 +1,348 @@
...
+
+* **Specified-amount** (`amount` present): the platform requests a
+  specific contribution, in ISO 4217 minor units.
+* **Open-amount** (`amount` omitted): the business determines the
+  instrument's contribution at processing time (e.g., by querying a
+  gift card's available balance).
+
+Instruments are submitted in **allocation priority order** — the first
+instrument gets first claim on the checkout total, the second gets next
+claim, and so on. The business MAY process payment authorizations in
```
ihoosain, line 58: nice this statement and the open amount works really well with how etsy processes giftcard balances in a split payment scenario!

docs/specification/split-payments.md
```diff
@@ -0,0 +1,348 @@
...
+instrument in a single checkout. Businesses declare the instrument
+combinations they support in `allowed_combinations`.
+
+Each instrument is submitted in one of two modes:
+
+* **Specified-amount** (`amount` present): the platform requests a
+  specific contribution, in ISO 4217 minor units.
+* **Open-amount** (`amount` omitted): the business determines the
+  instrument's contribution at processing time (e.g., by querying a
+  gift card's available balance).
```
drewolson-google, line 43: This seems like a reasonable starting point. Do you think in the future we'd want to be able to support "strategies"? For example, the platform may want to specify something like "split evenly" without knowing the example dollar amounts. This may be particularly useful in cases where AP2 is used. I could see wanting to define a splitting strategy at the time of cart mandate, where the full price of future purchases may not yet be known.

docs/specification/split-payments.md
```diff
@@ -0,0 +1,348 @@
...
+
+#### `allowed_combinations`
+
+An array of valid instrument combinations. Each combination is an array of
+**instrument groups** -- constraints that together define one valid way to
+split a payment.
+
+A set of instruments is valid if it matches **any** combination in the array.
+
+#### Instrument Group
```
drewolson-google, line 88: I'm struggling a bit to understand the need to be comprehensive with group definitions up-front. I worry you'll end up in one of two worlds:

1. You attempt to be very specific with your configuration, which runs into an `N x M` problem with many combinations.
2. You just have a single group with a min of 0 and a max of "some sufficiently large N" for each instrument type, effectively a free-for-all.

Might be worth talking through this tension live at TC.

drewolson-google (commented):
A very straight-forward proposal, thanks @raginpirate. I left a few comments below, but the area I'm most unsure of is pre-defining of instrument groups.

drewolson-google (commented):
A very straight-forward proposal, thanks @raginpirate. I left a few comments below, but the area I'm most unsure of is pre-defining of instrument groups.

docs/specification/split-payments.md
raginpirate: Thanks for the quick review Drew! Here's my POV:
1. Up-front declaration is a requirement of the extension.

The dominant real-world pattern is narrow: "I support splitting a credit card with redeemables, and nothing else." That's a specific, declarative capability, and the platform needs it explicitly so it can render the right experiences. If a business doesn't declare combinations, the platform has to guess. The question isn't whether to declare, but how expressively.

2. We shouldn't be dealing with an NxM problem for standard integrations, because of how the groups compose.

A group's types field is OR'd: one group can accept any of N instrument types. That collapses most of what would be N×M into a single combination. For example, "one primary instrument + up to 2 redeemables" with 4 primaries and 3 redeemable types is a single combination, not 12:

 ```
 [
    { "types": ["card", "paypal", "apple_pay", "google_pay"], "min": 1, "max": 1 },
    { "types": ["gift_card", "store_credit", "loyalty"], "max": 2 }
  ]
```
Did you have a perspective of standard use cases this fails to keep concise?

Let me know if this was helpful! I'd be more than happy to chat live during a TC; sadly I will have to miss the one this Friday I think but we can always review during the next if this is not resolved async!

**Comment by prasad-stripe** [comment]
Great work on this, @raginpirate. The capability-gated pattern is clean, allowed_combinations is a solid primitive, and the open-amount model maps naturally to how redeemables work in practice (as @ihoosain noted, Etsy's gift card flow fits this well). A few thoughts on scope and failure handling that I think would make this more robust:

**Scoping to redeemables + fallback**
The stated motivation is unlocking redeemable instruments, and every example in the spec follows the same shape: one or more redeemables (gift card, loyalty, store credit) with an open-loop fallback (card). In my experience, this is also the only multi-instrument pattern that sees real production traffic in online commerce. Arbitrary combinations (two credit cards, card + ACH) introduce a different class of problems: independent dispute timelines, partial authorization handling, cross-network retry semantics, etc.

Would it make sense to explicitly scope this version to redeemable + open-loop fallback? In my experience, the natural processing order is open-loop first: you authorize the card (the uncertain, externally-controlled leg), and only then debit the redeemable (which the business platform owns and can guarantee). If the card fails, nothing has been touched. This avoids the reversal problem entirely for the common case, keeps the complexity bounded, and gives us room to tackle the harder multi-network scenarios in a follow-up once we have answers for partial auth and dispute routing.

**Relaxing the atomicity requirement**
The current MUST ("void or reverse all previously successful authorizations and return payment_failed") concerns me. A few reasons:

1. It conflicts with UCP's own principle that the business is MoR and authoritative over payment processing. The protocol is prescribing an operational policy where it should be defining a communication contract.
2. In practice, when a second instrument soft-declines, the business may want to retry (with backoff) or escalate it to the buyer for an alternative. Mandating immediate void turns recoverable situations into hard failures. Worse, the void itself can fail or hit acquirer rate limits, creating a cascading problem.
4. For the redeemable case specifically, if a gift card has been debited and the card auth fails, the merchant might prefer to hold the transaction open and let the buyer provide a different card, rather than refund the gift card and lose the sale.

IMO the protocol should require the business to report per-instrument status (succeeded, failed, pending) so the platform has visibility, but leave the failure policy to the business. A SHOULD default of void-all is reasonable, but a MUST removes agency from the party best positioned to make that call. Something like a failure_policy field in the business config ("void_all", "partial_allowed", "escalate") would let businesses declare their behavior without the protocol dictating it.

**Processing order as a spec concern**
The current spec says instruments are submitted in "allocation priority order" and the business MAY process in any order. I'd suggest the business platform should make the decision about the ordering, they are best positioned to make the call and remove the complexity of the "allocation priority order". This could be addressed in a fast follow-up by encoding in a split-strategy as suggested by @drewolson-google. 

Excited to see this move forward. Happy to discuss further

**Comment by raginpirate** [comment]
Thanks as well for your quick feedback @prasad-stripe! Great thoughts, let me clarify and I also just pushed up a fix based on your feedback ❤️ 

> Scoping to redeemables + fallback

> Arbitrary combinations (two credit cards, card + ACH) introduce a different class of problems: independent dispute timelines, partial authorization handling, cross-network retry semantics, etc.

I think the advertised capability enables a business who cannot abstract these concerns from the platform to not support anything but redeemables. I also am of the opinion that all of these concerns _**can**_ be abstracted, and is only a concern to empower businesses who wants to do some potentially smarter things, some of which you touch on in your next paragraph...

> The current MUST ("void or reverse all previously successful authorizations and return payment_failed") concerns me.

Lets dig into this! 

> It conflicts with UCP's own principle that the business is MoR and authoritative over payment processing. The protocol is prescribing an operational policy where it should be defining a communication contract.
> In practice, when a second instrument soft-declines, the business may want to retry (with backoff) or escalate it to the buyer for an alternative. Mandating immediate void turns recoverable situations into hard failures. Worse, the void itself can fail or hit acquirer rate limits, creating a cascading problem.
> For the redeemable case specifically, if a gift card has been debited and the card auth fails, the merchant might prefer to hold the transaction open and let the buyer provide a different card, rather than refund the gift card and lose the sale.


**_I agree theres tension here... Thank you for the pushback!_**

I think this is a disconnect between my intention vs. how I wrote this. Although I believe we should dictate this as a MUST, it is not a MUST IN REALTIME; the effect after a batch of UCP checkout submissions needs to behave this way, but it does not have to happen immediately. Let me take a stab at improving that phrasing!

With that, I think the above perspective might change? Hopefully we can see we're not mandating the payment processing layer, but mandating the side-effect of abandoned charges is NOT ok. And that businesses SHOULD have the flexibility to do any smart caching they need to 😄

> IMO the protocol should require the business to report per-instrument status (succeeded, failed, pending) so the platform has visibility, but leave the failure policy to the business.

Great thinking, I definitely pondered this as well!
Right now, the spec mentions pushing error codes which help convey which instrument was problematic, and a successful instrument; is it actually important to use anything other than messages to convey this to the platform as well? It will be surfaced back with a fixed amount ready for the next submission, and if the buyer were to now remove it and pick a different instrument to cover the checkout that seems like acceptable commerce practice.
With that said, I'm not positive this makes sense to add to UCP, but if it does it might make sense as an improvement beyond the scope of this PR; what do you think?

> remove the complexity of the "allocation priority order"

This feature is pretty key to the spec though; when you submit N instruments without an amount as a platform, the order you submit them in determines how much they would be charged. We'd go from allowing a customer at our register to say "charge gift card A, then B, then my card for the rest", to only say "charge these (gift card A, B, and card), I don't care how".

_**Greatly appreciate your input here**_ and PTAL at my change to the MUST; does this move the needle for you or lmk some examples this is still problematic for!

File: docs/specification/split-payments.md (hunk 2 above)
Comment (raginpirate) [reply docs/specification/split-payments.md:43]: Interesting prompt @drewolson-google! I'm not positive about these different kinds of split strategies... but you've made me recognize an interesting limitation here.
If as a platform, I want to charge a gift card for whatever is on it, then split 50/50 the remainder between two cards, we can't.
I think the right solution might be more of a tree structure to submission, where we can indicate the root and a division of the remaining total amongst N nodes (which could also branch). But again, these scenarios seem a bit fictitious to me right now... but I'm open to them!

**Comment by prasad-stripe** [comment]
Thanks @raginpirate, really appreciate the thoughtful response and the quick update. The eventual-consistency framing lands well for me. "Buyer MUST NOT remain charged for an incomplete split after checkout" is the right invariant, and allowing async reversal with retry/rate-limiter flexibility is exactly what I was hoping for.

I'm aligned on allocation priority as buyer intent (charge A first, then B, then card for the rest) being distinct from processing order. That separation makes sense.

One thing I'd like to push on: the re-submit flow. Your updated language mentions "wait and see if the buyer re-submits partially captured instruments," which is great, but from the platform's perspective there's no structured signal to act on. Today the only response is payment_failed in messages, which doesn't distinguish between:

  - "Failed permanently, I'm cleaning everything up"
  - "One leg failed, I'm holding the successful auth, send me a replacement instrument"

For the second case, the platform (or agent) needs to know: which instruments succeeded, which failed, and what amount is still outstanding. Without that, the agent can't prompt the buyer with "your card was declined, you still owe $45, want to try another?"

My suggestion: on a partial failure where the business is holding the checkout open, return per-instrument status (succeeded/failed/pending) and an amount_pending field indicating what's left to cover. This gives the platform everything it needs to drive the recovery flow without the protocol dictating how the business handles it internally. And importantly, the consumer should be able to cancel the open checkout at any point, which would trigger reversal of any instruments that already succeeded in earlier stages.

I think there's real value in addressing this alongside this PR since split payments without a structured recovery path will push every implementer to invent their own. But if you'd rather keep this PR focused and track it as an immediate follow-up, I'm OK with that too as long as it's sequenced before this extension moves beyond working draft.

Thanks again for the collaboration here, this is shaping up nicely.

**Comment by jamesandersen** [comment]
Great discussion on the recovery flow, @prasad-stripe and @raginpirate. I think there's a low-complexity path to structured per-instrument status that doesn't require new schema fields.

+1 on the importance of platforms being able to convey per-instrument status to buyers — without it, the platform can't guide the user on what specifically needs to change to complete the transaction.

The existing message schemas already have a `path` field (RFC 9535 JSONPath). If the spec required the business to return **one message per instrument** on partial failure — `payment_failed` errors for failed instruments and `info` messages for successfully processed ones — the platform gets everything it needs to drive recovery:

```json
{
  "status": "incomplete",
  "payment": {
    "instruments": [
      { "id": "pi_gc_1", "type": "gift_card", "amount": 1000, "..." : "..." },
      { "id": "pi_card_1", "type": "card", "..." : "..." }
    ]
  },
  "messages": [
    {
      "type": "info",
      "path": "$.payment.instruments[0]",
      "content": "Gift card authorized for $10.00"
    },
    {
      "type": "error",
      "code": "payment_failed",
      "path": "$.payment.instruments[1]",
      "content": "Card declined — insufficient funds",
      "severity": "recoverable"
    }
  ]
}
```

The platform can reconstruct the full picture from what's already in the response:
- Instruments with an `info` message → succeeded (`amount` reflects actual charge)
- Instruments with a `payment_failed` error → failed, with `severity` signaling whether recovery is possible
- Outstanding amount = `total - sum(succeeded amounts)`

This is purely a spec-language clarification — no schema changes needed. It also pairs well with the updated eventual-consistency language: the business holds the checkout open, returns `severity: recoverable` on the failed leg, and the platform knows to prompt the buyer for a replacement rather than showing a terminal error.

docs/specification/split-payments.md
jamesandersen: @drewolson-google interesting thought about a split strategy; that total amount not yet known use case feels like it could become more tangible in a future of human-not-present checkout scenarios limited by agentic payment token guardrails and/or AP2.   

I feel like the PR has the most common production use case scenarios covered.  Just as a quick thought exercise to pressure test that this *could* extend to cover more interesting split strategies, a future PR might consider updating the `amount` field to support a few types of values e.g. fixed amount (ISO 4217), percentage, etc. almost like a [mixed unit grid layout](https://mozilladevelopers.github.io/playground/css-grid/05-mixing-units/) - pardon the crude analogy ;-)  So it feels like this is a good foundation - no one-way door

drewolson-google: @jamesandersen good points. I agree this covers the most common cases and I agree there are ways to extend it in the future for more complicated use cases.

docs/specification/split-payments.md
drewolson-google: I don't have a specific example other than worrying this will mostly devolve into people having a single "everything" entry in the most common cases.

One thought -- would it be easier to model this as a deny-list rather than an allow-list?

Totally open to discussion here, async is fine for sure.

raginpirate:
That’s exactly how I viewed it as well @jamesandersen! I think messages and paths work perfectly and I should have had more intentional examples of how to form those messages in the proposal.

Let me see what I can push to the PR later today and I’ll tag folks again when I do! 😄

docs/specification/split-payments.md
raginpirate: Greatly appreciate the push here Drew!

I think the deny-list pattern could be worse ergonomics for the ecosystem as businesses easily add new method types and fail to introduce a denied_combinations that should have existed, instead of properly exposing their method for different kind of applicable payment splits. This leads to poor UX as splits repeatedly get rejected only at complete time. I'll also note that we have no example capability leaning on a deny vs. allow list pattern, and that the combinatorial complexity of the solution does not change regardless of if we pick deny vs. allow.

At one point while writing this I did allow businesses to omit the allowed_combinations, which defaults to "everything works" if merchants wanted it; it was removed because it seems like a dangerous tool (like the pushback above), and somewhat impractical for a first deliverable (who in commerce supports splitting everything by anything?).

raginpirate:
Just pushed an update covering the open feedback from this round:

  For @jamesandersen & @prasad-stripe  — implemented the per-instrument status approach clarified above, pulling inspiration from the discounts capability! The Error Handling section now requires a payment_failed error per failed instrument with path pointing at the instrument, and
  documents info messages as a recommended pattern for succeeded instruments to convey positive context. Added a "Partial Failure with Recovery" example end-to-end so implementers can see the full shape. @prasad-stripe lmk do you still have open concerns here about how we convey this context to platforms!
  For @drewolson-google — hoping the deny-list discussion above and the strategies thread both feel resolved. Let me know if anything is still open on your end!

  Thanks all for the thoughtful review — this is much sharper than what I originally opened!

docs/specification/split-payments.md
drewolson-google: Thanks for the reply. I think we can take a "wait and see" approach on omitting `allowed_combinations` -- if a lot of folks are effectively doing it then we can make it easier. If not, no need.

drewolson-google (approved):

jamesandersen (approved):

igrigorik (commented):
@raginpirate with pedantic spec-authoring hat on: https://github.com/Universal-Commerce-Protocol/ucp/pull/409/changes/f97dca9451e9ef460060a59346e322f15713cd8b - please sanity check.  

---

Also, a few questions before we land this...

1/ **`amount` asymmetry:** schema has `amount` as single optional; prose says required-on-response. I think we can split into `requested_amount` (request) + `amount` (response, `ucp_request: "omit"`). Am I holding it right?

2/ **No `platform_schema` defined**: this is expected — correct?

> **Instrument Cardinality:** A checkout submission **MUST** contain exactly one
payment instrument unless the `dev.ucp.shopping.split_payments` capability is
active. Businesses **MUST** reject submissions that violate this constraint with
a `payment_failed` error in `messages[]`.

3/ **^ this is a validation error caught before the business attempts any authorization.** Should this be a distinct code (e.g. `payment_invalid`) to help differentiate instead of overloading?

4/ **Matching algorithm ambiguous for overlapping group types.** Perhaps...

```diff
   -**Matching algorithm:** for a given combination, each submitted instrument must
   -be assignable to exactly one group whose `types` list includes that
   -instrument's type. After assignment, every group must have between `min` and
   -`max` instruments (inclusive). If all constraints are satisfied, the
   -combination matches.
   +**Matching algorithm:** a submission matches a combination if there exists
   +an assignment of each submitted instrument to exactly one group such that
   +(a) the instrument's `type` is in the assigned group's `types`, and
   +(b) every group's instrument count falls within its `min` and `max`
   +(inclusive). The business MAY use any algorithm to find such an assignment;
   +if any valid assignment exists, the combination matches.
```

**5/ Any gotchas around multi-currency?** What if a gift card is denominated in CAD vs a USD checkout? Do we need to enforce consistency with `checkout.currency`? This foreshadows point redemption as well...

raginpirate:
Thanks for pushing those small improvements! Aligned with all of them 👍 
And great questions as always Ilya!

> 1/ amount asymmetry: schema has amount as single optional; prose says required-on-response. I think we can split into requested_amount (request) + amount (response, ucp_request: "omit"). Am I holding it right?

Good callout, let me share some thoughts.

The spec right now asks for businesses to always respond with the amounts for open-amount instruments on each response. I think this is the source of practical asymmetry; if a platform forgets its state, the business echos an amount and we can't tell by inspecting what state we're actually considering for that instrument: is it open or not?

To resolve this, I pushed up a diff to the spec language. I propose to instead ask for the business and platform to echo a symmetric value for amount _**until you successfully auth/charge an instrument**_, which is when the business should cement the final charged values on the amount field. WDYT? This makes it very clear that the business has the ability to force an open-amount instrument into a specified-amount one. So for the case of a payment failure, the business can keep a hold on the $10 captured from a gift card while the $40 for the card failed, and communicate to the platform "hey this gift card is now charging exactly $10 instead of open ended, the card failed, if you want the gift card open-ended or to charge more/less just clarify that on the next submit; the state is clear".

Please re-review this and lmk your thoughts; I think this looks really elegant now!

> 2/ No platform_schema defined: this is expected — correct?

Yup, there is no context for a platform to advertise except the extension namespace itself!

> 3/ ^ this is a validation error caught before the business attempts any authorization. Should this be a distinct code (e.g. payment_invalid) to help differentiate instead of overloading?

I agree with you but I don't want to start introducing more standard error codes to the spec without thinking a bit deeper on the taxonomy here. Thinking we can save this one for a follow up PR, wdyt?

> 4/ Matching algorithm ambiguous for overlapping group types. Perhaps...

Thanks, applied!

> 5/ Any gotchas around multi-currency? What if a gift card is denominated in CAD vs a USD checkout? Do we need to enforce consistency with checkout.currency? This foreshadows point redemption as well...

Thanks for flagging that again; the spec did not clarify what currency the amount must be in, updated! And if we get to multi-currency checkouts, we have a simple path to extend this.

igrigorik:
Hmm... Pulling on the `amount` thread surfaced something bigger. Current design introduces a stateful protocol pattern that I want to relitigate (sorry!): the business holds partial authorizations across UCP turns, and the platform's next request carries preservation/release semantics in the value of `amount`. This doesn't exist anywhere else in UCP. Every other recoverable error is "fix input, re-submit fresh, atomic per turn". This was deliberate.

> the merchant might prefer to hold the transaction open and let the buyer provide a different card, rather than refund the gift card and lose the sale

I'm dubious of the premise for how we got here... @prasad-stripe's original framing (quote above) makes sense in a UI-rendered mode, where buyer is progressively filling-in state. In an agent-to-business API, the agent retains all context across a failure, including the gift card credential. Re-submitting both instruments is no harder than submitting them the first time. We debated long-and-hard PATCH/PUT and right semantics for `complete`, and aligned on full-submit. The held-state pattern saves at most one void + re-debit cycle on the merchant's ledger, in exchange for introducing TTL semantics, cross-turn idempotency, state race conditions, and a permanent architectural tax.

I'd propose we restore the original atomic invariant (void immediately on failure; reversal MAY retry transiently), keep James's per-instrument messages for recovery signaling, and drop the cross-turn `amount` semantics entirely. That makes the field design trivial (split `requested_amount` + `amount`, response is purely informational) and keeps split-payments aligned with the rest of UCP's contract. 

Please feel free to push back if I'm holding it wrong.

raginpirate:
@igrigorik might be worth a sync to validate if I am grasping your concern correctly!

I'm aligned that a platform should indeed not care at all about what state is pushed out by the business, and businesses should have no concern resolving any state pushed in by a platform. It is PUT across the board. It is important that businesses retain a right to keep auths open longer to avoid spamming providers when only the Nth credential is being retried / replaced in the list, but its not vital to share if the auth is or is not still held.

My previous revision leveraged the amount as a way to indicate held auth state; but truthfully that was me over-stepping to discuss caching strategies that are an implementation detail and _**not**_ a requirement for businesses or platforms using the protocol. I've pushed to try and clarify language here. 

Businesses should always return an amount when they derive it, reflecting the true amount state of checkout. I hope this resolves your concern without needing to split requested_amount from amount.

igrigorik:
@raginpirate I think we're aligned on desired state, but current prose is tripping me up. PTAL at https://github.com/Universal-Commerce-Protocol/ucp/pull/409/commits/e3961fe99abd741b3a6c6f516930db30710e89bf as an attempt to close the gap -- if I'm holding it wrong, we'll revert.

raginpirate:
Thanks for the push @igrigorik!
Overall I'm aligned with this final PR state, appreciate the small wording tweaks and I'll just touch on the functional diff you've pushed for.
Rather than having the business echo back state of amount each time, the business fundamentally drops it until it leverages the fields to cement the final charged amounts, acting like a receipt.
This is completely functional, easy for all parties to implement, and I'm happy to ship it.

I will note; I am still not fully grasping where the pushback lived against having the business always echo back state (and thus moving from my previous commit to this version). I found it to be quite graceful as businesses always have the right to change the state (move from open-amount to fixed amount because they have cemented final charges). I really liked that previous commit because it made the field's use symmetrical between both parties; this final version we're shipping means that nil inbound has a unique meaning to nil outbound, which I believe was your original push for splitting the field.

If we ack that being fine; lets ship it!

igrigorik:
Synced offline with @raginpirate. Aligned to ship current version. Quick recap for the record:

The substantive difference came down to what "nil" means on `amount`:
 - **Proposed**: `amount` present on response = funds committed; absent = not committed (voided/failed/not attempted). Presence is self-disambiguating — you never need to check `status` to interpret a value.
 - **Alternative**: business echoes inbound amounts on failure too, making the wire shape match across success/failure; consumers cross-reference `status` to distinguish "committed" from "would have been."

Both designs move one bit of asymmetry around: direction-discriminated (shipped) vs status-discriminated (alternative). I lean toward former because it keeps processing logic as simple as possible for payment-related fields: "if present, then it is committed, full stop." Smaller surface area for misinterpretation.

Either design works, just different trade-offs.

igrigorik (approved):
Good to land.

source/schemas/shopping/types/business_split_payments_config.json
```diff
@@ -0,0 +1,24 @@
...
+        "description": "A single valid combination: an array of instrument groups that together define the constraints. All groups must be satisfied (AND logic).",
+        "items": {
+          "$ref": "instrument_group.json"
+        },
+        "minItems": 1
+      },
+      "minItems": 1
+    }
+  },
+  "additionalProperties": false
```
amithanda: Both `types/business_split_payments_config.json` and `types/instrument_group.json` currently define `"additionalProperties": false`. 

Per the [UCP Schema Authoring Guide (Extensibility and Forward Compatibility)](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/docs/documentation/schema-authoring.md#open-objects-additionalproperties), marking configuration schemas as closed is considered an **anti-pattern**:

> *“Marking an object as closed preemptively prevents any future non-breaking additions to the schema... By default, JSON Schema is open and ignores unknown properties. Authors should leave this keyword omitted except in rare circumstances: polymorphic discriminators, security-critical payloads, or protocol envelopes.”*

Since these configuration structures do not fall under any of the listed exceptions (they are not polymorphic discriminators, security-critical payloads, or protocol envelopes), keeping them closed restricts future extensible configuration additions (e.g., adding `max_amount` or `allow_cross_currency` in a minor version update) and will cause client-side validation failures on legacy systems.


Am I missing some strong reason for keeping it or should we remove `"additionalProperties": false`  to keep it consistent with our schema authoring guidelines:
* `source/schemas/shopping/types/business_split_payments_config.json`
* `source/schemas/shopping/types/instrument_group.json`

igrigorik: Great catch, agree, I think we can drop both of these. 

@raginpirate did you have a explicit reason to mark these as closed for extension?

docs/specification/split-payments.md
```diff
@@ -0,0 +1,500 @@
...
+`payment_failed` in `messages[]` and MUST void or reverse any
+authorizations it made. Reversal MAY retry transiently (e.g., to work
+around acquirer rate limits or a failing void call), but the buyer
+MUST NOT remain charged for an incomplete split.
+
+At the protocol layer, each request is processed independently —
+split-payments state does not persist between requests. The business
+MUST process each request as a fresh, full submission, without
+reference to prior requests or responses.
+
```
amithanda, line 174: **Minor Suggestion:**

Should we add a guidance note in the `Error Handling` section of the spec to clarify how protocol correlation IDs (`checkout_id` or similar) and idempotency keys behave during split payment retries.

**Why this may be a good idea:**
*   When a platform retries a partially failed split payment with a modified instrument set (e.g., replacing a declined card), the request payload changes and therefore **must** use a new idempotency key.
*   Because the idempotency key changes, merchants caching authorizations (like gift cards) across these attempts cannot rely on the idempotency key for correlation. They must use the protocol's stable session identifiers (e.g., Checkout Session ID) to correlate these distinct attempts.
*   The note should focus on UCP protocol field semantics rather than prescribing specific merchant database or caching implementation designs.
**Proposed Spec Modification:**
Add the following note to the `Error Handling` section in `docs/specification/split-payments.md`:

```suggestion
> [!NOTE]
> **Idempotency & Correlation during Recovery Retries:**
> When retrying a partially authorized split payment with a modified set of payment instruments (e.g., replacing a declined card), the request payload changes and therefore **MUST** use a new idempotency key. Downstream systems cannot rely on the idempotency key to correlate the retry with previous partial authorizations; they must use the stable Checkout Session `id` to correlate and reconcile these distinct attempts.
```

amithanda:
Great work @raginpirate and everyone on this PR! The stateless execution model, eventual consistency guarantees, and JSONPath-based error reporting resolved the key complexities beautifully. 

I've approved the PR, but left two minor suggestions:
1. **Extensibility:** Consider removing `additionalProperties: false` from the `instrument_group` and `business_split_payments_config.json`  to keep it forward-compatible in line with our schema authoring guidelines.
2. **Idempotency Note:** Suggest adding a quick note in Error Handling clarifying that retries with modified payloads require a new idempotency key, and to use the checkout ID for correlation.

Ready to land once these minor edits are addressed!

amithanda (approved):

source/schemas/shopping/types/business_split_payments_config.json
raginpirate: Nope I think this was just a bug that flew under the radar from months ago; thanks for the sharp eyes @amithanda!

docs/specification/split-payments.md
raginpirate: Thanks for the suggestion, applied!

igrigorik: @amithanda I don't think we need this, it's already defined at the transport layer / implied by the idempotency key contract. Whenever a payload or parameters change, the idempotency key must change. I also wouldn't do belt-and-suspenders here because it actually confuses as to why we don't do this for any other case when the payloads change -- but that's normatively defined and expected behavior.

Issue #355, opened 2026-04-13
[RFC] Proposal: Identity Linking, Identity Management, and Loyalty


amithanda:
# RFC: Identity Linking, Identity Management, and Loyalty

| Field           | Value                                                    |
| :-------------- | :------------------------------------------------------- |
| **Status**      | Proposal                                                 |
| **Authors**     | amithanda                                                |
| **Created**     | 2026-04-12                                               |
| **TC Decision** | Pending                                                  |
| **Relates to**  | #354, #337, #251, #340, #330, #265, #329, #287, #343     |

---

## Summary

This proposal defines a unified, incrementally-deliverable roadmap for identity
and loyalty support in UCP. It sequences a set of focused PRs — from the
foundational OAuth 2.0 authority model through delegated identity providers,
extended authentication mechanisms, loyalty benefit recognition, program
discovery, membership operations, and buyer profile management — so that each
PR is independently reviewable, non-breaking, and delivers real value on its
own while advancing the overall vision.

The proposal is motivated by three clusters of open work that are related to each other:

- **Identity linking** (#265, #329, #330, #354): Redesign iterations,
  a reverted breaking change, and need for discussion around delegated IdPs and
  multi-mechanism negotiation.
- **Identity management** (#337): A well-intentioned but still to be discussed consumption
  of `identity_linking` with `identity_management` that mixes the OAuth
  authority layer with the profile data layer.
- **Loyalty** (#251, #340): Two proposals — a comprehensive one with
  scope creep concerns from reviewers and a focused baseline version.

This RFC provides the architectural framing that lets all three clusters land
cleanly.

---

### The Four-Layer Model

We propose breaking overall Identity and loyalty work in a four-layered stack. Each layer depends on
the one below it; landing them in order is the core idea
of this RFC:

```
┌────────────────────────────────────────────────────────────┐
│  Layer 4: Loyalty Management (Phase 6)                 │
│  Write operations on membership state: enrollment, tier    │
│  progression, reward redemption, cancellation — Phase 6    │
│  (`memberships` capability). Account lifecycle: creation,  │
│  deletion, profile management                                               |
├────────────────────────────────────────────────────────────┤
│  Layer 3: Loyalty Data (Phases 4–5)                       │
│  Loyalty benefit recognition in checkout/cart — Phase 4    │
│  (`loyalty` extension). Program structure, tier catalog,   │
│  and enrollment conditions — Phase 5 (`memberships`        │
[...]

Issue #287, opened 2026-03-21
Wallet Attestation — Identity Mechanism + Eligibility Extension


douglasborthwick-crypto:
## Summary

Add wallet attestation as a first-class pattern in UCP, covering two surfaces:

1. **Identity mechanism** (`wallet_attestation` type in the mechanism registry) — for commerce flows where the wallet *is* the identity, no OAuth redirect or account linking required.
2. **Eligibility extension** (`attestations` map on Cart/Checkout) — for attaching cryptographically signed proofs alongside `context.eligibility` claims.

Both use the same verification primitive: a third-party provider signs an attestation payload, the business verifies offline via JWKS (`provider_jwks` → `kid` → `sig`). This approach was discussed in #203 and invited as a capability extension in [#250](https://github.com/Universal-Commerce-Protocol/ucp/pull/250#issuecomment-4048233895).

## Motivation

OAuth standardized the token format, the verification flow, and the trust boundaries — not how identity providers store passwords or validate sessions. That separation of concerns is what made interoperability possible: any provider that emits the right output participates in the ecosystem.

On-chain eligibility verification is missing that same layer. Every system that checks wallet state defines its own response shape — some return raw balances, others booleans; some are signed, others unsigned. Businesses integrating multiple providers must write bespoke parsing for each one.

UCP's eligibility contract (#250) intentionally leaves verification open — "UCP does not prescribe how verification occurs." For instrument-bound claims (BIN check) or authenticated session claims (loyalty), existing mechanisms suffice. But a third category — on-chain state claims (token holdings, NFT ownership, on-chain credentials) — has no verification path today.

These claims can't be verified from the payment credential or an OAuth session. They require a signed attestation from a third-party verifier that the business can check independently. Standardizing the *output format* — a cryptographically signed, time-bounded attestation — lets any verification method (direct state reads, zero-knowledge proofs, delegated verification) participate, the same way OAuth lets any identity provider participate.

### Goals

* Enable businesses to verify on-chain eligibility claims without trusting the platform's assertion alone
* Provide a wallet-based identity mechanism for commerce flows that don't need account linking (token-gated merchandise, NFT-gated access)
* Reuse the `sig` / `kid` / `provider_jwks` verification primitive already in the spec (#203)

### Non-Goals

* Prescribing specific attestation providers or on-chain verification methods — the spec standardizes the output format, not the computation behind it
[...]

PR #415, opened 2026-05-06
feat: identity linking — mechanism extensibility with wallet attestation

douglasborthwick-crypto:
## Overview

This PR adds `wallet_attestation` as a non-OAuth identity mechanism for identity
linking, materializing the `config.providers` extension point reserved in #354
and corresponding to RFC #355 Phase 3 (Mechanism Extensibility).

**Context.** #354 established the OAuth 2.0 foundation and reserved
`config.providers` as a single extension point — a map keyed by reverse-domain
provider identifier, with a `type` discriminator defaulting to `oauth2`,
explicitly anticipating "future non-OAuth mechanisms such as wallet attestation"
in the schema's `$comment`. RFC #355's dependency graph cites EP #287 as the
canonical input for this phase. Per RFC #355 lines 285 and 651–655, Phase 3 is
independent of Phase 2 and follows the normal proposal lifecycle against `main`
(not the `v2026-04-08` backport).

This PR is non-breaking. Businesses that do not declare `config.providers`
continue to use OAuth 2.0 with RFC 8414 discovery on the business domain.

**Agent-as-buyer.** Wallet attestation also enables flows where OAuth's
machinery is structurally unavailable: OAuth presupposes a prior human
consent event and an account to authenticate against, and an autonomous
agent transacting with no human in the loop at transaction time has
neither. Wallet attestation needs neither, because the entitlement is
recorded in chain state rather than in an account. For agent-as-buyer flows the wallet is the subject of the proof — the business supplies the predicate, and the attestation reports whether chain state satisfies it at the attested block, with no account to log into and no token to issue. It complements account-based mechanisms rather than replacing them; those remain the right tool for entitlements that live in a merchant or IdP account.

The mechanism proves the predicate, not the presenter. A business should not treat an attestation alone as proof that the requester controls the attested wallet — establishing that binding sits with the platform presenting it, and is out of scope for this PR.

**Schema-shape note.** `config.providers` is reachable today via the existing
`additionalProperties: true` rule on `config` — businesses can declare a
`providers` map at runtime without any schema-level change, and the
schema-level `$comment`'s normative ignore-and-fall-back rule covers
discovery semantics for unrecognized `type` values. This PR adds the
`wallet_attestation_provider` `$def` to the `$defs` catalog so spec text and
future structural work have a typed shape to reference. Phase 2 (#330) will
formalize `config.providers` as a structural typed field with a `oneOf`
discriminator over the catalog; that formalization is intentionally out of
scope here per RFC #355 line 285.

---

## Why a different mechanism *category* — not "wallet-flavored OAuth"

OAuth assembles four pieces of infrastructure to manage entitlements:
consent UI, token expiry, revocation lists, and introspection endpoints.

Wallet attestation collapses three of them into chain state plus a signed
boolean, and does not require the fourth in the same form — it grants no
scope and creates no session, so there is no delegated authority for a
consent screen to authorize:

| Operation | OAuth | wallet_attestation |
| :--- | :--- | :--- |
| Issuance | Account creation + scope grant | Chain mutation (mint, grant, transfer) |
| Consent | Per-request consent screen | Out of scope for the mechanism — no scope granted, no session created; the platform binds wallet to requester |
| Revocation | Revocation list / introspection | Chain mutation (burn, transfer-out, downgrade) |
| Verification | Token introspection or JWKS lookup | JWKS signature check + freshness check on signed attestation |

Issuance, revocation and verification are all anchored in chain state and
verifiable offline against the provider's published JWKS. Revocation is
native: the next attestation reads current state and returns `false`
automatically. What the attestation asserts is an eligibility predicate over
chain state at a block height — not an authorization, and not a claim about
who is presenting it.

This is the framing for why Phase 3 exists as a distinct mechanism class
rather than as a configuration variant of `oauth2`.

---

## Complementarity with OAuth Identity Linking

This is not a replacement for OAuth Scenario 4 (custom roles via `continue_url`)
described in the RFC #355 Appendix A walkthrough. The two mechanisms cover
different layers of entitlement representation:

- **OAuth (Scenario 4)** — account-based roles where entitlement lives in the
  business's database (e.g., a CRM tier, an employment role, a fraud-risk
  profile). The merchant authoritatively declares the role.

- **wallet_attestation** — wallet-bound roles where entitlement lives in chain
  state (e.g., holding a VIP NFT contract, holding a tier token, an allowlist
  membership). Chain state is the source of truth; the verifier signs a boolean
  over a predicate.

A business can legitimately declare both `oauth2` and `wallet_attestation`
provider entries in `config.providers` for the same identity-linking capability,
covering the full set of entitlements its commerce flow needs.

---

## Pre-auth signal accountability

The two-comment exchange between @jamesandersen and @gsmith85 on RFC #355 (May 4
2026, [4374487442](https://github.com/Universal-Commerce-Protocol/ucp/issues/355#issuecomment-4374487442)
+ [4374493631](https://github.com/Universal-Commerce-Protocol/ucp/issues/355#issuecomment-4374493631))
articulated a structural OAuth gap: the advertisement of benefits
(`scopes_supported.description`: *"log in for VIP pricing"*) and the delivery
of benefits are two unrelated promises. A user authenticates expecting a
discount, the business evaluates eligibility post-auth against state the
platform never sees signed, and the gap is a trust hole.

`wallet_attestation` collapses this. The signed boolean over the predicate IS
the contract — if the verifier signs `holds VIP NFT contract X ≥ 1` with a
JWKS-verifiable signature, eligibility is met at the attested block height,
period. There is no two-step advertisement-then-delivery. The signature is the
trust signal *to the platform, the user, and any post-hoc dispute-resolution
layer* that the merchant's claim is verifiable independent of the merchant's
backend.

---

### New: `wallet_attestation_provider` definition (`source/schemas/common/identity_linking.json`)

Single additive change to the schema: a new `$defs` entry materializing the
`wallet_attestation` mechanism shape reserved in #354's schema-level `$comment`.

```json
"wallet_attestation_provider": {
  "type": "object",
  "title": "Wallet Attestation Provider",
  "description": "Trusted third-party verifier that evaluates a business-supplied predicate against a wallet's on-chain state and signs the boolean result. Stateless: each attestation is a self-contained JWS-signed payload, verifiable offline against the provider's published JWKS.",
  "required": ["type", "provider_jwks"],
  "properties": {
    "type": { "const": "wallet_attestation" },
    "provider_jwks": {
      "type": "string",
      "format": "uri",
      "description": "HTTPS URI of the provider's JWKS (RFC 7517). Businesses fetch this URL to obtain the public keys used to verify attestation signatures."
    },
    "attestation_endpoint": {
      "type": "string",
      "format": "uri",
      "description": "Optional HTTPS URI where platforms POST attestation requests. The endpoint accepts a wallet address plus a predicate, evaluates the predicate against current chain state, and returns a JWS-signed payload (ES256 recommended per RFC 7518)."
    }
  },
  "additionalProperties": false
}
```

**Scope discipline:** This is the only schema change in Phase 3. The structural
formalization of `config.providers` as a typed field (`oneOf` discriminator,
`propertyNames` pattern, mapping each `type` value to its `$def`) is Phase 2's
deliverable per RFC #355 and PR #330 (igrigorik). Phase 3 relies on:

1. The existing `"additionalProperties": true` on `config` — businesses can
   include a `providers` map runtime without a schema change
2. The existing schema-level `$comment` normative rule — *"Platforms MUST
   ignore unrecognized fields in `config` and fall back to this default
   behavior"* — already covers the discovery-and-fall-back semantics for
   unknown `type` values
3. The new `$def` — catalog-available for Phase 2 to reference when it
   formalizes `providers` as a structural field

Per RFC #355 line 285, *"Phase 3 can land independently of Phase 2 (Phase 2
recommended but not required for mechanism extensibility)."* Keeping Phase 3's
schema delta minimal preserves that independence.

### New: `## Wallet Attestation` section (`docs/specification/identity-linking.md`)

Adds a normative section after the OAuth 2.0 sections covering:

- **How it works.** The provider reads current chain state, evaluates a
  business-supplied predicate against that state, and returns a JWS-signed
  boolean attestation. The business verifies the signature offline against the
  provider's published JWKS. The attestation is stateless and self-contained
  — no callback, no session, no token exchange.
- **Chain coverage.** UCP's schema is chain-agnostic. The predicate carries
  chain context; the set of chains a provider supports is documented at the
  provider's own discovery surface, not in `config.providers`. Different
  providers may cover different chain sets, and a business may declare
  multiple `wallet_attestation` provider entries (keyed by reverse-domain) to
  cover the union of chains it accepts.
- **When to use.** Wallet-bound entitlements (token-gated catalog access,
  NFT-gated merchandise, allowlist gating, soulbound credentials). Not a
  replacement for OAuth account linking — complementary mechanism.
- **Provider declaration.** `config.providers` entry shape with `type:
  "wallet_attestation"`, `provider_jwks`, `attestation_endpoint`. Reverse-domain
  key naming.
- **Verification procedure** (normative MUSTs):
  1. Platforms MUST fetch the provider's JWKS from `provider_jwks` and select
     the verification key by `kid` from the attestation header.
  2. Platforms MUST verify the JWS signature over the attestation payload using
     the algorithm declared in the JWKS entry (`ES256` recommended).
  3. Platforms MUST check the attestation's freshness window before relying on
     `payload.pass` (e.g., `expires_at` claim or `attestedAt` + max-age).
  4. Platforms MUST treat the signed boolean (`payload.pass`) as the
     entitlement decision; raw chain state in the payload is informational.
- **Privacy property.** The attestation answers *"does this wallet meet the
  predicate?"* rather than *"what does this wallet hold?"* — a buyer proving
  they hold ≥1 of a token never reveals the actual balance or the rest of
  their portfolio to the business. This is a structural distinction from raw
  on-chain queries (which leak full balances) and from credential-style
  approaches (which require buyer-side wallet UX).
- **Relationship to OAuth scopes.** `wallet_attestation` contributes zero
  scopes. The mechanism is stateless: each attestation is self-contained.
  Operations gated by chain-state predicates do not appear in
  `config.scopes`; they are evaluated at request time against fresh
  attestations. Operations gated by both account-state AND chain-state
  predicates may appear in `config.scopes` (for the OAuth side) AND require
  a `wallet_attestation` provider entry (for the chain-state side).

### New: `wallet_state_required` and `wallet_state_optional` info codes

Defined normatively in spec text (in the new `## Wallet Attestation` section of
`docs/specification/identity-linking.md`), mirroring how `identity_required`
and `insufficient_scope` shipped in #354 (normative definitions in spec
sections, not in the schema files). The two strings are also added to the
`examples` array in `source/schemas/shopping/types/info_code.json` as
documentary entries (mirroring how `identity_optional` is listed there). The
`info_code.json` schema is `type: string` — freeform-permitted, no enforced
enum; the additions are documentary only and do not change validation
behavior.

Spec semantics:

- `wallet_state_required` — operation requires a valid wallet_attestation; the
  business has declared the predicate the attestation must satisfy. Platforms
  MUST emit a `WWW-Authenticate: WalletAttestation` challenge on the 401
  response when this code is set (see "WalletAttestation HTTP authentication
  challenge scheme" below).
- `wallet_state_optional` — wallet attestation unlocks specific benefits; the
  business advertises a chain-state-bound benefit (e.g. *"verify your wallet
  for VIP pricing"*) and the platform may surface the option to the user.

These codes compose with `identity_*` rather than replacing them — a merchant
can legitimately emit BOTH `identity_optional` AND `wallet_state_optional` on
the same operation, letting the agent choose which path to offer the user.

### New: `WalletAttestation` HTTP authentication challenge scheme  *[Provisional — open for TC review]*

#354 adopted RFC 6750 §3 `WWW-Authenticate: Bearer` challenges for OAuth
identity linking. Wallet attestation is not OAuth and cannot use the
`Bearer` scheme verbatim — the challenge advertises a different proof
mechanism.

**Why a new scheme rather than extending `Bearer` with a parameter?** `Bearer`
semantics imply a token-issuance flow: the platform exchanges the challenge
for a token via an authorization endpoint, then presents that token on
subsequent requests. `wallet_attestation` issues no token. The challenge
advertises a predicate-evaluation request: the platform calls the provider's
`attestation_endpoint`, receives a JWS-signed payload, and presents the
payload directly. A new scheme name keeps `Bearer` semantics intact for OAuth
and avoids overloading the registered scheme with an alternative flow.

Phase 3 defines:

```
WWW-Authenticate: WalletAttestation
                  realm="https://merchant.example.com/",
                  predicate="<URL-safe-encoded predicate>",
                  expected_kid="<provider-kid>",
                  resource_metadata="https://merchant.example.com/.well-known/ucp-protected-resource"
```

- `realm` (MUST) — issuer URI per RFC 6750 §3 conventions
- `predicate` (MUST) — flat URL-safe encoded representation of the predicate
  the attestation must satisfy (e.g., `chain=base&erc20=0xABC&op=gte&value=1`).
  Suitable for single-condition predicates; complex composition (multi-clause
  AND/OR) is out of scope for this PR — see "Predicate encoding follow-up"
  below.
- `expected_kid` (SHOULD) — JWKS key identifier (the `kid` value the business
  expects to find in the attestation header). Allows platforms to validate
  provider identity before issuing the request.
- `resource_metadata` (SHOULD) — URI of the business's `ucp-protected-resource`
  metadata document, parallel to the `Bearer` challenge convention from #354

**Scheme name registration.** Scheme names are governed by IANA's HTTP
Authentication Scheme Registry (RFC 7235). UCP defining the scheme inline
here, with intent to register if Phase 3 ratifies, mirrors how `Bearer` was
specified in OAuth 2.0 before its later IANA registration. Open to
alternative naming proposals during review.

**Predicate encoding follow-up.** The flat query-string encoding above is
sufficient for single-clause predicates (a balance check, an NFT holding,
an allowlist membership). Complex composition (multi-clause AND/OR, nested
conditions, schema-versioned predicate dialects) will need either a richer
encoding or a discovery-URI pattern (e.g., `predicate_uri="https://merchant.example.com/.well-known/predicate/<id>"`)
where the predicate is published at a stable URL the platform fetches and
caches. Resolution of the encoding format is deferred to a follow-up Phase 3.x
PR so this PR's review surface stays focused on the mechanism itself.

---

## Forward compatibility

This is a non-breaking addition.

- Businesses without `config.providers` continue to use OAuth 2.0 with RFC 8414
  discovery on the business domain.
- Platforms that do not implement `wallet_attestation` MUST ignore entries with
  unrecognized `type` values per the schema-level `$comment` rule, falling
  back to the OAuth 2.0 default.
- The `wallet_attestation_provider` `$def` is catalog-available for Phase 2
  (delegated IdP, #330) to reference when it formalizes `config.providers` as
  a structural field with a typed `oneOf` discriminator. Phase 3 leaves that
  formalization to Phase 2 by design.

---

## Out of scope

- **Issuance.** Phase 3 defines a verification mechanism — given a wallet and a
  condition, return a signed boolean. How the wallet's chain state got into
  that condition is the issuance layer (ERC-721, ERC-1155, ERC-5114 soulbound,
  ERC-4671 non-transferable badges, allowlist mints, paid mints, snapshot
  drops). Issuance is vendor- and chain-specific by design; the verifier reads
  chain state regardless. This cut is consistent with #355's Layer 2
  separation.
- **Wallet connection / signing UX.** Platform-side concern, outside UCP scope.
- **Attestation provider selection.** UCP does not endorse specific verifiers.
  Businesses choose which `provider_jwks` URLs they trust, the same way they
  choose which OAuth issuers they accept.
- **Phase 2 (delegated IdP).** Out of scope for this PR; tracked separately
  per RFC #355 dependency graph.
- **Cart/Checkout `attestations` eligibility extension.** EP #287 covers two
  surfaces: (1) the `wallet_attestation` mechanism in the identity-linking
  registry (Layer 2 — this PR), and (2) an `attestations` map on Cart/Checkout
  for attaching signed proofs alongside `context.eligibility` claims (Layer 3,
  prior implementation attempt #264). The eligibility-extension surface is a
  separate workstream and out of scope for this PR. The mechanism defined
  here can be used independently of any Cart/Checkout extension; the two
  compose cleanly when both are present, but neither depends on the other.

---

## Type of change

- [x] New feature (non-breaking addition of wallet_attestation provider type)
- [x] Documentation update

---

## References

- **RFC #355** ([Identity Linking, Identity Management, and Loyalty](https://github.com/Universal-Commerce-Protocol/ucp/issues/355))
  — Phase 3 (Mechanism Extensibility) per the dependency graph
- **EP #287** ([Wallet Attestation — Identity Mechanism + Eligibility Extension](https://github.com/Universal-Commerce-Protocol/ucp/issues/287))
  — canonical schema input for this PR
- **#354** ([feat!: identity linking OAuth 2.0 foundation](https://github.com/Universal-Commerce-Protocol/ucp/pull/354))
  — Phase 1 foundation; reserved the `config.providers` extension point this
  PR materializes
- **#330** ([igrigorik's delegated IdP proposal](https://github.com/Universal-Commerce-Protocol/ucp/pull/330))
  — Phase 2 (separate workstream; this PR is independent per RFC #355 line 285)
- **#280** ([feat: add wallet_attestation mechanism type to identity linking registry](https://github.com/Universal-Commerce-Protocol/ucp/pull/280))
  — prior implementation attempt for the wallet_attestation mechanism, opened
  Mar 19 2026, predating #354's `config.providers` extension point. This PR
  supersedes #280 by re-cutting the mechanism against the merged Phase 1
  schema and the `config.providers` map shape ratified in #354's
  `$comment`. Recommend closing #280 in favor of this PR.
- **#264** ([feat: attestation extension for eligibility claims](https://github.com/Universal-Commerce-Protocol/ucp/pull/264))
  — prior implementation attempt for the Cart/Checkout `attestations`
  eligibility-extension surface. Out of scope for this PR (see Out of scope
  above). #264 remains the open workstream for that surface.
- **JWKS reference**: `https://api.insumermodel.com/.well-known/jwks.json`
  (ES256, P-256, kid `insumer-attest-v1`) — InsumerAPI's published verification
  key set, demonstrating the verifier shape this PR specifies

---

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation

### DIFF
--- a/docs/specification/common/identity-linking/index.md
+++ b/docs/specification/common/identity-linking/index.md
@@ -422,9 +422,11 @@ Each entry is described by its `type`:
 
 | Field | Type | Required | Description |
 | :---- | :--- | :------- | :---------- |
-| `type` | string | Yes | Provider mechanism discriminator. `oauth2` is the only type defined in this version; future versions **MAY** define additional types as non-breaking extensions. |
+| `type` | string | Yes | Provider mechanism discriminator. `oauth2` (delegated OAuth 2.0 / OIDC) and `wallet_attestation` (chain-state proof — see [Wallet Attestation](#wallet-attestation)) are defined in this version; future versions **MAY** define additional types as non-breaking extensions. |
 | `auth_url` | string (URI) | Yes, for `oauth2` | Base URL for authorization server metadata discovery. |
 | `required_claims` | array of strings | No, for `oauth2` | OIDC Core §5.1 claim names the business requires in the JWT authorization grant (e.g. `email`). Optional pre-filter hint — see [Provider Selection](#provider-selection). |
+| `provider_jwks` | string (URI) | Yes, for `wallet_attestation` | HTTPS URI of the provider's JWKS ([RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517){ target="_blank" }) used to verify attestation signatures. See [Wallet Attestation](#wallet-attestation). |
+| `attestation_endpoint` | string (URI) | No, for `wallet_attestation` | HTTPS URI where platforms POST attestation requests. See [Wallet Attestation](#wallet-attestation). |
 
 The `type` value is an open string, not a closed enumeration. Platforms
 **MUST** treat provider entries whose `type` they do not support as filtered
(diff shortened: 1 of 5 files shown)


damaz91:
@kmcduffie @GarethCOliver can you please take a look at this PR?

douglasborthwick-crypto:
@damaz91 @kmcduffie @GarethCOliver — thanks for picking this up. I've rebased this onto current `main` so it sits cleanly on the merged `config.providers` extension point from #423: `wallet_attestation` is now a `type` arm on the shared `provider` schema rather than a standalone `$def`, and the docs + example follow the array-form providers map. Should be a clean diff to review now — happy to adjust anything.

westonale:
Supportive of the extension point, and the agent-as-buyer motivation is the right one: OAuth's redirect is structurally unsatisfiable for an autonomous agent, so a non-OAuth arm belongs here. Two things in the current framing are worth separating before this lands, because they are doing different jobs.

**A wallet signature proves key control, not authorization scope.** OAuth's token is not only an authentication artifact, it carries scope, an expiry, and a subject the business can reason about. A bare `wallet_attestation` proves that whoever signed holds the key right now. It does not say what this agent may spend, on whose behalf, within what limits, or for how long. If `wallet_attestation` is admitted as a peer of `oauth2` without somewhere to carry those, businesses lose something they currently get for free, and every implementation will reinvent it in its own namespace.

**"Prior on-chain acquisitions are durable proof of consent" is the part I would not put in the spec.** Past purchases are evidence of past intent. They are not authorization for a new transaction, and treating them as standing consent means a wallet that bought once has consented to everything afterwards. What the mechanism actually evaluates is an eligibility predicate against chain state: does this wallet hold X, did it acquire Y. That is a useful and honestly checkable claim, and it is worth stating in those terms rather than as consent, because an auditor can verify the predicate and cannot verify the consent.

The shape that has worked for us is to keep the wallet as the key-control root and put authorization in a short-lived signed credential bound to that wallet, carrying scope and an expiry, verified against a published key set. That keeps the PR's core claim intact, the wallet is the credential, while making the authorization explicit, expiring and auditable rather than inferred from history.

Happy to review the schema arm concretely if that helps unblock the stale review.

douglasborthwick-crypto:
Thank you — this is a careful read, and both points land.

**On consent: the accurate description is the one you gave.** What the mechanism evaluates is an eligibility predicate against chain state — does this wallet hold X, did it acquire Y, at this block. "Durable proof of prior consent" claims more than that, and more than an auditor can check; your test is the right one. The narrow thing that paragraph reaches for is only that OAuth's *interactive* consent step is structurally unavailable to an autonomous agent — a statement about the OAuth flow, not a claim that funding a wallet constitutes consent to a later transaction. Predicate language says that without the overreach.

**On scope, I'd separate the two things you're pointing at.** `wallet_attestation` is deliberately stateless and contributes zero scopes — the PR says so explicitly in *Relationship to Scopes* — because a single attestation answers exactly one question: does this wallet satisfy this business-supplied predicate at this block, signed, verifiable offline against the published JWKS. It is a proof about chain state, not a grant of authority, and it is not trying to be one.

So I agree with your premise and draw a different conclusion from it. What may this agent spend, on whose behalf, within what limits, for how long — those are mandate semantics, and they sit in a layer above identity linking rather than inside the linking mechanism itself. Putting them inside the mechanism is what would produce the per-implementation namespaces you're worried about: every business would encode its own limit model into a slot that was only ever meant to carry a predicate result. Keeping the mechanism to one checkable question is what makes it composable with whatever mandate layer a business chooses.

Where you're straightforwardly right is that the spec should not let a reader infer authorization from an eligibility proof.

Douglas

westonale:
Agreed on the layering, and I think you have the better of that argument. Mandate semantics belong above the linking mechanism, and pushing limits into a slot meant to carry a predicate result is exactly how you get the per-implementation namespaces.

The thing I would still name in the text is what happens at the seam, because your last sentence is the whole issue and it currently has no normative form. If `wallet_attestation` sits in the same mechanism enum as `oauth2`, then a Business that branches on mechanism reads scopes from one arm and an empty set from the other. An empty scope set is not a defined value in most stacks: some read absent scopes as unrestricted, others as deny all, and both are silent. The mechanism is correct and the failure lands one layer up, in code that assumed every mechanism in the enum answers the same question.

One sentence closes it. Either the spec says a mechanism that contributes no scopes MUST NOT by itself satisfy a scope-gated operation, or the enum carries a marker distinguishing mechanisms that prove a predicate from mechanisms that grant authority. The first is smaller and I think sufficient. Without it, "does not grant authorization" is true of the mechanism and unenforceable against the reader, which is the gap you already identified.

douglasborthwick-crypto:
Appreciated on the layering — and the failure mode you're describing is real code. But I don't think the spec leaves it open. The closure is in the identity linking base text rather than in the wallet attestation section, which is probably why it doesn't read as closed from where you're standing.

Two sentences already stand, neither of them new in this PR:

*Scopes* — "Listing a scope in `config.scopes` declares that the corresponding operations require a user identity token."

*Error Handling* — "When an operation is gated by a scope listed in `config.scopes` and the request arrives with an absent, expired, invalid, or unverifiable user identity token, the business **MUST** return" 401 with `identity_required`.

A `wallet_attestation` entry issues no token — that's stated in the PR, and it's the reason it contributes zero scopes. So for any operation gated by a listed scope, the identity token is absent and the business MUST challenge, whatever its code reads an empty scope set to mean. The unrestricted reading isn't permitted-but-undefined; it's already non-conformant. And where no scope is listed, there's nothing scope-gated to satisfy, so a new rule wouldn't bite there either.

So I'd read what you've found as a navigation problem rather than a normative one: those two sentences live in *Scopes* and *Error Handling*, and a reader who arrives at the wallet arm first has no particular reason to go looking for them.

Douglas

westonale:
Agreed, and the sentence that settles it is "Advertised scopes MUST apply uniformly across identity paths", together with `config.scopes` being a single business-level map on the `dev.ucp.common.identity_linking` entry rather than a per-mechanism one. One map governing both arms is exactly what makes the branch I described non-conformant rather than undefined, so a new MUST NOT would be redundant. Withdrawn.

The consequence seems worth stating somewhere, though, because it bounds what the mechanism is for. If `wallet_attestation` issues no token, then it can never satisfy an operation gated by any listed scope: that path always arrives with the identity token absent, and the business must challenge. So the wallet arm reaches exactly the operations that are not scope-gated, by construction and not by policy. A business wanting wallet-attested agents to reach a gated operation is not looking for a different attestation, it needs a user identity token from another path.

That reading also makes your layering argument concrete. Limits do not belong inside the mechanism, because the mechanism cannot express them: the scope map already does, one level up, for every path at once.

On the navigation point, a cross reference from the wallet arm to Scopes and Error Handling would put the closure where a reader who starts there is standing. That is where I ended up looking for it and did not find it.

PR #423, opened 2026-05-08
feat: delegated identity providers and accelerated IdP flow

igrigorik:
The v1 design (#354) only supports direct OAuth between platform and business: every business requires a fresh redirect, fresh consent, fresh credentials. In practice, many businesses already offer delegated IdPs (Sign in with Google, Shop, Apple, etc.) to give their users accelerated sign-in and account creation+linking. Platforms that already hold an identity at one of those IdPs benefit too: they can complete the link without re-prompting the user, which is essential for agentic flows where redirects and per-merchant consent screens are prohibitive.

This PR lets a business advertise the delegated IdPs it trusts. Platforms that already hold a token at one of those IdPs chain identity to the business via a back-channel token exchange (RFC 8693 + RFC 7523) — no redirect, no extra prompt — and the business still issues its own access tokens under its own authority. Direct OAuth against the business's own AS remains always available; `providers` is an additive allowlist for chaining only, and businesses **MUST NOT** self-list (chaining-to-self is degenerate).

## How a business advertises providers

**Single delegation**: business trusts one external IdP. Platforms chain identity through that IdP; direct OAuth on the business domain is implicitly available as a fallback.

```json
"config": {
  "providers": {
    "com.shopify.accounts": {
      "type": "oauth2",
      "auth_url": "https://accounts.shopify.com/"
    }
  },
  "scopes": { "dev.ucp.shopping.order:read": {} }
}
```

**Multiple IdPs with claim pre-filter**: business accepts several IdPs and declares required OIDC §5.1 claims so platforms can skip providers whose upstream tokens can't satisfy them (reactive enforcement at the token endpoint remains mandatory):

```json
"config": {
  "providers": {
    "com.shopify.accounts": {
      "type": "oauth2",
      "auth_url": "https://accounts.shopify.com/",
      "required_claims": ["email"]
    },
    "com.google": {
      "type": "oauth2",
    }
  },
  "scopes": { "dev.ucp.shopping.order:read": {} }
}
```

The `type` discriminator is future-extensible (e.g., wallet attestation) as a non-breaking extension. AS discovery anchors on [RFC 9728 protected-resource metadata](https://datatracker.ietf.org/doc/html/rfc9728), so the business's AS can live on a different origin than its API domain.

---

## Checklist


docs/specification/identity-linking.md
```diff
@@ -309,6 +327,227 @@ response.
 where `code_verifier` is absent or does not verify against the stored
 `code_challenge`.
 
+## Identity Providers
+
+The `config.providers` map declares the trusted identity providers a
+business accepts for authentication.
+
+* **When absent:** by advertising the `dev.ucp.common.identity_linking`
```
drewolson-google: Is this for backwards compatibility? Why have this implicit behavior?

docs/specification/identity-linking.md
```diff
@@ -309,6 +327,227 @@ response.
...
+to the business via the [Accelerated IdP Flow](#accelerated-idp-flow) — no
+fresh browser-based OAuth flow required.
+
+### Provider Configuration
+
+Each provider entry specifies the discovery base URL:
+
+| Field | Type | Required | Description |
+| :---- | :--- | :------- | :---------- |
+| `type` | string | No (default: `oauth2`) | Provider type discriminator. |
```
drewolson-google: Again, why a default here?

docs/specification/identity-linking.md
```diff
@@ -309,6 +327,227 @@ response.
...
+with the following UCP-specific requirements:
+
+* `iss` **MUST** identify a provider listed in `config.providers`.
+* `aud` **MUST** match the business's own AS issuer URI exactly.
+* The JWT signature **MUST** be verified against the IdP's `jwks_uri`; if
+  JWKS cannot be retrieved, the business **MUST** fail closed.
+
+Once the assertion is validated, the business resolves the user from `sub`
+(auto-provisioning permitted) and issues an access token scoped to the
+requested UCP scopes. If user interaction is required (terms acceptance,
```
jamesandersen: "If user interaction is required... " should UCP require businesses to issue all advertised scopes _regardless_ of which identity provider is used? e.g. would a business have any cause to require use of _it's own_ OAuth endpoints to acquire certain scopes? ... I don't have a specific scenario in mind but possibly useful to cover this expectation in the docs?

docs/specification/identity-linking.md
```diff
@@ -309,6 +327,227 @@ response.
...
+[RFC 7523 §3](https://datatracker.ietf.org/doc/html/rfc7523#section-3){ target="_blank" },
+with the following UCP-specific requirements:
+
+* `iss` **MUST** identify a provider listed in `config.providers`.
+* `aud` **MUST** match the business's own AS issuer URI exactly.
+* The JWT signature **MUST** be verified against the IdP's `jwks_uri`; if
+  JWKS cannot be retrieved, the business **MUST** fail closed.
+
+Once the assertion is validated, the business resolves the user from `sub`
+(auto-provisioning permitted) and issues an access token scoped to the
```
jamesandersen: Worth _considering_ whether some guidance on identity claims would help implementers here — a couple of scenarios where opaque `sub` could be tricky:
                                                                                                                               
1. Cross-IdP user resolution. A business that accepts both com.shopify.accounts and com.google receives sub=abc123 from one and sub=xyz789 from the other. If these correspond to the same end user, the business has no way to correlate them without a shared identifier like email. Without that, a user who links via Google and later links via the business oauth could end up with two disjoint accounts at the same business.   This has happened to me a time or two in fact ;-)
                          
2. Auto-provisioning a new account. A business receives a chained grant for a sub it has never seen. To create a useful account it likely needs at minimum a contact identifier (email, phone) — otherwise it provisions an empty record it can't communicate with or match to an existing customer.
                                                                                                                               
[OIDC Core §5.1](https://openid.net/specs/openid-connect-core-1_0-31.html#StandardClaims) defines standard claims (email, email_verified, name, etc.) that seem like a natural fit here. Not necessarily as a MUST, but perhaps a SHOULD for IdPs to include email and email_verified when available, or a note in the spec acknowledging that businesses will likely need claims beyond sub for practical user resolution. The building blocks already exist via OIDC — it might just be a matter of connecting the reference.

docs/specification/identity-linking.md
shopdave: > Is this for backwards compatibility? Why have this implicit behavior?

Correct, the idea is to retain the same semantics for the regular identity linking flow as defined in the very first version of the specification.

docs/specification/identity-linking.md
shopdave: > Again, why a default here?

Purely convenience. Want to make clear that unless otherwise noted, OAuth is the protocol for identity linking.

source/schemas/common/identity_linking.json
```diff
@@ -26,6 +26,26 @@
       "pattern": "^[a-z][a-z0-9]*(?:\\.[a-z][a-z0-9_]*)+:[a-z][a-z0-9_]*$"
     },
 
+    "provider": {
+      "type": "object",
+      "title": "Identity Provider",
+      "description": "A trusted identity provider for delegated authentication. The 'type' discriminator defaults to 'oauth2' when absent; future versions may define additional types (e.g., 'wallet') without breaking the shape.",
+      "required": ["auth_url"],
+      "properties": {
```
jamesandersen, line 33: Related to the comment about about other OIDC claims, I'm wondering if there's a use case for the business to include in the `provider` certain well-known claims it needs to perform identity transformation between the IdP domain and it's own domain e.g. to advertise I minimally need an email or phone claim in the JWT from the IdP to either resolve an existing account or create a new one

Platforms may need to understand this when acquiring tokens from the identity provider the first time

jamesandersen (commented):
This is a great step forward in helping remove end user friction in the commerce journey; left a few comments for discussion.

douglasborthwick-crypto:
Strong +1 on lifting `provider.type` into the schema `$def` with the `oauth2` default — that turns `providers` from an OAuth-only list into a genuine mechanism registry, and the "filter unsupported `type`, apply the whitelist, abort if nothing remains" selection rule is the right shape for letting a business offer more than one mechanism class side by side.

It's worth stress-testing the extensibility model against a concrete non-`oauth2` consumer, since that's where the abstraction either holds or leaks. The wallet attestation mechanism I proposed in #415 is the natural first one, and it surfaces a few assumptions worth loosening now while this section is being written.

A wallet attestation provider issues **no token**. The platform presents the business a signed boolean ("wallet satisfies predicate P at block N"), the business verifies it offline against the provider's published JWKS, and nothing is minted — no access token, no session, and it contributes no entries to the scopes map under `config`. That doesn't fit the RFC 8693 → RFC 7523 grant-exchange chain, which assumes the flow terminates in a business-issued bearer token. Three additive adjustments would keep verification-only mechanisms non-breaking:

1. Generalize the deferred-mechanics phrase from *"grant exchange mechanics"* to something like "discovery and proof-presentation mechanics," so a `type` that presents a verifiable assertion (rather than exchanging a grant for a token) is in scope by construction.
2. Note that a provider entry's `type` **MAY** contribute no scopes — presence in `providers` doesn't imply participation in the scope/token model. The selection and abort rules work unchanged; they turn on whether the platform *supports* the `type`, independent of whether that `type` issues a token.
3. Scope `auth_url` to the mechanism that needs it. The `provider` `$def` currently sets `required: ["auth_url"]`, but the `$def` description already promises *"additional types (e.g., 'wallet') without breaking the shape"* — and a verification-only type discovers via `provider_jwks`, not an authorization-server URL, so it has no `auth_url` to supply. As written, any non-`oauth2` entry fails validation against the `$def`. Suggest either an `if`/`then` on the discriminator (`auth_url` required only when `type` is `oauth2`), or relaxing `required` at the `$def` level and letting each `type` declare its own required fields. This is the one structural change that lets #415 land as a pure extension rather than a retrofit.

Happy to rebase #415 onto this branch's selection semantics once it merges, and to fold these clarifications into that PR if you'd rather keep #423 focused on the OAuth-chaining path.

And agreed on the agentic-flow framing — that's the strongest case for keeping the registry mechanism-plural.

Douglas

source/schemas/common/identity_linking.json
```diff
@@ -26,6 +26,26 @@
       "pattern": "^[a-z][a-z0-9]*(?:\\.[a-z][a-z0-9_]*)+:[a-z][a-z0-9_]*$"
     },
 
+    "provider": {
```
raginpirate, line 28: This feels a lot like advertising services but we're doing it under a capability. I imagine that we will run into this a lot in the future; thinking about fulfillment providers, 3p catalog providers, etc.

I'm thinking we could instead extend the base service def for dev.ucp.identity_linking, such that you advertise:

```
"services": {
    "dev.ucp.common.identity_linking": [
      {
        "id": "{{id}}",
        "version": "{{version}}",
        "spec": "https://ucp.dev/{{version}}/specification/identity",
        "transport": "rest",
        "endpoint": "{{3p_url}}",
        "config": {
          "type": "oauth2"
        }
      },
      {
        "id": "{{id}}",
        "version": "{{version}}",
        "spec": "https://ucp.dev/{{version}}/specification/identity",
        "transport": "rest",
        "endpoint": "{{buisness_url}}",
        "config": {
          "type": "oauth2"
        }
      }
    ]
  }
```

The somewhat unique thing here is that right now it seems services have been described as defining verticals, but they can be used under specific product namespaces as well to represent a thinner slice of support, right? Such that you could advertise your checkout operations under a fundamentally unique url from cart, right?

raginpirate (commented):
high level thought about restructuring where we advertise this set of providers, wdyt?

source/schemas/common/identity_linking.json
raginpirate: fwiw this is non-blocking, if we wanted to ship this and later deprecate to advertise through a more unified channel in a follow up PR, we could! Depends on if we see this as a small diff or a larger refactor that needs debates and structural reform. While squinting this looked like it didn't need a reform, but I might be missing something.

source/schemas/common/identity_linking.json
```diff
@@ -26,6 +26,26 @@
...
+      "type": "object",
+      "title": "Identity Provider",
+      "description": "A trusted identity provider for delegated authentication. The 'type' discriminator defaults to 'oauth2' when absent; future versions may define additional types (e.g., 'wallet') without breaking the shape.",
+      "required": ["auth_url"],
```
amithanda: As @douglasborthwick-crypto  highlighted in his suggestion -  Should we relax the global `required: ["auth_url"]` constraint on the `provider` object to accommodate non-OAuth mechanisms (such as wallet attestation) that do not use authorization endpoints? 

Instead of a single flat object, should we define `provider` using UCP's standard `oneOf` pattern, delegating to specific sub-schemas (e.g. `#/$defs/oauth2_provider` which requires `auth_url`, and `#/$defs/wallet_provider` which does not)? This is consistent with other polymorphic patterns in UCP (like `fulfillment_destination.json`) and allows clean extensibility for future provider types without breaking backwards compatibility.

docs/specification/identity-linking.md
```diff
@@ -309,6 +327,227 @@ response.
...
+| Field | Type | Required | Description |
+| :---- | :--- | :------- | :---------- |
+| `type` | string | No (default: `oauth2`) | Provider type discriminator. |
+| `auth_url` | string (URI) | Yes | Base URL for authorization server metadata discovery. |
+
+Platforms **MUST** discover the provider's authorization server metadata from
+`auth_url` using the same two-tier hierarchy as [Discovery](#discovery)
+(RFC 8414 primary, OIDC fallback on 404 only).
+
+### Provider Selection
```
amithanda, line 402: Should we introduce an explicit differentiator (such as a reserved provider key `"self"` or a boolean flag `"self": true` / `"direct": true`) in the provider schema to help platforms distinguish between self-listed direct OAuth endpoints and external delegated IdPs?

It could be argued that platforms can identify the direct endpoint by matching the storefront domain with the provider's `auth_url`, or by treating all listed providers uniformly (running Token Exchange for every provider). However, both the direct OAuth fallback and the proposed provider schema together are insufficient:
1. **Strict Issuer Validation Forces Self-Listing:**
   UCP mandates that: *\"The `issuer` value in the discovered metadata **MUST** match the discovery base URI exactly (per RFC 8414 §3.3). Platforms **MUST NOT** normalize (e.g., strip trailing slashes) before comparison — the value must be a byte-for-byte match.\"*
   If a merchant separates storefront hosting from customer identity (e.g., storefront on `brand.com` but AS on Okta at `brand.okta.com`), direct OAuth discovery on the storefront domain will fail this strict issuer check. To bypass this domain mismatch, the merchant is forced to self-list their AS domain in the `providers` map.
2. **Structural Ambiguity and Circular Requests:**
   Once self-listed, the merchant's endpoint is structurally identical to external IdPs (such as Google). Because the domains do not match the storefront, platforms cannot use domain-matching heuristics to identify the direct endpoint. Without a clear differentiator, the platform cannot determine which endpoint requires direct redirects and which enables back-channel token exchange. If the platform treats the self-listed endpoint as an external IdP, it will attempt a redundant, circular token exchange (RFC 8693) on the merchant's own AS, causing latency and request failures.
An explicit flag resolves this structural ambiguity and ensures platforms can execute direct OAuth redirects for self-hosted authorization and reserve token exchange logic solely for external federators.

amithanda, line 402: Should we introduce an explicit differentiator (such as a reserved provider key `"self"` or a boolean flag `"self": true` / `"direct": true`) in the provider schema to help platforms distinguish between self-listed direct OAuth endpoints and external delegated IdPs?

It could be argued that platforms can identify the direct endpoint by matching the storefront domain with the provider's `auth_url`, or by treating all listed providers uniformly (running Token Exchange for every provider). However, both the direct OAuth fallback and the proposed provider schema together are insufficient:
1. **Strict Issuer Validation Forces Self-Listing:**
   UCP mandates that: *\"The `issuer` value in the discovered metadata **MUST** match the discovery base URI exactly (per RFC 8414 §3.3). Platforms **MUST NOT** normalize (e.g., strip trailing slashes) before comparison — the value must be a byte-for-byte match.\"*
   If a merchant separates storefront hosting from customer identity (e.g., storefront on `brand.com` but AS on Okta at `brand.okta.com`), direct OAuth discovery on the storefront domain will fail this strict issuer check. To bypass this domain mismatch, the merchant is forced to self-list their AS domain in the `providers` map.
2. **Structural Ambiguity and Circular Requests:**
   Once self-listed, the merchant's endpoint is structurally identical to external IdPs (such as Google). Because the domains do not match the storefront, platforms cannot use domain-matching heuristics to identify the direct endpoint. Without a clear differentiator, the platform cannot determine which endpoint requires direct redirects and which enables back-channel token exchange. If the platform treats the self-listed endpoint as an external IdP, it will attempt a redundant, circular token exchange (RFC 8693) on the merchant's own AS, causing latency and request failures.
An explicit flag resolves this structural ambiguity and ensures platforms can execute direct OAuth redirects for self-hosted authorization and reserve token exchange logic solely for external federators.

docs/specification/identity-linking.md
```diff
@@ -309,6 +327,227 @@ response.
...
+
+1. Platform discovers `config.providers` in the business's identity linking
+   capability and selects a provider it already holds a valid token for.
+2. Platform requests a JWT authorization grant from the IdP via token
+   exchange ([RFC 8693](https://datatracker.ietf.org/doc/html/rfc8693){ target="_blank" })
+   at the IdP's token endpoint:
+    * `grant_type`: `urn:ietf:params:oauth:grant-type:token-exchange`
+    * `subject_token`: the platform's existing IdP access token
+    * `subject_token_type`: `urn:ietf:params:oauth:token-type:access_token`
+    * `resource`: the business's authorization server issuer URI (the IdP
```
amithanda: Should we relax the strict requirement to use the `resource` parameter and instead permit the use of `audience`, `resource`, or both (containing the business AS issuer URI or logical identifier) during token exchange?
**Why:**
Strictly mandating `resource` and prohibiting `audience` (lines 433-436) introduces a critical interoperability barrier with several widely deployed Identity Providers (IdPs) and Security Token Services (STS):
1. **Google Cloud STS:** Google’s Workload Identity Federation STS endpoint strictly requires the `audience` parameter (mapped to the workload pool provider URI) and does not support or recognize the `resource` parameter.
2. **Keycloak:** In Keycloak's RFC 8693 token exchange implementation, downscoping and client targets are resolved via the `audience` parameter (mapping to Keycloak client IDs). Keycloak historically does not natively resolve the `resource` parameter out-of-the-box.

**RFC 8693 Semantics:**
While it may be argued that `resource` is defined in RFC 8693 as a URI and `audience` as a logical name, Section 2.1 of the RFC explicitly states that the `audience` value *\"MAY be a URI\"*. Therefore, passing the merchant's AS issuer URI in the `audience` parameter is fully compliant with the RFC and ensures out-of-the-box interoperability with standard OIDC and STS implementations without compromising security.

source/schemas/common/identity_linking.json
amithanda: While a unified vertical discovery channel is a valuable concept, I would not recommend moving IdPs into the global `services` registry here due to UCP's architectural separation:
1. **UCP-Compliant Interfaces vs. Generic Protocols:** 
   In UCP, advertising a service with a `rest` transport implies that the target endpoint conforms to UCP-governed schemas and API contracts (as will be the case with future 3P integrations like Fulfillment or Catalog adapters). In contrast, external IdPs (like Google OIDC or Shop Pay) do not implement UCP-governed schemas; they conform to standard OIDC and OAuth 2.0 specifications.
2. **Registry Pollution:** 
   Listing generic OIDC auth endpoints in the `services` registry would set a precedent of using the service registry for utility configurations, requiring the addition of auth-specific protocols (like `oauth2` or `oidc`) to the service transport enum, which is otherwise reserved for UCP vertical API transport bindings.
3. **Independent Negotiation:** 
   Independent routing and negotiation are already supported under the capability model. `dev.ucp.common.identity_linking` is negotiated as its own independent capability in the profile. Split-domain routing (direct OAuth vs. external IdP) is natively resolved by standard OAuth 2.0 endpoint configurations within the capability's `providers` config map.
Keeping the provider map encapsulated inside the capability configuration remains the correct abstraction. It avoids polluting the service transport registry and keeps commerce operations separated from generic identity mechanics.

source/schemas/common/identity_linking.json
douglasborthwick-crypto: The `oneOf` split is the right call, and the `fulfillment_destination.json` precedent you cited is the one to follow — `oneOf` over `$ref`'d member schemas, as in that file, with these members told apart by the `type` field (the `oauth2` default vs `const: "wallet_attestation"`). Most of the wallet arm is already drafted: #415 defines a `wallet_attestation_provider` $def in `identity_linking.json` with `required: ["type", "provider_jwks"]` and no `auth_url` (plus an optional `attestation_endpoint`). That's exactly the non-oauth2 member you're describing — it drops straight into #423's `oneOf`, with `oauth2_provider` keeping `required: ["auth_url"]` and the `type` default of `oauth2` preserving every existing config (non-breaking on the OAuth path).

Small naming nit: I'd suggest `wallet_attestation_provider` over `wallet_provider`. The arm is specific to the signed-predicate mechanism, and a bare `wallet_*` could read as covering wallet-based OAuth too.

The one thing worth deciding is sequencing: do you want #423 to absorb the wallet arm into its `oneOf` now, or land `providers` here as oauth2-only and let #415 add the second arm on top once #423 merges? Either works on my end — #415 also carries the spec prose, the offline verification procedure, the `wallet_state_required` / `wallet_state_optional` info codes, and a provisional `WalletAttestation` `WWW-Authenticate` scheme, so if #423 takes the schema arm I'll trim #415 down to that prose layer.

Douglas

source/schemas/common/identity_linking.json
raginpirate: Appreciate the pushback @amithanda. Let me pull at a couple of these though to make sure I'm reading the concern right, as my proposal might have been ambiguous.

For 1 - Is there a strong gain from enforcing that a REST API service speaks a UCP dictated schema? From my example, the service is advertised under a UCP capability with a UCP spec that explains how to speak to it over the provided transport and how to use its response to build UCP-owned schemas. I also want to challenge this a bit with a2a. The a2a service points at a metadata document published at a well-known path, whose contract is owned by an external standards body, which the client parses to discover the real endpoints. Understandably the transport then communicates UCP schemas over a2a (hence why its a transport and tied deeper into UCP), but I think of this akin to how after performing Oauth you speak UCP schemas back to our core capabilities to resolve the flow.

For 2 — I notice you started pushing in the direction of claiming oauth2 as a transport. In my initial thinking I saw REST remaining as the transport and communicating a distinction of the auth protocol as a config underneath a "dev.ucp.common.identity_linking" service. Is there a reason you tended in the direction of seeing oauth2 as a transport rather than letting the associated UCP spec inform that context? I agree that I don't want to see us proliferate N unique protocols we are not governing bodies for as transports in our spec.

For 3 - because I'm proposing to keep the services under the capability namespace ("https://ucp.dev/{{version}}/specification/identity"), if the capability is not to be used the services can be discarded. This is akin to leveraging UCP but choosing to not negotiate anything under the shopping vertical; we expect this to be valid and possible to navigate by pruning, correct? I think I see the same pollution problem here.

What I keep coming back to, and why I am proposing this, is because services are a great base schema to advertise services in a common way for all capabilities. What we're setting ourselves up for is N capabilities each over time advertising services in independent ways with fragmented schemas. If we standardize that service endpoints get advertised under services, associated to their capabilities, we can avoid this fragmentation of discovery processes.

docs/specification/identity-linking.md
igrigorik: Great catch. I think the cleaner approach is to flip the contract: providers is the delegated+approved set. PTAL at the new algorithm.

source/schemas/common/identity_linking.json
igrigorik: Good catch. Updated but not using oneOf as that doesn't allow clean extensibility. Instead using if/then to force appropriate required fields, ptal.

source/schemas/common/identity_linking.json
igrigorik: updated

docs/specification/identity-linking.md
igrigorik: Good catch, updated.

docs/specification/identity-linking.md
igrigorik: Clarified.

docs/specification/identity-linking.md
igrigorik: Updated the contract: providers is for delegated allowlist. This clause is no longer necessary.

igrigorik:
@amithanda @jamesandersen @douglasborthwick-crypto ty for thorough review and comments. Pushed a set of commits that should address all the flags and clarifications, ptal.

docs/specification/identity-linking.md
```diff
@@ -309,6 +331,290 @@ response.
...
 `code_challenge`.
 
+## Identity Providers
+
+The `config.providers` map declares external trusted identity providers
+from which the business will accept chained identity via JWT bearer
+assertions for the [Accelerated IdP Flow](#accelerated-idp-flow). It is
+additive metadata on top of the always-available direct OAuth path
+against the business domain (see [Discovery](#discovery)) — never a
+closed whitelist.
```
shopdave: Nit: The text calls `providers` "never a closed whitelist," but for chaining it is one. Lines around 511 requires `iss` to be a listed provider. Suggest distinguishing the two paths explicitly.

```suggestion
against the business domain (see [Discovery](#discovery)); for the chaining
path, it is a closed allowlist — a business **MUST** reject a JWT
authorization grant whose `iss` is not a listed provider (see
[Business Token Issuance](#business-token-issuance)).
```

docs/specification/identity-linking.md
```diff
@@ -618,42 +931,41 @@ field conveys the business's value prompt to the platform (e.g.,
...
+  authorization grants received from platforms. Grants are bearer
+  credentials scoped to a single audience (`aud`) and a single use.
+* **Cross-IdP account linking.** Federation collapses trust boundaries:
+  a business that auto-links accounts across listed providers based on
+  any IdP-asserted claim (email, phone, name) extends each provider's
+  verification process into account-takeover risk. A provider that
+  issues `email_verified=true` for an email the user does not control
+  can hijack any account at the business sharing that email. The stable
+  per-IdP identifier is `(iss, sub)`; other claims are advisory.
+  Businesses **SHOULD** require user-mediated linking — the user
```
jamesandersen, line 995: 👍

source/schemas/common/identity_linking.json
```diff
@@ -26,6 +25,35 @@
...
+      "title": "Identity Provider",
+      "description": "A trusted identity provider for delegated authentication, keyed by the 'type' discriminator. 'oauth2' denotes an OAuth 2.0 / OIDC authorization server. Future versions MAY define additional types (e.g. wallet attestation) as non-breaking extensions; platforms MUST treat entries whose 'type' they do not support as filtered out (see Provider Selection).",
+      "required": ["type"],
+      "properties": {
+        "type": {
+          "type": "string",
+          "description": "Provider mechanism discriminator. 'oauth2' for OAuth 2.0 / OIDC authorization servers. Additional values MAY be defined by future versions; the value is an open string, not a closed enum, so unrecognized types remain valid and are filtered at runtime."
+        }
+      },
+      "allOf": [
```
jamesandersen, line 39: non-blocking... but I do think allowing businesses to advertise OIDC Core §5.1 claims they require from identity providers could be advantageous to avoid an attempt to authenticate followed by an `invalid_grant` response which may or may not reveal the (via an error message) what claim is missing.  It would give platforms the options to skip attempts at using identity linking providers if they are aware that an identity provider cannot satisfy those claims

jamesandersen (approved):

docs/specification/identity-linking.md
shopdave: This suggestion anchors discovery on the RFC 9728 PRM's `authorization_servers` instead of hard-coding issuer = business domain. The byte-match must target the `authorization_servers` entry and the RFC 8414 URL must use sect 3.1 path insertion.

The business's own AS lives in `authorization_servers`, external IdPs stay in `config.providers`. This separates OAuth discovery for the trust domain from the chaining discovery in UCP. 

This also addresses @amithanda's [concerns](https://github.com/Universal-Commerce-Protocol/ucp/pull/423#discussion_r3297173348) on the case when the authorization server is different from the business domain.

```suggestion
A business exposes two OAuth roles, which **MAY** live on different hosts: the
**protected resource** (its UCP API, identified by the business domain in the
UCP profile) and the **authorization server** (AS) that issues tokens for it.
Platforms discover the AS starting from the protected resource.

**Step 1 — Resolve the authorization server.** Platforms determine the AS from
the resource's protected resource metadata
([RFC 9728](https://datatracker.ietf.org/doc/html/rfc9728){ target="_blank" }):

* Fetch `{business-domain}/.well-known/oauth-protected-resource`, or use the
  `resource_metadata` URL from a `WWW-Authenticate` challenge the resource
  returned
  ([RFC 9728 §5.1](https://datatracker.ietf.org/doc/html/rfc9728#section-5.1){ target="_blank" }).
* The metadata's `resource` value **MUST** identify the business resource the
  platform is accessing; platforms **MUST** reject metadata whose `resource`
  does not match
  ([RFC 9728 §3.3](https://datatracker.ietf.org/doc/html/rfc9728#section-3.3){ target="_blank" }).
* The platform **MUST** select an issuer from the non-empty
  `authorization_servers` array and use it as the **AS issuer** for the rest of
  discovery. This issuer is authoritative and **MAY** be hosted on a different
  origin than the business domain.
* If the resource publishes no protected-resource metadata, the AS issuer
  defaults to the business domain (single-host deployments).

**Step 2 — Fetch authorization server metadata.** Using the AS issuer from
Step 1, resolve metadata via a strict two-tier hierarchy. URLs are constructed
per [RFC 8414 §3.1](https://datatracker.ietf.org/doc/html/rfc8414#section-3.1){ target="_blank" }:
when the issuer has a path component, the well-known segment is **inserted
between the host and the path**, not appended.

1. **RFC 8414 (Primary):** Fetch
   `https://{issuer-host}/.well-known/oauth-authorization-server{issuer-path}`.
    * `2xx` response: use this metadata. Discovery complete.
    * `404 Not Found`: proceed to step 2.
    * Any other non-2xx response, network error, or timeout: **MUST** abort.
      **MUST NOT** proceed to step 2.

2. **OIDC Discovery (Fallback):** Fetch
   `{issuer}/.well-known/openid-configuration`.
    * `2xx` response: use this metadata. Discovery complete.
    * Any non-2xx response, network error, or timeout: **MUST** abort the
      identity linking process.

Platforms **MUST NOT** silently fall through on any error other than `404` in
step 1. This prevents partial or undefined behavior when a server is
misconfigured or temporarily unavailable.

**Step 3 — Validate.** The `issuer` value in the discovered metadata **MUST**
match the AS issuer selected in Step 1 exactly, byte-for-byte (per
[RFC 8414 §3.3](https://datatracker.ietf.org/doc/html/rfc8414#section-3.3){ target="_blank" }).
Platforms **MUST NOT** normalize (e.g., strip trailing slashes) before
comparison. The match target is the selected `authorization_servers` entry —
not the business domain.
```

docs/specification/identity-linking.md
shopdave: I believe this is a valid concern, and I have proposed changes to the Discovery section. The missing part here is how authorization servers get discovered from the resource server.

docs/specification/identity-linking.md
shopdave: We have now arrived at this future. :-) 

```suggestion
    domain conventions. The business **MUST** publish this metadata when
    the authorization server does not live on the business domain.
```

docs/specification/identity-linking.md
```diff
@@ -309,6 +331,290 @@ response.
...
+| `auth_url` | string (URI) | Yes, for `oauth2` | Base URL for authorization server metadata discovery. |
+
+The `type` value is an open string, not a closed enumeration. Platforms
+**MUST** treat provider entries whose `type` they do not support as filtered
+out (see [Provider Selection](#provider-selection)) rather than rejecting the
+business's configuration.
+
+For `oauth2` providers, platforms **MUST** discover the authorization server
+metadata from `auth_url` using the same two-tier hierarchy as
+[Discovery](#discovery) (RFC 8414 primary, OIDC fallback on 404 only).
```
shopdave: Align with the proposed changes to Discovery.

```suggestion
metadata from `auth_url` using the same two-tier metadata hierarchy as
[Discovery](#discovery) Step 2 (RFC 8414 primary with §3.1 path insertion,
OIDC fallback on 404 only), treating `auth_url` as the issuer. The
protected-resource step does not apply — `auth_url` is already the IdP
issuer — and the business validates the JWT grant's `iss` against it per
[Business Token Issuance](#business-token-issuance).
```

docs/specification/identity-linking.md
```diff
@@ -618,42 +931,41 @@ field conveys the business's value prompt to the platform (e.g.,
 * **Token revocation.** Platforms **MUST** revoke user identity tokens at the
   business's revocation endpoint (RFC 7009) when a user unlinks their account.
   Businesses **MUST** reject subsequent requests that present revoked tokens.
+* **JWT grant lifetime.** JWT authorization grants **MUST** be short-lived;
+  the `exp` claim **SHOULD** be no more than 60 seconds after `iat`. Short
+  lifetimes limit the window for grant theft and replay.
+* **JWT grant single-use.** Businesses **SHOULD** enforce single-use JWT
```
shopdave: We already mandate `jti` presence, so we might as well make enforcement mandatory.

```suggestion
* **JWT grant single-use.** Businesses **MUST** enforce single-use JWT;
  a short exp narrows the replay window, but only jti tracking closes it.
```

shopdave (commented):
Overall looks good. Suggest revamping the discovery section to explicitly use PRM as the mechanism for discovering where the authorization server is located. This also closely aligns us with the [MCP approach](https://modelcontextprotocol.io/docs/tutorials/security/authorization) and makes off-business domain authorization servers possible.

Also suggest updating the PR body to reflect the latest changes as to not confuse anyone in the future.

docs/specification/identity-linking.md
igrigorik: @shopdave 👍  on the aim but feels like we're replicating a good chunk of normative language defined by the RFCs we're citing. Attempt at a compressed form: https://github.com/Universal-Commerce-Protocol/ucp/pull/423/changes/0932c8e27d3be1d868a9cb3aec88891efdeaab2a -- does that look reasonable?

source/schemas/common/identity_linking.json
igrigorik: 👍  updated

docs/specification/identity-linking.md
shopdave: LGTM!

shopdave (approved):

source/schemas/common/identity_linking.json
raginpirate: Thanks for the conversation in the TC today @igrigorik, just sharing the aligned view: by moving this up to the services layer we create a disconnect capability and the capability endpoint itself. In this case its _**might**_ be fine but I can see scenarios needing your capability config to carry pointers to a service or vice-versa; the ergonomics are debatable.

So overall, we agreed that keeping this under the capability is the right shape, and where we might have room to grow in the future is defining a shared struct that can be leveraged here; I'd point to available_instruments on payment handlers as a good example of pulling common patterns out of the config-grab-bag. We can do this in the future if this becomes useful 👍

raginpirate (approved):

source/schemas/common/identity_linking.json
```diff
@@ -45,9 +79,14 @@
             "properties": {
               "config": {
                 "type": "object",
-                "$comment": "Reserved extension point: 'providers' (map of trusted identity providers keyed by reverse-domain). See schema-level $comment for details.",
                 "required": ["scopes"],
                 "properties": {
+                  "providers": {
```
raginpirate, line 84: Sorry for a re-review, but I just noticed one edge here compared to services that might feel off as we grow to support non-oauth idps. Example:
```
"providers": {
  "app.example.loginOauth": {
    "type": "oauth2",
    "auth_url": "https://accounts.example-login.app/",
    "required_claims": ["email"]
  },
  "app.example.loginWallet": {
    "type": "wallet",
    ... other fields
  }
},
```

vs.


```
"providers": {
  "app.example.login": [{
      "type": "oauth2",
      "auth_url": "https://accounts.example-login.app/",
      "required_claims": ["email"]
    }, {
      "type": "wallet",
      .. other fields
  }]
},
```

By having the keys of providers not be an array, we'd force all idp namespaces to key off their type of token acquisition. I'd push for us to mirror that existing pattern used in the base ucp obj, which alleviates this awkwardness and feels consistent.

igrigorik:
@raginpirate sharp eyes and great catch! Agree on the nudge, updated -- ptal.

raginpirate (approved):
🚀!

docs/specification/identity-linking.md
```diff
@@ -618,42 +975,42 @@ field conveys the business's value prompt to the platform (e.g.,
 * **Token revocation.** Platforms **MUST** revoke user identity tokens at the
   business's revocation endpoint (RFC 7009) when a user unlinks their account.
   Businesses **MUST** reject subsequent requests that present revoked tokens.
+* **JWT grant lifetime.** JWT authorization grants **MUST** be short-lived;
+  the `exp` claim **SHOULD** be no more than 60 seconds after `iat`. Short
+  lifetimes limit the window for grant theft and replay.
+* **JWT grant single-use.** Businesses **MUST** enforce single-use JWT;
```
amithanda, line 981: I think the "JWT grant single-use" bullet is broken mid-clause:

> Businesses **MUST** enforce single-use JWT; a short exp narrows the replay window, but only jti tracking closes it. authorization grants by tracking the `jti` claim within the grant's validity window.

This is probably a merge artifact from @shopdave's suggestion. I think, two fragments got spliced. 

```suggestion
* **JWT grant single-use.** Businesses **MUST** enforce single-use of JWT
  authorization grants by tracking the `jti` claim within the grant's
  validity window. A short `exp` narrows the replay window, but only
  `jti` tracking closes it.
```

amithanda:
## Approving — great work on this 🚀

Thanks for the thorough iteration here, @igrigorik. My earlier suggestions are all well addressed:

- **`resource`/`audience` interop** - the move to "`resource` and/or `audience`, MUST include at least one" cleanly unblocks existing implementations like Google STS and Keycloak without weakening the binding. 👍
- **Self-listed vs. external IdP ambiguity** - flipping `providers` to a delegated-only closed allowlist (plus the "ignore an entry whose `auth_url` matches the business's own issuer" rule) resolves the circular-token-exchange concern far more elegantly than a `self` flag would have.
- **Non-OAuth extensibility** - using `if/then` on the `type` discriminator keeps `auth_url` required only for `oauth2` while leaving the door open for wallet attestation and future mechanisms as non-breaking additions.

One small open item before merge - the **"JWT grant single-use"** bullet in Security Considerations is currently garbled (the original sentence's tail got left dangling when @shopdave's `jti` suggestion was applied). Left an inline suggestion to restore it to a single coherent MUST. Purely editorial.

This is a meaningful step forward for UCP identity linking capability. Nicely done. Approving with the one nit above.

amithanda (approved):

PR #424, opened 2026-05-08
feat!: Express instrument requirements with request constraints, split PAN and Network Token credential types

raginpirate:
Larger-scope continuation of #288, now stacked on #655. Earlier iterations are #580 (axis factoring), #626 (`constraints` / `options` partition), #564 (fulfillment application) and #565 (credential split).

**What changed since the original proposal.** This PR opened by adding three new primitives — `object_constraint.json`, `value_constraint.json`, `type_constraint.json` — and modelling `available_instruments[]` as an array of Type Constraints. Review established that a constraint primitive grafted onto every object that negotiates input is not a generic contract (#424 review), that closed enums and closed dispatch permanently exclude extensions because `allOf` can only intersect, and that the validation the complexity bought was never actually enforced. #655 then landed a single closed Constraint Expression grammar and bound it to request data.

So this PR now adds **no additional constraint primitive at all**. It makes the requirements this arc was about expressible with the grammar that already exists, by fixing the two things that were blocking it: credential variants that could not be addressed positionally, and a protocol member that could only carry one rule.

## Changes

   - **Distinct `pan_credential` and `network_token_credential`.** Different PCI
     scope, different required fields, different verification data.
     `card_credential` — which folded both behind `card_number_type` — is
     deprecated, not removed.
   - **`card_payment_instrument.network`**, the card scheme *elected* for a
     transaction. A submitted field, present when a co-badged selector was shown.
   - **`constraint_target.brand`**, representing the card scheme accepted. A value the Business
     *derives* from the account number, so it survives tokenization and is not a
     submitted field. This is the target for the standard constraint object on every available instrument.
   - **Outright removed the $def for available_instruments**; trying to maintain it through this significant prose change is tough, and this release is outright breaking all references to payments objects regardless. Its worth also renaming the constraint target now.
  
   ## Example: Require a CVC for a PAN but a cryptogram for a network token                                                                                           
                                                                                                                                                                      
   The anchoring ask from #288, and the one concern (C4) left open on #655. One path                                                                                  
   selects the submitted instrument and a nested `anyOf` (#757) expresses the two                                                                                     
   credential branches: splitting the credential types is what gives each branch a                                                                                    
   stable discriminator, so no rule has to branch on a sibling field.                                                                                                 
                                                                                                                                                                      
   ```json                                                                                                                                                            
   {                                                                                                                                                                  
     "type": "card",                                                                                                                                                  
     "constraints": { "properties": { "brand": { "enum": ["visa", "mastercard"] } } },                                                                                                  
     "ucp": {                                                                                                                                                         
       "request_constraints": {                                                                                                                                       
         "path": "$['payment']['instruments'][?@['handler_id'] == 'processor_1' && @['type'] == 'card']",                                                             
         "required": ["network", "billing_address", "credential"],                                                                                                    
         "properties": {                                                                                                                                              
           "billing_address": { "required": ["postal_code", "address_country"] },                                                                                     
           "credential": {                                                                                                                                            
             "anyOf": [                                                                                                                                               
               { "properties": { "type": { "const": "pan" } }, "required": ["cvc"] },                                                                                 
               { "properties": { "type": { "const": "network_token" } }, "required": ["cryptogram", "eci_value"] }                                                    
             ]                                                                                                                                                        
           }                                                                                                                                                          
         }                                                                                                                                                            
       }                                                                                                                                                              
     }                                                                                                                                                                
   }                                                                                                                                                                  
   ```                                                                                                                                                                
                                                                                                                                                                      
   One declaration covers all three of the use cases worked through on this PR:                                                                                       
   CVV on a card, AVS fields on the billing address, and a network-token cryptogram.                                                                                  
   Requirements shared by every card instrument are sibling members; the ones that                                                                                    
   differ per credential family are branches. A network token is never asked for a                                                                                    
   `cvc`, and a PAN is never asked for a `cryptogram`. `brand` sits in `constraints`                                                                                  
   because it is derived; every requirement above targets a field a Platform                                                                                          
   actually submits. 

   ## Where the review concerns land

   Numbering follows the summary on #655.

   | #   | Concern | Status |                                                                                                                                         
   | :-- | :-- | :-- |                                                                                                                                                
   | C1 | Learn requirements before submitting, not via submit-and-fail | Resolved. Response-carried, evaluated before the next request. |                            
   | C2 | Field-name typos (`billing_addr`) | Build-time lint. Documented, with a SHOULD on the analyzable path form so tooling can resolve the target. |             
   | C3 | Brittle named booleans (`requires_card_verification`) | Resolved. No named booleans anywhere. |                                                             
   | C4 | Polymorphic credential discrimination | **Resolved.** Structural split plus a nested `anyOf`; see the example above. |                                      
   | C5 | `type` discriminator didn't drive validation | Resolved where a handler references the instrument's `available_*` definition and declares the optional `ucp` property. Both now documented. |                                                                                                                               
   | C6 | `constraints` isn't truly generic | Resolved by not adding a primitive. The grammar is `constraint_expression.json`, unchanged. |                           
   | C7 | Credential inception — requirements on what a token represents | **Not addressed.** Off-wire; see below. |                                                  
   | C8 | Expressiveness and locality | Resolved by division: derived values stay in the domain's `constraints`, wire data goes to `request_constraints`. |           
   | C9 | `options` over-strict typing | Dissolved with the `options` axis. |                                                                                         
   | C10 | `properties` keyword collision | Resolved upstream in #655. |                                                                                              
   | — | Closed enums exclude extensions | Resolved. No enums; see example 3. |

   ## Not in scope

   **Requirements on what a token represents (C7).** A token credential carries an
   opaque token, so the credential it was minted from is not in the payload. This requires a bit of a taxonomy shift and the hardest part of the problem is wanting to evaluate "required": [] constraints for elements NOT being actively negotiated over the wire; requiring payments to commit to minting a potentially unique constraint application.

## Category (Required)

## Checklist

### DIFF
--- a/docs/specification/overview/index.md
+++ b/docs/specification/overview/index.md
@@ -244,6 +244,14 @@ that member is present, so a branch pinning a discriminator through
 discriminator in the branch's `required` makes the branch match only the shape
 it describes.
 
+The grammar is defined independently of the object it is bound to. Request
+Constraints bind it to objects in the next request and add `path`; other UCP
+schemas reuse it where a declaration already identifies the object it
+constrains, such as
+[`available_instruments[].constraints`](site:schemas/shopping/types/available_payment_instrument.json),
+whose object is the `constraint_target` declared by the instrument schema for
+that entry's `type`.
+
 ### Path
 
 Every `request_constraints` value has exactly one effective path. When `path`
@@ -1356,7 +1364,7 @@ Businesses publish their profile at `/.well-known/ucp`. An example:
             {
               "type": "card",
               "constraints": {
-                "brands": ["visa", "mastercard", "amex"]
+                "properties": { "brand": { "enum": ["visa", "mastercard", "amex"] } }
               }
             }
           ],
@@ -1517,7 +1525,7 @@ example:
           "spec": "https://example.com/specs/payments/processor_tokenizer-payment",
           "schema": "https://example.com/schemas/payments/delegate-payment.json",
           "available_instruments": [
-            {"type": "card", "constraints": {"brands": ["visa", "mastercard"]}}
+            {"type": "card", "constraints": {"properties": {"brand": {"enum": ["visa", "mastercard"]}}}}
           ]
         }
       ]
@@ -2709,7 +2717,7 @@ request a challenge.
             {
               "type": "card",
               "constraints": {
-                "brands": ["visa", "mastercard"]
+                "properties": { "brand": { "enum": ["visa", "mastercard"] } }
               }
             }
           ],
--- a/docs/specification/overview/index.md
+++ b/docs/specification/overview/index.md
@@ -558,8 +558,10 @@ payment policy.
 #### Alternative verification requirements on a submitted credential
 
 A Business accepts more than one credential shape and requires different
-verification data for each. In this example, a raw PAN must carry a `cvc`, and a
-network token must carry a `cryptogram` with its `eci_value`:
+verification data for each. In this example, a PAN must carry a `cvc`, and a
+network token must carry a `cryptogram` with its `eci_value`. Each credential
+family is its own schema, so every branch discriminates on the credential's own
+`type` and no rule has to branch on a sibling field:
 
 <!-- ucp:example schema=shopping/types/available_payment_instrument op=read direction=response -->
 ```json
@@ -573,12 +575,12 @@ network token must carry a `cryptogram` with its `eci_value`:
         "credential": {
           "anyOf": [
             {
-              "properties": {"card_number_type": {"const": "fpan"}},
-              "required": ["card_number_type", "cvc"]
+              "properties": {"type": {"const": "pan"}},
+              "required": ["cvc"]
             },
             {
-              "properties": {"card_number_type": {"const": "network_token"}},
-              "required": ["card_number_type", "cryptogram", "eci_value"]
+              "properties": {"type": {"const": "network_token"}},
+              "required": ["cryptogram", "eci_value"]
             }
           ]
         }
@@ -591,20 +593,26 @@ network token must carry a `cryptogram` with its `eci_value`:
 One path selects the submitted instrument, and one Object Constraint describes
 it. The sibling `required` applies to every matching instrument; the `anyOf`
 branches then apply to the nested `credential` object, which must satisfy at
-least one. Each branch pins `card_number_type` and also names it in `required`,
-so a branch matches only the credential shape it describes. A raw PAN without a
-`cvc` fails, as does a network token missing its `eci_value`.
+least one. Each branch pins `type` with `const`, so a branch matches only the
+credential family it describes;
+[`payment_credential.json`](site:schemas/shopping/types/payment_credential.json)
+already requires `type` on every credential, so no branch has to name it in
+`required`. A [PAN credential](site:schemas/shopping/types/pan_credential.json)
+without a `cvc` fails, as does a [network
+token](site:schemas/shopping/types/network_token_credential.json) missing its
+`eci_value`.
 
 Because every branch pins the discriminator, the branch set also closes the
-accepted values. A `dpan` credential is valid under
-[`card_credential.json`](site:schemas/shopping/types/card_credential.json) but
-satisfies neither branch, so this Business does not accept it at this path. A
-Business that later accepts a new variant adds a branch for it.
+accepted credential families. A handler
+[token credential](site:schemas/shopping/types/token_credential.json) is a valid
+credential at this position but satisfies neither branch, so this Business does
+not accept it at this path. A Business that later accepts another family adds a
+branch for it.
 
 Two separately targeted constraints cannot express this rule. Request
 Constraints conjoin, so one value requiring `cvc` and another requiring
 `cryptogram` would require both. Discriminating through the path filter instead
-— selecting `fpan` credentials in one value and `network_token` credentials in
+— selecting `pan` credentials in one value and `network_token` credentials in
 another — moves conditional logic into the selector, which paths do not carry.
 
 ## Actions
--- a/docs/specification/overview/index.md
+++ b/docs/specification/overview/index.md
@@ -184,11 +184,6 @@ Platform request. For example, it can constrain a Line Item quantity to exactly
 avoiding a round trip for request data the Business has already indicated it
 will reject.
 
-`ucp.request_constraints` is an array. Each element carries one optional `path`
-and one Constraint Expression, so a single scope can state several rules with
-different targets. Elements compose conjunctively, array order carries no
-meaning, and a consumer **MUST NOT** read precedence into position.
-
 `ucp.request_constraints` is used only in authoritative operation responses and
 has no effect in discovery profiles or operation requests.
 
@@ -372,14 +367,12 @@ The following Cart responses are alternatives that illustrate different scopes.
     {
       "ucp": {
         "version": "{{ ucp_version }}",
-        "request_constraints": [
-          {
-            "path": "$['line_items'][*]",
-            "properties": {
-              "quantity": {"const": 1}
-            }
+        "request_constraints": {
+          "path": "$['line_items'][*]",
+          "properties": {
+            "quantity": {"const": 1}
           }
-        ]
+        }
       },
       "id": "cart_123",
       "line_items": [
@@ -428,14 +421,12 @@ The following Cart responses are alternatives that illustrate different scopes.
             {"type": "total", "amount": 120000}
           ],
           "ucp": {
-            "request_constraints": [
-              {
-                "path": "$['line_items'][?@['id'] == 'line_123']",
-                "properties": {
-                  "quantity": {"const": 100}
-                }
+            "request_constraints": {
+              "path": "$['line_items'][?@['id'] == 'line_123']",
+              "properties": {
+                "quantity": {"const": 100}
               }
-            ]
+            }
           }
         }
       ],
@@ -478,19 +469,17 @@ derived path is `$`, which selects the next request root.
       ]
     },
     "payment_handlers": {},
-    "request_constraints": [
-      {
-        "required": ["discounts"],
-        "properties": {
-          "discounts": {
-            "required": ["codes"],
-            "properties": {
-              "codes": {"const": ["ACME-X7Q9-L2M4"]}
-            }
+    "request_constraints": {
+      "required": ["discounts"],
+      "properties": {
+        "discounts": {
+          "required": ["codes"],
+          "properties": {
+            "codes": {"const": ["ACME-X7Q9-L2M4"]}
           }
         }
       }
-    ]
+    }
   },
   "id": "checkout_123",
   "status": "incomplete",
@@ -541,12 +530,10 @@ matching submitted card instrument:
 {
   "type": "card",
   "ucp": {
-    "request_constraints": [
-      {
-        "path": "$['payment']['instruments'][?@['handler_id'] == 'processor_1' && @['type'] == 'card']",
-        "required": ["billing_address"]
-      }
-    ]
+    "request_constraints": {
+      "path": "$['payment']['instruments'][?@['handler_id'] == 'processor_1' && @['type'] == 'card']",
+      "required": ["billing_address"]
+    }
   }
 }
 ```
@@ -2867,6 +2854,10 @@ Most platform implementations can **avoid PCI-DSS scope** by:
 - Forwarding credentials without the ability to use them directly
 - Using PSP tokenization payment handlers where raw credentials never pass
     through the platform
+- Presenting pre-provisioned card network tokens
+    (`network_token_credential.json`) rather than an FPAN — the platform never
+    shares the underlying account number, and the token is unusable without a
+    matching cryptogram
 
 #### Business Scope
 
--- a/docs/specification/overview/index.md
+++ b/docs/specification/overview/index.md
@@ -184,6 +184,11 @@ Platform request. For example, it can constrain a Line Item quantity to exactly
 avoiding a round trip for request data the Business has already indicated it
 will reject.
 
+`ucp.request_constraints` is an array. Each element carries one optional `path`
+and one Constraint Expression, so a single scope can state several rules with
+different targets. Elements compose conjunctively, array order carries no
+meaning, and a consumer **MUST NOT** read precedence into position.
+
 `ucp.request_constraints` is used only in authoritative operation responses and
 has no effect in discovery profiles or operation requests.
 
@@ -367,12 +372,14 @@ The following Cart responses are alternatives that illustrate different scopes.
     {
       "ucp": {
         "version": "{{ ucp_version }}",
-        "request_constraints": {
-          "path": "$['line_items'][*]",
-          "properties": {
-            "quantity": {"const": 1}
+        "request_constraints": [
+          {
+            "path": "$['line_items'][*]",
+            "properties": {
+              "quantity": {"const": 1}
+            }
           }
-        }
+        ]
       },
       "id": "cart_123",
       "line_items": [
@@ -421,12 +428,14 @@ The following Cart responses are alternatives that illustrate different scopes.
             {"type": "total", "amount": 120000}
           ],
           "ucp": {
-            "request_constraints": {
-              "path": "$['line_items'][?@['id'] == 'line_123']",
-              "properties": {
-                "quantity": {"const": 100}
+            "request_constraints": [
+              {
+                "path": "$['line_items'][?@['id'] == 'line_123']",
+                "properties": {
+                  "quantity": {"const": 100}
+                }
               }
-            }
+            ]
           }
         }
       ],
@@ -469,17 +478,19 @@ derived path is `$`, which selects the next request root.
       ]
     },
     "payment_handlers": {},
-    "request_constraints": {
-      "required": ["discounts"],
-      "properties": {
-        "discounts": {
-          "required": ["codes"],
-          "properties": {
-            "codes": {"const": ["ACME-X7Q9-L2M4"]}
+    "request_constraints": [
+      {
+        "required": ["discounts"],
+        "properties": {
+          "discounts": {
+            "required": ["codes"],
+            "properties": {
+              "codes": {"const": ["ACME-X7Q9-L2M4"]}
+            }
           }
         }
       }
-    }
+    ]
   },
   "id": "checkout_123",
   "status": "incomplete",
@@ -530,10 +541,12 @@ matching submitted card instrument:
 {
   "type": "card",
   "ucp": {
-    "request_constraints": {
-      "path": "$['payment']['instruments'][?@['handler_id'] == 'processor_1' && @['type'] == 'card']",
-      "required": ["billing_address"]
-    }
+    "request_constraints": [
+      {
+        "path": "$['payment']['instruments'][?@['handler_id'] == 'processor_1' && @['type'] == 'card']",
+        "required": ["billing_address"]
+      }
+    ]
   }
 }
 ```
--- a/source/schemas/shopping/types/card_credential.json
+++ b/source/schemas/shopping/types/card_credential.json
@@ -2,7 +2,8 @@
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "$id": "https://ucp.dev/schemas/shopping/types/card_credential.json",
   "title": "Card Credential",
-  "description": "A card credential containing sensitive payment card details including raw Primary Account Numbers (PANs). This credential type MUST NOT be used for checkout, only with payment handlers that tokenize or encrypt credentials. CRITICAL: Both parties handling CardCredential (sender and receiver) MUST be PCI DSS compliant. Transmission MUST use HTTPS/TLS with strong cipher suites.",
+  "description": "Deprecated: use PAN Credential (`pan_credential.json`) or Network Token Credential (`network_token_credential.json`). A card credential containing sensitive payment card details including raw Primary Account Numbers (PANs). This credential type MUST NOT be used for checkout, only with payment handlers that tokenize or encrypt credentials. CRITICAL: Both parties handling CardCredential (sender and receiver) MUST be PCI DSS compliant. Transmission MUST use HTTPS/TLS with strong cipher suites.",
+  "deprecated": true,
   "allOf": [
     {
       "$ref": "payment_credential.json"
@@ -25,7 +26,7 @@
             "network_token",
             "dpan"
           ],
-          "description": "The type of card number. Network tokens are preferred with fallback to FPAN. See PCI Scope for more details."
+          "description": "Deprecated: the credential type now carries this distinction. The type of card number. Network tokens are preferred with fallback to FPAN. See PCI Scope for more details."
         },
         "number": {
           "type": "string",
--- a/source/schemas/shopping/types/card_credential.json
+++ b/source/schemas/shopping/types/card_credential.json
@@ -2,76 +2,64 @@
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "$id": "https://ucp.dev/schemas/shopping/types/card_credential.json",
   "title": "Card Credential",
-  "description": "A card credential containing sensitive payment card details including raw Primary Account Numbers (PANs). This credential type MUST NOT be used for checkout, only with payment handlers that tokenize or encrypt credentials. CRITICAL: Both parties handling CardCredential (sender and receiver) MUST be PCI DSS compliant. Transmission MUST use HTTPS/TLS with strong cipher suites.",
+  "description": "A card credential whose verification scheme is the card verification code (CVC). The `number` field MAY carry an FPAN, a network token, or a wallet-issued token — the credential type is defined by its verification fields, not by the underlying PAN form. CRITICAL: Both parties handling card credentials with raw FPANs MUST be PCI DSS compliant. Transmission MUST use HTTPS/TLS with strong cipher suites.",
   "allOf": [
-    {
-      "$ref": "payment_credential.json"
-    },
+    { "$ref": "payment_funding_source.json" },
     {
       "type": "object",
-      "required": [
-        "type",
-        "card_number_type"
-      ],
+      "required": ["type", "number"],
       "properties": {
         "type": {
           "const": "card",
-          "description": "The credential type identifier for card credentials."
-        },
-        "card_number_type": {
-          "type": "string",
-          "enum": [
-            "fpan",
-            "network_token",
-            "dpan"
-          ],
-          "description": "The type of card number. Network tokens are preferred with fallback to FPAN. See PCI Scope for more details."
+          "description": "The credential type identifier."
         },
         "number": {
           "type": "string",
-          "description": "Card number.",
-          "examples": [
-            "4242424242424242"
-          ]
+          "description": "Card number. May be an FPAN, network token, or wallet token; the discriminator is the credential type, not the number form.",
+          "examples": ["4242424242424242"]
         },
         "expiry_month": {
           "type": "integer",
-          "description": "The month of the card's expiration date (1-12)."
+          "description": "Card expiration month (1-12)."
         },
         "expiry_year": {
           "type": "integer",
-          "description": "The year of the card's expiration date."
+          "description": "Card expiration year."
         },
         "name": {
           "type": "string",
           "description": "Cardholder name.",
-          "examples": [
-            "Jane Doe"
-          ]
+          "examples": ["Jane Doe"]
         },
         "cvc": {
           "type": "string",
           "maxLength": 4,
-          "description": "Card CVC number.",
-          "examples": [
-            "223"
-          ]
-        },
-        "cryptogram": {
-          "type": "string",
-          "description": "Cryptogram provided with network tokens.",
-          "examples": [
-            "gXc5UCLnM6ckD7pjM1TdPA=="
-          ]
-        },
-        "eci_value": {
-          "type": "string",
-          "description": "Electronic Commerce Indicator / Security Level Indicator provided with network tokens.",
-          "examples": [
-            "07"
-          ]
+          "description": "Card verification code. Optional at the protocol level; merchants can require it via `required_fields` on the funding-source constraint.",
+          "examples": ["223"]
         }
       }
     }
-  ]
+  ],
+  "$defs": {
+    "constraint": {
+      "title": "Card Credential Constraint",
+      "description": "Constraint shape for the `card` credential type.",
+      "allOf": [
+        { "$ref": "constraint.json" },
+        {
+          "type": "object",
+          "properties": {
+            "required_fields": {
+              "type": "array",
+              "items": {
+                "type": "string",
+                "enum": ["number", "expiry_month", "expiry_year", "name", "cvc"]
+              },
+              "uniqueItems": true
+            }
+          }
+        }
+      ]
+    }
+  }
 }
--- a/source/schemas/shopping/types/network_token_credential.json
+++ b/source/schemas/shopping/types/network_token_credential.json
@@ -1,0 +1,68 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/network_token_credential.json",
+  "title": "Network Token Credential",
+  "description": "A card-network token credential verified with a transaction cryptogram. The `number` field carries the network token or wallet-provisioned token rather than the underlying FPAN.",
+  "allOf": [
+    {
+      "$ref": "payment_credential.json"
+    },
+    {
+      "type": "object",
+      "required": [
+        "type",
+        "number",
+        "cryptogram"
+      ],
+      "properties": {
+        "type": {
+          "const": "network_token",
+          "description": "The credential type identifier for network token credentials."
+        },
+        "number": {
+          "type": "string",
+          "description": "Network token or wallet-provisioned token replacing the underlying FPAN.",
+          "examples": [
+            "5204240000004242"
+          ]
+        },
+        "expiry_month": {
+          "type": "integer",
+          "description": "The month of the token's expiration date (1-12)."
+        },
+        "expiry_year": {
+          "type": "integer",
+          "description": "The year of the token's expiration date."
+        },
+        "name": {
+          "type": "string",
+          "description": "Cardholder name.",
+          "examples": [
+            "Jane Doe"
+          ]
+        },
+        "cryptogram": {
+          "type": "string",
+          "description": "Transaction cryptogram or dynamic CVC (dCVV), in the long or short form expected by the card network or processor.",
+          "examples": [
+            "gXc5UCLnM6ckD7pjM1TdPA=="
+          ]
+        },
+        "eci_value": {
+          "type": "string",
+          "description": "Electronic Commerce Indicator / Security Level Indicator associated with the transaction.",
+          "examples": [
+            "07"
+          ]
+        },
+        "token_requestor_id": {
+          "type": "string",
+          "description": "Payment network token requestor identifier, when required by the processor or network-token program.",
+          "examples": [
+            "12345678901"
+          ]
+        }
+      }
+    }
+  ]
+}
--- a/source/schemas/shopping/types/network_token_credential.json
+++ b/source/schemas/shopping/types/network_token_credential.json
@@ -1,0 +1,61 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/network_token_credential.json",
+  "title": "Network Token Credential",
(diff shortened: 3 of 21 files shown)


source/schemas/shopping/types/card_credential.json
```diff
@@ -2,76 +2,64 @@
...
-        "type",
-        "card_number_type"
-      ],
+      "required": ["type", "number"],
       "properties": {
         "type": {
           "const": "card",
-          "description": "The credential type identifier for card credentials."
-        },
-        "card_number_type": {
```
jamesandersen, line 21: Should we deprecate rather than removing to avoid a breaking change?

source/schemas/shopping/types/network_token_credential.json
```diff
@@ -0,0 +1,61 @@
...
+          "type": "integer",
+          "description": "Token expiration month (1-12)."
+        },
+        "expiry_year": {
+          "type": "integer",
+          "description": "Token expiration year."
+        },
+        "cryptogram": {
+          "type": "string",
+          "description": "Per-transaction cryptogram. Structurally required — a network token credential without a cryptogram is a CVC-verified card; submit it as `card_credential` instead."
```
jamesandersen: nit: Could we indicate this may be either long form or short form (e.g. dCVV)

source/schemas/shopping/types/network_token_credential.json
```diff
@@ -0,0 +1,61 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/network_token_credential.json",
+  "title": "Network Token Credential",
+  "description": "A card credential whose verification scheme is a per-transaction cryptogram and Electronic Commerce Indicator. Used for network-tokenized PANs (merchant tokens, wallet-provisioned DPANs, processor-issued network tokens). The `number` field carries the network/wallet token; verification is via `cryptogram` + `eci_value` rather than CVC. The agent selects this credential when the underlying instrument provides cryptographic verification; it selects `card_credential` instead when the verification is a CVC, regardless of whether the PAN itself is an FPAN or a token.",
+  "allOf": [
+    { "$ref": "payment_funding_source.json" },
+    {
+      "type": "object",
+      "required": ["type", "number", "cryptogram", "eci_value"],
+      "properties": {
```
jamesandersen: Thoughts about adding `token_requestor_id` (#296) onto `network_token_credential` while we're at it?  not widely used as discussed on that PR but we do have the [Braintree BYOT example](https://developer.paypal.com/braintree/docs/guides/network-tokens/bring-your-own-token/java/)

source/schemas/shopping/types/network_token_credential.json
```diff
@@ -0,0 +1,61 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/network_token_credential.json",
```
jamesandersen: This scratches that itch I had from #296  ;-)

jamesandersen (commented):
@raginpirate thanks for putting #424 together — I do think this proposal works well:

  - (+) Introduces a new credential type `network_token` 
    - Adds the distinction I was looking for originally in #296 ;-)
    - ... and probably will have implications for the doc changes proposed in #367                                                                                                                                                                                      
  - (+) Required fields are explicit via the new `constraint` primitive
    - e.g. under instrument -> credentials -> constraints -> required_fields  (more like where #288 started but now implemented at the right level)
    - In contrast to where #288 was trending with *implicit* constraints via instrument -> constraints -> requires_card_verification and docs to explain intent at the credential level                                                                             
  - (+) Cross-domain reuse e.g. `address_constraint.json`, `credential_constraint.json`                                                     
  - (-) Breaking change with removal of `card_number_type` ... left separate comment on this

On the "credential inception" point e.g. distinct constraint on a wire credential and - where applicable - the funding source underlying it. What do you think about taking this on in a separate PR? I'd like to better understand if there is a concrete scenario requiring both? e.g. can a business require a CVV on the underlying instrument while only accepting the network token?  It's been a winding road so far to get alignment on updated constraint modeling for just the "wire credential" ... hopefully a follow-on for the underlying funding source could be quicker and cleaner after landing this. 
                                         

TBH ... it took a while to grok this all (in large part for lack of good focus time) but I think the resolved examples are actually much easier than the schema suggests at first glance - just as another sanity check the multiple entries under `credentials` is how we address the hurdle that led us to the implicit bool and AVS enum on #288. Look right?
```json                                                                                                                      
  {                                                                                                                            
    "available_instruments": [                              
      {                                                                                                                        
        "type": "card",  // card-based instrument                                                                                         
        "constraints": {                                                                               
          "brands": ["visa", "mastercard"],                                                                                    
          "required_fields": ["billing_address"],                                                                     
          "billing_address": {                                                                                                 
            "required_fields": ["postal_code"] // e.g. AVS1, but adapts to other schemes                                                                    
          },                                                                                                                   
          "credentials": [
            { 
              "type": "card", // FPAN
              "constraints": { "required_fields": ["cvc"] } 
            },
            { 
              "type": "network_token", // DPAN / CPAN
              "constraints": { "required_fields": ["cryptogram"] } // could add ECI if a business needs it
            }
          ]                                                                                                                    
        }                                                   
      },                                                                                                                        
    ]                                                                                                                          
  }
```

TateLyman:
Schema read only.

One validation gap to keep an eye on if this moves from prototype to spec shape: `credential_constraint.json` describes `constraints` as being scoped to the named credential type, and the concrete credential schemas define narrowed `$defs.constraint` objects, but the actual reference is still the generic base constraint:

```json
"constraints": {
  "$ref": "constraint.json"
}
```

Because of that, JSON Schema validation will not use `type` to select `card_credential.json#/$defs/constraint` or `network_token_credential.json#/$defs/constraint`. For example, these look invalid by the intended semantics but appear valid against the current primitive:

```json
{ "type": "card", "constraints": { "required_fields": ["cryptogram"] } }
{ "type": "network_token", "constraints": { "required_fields": ["cvc"] } }
```

Same issue for `funding_sources`: the body says entries must reference schemas inheriting from `payment_funding_source.json`, but `type` is an unconstrained string in the schema itself.

Suggested direction: either make `credential_constraint` an explicit `oneOf` over known credential types (`if type == card then constraints -> card_credential#/$defs/constraint`, etc.), or keep it intentionally registry/runtime-validated and soften the schema language so platforms do not assume this is machine-checkable from the JSON Schema alone.

The interoperability risk is that merchants/platforms may think constraint payloads are statically validated, while cross-axis mistakes only exist in prose. That matters for checkout because a mistyped funding-source requirement can become a failed tokenization/payment path rather than a schema error.

jamesandersen:
Great catch @TateLyman — this is a real validation gap. As you noted, the `type` discriminator in `credential_constraint.json` doesn't drive schema selection for the `constraints` body, so cross-axis mistakes like `{ "type": "card", "constraints": { "required_fields": ["cryptogram"] } }` pass validation silently and only surface as payment failures. This matters particularly because UCP aims for a machine-verifiable schema — it's the foundation for SDK generation, so constraints that are only enforced in prose undermine that goal.

Your `if/then` suggestion would work well and has precedent in UCP — `total.json` already uses this exact pattern (discriminating on `type` to apply different validation rules to a sibling property), as does `pagination.json` (conditional field requirements based on a sibling value).

A couple of alternative approaches that achieve the same schema validation, for consideration:

**Per-credential `$defs` with `oneOf` dispatch:** Each credential schema (`card_credential.json`, `network_token_credential.json`) already defines a `$defs/constraint` narrowing `required_fields`. Each could additionally define a `$defs/credential_constraint` that bundles `type: { const: "card" }` with a `constraints` ref to its own `$defs/constraint`. The instrument schema would then use `oneOf` over those refs instead of referencing `credential_constraint.json` directly. This keeps each credential schema self-contained — no central file needs to know about all types.

**Single `$defs/constraint` (breaking change):** Take the above a step further — instead of adding a parallel `$defs/credential_constraint`, redefine `$defs/constraint` itself to be the full credential constraint entry (extending `credential_constraint.json` via `allOf`, adding the `type` const and narrowed `constraints`). One concept, one `$defs` entry per credential schema. Cleanest mental model, but this would be a breaking change to the main schema if merged. Worth weighing whether the value of a tighter, self-contained schema outweighs the inconvenience to implementations that have already adopted `$defs/constraint` in its current form.

All three approaches reject the same invalid payloads — the difference is where the enforcement lives: central dispatch (`if/then`), per-schema with two defs, or per-schema with one def.

@raginpirate WDYT?

raginpirate:
Thanks for the comments folks; totally agree with the schema validation concern. This actually goes back to the original payment handler design; we did not associate instruments with any credential types as an attempt to show they are open objects to be associated as needed. In that same vein, I did not originally try having machine-enforcable types for the constraints, but honestly I think this is just a gap we can close. In my latest revision I've actually taken the oneOf approach for both constraints as well as credentials themselves 👍

I wanted to get this into review this weekend but it took quite a bit of time shredding my original proposal apart here: one of the main contentious points is "funding sources" as you pointed out @jamesandersen, and in trying to isolate it I reverted my world view on the concept and moved it into a type of constraint specific to token credentials.

I'm feeling good about the work in this PR, and my steps to get this ready to publish are:
* Isolate each commit into its own PR
* Add documentation for schema-authoring to describe how to write constraints across UCP
* Rewrite documentation for payment handlers across the board to include our new constraints

One interesting note is still how we refactor the card credential: I think the right path is to actually entirely deprecate it and just introduce a PAN credential and a network token credential. Its technically non-breaking too because card is isolated and attached to nothing in the base spec; handlers in the wild using it can just keep referencing it from the old api versions without concern if they really want to. WDYT about this or the plan above?

jamesandersen:
> "funding sources" as you pointed out @jamesandersen, and in trying to isolate it I reverted my world view on the concept and moved it into a type of constraint specific to token credentials.

@raginpirate  makes sense to me ... if introducing distinct PRs I think this will become more digestible ;-)

> deprecate it and just introduce a PAN credential and a network token credential

I do think this is going to be easier to reason about in the long run

igrigorik (changes requested):
Catching up, great discussion. The direction makes sense, I agree that modelling this as booleans doesn't scale and pollutes the schema. Reading the current shape, though, I hit a number of gotchas and I'm wondering if we're over-promising here.

## Meta issue: `constraints` isn't actually generic

The pitch is a universal primitive ("every constraint object in UCP follows this shape"), but it isn't one — it's a shape you have to graft onto every object that negotiates input. That's a mixin re-attached per-object, not a generic contract, and it doesn't scale as the surface grows; we'd have to include this on every object at the limit.

#564 is the early evidence: it reuses the primitive as `fulfillment_available_method.constraints.destination`, but `destination` isn't a field of the fulfillment method, so it already breaks the base contract ("`required_fields` names properties of the constrained object").

## If it *is* meant to be generic, it has to be extensibility-friendly

This rules out enums / closed dispatch. `allOf` can only *intersect*, never widen — so any closed construct in the base permanently excludes extensions:

- **Enum:** `address_constraint.required_fields` is a closed enum of the 9 canonical fields. `postal_address.json` is open, so a handler can add `tax_id` — but can never require it (`"required_fields": ["tax_id"]` is rejected). A "declare what you need" primitive that can't name an extended field isn't extensibility-friendly.
- **Closed dispatch:** validating a typed family by enumerating known types in the base rejects extension types. The `credentials` `oneOf` here avoids that only by adding an open catch-all branch + `type_constraint` to stay open.

And the validation this complexity buys isn't actually enforced. The `brands` and `credentials` rules live on the card schema, but a handler declaration validates against the base `available_payment_instrument.json` — which never dispatches on `type == "card"` to pull those rules in. So they go unchecked: the documented `business_schema` example still passes the repo's own validator even with `brands` broken (change `["visa","mastercard"]` to a bare `"visa"` and it's still valid). Same gap @TateLyman/@jamesandersen flagged for credentials, one level up.

---

## Stepping back: are we over-generalizing this problem?

Reactive feedback is the established mechanism, and it already handles dynamic requirements. Submit what you have → `checkout.status: "incomplete"` + `messages[]` with `message_error.path` (RFC 9535 JSONPath) and `severity: "recoverable"` ("fix inputs and retry via API"). Zero new schema, works with extensions and responds to dynamic requirements of the current negotiation. This does put an obligation on handlers to validate and return messages *before* tokenizing (rather than failing at the PSP) — but that's a smaller, more local ask than a new schema primitive. The main gap is that it costs a roundtrip.

**I'm not against modeling *some* pre-submit requirements.** An upfront signal is genuinely useful, so a platform can collect complete input (and render the right fields per option) before the first submit. Also agree that bools were the wrong mechanism.

### Alt, what if: model it as `requires` array of JSONPaths...

```json
{
  "type": "card",
  "accepts": {
    "brands": ["visa", "mastercard"],
    "credentials": [
      { "type": "card",          
         "requires": ["$.payment.instrument.credential.cvc"] },
      { "type": "network_token", 
         "requires": ["$.payment.instrument.credential.cryptogram"] }
    ]
  },
  "requires": ["$.payment.instrument.billing_address.postal_code"]
}
```

- One vocabulary across `requires` and `message_error.path`
- `requires` is a minimum, not a completeness guarantee
- Models arbitrary requirements without introducing N booleans
- Applied to payments, can be selectively adopted in other places if/as necessary

Net: no `constraint.json` / `type_constraint.json` / `address_constraint.json`, no enum, no per-object algebra, requirements separated from capabilities, and nothing to misapply (so #564 doesn't arise). This pattern can be applied selectively to other places too, but we would use it surgically in select places, instead of promising a generic contract that we can't deliver.

jamesandersen:
Appreciate this @igrigorik — good callout that even the landed `brands` constraints aren't really being enforced today (a handler config validates against the base `available_payment_instrument.json`, which never dispatches on `type == "card"`, so the card-specific `brands` rule is never applied — a malformed `brands` passes validation). That's a fair reason to rethink the shape rather than build more on top of it.

On the JSONPath direction: worth naming that its main weakness — unvalidated paths (a `cvv`/`cvc` typo passes silently) — is the same one @alexpark20 raised on [#288](https://github.com/Universal-Commerce-Protocol/ucp/pull/288#issuecomment-4179417767) that originally pushed us toward a schema-based approach. That said, you've pointed out that the schema/`$defs` route can get cumbersome quickly (`constraints` applied at the base getting out of sync with future object extensions) and still not be fully functional, I'm supportive of _consciously_ going with `requires: [JSONPath]` — the consistency with existing UCP mechanisms (shared vocabulary with `message_error.path`) is worth more here than validation we're not actually delivering.

The one thing I'd hold firm on is the need for an **upfront** signal. The anchoring use case from #288 is a platform knowing *before it submits complete checkout* whether a business' payment handler requires CVV (a per-merchant setting on handlers like `dev.shopify.card`) — otherwise the platform guesses or eats a failed tokenization.  Your `requires` proposal would solve that.  

@raginpirate looks like this would still build on the distinct `card` / `network_token` credential types introduced here which I'm a fan of.  WDYT?

igrigorik:
@jamesandersen — agreed on both, and I think we can close the typo concern rather than eat it...

Upfront signal: yes, `requires[]` is exactly that — declared at discovery so the platform knows before it submits.

Credential types: `requires` is type-agnostic — it co-locates on each credential entry, so it rides on `card`/`network_token` (or whatever we land). 

On the typo gap: rather than an enum on the wire (which can't name extension fields), we can teach `ucp-schema` to resolve each `requires` path against the composed instrument/credential schema — a path to a non-existent property fails. That's build-time/CI validation (whoever authors the path runs it, same as any UCP payload), not a forced per-handshake operation, which I think is a nice balance.

raginpirate:
Thanks for the feedback @igrigorik and @jamesandersen! ❤️

* Totally right call to avoid enums, that was an overstep on my part where I looked at it from an "unlikely to change" perspective rather than an immutability perspective. Made the change on this to kill the address constraint and simplify the base constraint language. With this, I don't see any other extensibility concerns in the PR, but please flag if there still is.
* Also aligned on the typing concern: I recognize that no matter what, once we shift away from being able to use an enum on these branches, the type enforcement level I tried pursuing was off-base. Also made the changes on this front to just keep the typing simple across the board.

Now... hear me out, I am not sold on the shift from `constraints` as a mixin to this `requires` proposal. When I look at anywhere I'd want to place this requires field, I notice that I'm actually colocating it within an object that uses convention to accomplish the path your encoding there... examples being:
* available_payment_instrument already implicitly targets the payment instrument
* fulfillment_available_method already implicitly targets the fulfillment method
* the config of an extension
I think that having the fullpath on messages is important because the entirety of the message is bound to a single target point: but these objects are constraining a wide number of fields and subfields.

Maybe to help drive it home, the $path based proposal:                                                                                                                                                           
 ```json                                                                                                                                                   
   {                                                                                                                                                       
     "available_instruments": [                                                                                                                            
       {                                                                                                                                                   
         "type": "card",                                                                                                                                   
         "constraints": {                                                                                                                                      
           "brands": ["visa", "mastercard"],                                                                                                               
           "credentials": [                                                                                                                                
             {                                                                                                                                             
               "type": "pan",                                                                                                                              
               "requires": [                                                                                                                               
                 "$.payment.instruments[].credential.cvc"                                                                             
               ]                                                                                                                                           
             },                                                                                                                                            
             {                                                                                                                                             
               "type": "network_token",                                                                                                                    
               "requires": [                                                                                                                               
                 "$.payment.instruments[].credential.cryptogram"                                                                      
               ]                                                                                                                                           
             }                                                                                                                                             
           ],                                                                                                                                                
         "requires": [                                                                                                                                     
           "$.payment.instruments[].billing_address.postal_code",                                                                     
           "$.payment.instruments[].billing_address.address_country"                                                                  
         ]                                                                                                                                                
         }                                                                                                                                                
       }                                                                                                                                                   
     ]                                                                                                                                                     
   }                                                                                                                                                       
 ```

The localized required_fields proposal:

   ```json                                                                                                                                                 
   {                                                                                                                                                       
     "available_instruments": [                                                                                                                            
       {                                                                                                                                                   
         "type": "card",                                                                                                                                   
         "constraints": {                                                                                                                                  
           "brands": ["visa", "mastercard"],                                                                                                               
           "required_fields": ["billing_address"],                                                                                                         
           "billing_address": {                                                                                                                            
             "required_fields": ["postal_code", "address_country"]                                                                                         
           },                                                                                                                                              
           "credentials": [                                                                                                                                
             {                                                                                                                                             
               "type": "pan",                                                                                                                              
               "constraints": {                                                                                                                            
                 "required_fields": ["cvc"]                                                                                                                
               }                                                                                                                                           
             },                                                                                                                                            
             {                                                                                                                                             
               "type": "network_token",                                                                                                                    
               "constraints": {                                                                                                                            
                 "required_fields": ["cryptogram"]                                                                                                         
               }                                                                                                                                           
             }                                                                                                                                             
           ]                                                                                                                                               
         }                                                                                                                                                 
       }                                                                                                                                                   
     ]                                                                                                                                                     
   }                                                                                                                                                       
 ```
 
Very similar but I think the later is significantly less noisy, and your earlier feedback greatly improved the complexity and correctness of the final solution.

WDYT? I'm happy to be pushed back on if theres a benefit I've missed on the fullpath here. Also happy to rename required_fields to requires if you think thats a better name!

igrigorik:
@raginpirate good turn, I think we're converging on right shape. As a wildcard thought experiment, what if we push this one step further: https://github.com/Universal-Commerce-Protocol/ucp/pull/580

jamesandersen:
Really like where this landed, @raginpirate — dropping `address_constraint.json` and the closed enum in favor of the open `required_fields: string[]` addresses the extensibility concern cleanly, and the `if type == "token" then …` dispatch on `credentials` is exactly the right shape: it validates known credential types while letting handler-extended types fall through to the open base, no rejection. That's additive and extension-safe.

Good call on keeping local `required_fields` over the `requires: [JSONPath]` direction — a constraint object already implicitly targets its host, so full paths would be redundant here. Agreed.

The one item I don't think is closed yet is the **enforcement gap**:

- `payment_handler.json → available_instruments[].items` refs the **base** `available_payment_instrument.json`.
- The base doesn't dispatch on `type`, so the card-specific rules (`brands`, the `if type==token` credential dispatch) — which live in `card_payment_instrument.json#/$defs/available_card_payment_instrument` — are never pulled into the validated path.
- Concretely: a handler declaring `{ "type": "card", "constraints": { "brands": "visa" } }` (bare string, malformed) still validates clean, because it only ever hits the base's open `additionalProperties`.

That's the same "validation isn't actually enforced" point from earlier, still true on HEAD. The fix is the same `if/then` pattern you already used for token credentials, just applied one level up at the **instrument type** — but it needs to sit at the point-of-use (or a small dispatcher schema), *not* in the base itself, since `available_card_payment_instrument` already `allOf`s the base and a base→card ref would recurse. Roughly:

```json
// what available_instruments[].items validates against — a dispatcher, not the bare base
{
  "if":   { "properties": { "type": { "const": "card" } }, "required": ["type"] },
  "then": { "$ref": "card_payment_instrument.json#/$defs/available_card_payment_instrument" }
  // unknown instrument types fall through to the open base — extension-safe, same as your credential dispatch
}
```

That turns the `brands`/credential rules from documented-but-dormant into actually-enforced, and closes the gap without reintroducing enums or a base-level mixin.

raginpirate:
👋 Sorry for the slow loop here, @igrigorik @jamesandersen this is good for re-review!

TL;DR: I kept #424’s local constraint shape and pulled over the JSON Schema vocabulary from #580 that fit cleanly: required, plus enum and const. The proposal now has an explicit composition of Object, Value, and Type Constraints, while handler-specific literal constraints remain open.

@jamesandersen, I agree with the behaviour you observed, but not the proposed validation layer. A core if type == "card" dispatcher works mechanically, but it globally binds instrument type strings and requires payment_handler.json to enumerate UCP instrument schemas. The generic handler should remain open; the negotiated handler’s available_payment_instrument definition should compose and validate its concrete branches only, not the entire base spec's definitions.

I also don’t think the typing gap should remain unresolved. Some constraint-aware ucp-schema tools could resolve the target and concrete constraint schemas, check required names, and validate every constraint value against its declared type. Custom constraints such as brands remain semantically owned by their defining handler. I put together a small draft demonstrating that approach: https://github.com/Universal-Commerce-Protocol/ucp-schema/compare/raginpirate/constraint-evaluation <-- not advocating that even moves to draft yet, but a _**quick**_ AI generated validator covering both our constraint validations as well as validations of the submitted instrument using a produced json fragment.

docs/specification/payment-handler-guide.md
```diff
@@ -508,19 +572,20 @@ multiple instrument types for different payment flows.
...
+| [`available_payment_instrument.json`](site:schemas/shopping/types/available_payment_instrument.json) | Open type, `required`, address, and credential keys  |
+| `card_payment_instrument.json#/$defs/available_card_payment_instrument`                              | `type: "card"`, `brands`, and credential refinements |
 
-Handlers reference these instrument-defined schemas when declaring
-`available_instruments`. The **instrument schema authors** define what
-constraints are meaningful (e.g., `brands` for cards), and **platforms/businesses** use this to advertise what they support (e.g., `["visa", "mastercard"]`).
+The base Payment Handler intentionally remains open: an instrument `type` does
+not globally select a schema. After negotiation, consumers use
+`$defs.{handler_name}.available_payment_instrument` from the handler schema.
+Handlers that add no availability-specific keys can omit this definition.
```
jamesandersen, line 588: A few concrete examples would help here — this is how the three use cases map onto the primitives. Suggested insertion:

````suggestion
Handlers that add no availability-specific keys can omit this definition.

**Common constraint examples**

The same three primitives express the requirements handlers most often advertise
upfront. Each is an entry a handler publishes under `available_instruments[]`.

Require a card verification value (CVV) for card payments:

<!-- ucp:example skip reason="instrument availability instance" -->
```json
{
  "type": "card",
  "constraints": {
    "credentials": [
      { "type": "card", "constraints": { "required": ["cvc"] } }
    ]
  }
}
```

Require Address Verification System (AVS) fields from the billing address:

<!-- ucp:example skip reason="instrument availability instance" -->
```json
{
  "type": "card",
  "constraints": {
    "billing_address": { "required": ["street_address", "postal_code"] }
  }
}
```

Require a network-token cryptogram (and ECI value) on card credentials:

<!-- ucp:example skip reason="instrument availability instance" -->
```json
{
  "type": "card",
  "constraints": {
    "credentials": [
      {
        "type": "card",
        "constraints": {
          "card_number_type": { "const": "network_token" },
          "required": ["cryptogram", "eci_value"]
        }
      }
    ]
  }
}
```
````

**Review by jamesandersen — COMMENTED** [review]
Stepping back to what #288 was originally about — letting a handler say up front what it needs for an instrument so the agent can collect it before checkout — I think this version covers it. I took three practical use cases we'd want these changes to express and worked them through the current shape: requiring a CVV on card, requiring AVS fields from the billing address, and requiring a network-token cryptogram. All three express cleanly, and it's more flexible than #288's old `requires_card_verification` boolean. I've left them as a committable docs suggestion below so they're easy to fold into the guide.

On the base coupling: you're right that an `if type == "card"` dispatcher doesn't belong in the base. It would make the generic handler enumerate instrument schemas, which breaks the openness we wanted to keep. Dropping that.

On the enforcement gap: it's only about catching malformed constraints, not the well-formed ones expressing these use cases. Something like `{ "type": "card", "constraints": { "brands": "visa" } }` (string instead of array) still validates when it shouldn't, but the valid declarations above work regardless.

@igrigorik — WDYT?

**Review by gsmith85 — APPROVED** [review]

File: source/schemas/shopping/types/network_token_credential.json [hunk 6]
```diff
@@ -0,0 +1,68 @@
...
+            "Jane Doe"
+          ]
+        },
+        "cryptogram": {
+          "type": "string",
+          "description": "Transaction cryptogram or dynamic CVC (dCVV), in the long or short form expected by the card network or processor.",
+          "examples": [
+            "gXc5UCLnM6ckD7pjM1TdPA=="
+          ]
+        },
```
jamesandersen: `token_credential.json` sets `ucp_response: "omit"` on `token`; these two need the same. Not a regression from `card_credential`, but this is the PR that makes a network token a legitimate checkout credential, so it can now sit in a session and be echoed back in a response.

```suggestion
        "number": {
          "type": "string",
          "description": "Network token or wallet-provisioned token replacing the underlying FPAN.",
          "ucp_response": "omit",
          "examples": [
            "5204240000004242"
          ]
        },
        "expiry_month": {
          "type": "integer",
          "description": "The month of the token's expiration date (1-12)."
        },
        "expiry_year": {
          "type": "integer",
          "description": "The year of the token's expiration date."
        },
        "name": {
          "type": "string",
          "description": "Cardholder name.",
          "examples": [
            "Jane Doe"
          ]
        },
        "cryptogram": {
          "type": "string",
          "description": "Transaction cryptogram or dynamic CVC (dCVV), in the long or short form expected by the card network or processor.",
          "ucp_response": "omit",
          "examples": [
            "gXc5UCLnM6ckD7pjM1TdPA=="
          ]
        },
```

source/schemas/shopping/types/pan_credential.json
```diff
@@ -0,0 +1,54 @@
...
+          ]
+        },
+        "cvc": {
+          "type": "string",
+          "maxLength": 4,
+          "description": "Card verification code.",
+          "examples": [
+            "223"
+          ]
+        }
```
jamesandersen: Same omission here — a raw PAN and a CVC should never come back in a response. `display.last_digits` already carries what the buyer needs to see.

```suggestion
        "number": {
          "type": "string",
          "description": "Funding primary account number (FPAN).",
          "ucp_response": "omit",
          "examples": [
            "4242424242424242"
          ]
        },
        "expiry_month": {
          "type": "integer",
          "description": "The month of the card's expiration date (1-12)."
        },
        "expiry_year": {
          "type": "integer",
          "description": "The year of the card's expiration date."
        },
        "name": {
          "type": "string",
          "description": "Cardholder name.",
          "examples": [
            "Jane Doe"
          ]
        },
        "cvc": {
          "type": "string",
          "maxLength": 4,
          "description": "Card verification code.",
          "ucp_response": "omit",
          "examples": [
            "223"
          ]
        }
```

source/schemas/shopping/types/pan_credential.json
```diff
@@ -0,0 +1,54 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/pan_credential.json",
+  "title": "PAN Credential",
+  "description": "A card credential carrying a funding primary account number (FPAN). This credential type MUST NOT be used for checkout, only with payment handlers that tokenize or encrypt credentials. CRITICAL: Both parties handling a PAN credential (sender and receiver) MUST be PCI DSS compliant. Transmission MUST use HTTPS/TLS with strong cipher suites.",
```
jamesandersen: Worth stating which credential an autofill-style network token uses. It's PAN-shaped with a dynamic code standing in for the cryptogram, so by wire shape it lands here — but the description currently says "funding primary account number (FPAN)," which reads as excluding it.

The protocol question this surfaces: routing it here also makes it non-checkout-eligible under this schema's "MUST NOT be used for checkout." Is that intended? Fine either way — just flagging so it's a decision rather than a side effect.

```suggestion
  "description": "A card credential carrying a funding primary account number (FPAN). Credential selection follows the shape of the value on the wire rather than its provenance: a network token surfaced in PAN form and verified with a `cvc` — as with credentials where a dynamic verification code proxies the cryptogram — is carried here, while a token verified with a discrete `cryptogram` uses Network Token Credential. This credential type MUST NOT be used for checkout, only with payment handlers that tokenize or encrypt credentials. CRITICAL: Both parties handling a PAN credential (sender and receiver) MUST be PCI DSS compliant. Transmission MUST use HTTPS/TLS with strong cipher suites.",
```

source/schemas/shopping/types/network_token_credential.json
jamesandersen: `token_credential.json` sets `ucp_response: "omit"` on `token`; these two need the same.

The spec already covers this in prose — *Credential Flow & PCI Scope* item 1: "Credentials flow **Platform → Business** only. Businesses **MUST NOT** echo credentials back in responses." But nothing enforces it mechanically. `payment_instrument.credential` carries no `ucp_response`, so the credential object survives into resolved response schemas, and `token_credential.token` is the only credential field anywhere that `resolve` actually strips — every field on `card_credential` is projected into response views today despite that MUST NOT.

Not a regression from `card_credential`, but this is the PR that makes a network token a legitimate checkout credential, so it can now sit in a session and be echoed back.

```suggestion
        "number": {
          "type": "string",
          "description": "Network token or wallet-provisioned token replacing the underlying FPAN.",
          "ucp_response": "omit",
          "examples": [
            "5204240000004242"
          ]
        },
        "expiry_month": {
          "type": "integer",
          "description": "The month of the token's expiration date (1-12)."
        },
        "expiry_year": {
          "type": "integer",
          "description": "The year of the token's expiration date."
        },
        "name": {
          "type": "string",
          "description": "Cardholder name.",
          "examples": [
            "Jane Doe"
          ]
        },
        "cryptogram": {
          "type": "string",
          "description": "Transaction cryptogram or dynamic CVC (dCVV), in the long or short form expected by the card network or processor.",
          "ucp_response": "omit",
          "examples": [
            "gXc5UCLnM6ckD7pjM1TdPA=="
          ]
        },
```

Worth weighing the more general fix instead: a single `"ucp_response": "omit"` on `payment_instrument.credential` would enforce the rule for every credential type at once, including ones handlers define later, rather than per-field across these two schemas. That touches the generic base, which you've pushed back on before here — so your call which shape you prefer. The field-level version above just follows the `token_credential` precedent.

source/schemas/shopping/types/pan_credential.json
jamesandersen: Same omission here — a raw PAN and a CVC should never come back in a response. `display.last_digits` already carries what the buyer needs to see. See the note on `network_token_credential.json` re: doing this once on `payment_instrument.credential` instead.

```suggestion
        "number": {
          "type": "string",
          "description": "Funding primary account number (FPAN).",
          "ucp_response": "omit",
          "examples": [
            "4242424242424242"
          ]
        },
        "expiry_month": {
          "type": "integer",
          "description": "The month of the card's expiration date (1-12)."
        },
        "expiry_year": {
          "type": "integer",
          "description": "The year of the card's expiration date."
        },
        "name": {
          "type": "string",
          "description": "Cardholder name.",
          "examples": [
            "Jane Doe"
          ]
        },
        "cvc": {
          "type": "string",
          "maxLength": 4,
          "description": "Card verification code.",
          "ucp_response": "omit",
          "examples": [
            "223"
          ]
        }
```

source/schemas/shopping/types/pan_credential.json
jamesandersen: Shall we clarify that this credential type is intended for anything "PAN-shaped" e.g. an "autofill" CPAN with a dynamic code standing in for the cryptogram?

Routing it here also makes it non-checkout-eligible under this schema's "MUST NOT be used for checkout." Just flagging so it's a decision rather than a side effect.

```suggestion
  "description": "A card credential carrying a funding primary account number (FPAN). Credential selection follows the shape of the value on the wire rather than its provenance: a network token surfaced in PAN form and verified with a `cvc` — as with credentials where a dynamic verification code proxies the cryptogram — is carried here, while a token verified with a discrete `cryptogram` uses Network Token Credential. This credential type MUST NOT be used for checkout, only with payment handlers that tokenize or encrypt credentials. CRITICAL: Both parties handling a PAN credential (sender and receiver) MUST be PCI DSS compliant. Transmission MUST use HTTPS/TLS with strong cipher suites.",
```

docs/specification/overview.md
jamesandersen, line 2872: The other bullets in this list describe mechanisms and let the reader draw the scope conclusion; this follows that pattern rather than asserting an outcome.
```suggestion
- Using PSP tokenization payment handlers where raw credentials never pass
through the platform
- Presenting pre-provisioned card network tokens
      (`network_token_credential.json`) rather than an FPAN — the platform never
      shares the underlying account number, and the token is unusable without a
      matching cryptogram
```

docs/specification/tokenization-guide.md
jamesandersen, line 63: The flow diagram still names `card/fpan` and `card/dpan`, the old `card_number_type` values

```suggestion
|   +-----------------+                      +-------------------------+  |
|   | source_         |                      | checkout_               |  |
|   | credentials     |    What goes IN      | credentials             |  |
|   |                 |<---------------      |                         |  |
|   | * pan           |                      | What comes OUT          |  |
|   | * network_token |                ----->| * token                 |  |
|   |                 |                      |                         |  |
|   +-----------------+                      +-------------------------+  |
```

jamesandersen: An additional suggestion drawing from #367

```suggestion
Tokenization handlers accept source credentials (e.g., card with FPAN) and
produce checkout credentials (e.g., tokens).

A network token does not always need this round-trip. A handler MAY accept
`network_token_credential.json` as a source credential for tokenization, or MAY
accept it directly as a checkout credential, since the token is unusable without
a matching cryptogram. The handler's specification MUST state which of the two
it accepts.
```

jamesandersen (approved):

raginpirate:
Good doc suggestion @jamesandersen, thanks and adopted!

This is now based on top of the anyOf PR which simplifies the shape in our examples 😄

igrigorik:
Still working through the details, but quick sanity check: `constraints` isn't executable grammar.

From the guide example — the two constraint slots sit side by side (trimmed):

```jsonc
{
  "type": "card",
  "constraints": { "brand": { "enum": ["visa", "mastercard"] } },   // ← bare map
  "ucp": {
    "request_constraints": {                                        // ← Constraint Expression
      "path": "$['payment']['instruments'][?@['type'] == 'card']",
      "required": ["billing_address", "credential"],
      "properties": {
        "billing_address": { "required": ["postal_code", "address_country"] }
      }
    }
  }
}
```

`request_constraints` is a schema — a validator resolves `path` to an instance and evaluates it. `constraints` is `{"type":"object","additionalProperties":true,"minProperties":1}`, a map of domain keys onto `value_constraint` objects. It borrows the grammar's *value* slot without its *object* slot, so nothing can execute it. 

If we're reusing the grammar, `constraints` ought to be one too:

```json
"constraints": {
  "properties": { "brand": { "enum": ["visa", "mastercard"] } }
}
```

I believe the intent **was** to reuse the grammar?

igrigorik (approved):
Nice work! 👍

docs/specification/overview/index.md
amithanda, line 558: This section still discriminates on `credential.card_number_type`, and this PR
deprecates that field in the same diff. After merge the flagship `anyOf` example in the core
overview would teach the sibling-field discriminator that this PR's own rationale is
written to remove: "splitting the credential types is what gives each branch a stable
discriminator, so no rule has to branch on a sibling field."

Three things go stale:

- both branches pin `card_number_type`, which now reads "Deprecated: the credential
  type now carries this distinction"
- the closure paragraph turns on `dpan`, which the split model no longer has
- the `site:` link sends readers to `card_credential.json`, now `"deprecated": true`

The corrected version already exists in this PR's own description, using
`credential.type` with `{"const": "pan"}` and `{"const": "network_token"}`, which
also removes the need for `card_number_type` in each branch's `required`, since
`payment_credential.json` already requires `type`.

This is the same sweep you started on the tokenization guide flow diagram for the
`card/fpan` and `card/dpan` values. Could we pull the description's version of the
example into this section so the credential split and the example that depends on it
land together?

amithanda:
Added a minor non-blocking comment, also seems like there is a merge conflict with `/overview/index.md` that needs to be take care of before we can merge.  PTAL.

raginpirate:
Ah, documentation whiplash! Good catch @amithanda, ready for a final review!

amithanda (approved):

PR #267, opened 2026-03-16
feat: extend warning to support disclosure contract

igrigorik:
Resolves #222

Extends `message_warning` with a `presentation` field and two new optional fields (`image_url`, `url`). When `presentation: "disclosure"`, the warning carries a notice — safety warnings, allergen declarations, compliance content, etc. — that platforms MUST display in proximity to the referenced component and MUST NOT hide, collapse, or auto-dismiss.

The `presentation` field defines the rendering contract axis for warnings. Two values are defined today:

- `"notice"` (default) — current warning behavior: MUST display, MAY dismiss
- `"disclosure"` — MUST display in proximity to `path`, MUST NOT dismiss, MUST escalate via `continue_url` if cannot honor

### Why extend warning rather than a new message type or `line_item.disclosures[]`

This is an alternative to both `line_item.disclosures[]` (#222) and a separate `type: "disclosure"`:

- `messages[]` already works across all surfaces (checkout, cart, catalog_lookup, catalog_search)
- Path-based targeting enables product-level inheritance: a disclosure on `$.products[0]` applies to all variants
- Graceful degradation: platforms that don't recognize the `presentation` field still render the underlying warning
- Preserves the established error/warning/info model without setting a precedent for type proliferation

### Why `presentation` (string) rather than `disclosure` (boolean)

- `code` = **what** the warning is about (allergens, prop65, etc.)
- `presentation` = **how** the platform must render it (notice, disclosure)

Provides a natural home for future rendering contracts. For example, an `acknowledge` value could fill the gap between display-only (`disclosure`) and full escalation (`requires_buyer_review`) — where the platform natively collects buyer confirmation without handing off to merchant UI. That value isn't needed today, but the axis is ready for it without migration.

### Examples

**Catalog — allergen disclosure targeting a product (all variants inherit):**

```json
{
  "ucp": { "version": "2026-01-11", "status": "success" },
  "products": [
    {
      "id": "prod_nut_butter",
      "title": "Artisan Nut Butter Collection",
      "variants": [
        {
          "id": "var_almond",
          "title": "Almond Butter",
          "price": { "amount": 1299, "currency": "USD" },
          "availability": { "available": true }
        },
        {
          "id": "var_cashew",
          "title": "Cashew Butter",
          "price": { "amount": 1499, "currency": "USD" },
          "availability": { "available": true }
        }
      ]
    }
  ],
  "messages": [
    {
      "type": "warning",
      "code": "allergens",
      "path": "$.products[0]",
[...]

PR #428, opened 2026-05-10
docs(messages): rename presentation: "disclosure" to "compliance"

ShopKlaw:
## Summary

Rename the `presentation` enum value `"disclosure"` to `"compliance"` so the mode name describes the category of message (regulatory compliance) rather than the rendering verb (disclosing). Reads more naturally alongside the default `"notice"` mode for general advisories.

Follow-up to #267, which introduced the strict-display `"disclosure"` mode. Rendering contract is unchanged — only the value name.

## Why

`disclosure` reads as a verb / general concept; almost every warning is technically "disclosing" something. The mode is specifically for messages that exist to satisfy a regulatory display requirement (Prop 65, GPSR responsible-person, allergen declarations, energy labels). `compliance` names that intent directly.

## What changes

- `source/schemas/shopping/types/message_warning.json` — description text on the `presentation` field
- `docs/specification/checkout.md` — table column header, the `#### \`compliance\`` section heading, prose references to the value, JSON example
- `docs/specification/catalog/index.md` — prose references to the value, JSON example

## What is intentionally NOT changed

- `message_warning` type itself stays — `presentation: "compliance"` is a mode of the warning slot, not a replacement for it. The message-type axis (error/info/warning) is severity-based and remains orthogonal to category.
- References to "disclosure" as a real-world concept (e.g. "allergen disclosure", "the disclosure rendering contract", "Disclosure and Acknowledgment") are preserved. Only the literal enum value is renamed.

## Migration

`presentation` has no enum constraint in the schema (free-form string with documented values), so this is a documentation-level change. Implementations currently emitting `presentation: "disclosure"` should update to `"compliance"`. There is no value-coexistence story; the rename is a clean replacement.

Opening as draft for discussion.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/specification/catalog/index.md
+++ b/docs/specification/catalog/index.md
@@ -165,7 +165,7 @@ Messages communicate business outcomes and provide context:
 | `warning` | Important conditions affecting purchase | `DELAYED_FULFILLMENT`, `FINAL_SALE` |
 | `info` | Additional context without issues | `PROMOTIONAL_PRICING`, `LIMITED_AVAILABILITY` |
 
-Warnings with `presentation: "disclosure"` carry notices (e.g., allergen
+Warnings with `presentation: "compliance"` carry notices (e.g., allergen
 declarations, safety warnings) that platforms must not hide or dismiss. See
 [Warning Presentation](../checkout.md#warning-presentation) for the full
 rendering contract.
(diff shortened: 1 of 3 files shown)


ShopKlaw:
Closing. Legal is sticking with the name disclosure

PR #435, opened 2026-05-11
feat(schema): add one-time use hint for shipping destinations

gil--:
## Motivation

Today there's no way for a platform to signal that a shipping destination
should not be retained beyond the immediate transaction — e.g., a gift,
gift-registry shipment, or care-of address that belongs to someone other than
the buyer. Without a hint, businesses default to persisting every checkout
address into the buyer's address book, marketing audience, or personalization
profile.

## Changes

- Add optional `one_time_use: boolean` (default `false`) to `shipping_destination`
- Document behavior contract in `fulfillment.md` as a top-level section with a
  Gift Example under `## Examples`
- No separate business-profile support declaration; businesses apply the hint
  on a best-effort basis

```json
{
  "id": "dest_recipient",
  "street_address": "123 Recipient St",
  "address_country": "US",
  "one_time_use": true
}
```

Orthogonal to `dev.ucp.shopping.buyer_consent`, which carries per-buyer
consent signals (marketing, analytics, etc.). `one_time_use` is per-destination
and applies to that address alone.

## Validation

- `ucp-schema lint source/schemas/shopping/types/shipping_destination.json`
- `ucp-schema resolve --request --op create --pretty source/schemas/shopping/types/shipping_destination.json`
- `ucp-schema resolve --response --op read --pretty source/schemas/shopping/types/shipping_destination.json`
- `uv run --frozen mkdocs build`


gil--:
Two open design questions I'd like input on before merge:

### 1. Should `one_time_use` live on `postal_address` instead of `shipping_destination`?

The privacy attribute is about *the address*, not its shipping role. If we ever add billing addresses, pickup contacts, or other `postal_address` consumers that benefit from a "don't retain" hint, they'd each need a parallel field. Lifting `one_time_use` to `postal_address.json` is a small schema move (the field is already optional with `default: false`) and future-proofs reuse.

Trade-off: the privacy hint becomes available everywhere `postal_address` is referenced, including places where it may not be meaningful today (e.g., business return addresses).

### 2. Core field vs `dev.ucp.shopping.*` extension?

`dev.ucp.shopping.buyer_consent` — the only other privacy-adjacent feature — is structured as an extension with capability advertisement. This PR is proposed for core fulfillment with no capability bit, on the rationale that it's a best-effort hint and doesn't gate behavior. Both framings are defensible:

- **Core**: keeps the field universally available; matches Shopify's `DeliveryAddressInput.oneTimeUse` placement; avoids capability negotiation for what is effectively a passive flag.
- **Extension**: parallel to `buyer_consent`; lets businesses advertise compliance; preserves a consistent governance pattern for privacy signals.

If there is preference, happy to restructure before merge. Otherwise I'll proceed with current shape (core + on `shipping_destination`).

karangoel16 (commented):
Thanks for putting this together — the privacy intent is clearly the right thing to add. I'd push back on the *shape* though: I think `one_time_use: boolean` should be an enum like `address_use: "default" | "gift" | "care_of" | "one_time"`.

A few reasons:

**The motivation already lists distinct cases.** Gift, gift-registry, and care-of all happen to collapse to "don't persist" today, but the intent differs in ways businesses may want to act on — gift suggests suppressing *recipient* marketing entirely, care-of is "not the buyer's address but may be a known recurring relationship," one-time is a pure transactional hint. A boolean throws away that signal at the schema layer, so businesses can never recover it even if they want to behave differently.

**Booleans don't extend.** If a fourth retention mode shows up later (e.g., "retain for fulfillment/fraud only, never enrich profile"), there's no clean way to add it without introducing a second boolean and spec'ing the combinations. An enum grows by adding a value; the field stays one field.

**`default: false` encodes "persist" as a silent contract.** That's a load-bearing default expressed as an absence. An explicit `"default"` enum value makes the persistence policy a deliberate choice on every destination, which feels more honest given how sensitive the underlying behavior is (address-book writes, audience enrichment).

**Industry precedent.** Stripe and others have publicly walked back boolean state/mode fields (`paid`, `refunded`, etc.) toward enums for exactly this reason — once a boolean ships, you're stuck with two states forever.

Concrete proposal:

```json
"address_use": {
  "type": "string",
  "enum": ["default", "gift", "care_of", "one_time"],
  "default": "default",
  "ucp_request": "optional",
  "description": "Intent for this destination. Businesses SHOULD apply persistence and marketing-enrichment policy based on the declared intent; values other than `default` are best-effort hints that the address should not be retained beyond the immediate transaction or used to enrich the buyer's profile."
}
```

Behavior contract in the prose stays almost the same — just keyed off `address_use != "default"` instead of `one_time_use == true`, with the spec free to layer per-value guidance later (e.g., "for `gift`, businesses SHOULD also suppress order-status marketing CTAs to the recipient").

Happy to be wrong if there's prior art on the boolean shape I'm missing — but if not, I'd rather pay the small upfront cost of an enum than lock the field into a binary it can't grow out of.

PR #436, opened 2026-05-12
chore: Refactor schema references to common types

amithanda:
# refactor: move generic types from `shopping/` to `common/types/`

## Rationale

All types currently live under `shopping/types/`, but many of them carry no shopping-specific semantics. As UCP is envisioned to expand to other verticals beyond shopping, those verticals will need the same primitives — money amounts, addresses, pagination, messaging, rich text — without taking a dependency on the shopping namespace. This refactor establishes `common/types/` as the canonical home for types that belong to the protocol, not to a vertical.

This also fixes a pre-existing namespace violation: `common/identity_linking.json` was referencing `../shopping/types/description.json`, which is the exact problem this refactor corrects structurally.

---

### Phase 1 — Pure primitives
No vertical semantics. Any future capability will need these.

`amount`, `signed_amount`, `price`, `description`, `media`, `postal_address`, `pagination`, `reverse_domain_name`, `link`, `buyer`

### Phase 2 — Protocol messaging infrastructure
UCP's error/warning/info messaging pattern. Every vertical reuses this plumbing verbatim.

`message`, `message_error`, `message_warning`, `message_info`, `error_response`, `error_code`, `warning_code`, `info_code`

### Phase 3 — Cross-vertical candidates *(future PR)*
Strong candidates with clear cross-vertical semantics, deferred to validate design stability first.

`attribution`, `signals`, `context`, `rating`

### Phase 4 — Transactional primitives *(future PR)*
Financial breakdown types whose schema may need to evolve per vertical before being promoted.

`total`, `totals`, `price_range`, `price_filter`

### Phase 5 - Payment types *(future PR)*
`payment_instrument`, `payment_credential`, `payment_identity`, `card_credential`, `card_payment_instrument`, `token_credential`, `available_payment_instrument` — these are cross-vertical but domain-specific. Should they be in their own `payment/types/` namespace?

---

## This PR (Phase 1 + Phase 2)

- Creates `source/schemas/common/types/` with 18 new files, each with updated `$id` (`https://ucp.dev/schemas/common/types/...`)
- Deletes the 18 originals from `source/schemas/shopping/types/`
- Updates all `$ref` paths in:
  - `shopping/types/*.json` (remaining shopping types that reference moved types)
  - `shopping/*.json` (capability schemas: cart, checkout, order, fulfillment, etc.)
  - `source/schemas/ucp.json`
  - `source/schemas/common/identity_linking.json` (fixes pre-existing violation)
  - `source/services/shopping/` (OpenRPC and OpenAPI service definitions)
- Adds `source/schemas/common` to `SCHEMAS_DIRS` in `main.py` so the docs macro system resolves moved types correctly

No schema semantics were changed — this is a pure relocation with path updates.


source/schemas/common/types/buyer.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://ucp.dev/schemas/shopping/types/buyer.json",
+  "$id": "https://ucp.dev/schemas/common/types/buyer.json",
```
yanheChen: Is `buyer` really vertical-agnostic? I feel that a non-shopping vertical wouldn't naturally call the human interacted with it as a `buyer`. I knew that we use buyer as a protocol layer word, e.g. `requires_buyer_input` but maybe if we want reconsider those as well.

Worth considering `user`, or `party` (or keeping it shopping-only and reintroducing as a more neutral type when cross-vertical work happens). 

If people feel is comfortable that `buyer` reads naturally across verticals, fine to keep — just want to flag it as a one-time decision.

yanheChen (approved):

drewolson-google (approved):

jamesandersen (approved):
thanks @amithanda for setting these types up for more broad use

source/schemas/common/types/buyer.json
amithanda: Good observation @yanheChen . Trying to keep this PR focussed on pure refactoring avoiding broader discussion on changing names (as changing the name might have some other side effects) but we can probably bring this up in a follow up PR. Let me know if that is ok.

yanheChen: yes that makes sense! Thanks for taking consideration of my comment.

source/schemas/common/identity_linking.json
```diff
@@ -13,7 +13,7 @@
       "description": "Per-scope policy and metadata — auth constraints (e.g. min_acr, max_token_age), declarative metadata (e.g. claims produced, consent descriptions), or any other scope-specific configuration. An empty object means user authentication is required with no additional policy. Open for non-breaking extension.",
       "properties": {
         "description": {
-          "$ref": "../shopping/types/description.json",
+          "$ref": "./types/description.json",
```
jingyli: nit: We probably don't need the `./` prefix to be consistent with other existing references to the same directory level (i.e. https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/shopping/order.json#L59).

source/schemas/common/types/error_code.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/common/types/error_code.json",
+  "title": "Error Code",
+  "description": "Error code identifying the type of error. Standard errors are defined in capability specifications and have standardized semantics; freeform codes are permitted.",
+  "type": "string"
```
jingyli: Looks like we have dropped some of the [example strings](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/shopping/types/error_code.json#L7) that are present in error_code.json today, is this intended? It might be good to preserve since it's just some common well-known examples.

Same comment to `warning_code.json` and `info.code.json`: we probably should move all the examples as part of this refactoring.

jingyli (approved):

source/schemas/common/types/error_code.json
amithanda: Thanks for catching that. Fixed.

source/schemas/common/types/buyer.json
jingyli: +1 on the comment thread here: I would vouch for the approach of keeping `buyer` as a shopping-only concept instead of including it in the refactoring since buyer does introduce the connotation that some kind of goods is being purchased/exchanged. But zooming out to non-shopping verticals, the concept of purchasable goods may not always exist.

main.py
```diff
@@ -33,6 +33,8 @@
 SCHEMAS_DIRS = [
   Path("source/handlers/google_pay"),
   Path("source/schemas"),
+  Path("source/schemas/common"),
+  Path("source/schemas/common/types"),
```
igrigorik: IMO, we should enforce the basename-uniqueness invariant at build time rather than leaving it implicit. 

Right now the tool resolves schemas by basename across `SCHEMAS_DIRS` (`_load_json_file`, the redirect logic in `create_link`), but JSON Schema's `$ref` resolver resolves by path. That divergence means a future collision — e.g. someone adds `shopping/types/buyer.json` while `common/types/buyer.json` exists — passes JSON Schema validation but silently misbinds in the rendered docs (wrong anchors, wrong "see reference" redirect target). No error, just wrong links.

Suggest something like:

```python
def _validate_schema_basename_uniqueness() -> None:
    seen: dict[str, Path] = {}
    for d in SCHEMAS_DIRS:
        if not d.exists():
            continue
        for p in d.glob("*.json"):
            if p.name in seen:
                raise RuntimeError(
                    f"Schema filename collision: '{p.name}' exists in "
                    f"both {seen[p.name]} and {p}. "
                    f"Filenames must be unique across SCHEMAS_DIRS — "
                    f"doc resolution is basename-based."
                )
            seen[p.name] = p

_validate_schema_basename_uniqueness()
```

Cheap, fails fast at build time, and the error message tells the next person exactly what they did. If we ever want a vertical to intentionally shadow a common type, that's a real design conversation worth surfacing — not something to allow by accident.

source/schemas/common/identity_linking.json
igrigorik: +1

igrigorik:
TY for restoring error codes, that was my main flag on firt pass. 

Pedantic nice: technically this change _may_ be breaking because we're changing $id pointers and URLs from published schemas. If any code or validators cached that state, they'll get 404's. My intuition is that we should be OK, but this is something to watch out for... let's see if and what breaks.

Question: how do we want to distinguish, or not, common vs domain-specific schemas? Right now they're indistinguishable in how we render then in our tables. Should we separate and note that some are common vs not? This also connects to my flag on `main.py` behavior where we can end up in shadow / duplicate schemas unless we put some strict guards around it.

source/schemas/common/identity_linking.json
amithanda: updated, thanks for catching this.

main.py
amithanda: Great call @igrigorik — I implemented a guard, but scoped it more narrowly than your original suggestion
after thinking through the multi-vertical future.

## Why not the global basename uniqueness check

Your suggested guard enforces uniqueness across **all** of `SCHEMAS_DIRS`. That would break as soon as a
second vertical is introduced. Consider:
source/schemas/shopping/checkout.json ← shopping checkout flow
source/schemas/new-vertical/checkout.json ← new-vertical booking checkout flow

These are legitimately different schemas that share a filename because they represent the same protocol
concept in different verticals. A global guard would either fail the build or force an artificial rename
like `new-vertical_checkout.json` — both wrong outcomes. Multi-vertical same-filename is intentional design, not a collision.
## The invariant that actually matters: common vs vertical
The dangerous collision is specifically between `common/types/` and a vertical's `types/` directory:
source/schemas/common/types/amount.json ← canonical common type
source/schemas/shopping/types/amount.json ← shadow / duplicate

This is always a mistake. The docs macro system resolves schemas by basename (first match in
`SCHEMAS_DIRS`), while JSON Schema's `$ref` resolver resolves by full path. A common/vertical collision means `_load_json_file("amount")` and `create_link("amount.json", ...)` silently pick whichever appears first in `SCHEMAS_DIRS` — wrong anchors, wrong "see reference" redirects, no error. Unlike two verticals sharing a name, there is no legitimate design intent behind having the same type in both `common/` and a vertical.
## What we implemented
```python
COMMON_TYPES_DIR = Path("source/schemas/common/types")
VERTICAL_TYPES_DIRS = [
  SHOPPING_SCHEMAS_DIR / "types",
  # Add new vertical types directories here as verticals are introduced.
]
def _validate_common_vertical_uniqueness() -> None:
    if not COMMON_TYPES_DIR.exists():
        return
    common_names = {p.name for p in COMMON_TYPES_DIR.glob("*.json")}
    for vertical_dir in VERTICAL_TYPES_DIRS:
        if not vertical_dir.exists():
            continue
        for p in vertical_dir.glob("*.json"):
            if p.name in common_names:
                raise RuntimeError(
                    f"Schema filename collision between common and vertical: "
                    f"'{p.name}' exists in both {COMMON_TYPES_DIR} and {vertical_dir}. "
                    f"A type cannot be defined in both common/types/ and a vertical — "
                    f"remove it from one or rename it."
                )
_validate_common_vertical_uniqueness()
Runs at module load time — the build fails immediately with a message that names both conflicting paths.
```
Let me know if you see any issues or have more feedback.

amithanda:
> Question: how do we want to distinguish, or not, common vs domain-specific schemas? Right now they're indistinguishable in how we render then in our tables. Should we separate and note that some are common vs not? This also connects to my flag on `main.py` behavior where we can end up in shadow / duplicate schemas unless we put some strict guards around it.

You're right that the reference page currently renders all types indistinguishably:
Type Schemas
Amount ← common/types/ (no label)
Buyer ← shopping/types/ (no label)
Message ← common/types/ (no label)
Product ← shopping/types/ (no label)
This matters for implementors building non-shopping verticals who need to know at a glance which types
are safe to reuse. Fixing it requires changes to `reference.md` and `auto_generate_schema_reference`
(section split, or a "Common" badge, or both). Intentionally leaving this for a follow-up so the
presentation choice gets proper discussion rather than being decided quietly inside a refactor PR.
## Shadow/duplicate schema guard
Implemented — but we scoped the guard more narrowly than your suggestion after thinking through the
multi-vertical future. PTAL at my other comment.

amithanda:
Following up on @yanheChen and @jingyli's comments — agreed that `buyer` carries shopping-domain
semantics and shouldn't be declared a cross-vertical primitive prematurely.

Reverted `buyer` back to `shopping/types/` in this PR. The reasoning:

- The name encodes a purchase transaction — it may not generalize cleanly to other verticals 
- Promoting it to `common/` would lock in shopping-centric field names (`first_name`, `last_name`,
  `email`, `phone_number`) as the universal person model before UCP has any non-shopping verticals
  to validate that shape against
- The right pattern is to introduce a neutral type (e.g. `party` or `contact`) when the first
  non-shopping vertical is designed and its person-identity requirements are actually known — not
  to rename a shopping type preemptively and carry the debt forward

`common/types/` now contains 18 types, all of which have zero vertical-specific semantics. `buyer`
remains in `shopping/types/` unchanged.

The neutral cross-vertical person type is tracked as a follow-up for when the first non-shopping
vertical is scoped.

main.py
igrigorik: @amithanda agree on the reframe, the key is to ensure common has strict non-overlap contract. 

I don't see above code in the commit log though?

amithanda: Sorry, It was still sitting in my local branch, I thought I had pushed it. PTAL.

igrigorik (approved):
@amithanda pushed a small improvement to the validation logic, please sanity check.

LGTM.

amithanda:
> @amithanda pushed a small improvement to the validation logic, please sanity check.
> 
> LGTM.

@igrigorik - Thanks for the cleanup and auto-discovery in `72eca50`! 

However, zooming out to the multi-vertical design, there is a fundamental gap in enforcing global filename uniqueness across all vertical namespaces:

1. **Top-Level Capabilities Remain Unprotected:** The current check only validates filenames under the `types/` subdirectories. It does not protect capability schemas (e.g., `checkout.json`, `order.json`, `fulfillment.json`). If a second vertical introduces `new-vertical/checkout.json` alongside `shopping/checkout.json`, `main.py` will still suffer from the exact same silent resolution bug (loading whichever vertical appears first in `SCHEMAS_DIRS` for all pages) without triggering this guard.
2. **Restrictions on Vertical Autonomy:** Enforcing global name uniqueness forces different verticals to use different names for parallel concepts (e.g., we'd have to use `new-vertical_status.json` vs. `shopping_status.json` instead of clean, vertical-scoped `types/status.json` files).

### Proposal: A Path-Aware Resolver

Instead of using strict filename constraints to workaround the resolver's flat-search design, we should make the resolver path-aware. We can inspect the MkDocs compilation context (`env.page.file.src_path`) to identify which vertical's documentation page is being compiled, and prioritize that vertical's folders in the lookup path.

Here is a quick sketch of how we can achieve this in `main.py`:

```python
def get_active_vertical() -> str:
  try:
    # src_path is like "specification/new-vertical/checkout.md" 
    # or "specification/checkout.md" (defaults to shopping)
    src_path = Path(env.page.file.src_path)
    parts = src_path.parts
    if len(parts) > 2 and parts[0] == "specification":
      vertical = parts[1]
      # Verify against discovered verticals
      if vertical in [d.name for d in VERTICAL_DIRS]:
        return vertical
  except AttributeError:
    pass
  return "shopping"  # Default vertical fallback
```

Then, in `_load_json_file(entity_name)`:

```python
def _load_json_file(entity_name):
  """Try loading a JSON file prioritizing the active vertical."""
  active_vertical = get_active_vertical()
  
  # Construct prioritized search paths for the active vertical
  active_schemas_dir = SCHEMAS_DIR / active_vertical
  active_types_dir = active_schemas_dir / "types"
  
  search_dirs = []
  if active_types_dir.exists():
    search_dirs.append(active_types_dir)
  if active_schemas_dir.exists():
    search_dirs.append(active_schemas_dir)
    
  # Append remaining directories in standard order
  for d in schemas_dirs:
    if d not in search_dirs:
      search_dirs.append(d)
      
  # Resolve file
  for schemas_dir in search_dirs:
    full_path = Path(schemas_dir) / (entity_name + ".json")
    try:
      with full_path.open(encoding="utf-8") as f:
        return json.load(f)
    except FileNotFoundError:
      continue
  return None
```

### Next Steps

If this looks good, we can:
1. Revert `72eca50` to remove the vertical-to-vertical type name constraints.
2. Keep the `common` vs. `vertical` uniqueness check (since a common type should never be shadowed by a vertical).
3. Implement this path-aware lookup, which guarantees that pages for any vertical will always resolve their own schemas correctly, without restricting filenames.

OR

We can come back and look at build time enforcement once we have some idea on the patterns that are emerging.

igrigorik:
>Restrictions on Vertical Autonomy: Enforcing global name uniqueness forces different verticals to use different names for parallel concepts (e.g., we'd have to use new-vertical_status.json vs. shopping_status.json instead of clean, vertical-scoped types/status.json files).

Right, I see this as a feature, not a bug. If we promote something to common, then let's not create a shadow override in sub-verticals -- that violates the contract of common.

amithanda:
I think there are 3 issues we should discuss:
---

### 1. How should we generate documentation clearly for each vertical? (Path-Aware Resolver)

* **The Problem:** 
The MkDocs macro system (`main.py`) resolves JSON schemas by searching for their basenames (e.g., `'checkout'` or `'types/status'`) across a flat list of directories configured in `SCHEMAS_DIRS`. The first folder match wins. When multiple verticals exist, this flat lookup causes silent cross-vertical rendering failures.

* **Example (Silent Doc Corruption):**
Suppose we have a new vertical `new-vertical` alongside `shopping`, and both define a capability schema `checkout.json`:
1. `source/schemas/shopping/checkout.json` (defines shopping-specific lines, totals, cart links)
2. `source/schemas/new-vertical/checkout.json` (defines new-vertical-specific reservation details, identifiers)

Since `new-vertical` comes first alphabetically (or is ordered before `shopping`), `source/schemas/new-vertical` appears before `source/schemas/shopping` in `SCHEMAS_DIRS`. 
When compiling the **shopping** checkout page, calling `{{ schema_fields('checkout', 'checkout') }}` will scan `SCHEMAS_DIRS`, find `new-vertical/checkout.json` first, and silently render the new-vertical schema on the shopping documentation page without any build warnings or errors.

* **Recommended Solution:**
Implement a **Path-Aware Resolver**. We inspect the MkDocs compiler context (`env.page.file.src_path`) to identify which vertical's page is being built and prioritize that vertical's schema and types directories at the top of the search path.
* For page `specification/new-vertical/checkout.md`, search path starts with: `[new-vertical/types, new-vertical]`
* For page `specification/checkout.md` (shopping default), search path starts with: `[shopping/types, shopping]`

---

### 2. Should we enforce that vertical and common namespaces cannot have overlapping names?

* **The Question:**
Should we allow a vertical to define a type name that already exists in the core protocol namespace (e.g., `common/types/`)?

* **Example:**
If `common/types/amount.json` exists as the protocol's canonical definition of money, should a vertical be allowed to define `source/schemas/shopping/types/amount.json`?

* **Recommendation:**
**Enforce strict uniqueness (no overlaps).**
The common namespace defines universal protocol primitives. Shadowing a core common type with a vertical-specific version of the same name introduces severe developer confusion and breaks specification predictability. If a vertical needs a different shape, it should define a distinctly named type (e.g., `types/billing_amount.json`) or propose an update to the common type. We should maintain a build-time guard that fails fast on any common-to-vertical name collision.

---

### 3. Should we enforce that filenames be unique between different verticals?

* **The Question:**
Should we force sibling verticals to have globally unique filenames across the entire codebase? 

* **Example:**
If the `shopping` vertical defines a type `shopping/types/guest.json`, should the `new-vertical` vertical be forbidden from defining a type `new-vertical/types/guest.json`?

* **Recommendation:**
**Allow overlapping names between verticals (do NOT enforce uniqueness).**
Each vertical represents a separate, modular application domain. Forcing different verticals to coordinate filenames globally goes against vertical autonomy and adds unnecessary prefix boilerplate (e.g., renaming files to `shopping_guest.json` and `new-vertical_guest.json`). 

Once the **Path-Aware Resolver (Point 1)** is implemented, the documentation build will resolve vertical-specific files correctly based on the compilation context. Sibling verticals should have the freedom to design their namespaces independently using clean, intuitive terms.

igrigorik:
@amithanda updated per our TC discussion, PTAL -- right shape?

PR #443, opened 2026-05-16
docs(specification): centralize shared types per reference.md (#412)

### LINKED ISSUE #412: Complete the shared-type centralization started in #226
[PR #226](https://github.com/Universal-Commerce-Protocol/ucp/pull/226) (Ryan C, 2026-03-10) introduced the convention that `reference.md` is the canonical render destination for all `types/` schemas. The fix:

1. Added a link checker to CI
2. Expanded `reference.md` to render shared types (Pagination, Error Code, etc.) natively
3. Added a redirect rule in `main.py:311-313` so any cross-reference to a `types/` schema resolves its anchor to `reference.md`, regardless of which capability page is rendering it

Before #226, capability pages each rendered their own `types/` schemas inline, with intra-page anchors. Cross-references between types could break because not every page had every anchor. #226's fixes this.

## Gap

#226 made the cross-references resolve to `reference.md`, but did not remove the inline schemas from capability pages. So today:

- `reference.md` is the canonical destination for all `types/` cross-references — anchors and links go there
- `cart.md`, `checkout.md`, `catalog/index.md` still render the same `types/` tables inline (Buyer, Context, Signals, Attribution, Total, Line Item, Message, Link, etc), but they're not used in internal refs, everything gets routed to `reference.md`.
  - The same tables get rendered three times: once in each capability page, once in `reference.md`
  - The macro's second `spec_file_name` argument is dead weight for `types/...`, can/should be removed.

## Desired end state

1. **Capability pages stop rendering `types/...` tables inline.** Each entity section keeps its heading and brief prose context (the "what is this and why is it here" paragraph), but links to `reference.md#<anchor>` instead of duplicating the table render.
2. **`reference.md` is the single source of truth** for shared-type field tables. It already is functionally — this just makes it consistent.
3. **`schema_fields` macro signature simplifies.** The second arg either:
   - Gets removed entirely (and the redirect logic becomes the default), or
   - Gets enforced (callers passing `'cart'` or `'checkout'` for a `types/...` schema fail loudly), making misuse visible.

## Tradeoff to acknowledge

Cross-linking out to `reference.md` from capability pages **breaks the flow of reading a single page end-to-end.** A reader on `cart.md` who wants to understand what fields `signals` carries has to navigate to `reference.md`, find the section, and navigate back. Self-contained capability pages are nicer to read.

The counter-argument — and the reason #226 went this way in the first place — is that UCP has many shared types, and replicating all of them across every capability page is tedious to maintain and creates anchor-resolution problems. Centralizing into `reference.md` was the pragmatic call; we just didn't finish the cleanup.

sriakula1:
## Description

Closes #412.

[PR #226](https://github.com/Universal-Commerce-Protocol/ucp/pull/226) made `reference.md` the canonical render destination for all `types/` schemas — cross-references already resolve there via the redirect in `create_link`. The gap @igrigorik called out: capability pages still re-render the same `types/...` field tables inline, so each shared type (Buyer, Signals, Attribution, Total, Line Item, Message, Link, …) is rendered three times — once each in `cart.md`, `checkout.md`, `catalog/index.md` — alongside the canonical render in `reference.md`.

From the issue, the desired end state:

> 1. **Capability pages stop rendering `types/...` tables inline.** Each entity section keeps its heading and brief prose context (the "what is this and why is it here" paragraph), but links to `reference.md#<anchor>` instead of duplicating the table render.
> 2. **`reference.md` is the single source of truth** for shared-type field tables.

This PR implements that without churning a single `.md` file.

## What changed

`main.py` only. Two macros now detect when the requested schema lives under a `types/` directory (`common/types/` or any `<vertical>/types/`) and, when the caller is anything other than `reference.md`, emit a Markdown link to the canonical entry instead of rendering the table inline:

* `schema_fields(entity, spec)` — handles calls like `schema_fields('types/signals', 'cart')`, `schema_fields('buyer', 'checkout')`, `schema_fields('types/line_item_create_req', 'cart')`, etc. Renders:

  ```
  See [Signals](site:specification/reference/#signals) in the [Schema Reference](reference.md) for the canonical field definition.
  ```

* `extension_schema_fields(entity, spec)` — handles calls like `extension_schema_fields('types/pagination.json#/$defs/response', 'cart')` and `extension_schema_fields('types/payment_instrument.json#/$defs/selected_payment_instrument', 'checkout')`. Renders a single anchor link.

* `reference.md` still renders the full tables (via `auto_generate_schema_reference` and direct macro calls with `spec_file_name='reference'`); the source of truth is unchanged.

All existing capability-page macro calls (`{{ schema_fields('types/signals', 'cart') }}`, etc.) work unmodified — no `.md` edits required. The surrounding prose context already present on each capability page (e.g. cart.md's paragraphs above Signals/Attribution/Total) is preserved untouched.

### Tradeoff acknowledged

As @igrigorik noted in the issue: cross-linking out to `reference.md` does break the flow of reading a single capability page end-to-end. The counter-argument — that UCP has many shared types and replicating them across every capability page creates anchor-resolution problems and maintenance churn — is the reason #226 went this way, and is the same reason here.

## Files changed

* `main.py` — added `_resolves_to_shared_type()` and `_render_shared_type_link()` helpers; one early-return branch in `schema_fields`; one early-return branch in `extension_schema_fields`.

## Local build

`mkdocs build` exercises the modified macros and they short-circuit cleanly for every `types/...` capability-page caller. The full build still needs the `ucp-schema` Rust CLI for the remaining (non-types) macro calls on `reference.md` itself — that's unchanged from `main`, and CI runs the canonical build.

Verified via a small harness (mock mkdocs env, call the macro directly):

* `schema_fields('types/signals', 'cart')` → link to `reference.md#signals` ✓
* `schema_fields('buyer', 'checkout')` → link to `reference.md#buyer` ✓
* `schema_fields('types/line_item_create_req', 'cart')` → link to `reference.md#line-item-create-request` ✓
* `extension_schema_fields('types/pagination.json#/$defs/response', 'cart')` → `reference.md#pagination-response` ✓
* `schema_fields('types/signals', 'reference')` → falls through to inline render (correct) ✓

## Deferred (worth a separate issue)

@wry-ry suggested replacing the "Required" column with a "Visibility" column (`required` / `optional` / `omit`) on rendered tables. That's a great direction but needs an explicit `omit` annotation in the schemas, which isn't in the data model today. I left it out of this PR so the diff stays narrow and reviewable — happy to follow up with a separate proposal once the schema-side question is settled.

## Type of change

- [x] Documentation update
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Reviewer notes

Tagging @igrigorik and @wry-ry since you both engaged on the issue. Per `MAINTAINERS.md`, `docs/specification/` ultimately wants tech-council eyes — please route as appropriate.

### DIFF
--- a/main.py
+++ b/main.py
@@ -903,6 +903,44 @@ def define_env(env):
       f" directory{get_error_context()}."
     )
 
+  def _resolves_to_shared_type(base_name):
+    """Return the resolved path if base_name lives in a `types/` directory.
+
+    A "shared type" is any schema under `source/schemas/common/types/` or
+    `source/schemas/<vertical>/types/`. These are canonically rendered in
+    `reference.md` per the convention established in PR #226. Callers from
+    other pages should link out rather than re-render the table.
+
+    Returns the resolved Path on match, or None otherwise.
+    """
+    # Strip leading 'types/' if present so we can match the basename
+    # against the types directories directly.
+    name = base_name.split("/")[-1]
+    for types_dir in (COMMON_TYPES_DIR, *VERTICAL_TYPES_DIRS):
+      candidate = types_dir / (name + ".json")
+      if candidate.exists():
+        return candidate
+    return None
+
+  def _render_shared_type_link(entity_name, spec_file_name):
+    """Emit a Markdown link to `reference.md` for a shared type.
+
+    Used in place of inline-rendering a `types/...` table on capability
+    pages (#412). `create_link` already routes the anchor to
+    `reference.md`, so we reuse it for consistency.
+    """
+    # create_link expects a ref-style path; normalize entity_name so the
+    # anchor logic in create_link matches the rendered heading in
+    # reference.md (which is auto-generated from the schema filename).
+    ref = entity_name if entity_name.startswith("types/") else f"types/{entity_name}"
+    if not ref.endswith(".json"):
+      ref = ref + ".json"
+    link = create_link(ref, spec_file_name)
+    return (
+      f"See {link} in the [Schema Reference](reference.md) "
+      f"for the canonical field definition."
+    )
+
   # --- MACRO 1: For Standalone JSON Schemas ---
   @env.macro
   def schema_fields(entity_name, spec_file_name):
@@ -914,6 +952,11 @@ def define_env(env):
     - 'cart_resp' -> resolves cart.json as response schema
     - 'cart_create_req' -> resolves cart.json as request schema (op=create)
     - 'buyer' -> resolves buyer.json as response schema (default)
+
+    For shared types (any schema under a `types/` directory), callers from
+    pages other than `reference.md` receive a Markdown link to the
+    canonical entry in `reference.md` instead of an inline table. This
+    finishes the centralization started in #226 (see #412).
 
     Args:
     ----
@@ -943,6 +986,16 @@ def define_env(env):
       else:
         base_name = entity_name[:-4]
         direction = "request"
+
+    # Shared-type redirect: when a capability page asks for a `types/...`
+    # schema, render a link to reference.md instead of duplicating the
+    # field table. reference.md itself still renders the full tables via
+    # auto_generate_schema_reference. See #412.
+    if (
+      spec_file_name != "reference"
+      and _resolves_to_shared_type(base_name) is not None
+    ):
+      return _render_shared_type_link(entity_name, spec_file_name)
 
     # Build context for downstream link generation
     context = {"io_type": direction, "operation_id": operation}
@@ -977,6 +1030,11 @@ def define_env(env):
 
     Usage: {{ extension_schema_fields('fulfillment_option') }}
 
+    For shared types referenced from a capability page (e.g.
+    `types/pagination.json#/$defs/response`), this macro emits a link to
+    the canonical entry in `reference.md` instead of rendering the table
+    inline. See #412.
+
     Args:
     ----
       entity_name: The name of the schema entity embedded in the extension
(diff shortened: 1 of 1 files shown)


sriakula1:
@googlebot I signed it!

sriakula1:
@googlebot I signed it!

sriakula1:
Refreshed this PR and added the **Visibility column** @wry-ry asked for in #412.

**Rebased** onto the latest `main` — the branch is now two clean commits (centralize + visibility), `main.py` only.

**Visibility column.** The schema-reference tables now render `Visibility` in place of the binary `Required` column, showing `required` / `optional` / `omit` per request operation **and** response. It's derived from the existing `ucp_request` annotation plus the base `required` array — no schema changes, rendering only. (Note: `omit` already exists in the schemas — `ucp_request: "omit"` is used throughout — so no new annotation was needed.)

Examples from the rendered reference:

| Field | Visibility |
| :--- | :--- |
| `Line Item.id` | **Required** in responses; omitted on create, optional on update |
| `Line Item.totals` | **Required** in responses; omitted in requests |
| `Checkout.line_items` | **Required** in responses; required on create & update, omitted on complete |
| `Checkout.buyer` | Optional in responses; optional on create & update, omitted on complete |

Fields with no request-specific annotation render as a simple **Required** / Optional, as before. Operation-specific request/response tables on capability pages (resolved via `ucp-schema`) keep showing the effective state for that operation.

The cell format is a judgement call since it wasn't specified — happy to adjust the wording/layout (e.g. collapse request clauses that match the response, or split into separate columns) if you'd prefer a different shape. Ready for review 🙏

sriakula1:
+1 to #536. I rendered it locally too, and narrowing the redirect predicate to `common/types/` only keeps the capability pages readable while still removing the cross-page duplication #412 flagged. Happy to defer to @igrigorik's #536 for the centralization and close this out once it lands.

One piece here is independent of that: the `Visibility` column (the per-operation required / optional / omit rendering @wry-ry asked for in #412). It doesn't depend on the redirect approach at all, so if it's useful I can rebase it into a small standalone PR on top of #536 so it isn't lost. Either way, thanks for the quick turn on this.

PR #451, opened 2026-05-19
feat!: buyer consent with per-segment extensibility

igrigorik:
Evolves the `buyer_consent` extension to support per-purpose and per-segment consent capture under a uniform object shape. Each consent decision carries the current `granted` state, the `source` identifying who asserted it (business default vs. platform-captured buyer preference), and human-readable context — keyed by reverse-DNS identifiers at both purpose and segment levels. This generalizes the per-channel consent capture explored in #407 into a primitive that supports open extensibility for purposes and segments without further schema changes.

Businesses advertise the current state of every supported consent option in cart/checkout responses:

```json
"buyer": {
  "consent": {
    "dev.ucp.consent.marketing": {
      "granted": false,
      "source": "business",
      "description": "Promotional communications across all channels",
      "links": [{ "type": "privacy_policy", "url": "https://example.com/privacy" }],
      "segments": {
        "dev.ucp.consent.marketing.email": {
          "granted": true,
          "source": "platform",
          "description": "Promotional emails and exclusive offers"
        },
        "dev.ucp.consent.marketing.sms": {
          "granted": false,
          "source": "business",
          "description": "Marketing text messages"
        }
      }
    },
    "dev.ucp.consent.analytics": {
      "granted": true,
      "source": "business",
      "description": "Site analytics and performance measurement"
    }
  }
}
```

This is a breaking change: the wire shape and field names differ from the prior `allowed` boolean. `granted` is a past-participle state-adjective ("consent is granted") that reads correctly in all four `{state × source}` combinations: business defaults and platform-captured buyer preferences alike. The `source` field carries causation; `granted` carries the state.

The same map shape carries both directions:

| Direction | Who populates | Fields populated |
| --- | --- | --- |
| **Advertise** (response) | Business | `granted`, `source`, `description`; optional `links` and `segments` |
| **Confirm** (request) | Platform | `granted`, `source`; `segments` when present |

Platforms submit the current state of every advertised purpose and segment back to the business:

```json
"buyer": {
  "consent": {
    "dev.ucp.consent.marketing": {
      "granted": true,
      "source": "platform",
      "segments": {
        "dev.ucp.consent.marketing.email": { "granted": true, "source": "platform" },
        "dev.ucp.consent.marketing.sms":   { "granted": true, "source": "platform" }
      }
    },
    "dev.ucp.consent.analytics": { "granted": true, "source": "business" }
  }
}
```

Per-purpose/segment data requirements (e.g., SMS marketing needs `buyer.phone_number`) are business-specific and not enforced by UCP. Missing dependencies surface via the standard message pattern with asymmetric severity: `warning` on cart/checkout `create`/`update`, `error` on `complete_checkout` (the business MUST NOT transition to `completed` while consent data dependencies are unmet).

```json
{
  "messages": [
    {
      "type": "warning",
      "code": "missing_consent_data",
      "content": "Phone number is required for SMS marketing.",
      "path": "$.buyer.phone_number"
    }
  ]
}
```

---

## Checklist

### DIFF
--- a/docs/specification/buyer-consent.md
+++ b/docs/specification/buyer-consent.md
@@ -27,15 +27,21 @@ Consent is modeled as a two-level structure:
 
 - **Purpose** — what the consent is for. Keyed at the top level by a
     reverse-DNS identifier (e.g., `dev.ucp.consent.marketing`). Each purpose
-    carries an `allowed` state, a human-readable `description`, and optional
-    `links`.
+    carries a `checked` state, a `source` identifying who asserted that state,
+    a human-readable `description`, and optional `links`.
 - **Segment** — an optional refinement scoping the parent purpose to a
     specific channel (e.g., email, SMS), vendor (e.g., a measurement provider),
-    or program. Each segment has the same shape (`allowed`, `description`,
-    optional `links`) and overrides the parent purpose for its specific scope.
+    or program. Each segment has the same shape (`checked`, `source`,
+    `description`, optional `links`) and overrides the parent purpose for its
+    specific scope.
 
 The structure is bounded to one level of nesting: purposes have segments;
 segments do not nest further.
+
+The protocol carries only the binary state and its source attribution.
+The operational meaning of `checked: true` vs `false` (granted, denied,
+enabled, subscribed) is defined by the consent context — not by the field
+name or by the protocol.
 
 ## Scope
 
@@ -47,13 +53,11 @@ displaying current subscription state with an unsubscribe affordance) are
 governed by business policy and applicable regulation, not by this
 specification.
 
-Businesses are responsible for selecting the initial `allowed` value according
-to applicable policy and regulation before advertising consent options. In
-contexts that require affirmative opt-in, the business SHOULD advertise
-`allowed: false` until the buyer chooses otherwise. In contexts that allow
-opt-out, the business MAY advertise `allowed: true` as a default the buyer can
-override. The protocol carries the precomputed `allowed` value; policy and
-regulatory reasoning remain the business's responsibility.
+Businesses are responsible for selecting the initial `checked` value according
+to their applicable policy before advertising consent options. The advertised
+value with `source: "business"` is the business's authoritative default;
+buyers may override it through the platform. Policy reasoning remains the
+business's responsibility.
 
 ## Discovery
 
@@ -137,15 +141,27 @@ channel segment identifiers:
 
 The same map shape carries two complementary directions of information:
 
-| Direction                | Who populates | Fields populated                                               |
-| ------------------------ | ------------- | -------------------------------------------------------------- |
-| **Advertise** (response) | Business      | `description`, `links`, current `allowed`, optional `segments` |
-| **Confirm** (request)    | Platform      | captured `allowed`, optional `segments`                        |
+| Direction                | Who populates | Fields populated                                                          |
+| ------------------------ | ------------- | ------------------------------------------------------------------------- |
+| **Advertise** (response) | Business      | `checked`, `source`, `description`; `links` and `segments` when present   |
+| **Confirm** (request)    | Platform      | `checked`, `source`; `segments` when present                              |
 
 `description` and `links` are response-only; the platform does not echo them on
-confirm. `allowed` is required in both directions: businesses send the current
-or default state when advertising; platforms send the captured decision when
-confirming.
+confirm. `checked` and `source` are required in both directions: businesses
+send the current state and its source when advertising; platforms send the
+current state and its source when confirming.
+
+The `source` field carries the authorship of the current `checked` value:
+
+- `source: "business"` means the value reflects the business's default; no
+  buyer preference applies.
+- `source: "platform"` means the value reflects the buyer's stated
+  preference, captured by the platform.
+
+The source signals how to treat the current value: `source: "business"`
+invites the platform to surface the choice for buyer engagement;
+`source: "platform"` indicates a recorded buyer preference the platform
+SHOULD respect unless the buyer initiates a change.
 
 ### Advertise example
 
@@ -163,42 +179,50 @@ state (either the buyer's prior decision or the business default):
   "buyer": {
     "consent": {
       "dev.ucp.consent.marketing": {
-        "allowed": false,
+        "checked": false,
+        "source": "business",
         "description": "Promotional communications across all channels",
         "links": [{ "type": "privacy_policy", "url": "https://example.com/privacy" }],
         "segments": {
           "dev.ucp.consent.marketing.email": {
-            "allowed": false,
+            "checked": true,
+            "source": "platform",
             "description": "Promotional emails and exclusive offers"
           },
           "dev.ucp.consent.marketing.sms": {
-            "allowed": false,
+            "checked": false,
+            "source": "business",
             "description": "Marketing text messages",
             "links": [{ "type": "terms_of_service", "url": "https://example.com/sms-terms" }]
           },
           "com.example.channel.marketing": {
-            "allowed": false,
+            "checked": false,
+            "source": "business",
             "description": "Marketing messages via a third-party channel"
           }
         }
       },
       "dev.ucp.consent.analytics": {
-        "allowed": true,
+        "checked": true,
+        "source": "business",
         "description": "Site analytics and performance measurement",
         "segments": {
           "com.example.analytics": {
-            "allowed": false,
+            "checked": false,
+            "source": "business",
             "description": "Third-party analytics measurement",
             "links": [{ "type": "privacy_policy", "url": "https://example.com/analytics-privacy" }]
           }
         }
       },
       "dev.ucp.consent.preferences": {
-        "allowed": true,
+        "checked": true,
+        "source": "business",
         "description": "Remember preferences and personalize the shopping experience"
       },
       "dev.ucp.consent.sale_or_sharing": {
-        "allowed": false,
+        "checked": false,
+        "source": "platform",
         "description": "Sale or sharing of personal data with third parties",
         "links": [{ "type": "privacy_policy", "url": "https://example.com/privacy" }]
       }
@@ -207,11 +231,19 @@ state (either the buyer's prior decision or the business default):
 }
 ```
 
+In this example, the buyer has previously opted in to promotional email
+(`marketing.email` shows `source: "platform"`) and previously opted out of
+data sale (`sale_or_sharing` shows `source: "platform"`). Every other choice
+remains at its business-asserted default and is a candidate for the platform
+to surface for explicit capture.
+
 ### Confirm example
 
-Platforms submit current consent decisions in a subsequent `update_checkout` or
-`complete_checkout` request, with buyer changes applied to the advertised
-settings:
+Platforms submit the current state of every advertised purpose and segment in a
+subsequent `update_checkout` or `complete_checkout` request. For each choice,
+the platform echoes the advertised `checked` and `source` when no buyer
+preference applies, or sets `source: "platform"` with the buyer's stated
+`checked` value when one does.
 
 <!-- ucp:example schema=shopping/buyer_consent def=consent op=complete direction=request extract=$.buyer.consent -->
 
@@ -220,34 +252,45 @@ settings:
   "buyer": {
     "consent": {
       "dev.ucp.consent.marketing": {
-        "allowed": false,
+        "checked": true,
+        "source": "platform",
         "segments": {
-          "dev.ucp.consent.marketing.email": { "allowed": true },
-          "dev.ucp.consent.marketing.sms": { "allowed": false },
-          "com.example.channel.marketing": { "allowed": false }
+          "dev.ucp.consent.marketing.email": { "checked": true,  "source": "platform" },
+          "dev.ucp.consent.marketing.sms":   { "checked": true,  "source": "platform" },
+          "com.example.channel.marketing":   { "checked": false, "source": "business" }
         }
       },
       "dev.ucp.consent.analytics": {
-        "allowed": true,
+        "checked": true,
+        "source": "business",
         "segments": {
-          "com.example.analytics": { "allowed": false }
+          "com.example.analytics": { "checked": false, "source": "business" }
         }
       },
-      "dev.ucp.consent.preferences": { "allowed": true },
-      "dev.ucp.consent.sale_or_sharing": { "allowed": false }
+      "dev.ucp.consent.preferences":     { "checked": true,  "source": "business" },
+      "dev.ucp.consent.sale_or_sharing": { "checked": false, "source": "platform" }
     }
   }
 }
 ```
 
+Here, the buyer opted in to marketing on email and SMS; other choices are
+echoed at their advertised values, including `sale_or_sharing` which retains
+a prior platform-captured choice.
+
 ## Data dependencies
 
 Some consent states depend on additional buyer or checkout data. For example,
-SMS marketing consent may require a buyer phone number. When required data is
-missing, the business SHOULD use standard `messages[]` warnings or errors to
-inform the platform or buyer.
+SMS marketing consent requires a buyer phone number. When the
+platform confirms a consent value whose required dependency is missing,
+businesses surface the gap through the standard [Checkout Status
+Lifecycle](checkout.md#checkout-status-lifecycle) and [Error
+Handling](checkout.md#error-handling) mechanisms.
 
-For example:
+On `create_cart`, `update_cart`, `create_checkout`, and `update_checkout`,
+businesses SHOULD surface missing dependencies as `warning` messages so the
+platform can collect the data on a subsequent operation. The advertised
+consent decisions remain valid; the warning is informational.
 
 <!-- ucp:example schema=shopping/checkout target=$.messages op=read -->
 
@@ -262,16 +305,25 @@ For example:
 ]
 ```
 
+On `complete_checkout`, businesses MUST NOT transition the checkout to
+`completed` while a confirmed consent decision has unmet data dependencies.
+Missing dependencies MUST be surfaced via the standard [Error
+Handling](checkout.md#error-handling) flow.
+
 ## Normative requirements
 
 1. **Use the advertised settings.** Platforms MUST initialize and present
-   consent using the advertised `description`, `links`, provided `allowed`
-   values, and purpose/segment grouping. Identifiers are opaque handles; agents
-   MUST NOT infer semantics from identifier paths alone.
+   consent using the advertised `description`, `links`, provided `checked`
+   values, `source` attribution, and purpose/segment grouping. Identifiers are
+   opaque handles; platforms MUST NOT infer semantics from identifier paths
+   alone.
 
-2. **Complete confirm.** When submitting consent, platforms MUST include every
-   advertised purpose and segment key, carrying either the advertised `allowed`
-   value or the buyer's captured change.
+2. **Confirm semantics.** The `consent` field is optional on requests;
+   omitting it provides no consent update and the business retains its prior
+   position. When submitting `consent`, platforms MUST include every advertised
+   purpose and segment key, carrying both `checked` and `source` for each.
+   Omitting an advertised key within a submitted `consent` map MUST NOT be
+   used to signal any value.
 
 3. **Advertised set is authoritative.** Businesses MUST ignore purposes and
    segments in a request that were not advertised in a prior response. Platforms
@@ -279,4 +331,18 @@ For example:
    advertise.
 
 4. **More-specific values win.** For an advertised segment, the segment's
-   `allowed` value overrides the parent purpose's `allowed` for that segment.
+   `checked` value overrides the parent purpose's `checked` for that segment.
+
+5. **Source attribution requires a buyer-stated preference.** Platforms MUST
+   set `source: "platform"` only when the `checked` value reflects the buyer's
+   stated preference. For choices to which no buyer preference applies,
+   platforms MUST echo the advertised `source` unchanged. The protocol does
+   not prescribe what counts as a buyer-stated preference; that determination
+   is made by the platform under applicable policy.
+
+6. **Per-business attribution.** A consent decision captured with a specific
+   business is attributable only to that business; platforms MUST NOT
+   propagate it as the basis for `source: "platform"` in a different business's
+   request. Broader buyer preferences that are not tied to a specific business
+   are independent of per-business decisions and may be applied to each
+   business's advertised choices on their own basis.
--- a/docs/specification/buyer-consent.md
+++ b/docs/specification/buyer-consent.md
@@ -18,30 +18,60 @@
 
 ## Overview
 
-The Buyer Consent extension enables platforms to transmit buyer consent choices
-to businesses regarding data usage and communication preferences. It allows
-buyers to communicate their consent status for various categories, such as
-analytics, marketing, and data sales, helping businesses comply with privacy
-regulations like CCPA and GDPR.
+The Buyer Consent extension lets businesses communicate the consent options they
+offer and lets platforms transmit buyers' consent decisions for data usage and
+communication preferences — purposes like marketing, analytics, preferences, and
+sale or sharing of data.
 
-When this extension is supported, the `buyer` object in checkout is extended
-with a `consent` field containing boolean consent states.
+Consent is modeled as a two-level structure:
 
-This extension can be included in `create_checkout` and `update_checkout`
-operations.
+- **Purpose** — what the consent is for. Keyed at the top level by a
+    reverse-DNS identifier (e.g., `dev.ucp.consent.marketing`). Each purpose
+    carries an `allowed` state, a human-readable `description`, and optional
+    `links`.
+- **Segment** — an optional refinement scoping the parent purpose to a
+    specific channel (e.g., email, SMS), vendor (e.g., a measurement provider),
+    or program. Each segment has the same shape (`allowed`, `description`,
+    optional `links`) and overrides the parent purpose for its specific scope.
+
+The structure is bounded to one level of nesting: purposes have segments;
+segments do not nest further.
+
+## Scope
+
+This extension defines the mechanism by which businesses and platforms establish
+and communicate buyer consent state. It does not dictate either party's behavior
+in response to that state — how a business acts on it (e.g., sending emails,
+sharing data) and how a platform surfaces existing decisions to the buyer (e.g.,
+displaying current subscription state with an unsubscribe affordance) are
+governed by business policy and applicable regulation, not by this
+specification.
+
+Businesses are responsible for selecting the initial `allowed` value according
+to applicable policy and regulation before advertising consent options. In
+contexts that require affirmative opt-in, the business SHOULD advertise
+`allowed: false` until the buyer chooses otherwise. In contexts that allow
+opt-out, the business MAY advertise `allowed: true` as a default the buyer can
+override. The protocol carries the precomputed `allowed` value; policy and
+regulatory reasoning remain the business's responsibility.
 
 ## Discovery
 
-Businesses advertise consent support in their profile:
+Businesses advertise consent support in their profile. The capability can extend
+cart, checkout, or both:
 
 <!-- ucp:example schema=profile def=business_schema extract=$.capabilities target=$.ucp.capabilities -->
+
 ```json
 {
   "capabilities": {
     "dev.ucp.shopping.buyer_consent": [
       {
         "version": "{{ ucp_version }}",
-        "extends": "dev.ucp.shopping.checkout"
+        "extends": [
+          "dev.ucp.shopping.cart",
+          "dev.ucp.shopping.checkout"
+        ]
       }
     ]
   }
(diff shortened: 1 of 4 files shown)


source/schemas/shopping/buyer_consent.json
```diff
@@ -3,55 +3,110 @@
...
+          "ucp_response": "required"
+        },
+        "links": {
+          "type": "array",
+          "items": { "$ref": "../common/types/link.json" },
+          "description": "Optional segment-specific links (e.g., channel disclosures, program details, vendor terms). Overlays the cart/checkout top-level `links[]` for this segment."
+        },
+        "allowed": {
+          "type": "boolean",
+          "description": "Current consent state. Businesses set the prior decision or default when advertising; platforms set the captured decision when confirming."
```
jamesandersen:
```suggestion
          "description": "Current consent state. Businesses set the prior decision or jurisdiction-specific default when advertising; platforms set the captured decision when confirming."
```

source/schemas/shopping/buyer_consent.json
```diff
@@ -3,55 +3,110 @@
...
   "name": "dev.ucp.shopping.buyer_consent",
   "title": "Buyer Consent Extension",
-  "description": "Extends Checkout with buyer consent tracking for privacy compliance via the buyer object.",
+  "description": "Extends the buyer object with per-category consent. Each category accepts either a boolean (blanket decision) or a map of per-segment consent keyed by reverse-DNS segment identifiers, enabling channel-, program-, or purpose-specific opt-in.",
   "$defs": {
+    "consent_segment": {
+      "type": "object",
+      "description": "A single consent segment within a category. Businesses populate `description`, `links`, and current `allowed` state when advertising. Platforms populate `allowed` with the captured decision when confirming.",
+      "required": ["description", "allowed"],
+      "properties": {
```
jamesandersen, line 12: If we were to add a `dev.ucp.consent.mobile_app_push` the business would need to collect some kind os specific push token that wouldn't be drawn from the existing cart/checkout request body.  Should we consider some flexible "metadata" dict here e.g. businesses can look for additional platform derived data where applicable in this dict and report via messages if a needed token is not found / invalid.

I'm just pressure testing a bit here; this could also wait until the need is more clear

source/schemas/shopping/buyer_consent.json
jamesandersen: Nit ... but want to suggest the idea that some jurisdictions should be modeled as opt-in, some as opt-out and it's the *business' responsibility* to own that decision

source/schemas/shopping/buyer_consent.json
```diff
@@ -3,55 +3,110 @@
...
+          "type": "boolean",
+          "description": "Current consent state. Businesses set the prior decision or default when advertising; platforms set the captured decision when confirming."
+        }
+      }
+    },
+    "consent_value": {
+      "description": "A consent decision for a category. Either a boolean for blanket consent, or a map of per-segment consent for finer-grained capture (e.g., separate email vs. SMS opt-in).",
+      "oneOf": [
+        {
+          "type": "boolean",
```
jamesandersen: If this will be a breaking change anyway... do we have some conviction that:
1. This alternate `bool` modeling is useful to keep around?
2. We've got the **right** four categories e.g. `analytics`, `marketing`, `preferences`, `sale_of_data`?   e.g. if  introducing a more richly modeled `consent_segment` are these four categories really meaningful at all or should we dispense with this intermediate strict hierarchy and just let businesses express an array of consent segments?  These four _may_ have been inspired by some major regulation ... but we all know regulation changes / evolves frequently ;-)   For implementors there could be some confusion trying to map to these four e.g. Google Tag platform [models](https://developers.google.com/tag-platform/security/guides/consent-debugging) consents for `ad_storage`, `ad_user_data` and `analytics_data` ... are those all `analytics` (because it's GTM) or are some `marketing` (ads)?

source/schemas/shopping/buyer_consent.json
jamesandersen: Discussed on #307 in response to @amithanda 's proposal _a bit_ but I'd like to see `"complete": "optional"` here as well (or better understand why that would be problematic).  In a common _happy path_ a platform... 
1. Calls create checkout
   - no business derived consent segments have been fetched so UX can't have rendered and collected buyer consent yet
   - Is there a scenario where passing consent on `create` makes sense? ... maybe if `cart` was used first and it's *assumed* the consent segments are the same? Is that safe?
2. renders checkout UX with a buyer's primary contact, shipping and payment information *and* applicable consent UX from the create response
3. user reviews checkout details and interacts with the consent UX as applicable before confirming checkout
4. Platform now calls `complete` checkout providing the user's selections

If we continue to `omit` the `buyer` object on `complete` then this happy path is required to add an extra `update` network call to communicate the consent state.   Can we avoid that latency?  Does anything else that lives on `buyer` potentially alter the pricing totals in a way that would make this challenging?

docs/specification/buyer-consent.md
```diff
@@ -88,46 +121,121 @@ POST /checkouts
...
+This extension defines the following well-known segment identifiers under
+the `dev.ucp.consent.*` namespace:
+
+| Identifier | Intent |
+| --- | --- |
+| `dev.ucp.consent.email` | Email channel (any provider) |
+| `dev.ucp.consent.sms` | SMS channel (any carrier) |
+
+All other identifiers are governed by their owners under their own
+reverse-DNS namespace — for example, `com.chatapp.marketing` for a
```
jamesandersen: 👌

jamesandersen (commented):
Another worthy alternate design alongside that from @amithanda on #407 ... this one going after all forms of consent. I can get behind this one also if we want to tackle the **broader** scope of consent (not just marketing) ... though honestly I'd recommend just dropping our pre-existing consent categories in the process if we go this route.   @amithanda , @vixdug thoughts?

amithanda:
Hi @igrigorik, @jamesandersen 

Thanks for putting this together. Evolving `buyer_consent` to support granular, per-segment consent is a very clean way to address regulatory requirements without locking the schema into a fixed set of channels.

I’d like to highlight 3 core architectural topics for us to discuss and align on, building upon some of `@jamesandersen`'s points:

### 1. Polymorphism (Boolean vs. Segment Map)
As `@jamesandersen` pointed out, since this is already a breaking change, we should evaluate if keeping the hybrid `boolean | object` modeling is the right path. 

*   **Implications:** Allowing a property to accept either a primitive `boolean` or a complex `object` introduces type polymorphism. This complicates runtime parsing and often generates messy/ambiguous union types in standard client-side model-generation tools.
*   **Discussion Point:** If we want to support granular consent, should we completely deprecate the raw boolean values and standardise on a single object-based configuration? Furthermore, should we dispense with the intermediate strict category hierarchy (`marketing`, `analytics`, etc.) and instead let businesses express a flat map/array of consent segments directly under `buyer.consent`?

### 2. Consent Flow and Lifecycle (`update` vs. `complete` & PII)
I want to echo `@jamesandersen`’s recommendation to allow the `buyer` object on checkout `complete` (making `"complete": "optional"`).

*   **The Latency Problem:** Surfacing shipping, contact, and consent forms on the final screen is a standard checkout layout. If the `buyer` object is omitted on `complete`, the platform is forced to run a blocking `update_checkout` API call to save consent states right before triggering the final payment. This blocking round-trip at the conversion point adds unnecessary latency.
*   **The PII Disclosure Boundary:** For privacy-centric platforms (e.g., assistants/wallets that withhold buyer contact PII until the user actively confirms the order), submitting `buyer.phone_number` or `buyer.email` is reserved for the final `complete` call. If consent dependencies (like SMS requiring a phone number) are validated during `update`, these privacy flows are broken. Allowing `buyer` and consent states to be submitted atomically on `complete` resolves both latency and PII boundary conflicts.

### 3. Default Consent Values & Jurisdictional Gaps
Building on `@jamesandersen`’s note that *"some jurisdictions should be modeled as opt-in, some as opt-out and it's the business' responsibility to own that decision"*:

*   **The Compliance Gap:** If a US-based merchant defaults a checkbox to checked (`"allowed": true`), but the platform renders the checkout for a user in the EU (where pre-checked marketing boxes are illegal under GDPR), the platform must override the merchant default and show the box unchecked.
*   **Lack of Intent Signaling:** The current schema only communicates the final `allowed` state. The merchant has no way of knowing if a `false` value represents a user **actively opting out** (unchecking a box) vs. a platform **default override** (GDPR opt-in rendering where the user just ignored the box).
*   **Discussion Point:** Should we:
    1.  Add explicit metadata to the segment payload (e.g., `consent_model: "opt_in" | "opt_out"`) so the platform knows how to legally render the defaults?
    2.  Introduce a signal like `explicit_choice: boolean` or `user_interacted: boolean` on confirm requests to help merchants audit and verify active user consent decisions?

source/schemas/shopping/buyer_consent.json
wsbrunson: > Is there a scenario where passing consent on create makes sense? ... maybe if cart was used first and it's assumed the consent segments are the same? Is that safe?

I could see another scenario where a platform that has default collection (email, sms) enabled and so when they make a request to create the checkout session, they also already have the consents collected.

But I agree, I think we would still want this to be available on `complete` though so that a two-shot, `create -> complete` for `checkout_session` could still be supported.

igrigorik:
@jamesandersen, @amithanda, @wsbrunson ty for the feedback!

Updated to allow consent on `complete` (https://github.com/Universal-Commerce-Protocol/ucp/pull/451/changes/dfc561965284776d34053a8bb7bfe0b4ca92e706). 

The big design question is on nesting. I _like_ the idea of a flat namespace and spent good chunk of time trying to iterate through how that would look. A key outcome and realization from running this exercise...

**`Consent > Purpose > Segment` is load-bearing structure**.

A consent decision is *about* a purpose (marketing, analytics, etc) that may be *qualified* by a channel / segment / vendor / program. One way or another, this relationship needs to be captured, and the question is by whom, as that determines where the grouping lives. Let's step through a few examples.

### Flat shape forces agent to own grouping

```jsonc
"consent": {
  "dev.ucp.consent.marketing.email": { ... },
  "dev.ucp.consent.marketing.sms":   { ... },
  "com.chatapp.marketing":           { ... },
  "dev.ucp.consent.analytics":       { ... },
  "com.merchant.purpose_or_channel": { ... }
}
```

Above yields a simple flat structure but, notably, it defers any semantics over purpose to the `description` of each object and there is no easy or logical grouping unless agent leans on substring matches (not guaranteed to work reliably) or spends LLM tokens to group and present a coherent choice matrix: as a user I may want to opt out from all marketing, or only consent to a particular segment. This places the burden on the agent to present and get the choice matrix right, which is an antipattern that lands everyone (agent, business, user) in non-deterministic (bad) outcome. 

#### Channels / segments at top level are footguns

As an example, `dev.ucp.consent.email` at top level carries a footgun. If buyer says no, does that mean that order confirmation and update emails are blocked? With segmen keys at the top, description becomes load-bearing to capture purpose. This is fragile and adds another complication on top of above concerns. 

### (Recommended) nested shape: Purpose > (optional) Segments

We can normatively spec that businesses must provide `purpose > segment` structure...

```jsonc
"dev.ucp.consent.marketing": {
  "allowed": false,
  "description": "Promotional communication...",
  "links": [...],
  "segments": {
    "dev.ucp.consent.marketing.email": { "allowed": true, "description": "...", links: [] },
    "dev.ucp.consent.marketing.sms":   { "allowed": false, "description": "...", links: [] },
    "com.chatapp.channel.marketing":  { "allowed": false, "description": "...", links: [] }
  }
}
```

The agent walks the tree. UI structure IS protocol structure: render purpose description, nest segments under with appropriate state and toggles. Channels / vendors nest naturally -- ChatApp owns the segment identifier, merchant does the work of logically grouping under a purpose key, which itself carries a description and links. No guessing for the agent required, and this mirrors structure that agents already model in their other channels and UIs. Further, the choice of how it's presented and group is material.

In effect, protocol carries the grouping the user-facing UI needs, so the agent doesn't reconstruct it from heuristics.

---

This structure yields roughly...

```jsonc
"buyer": {
  "consent": {
    "dev.ucp.consent.marketing": {
      "allowed": false,
      "description": "Promotional communications across all channels",
      "links": [...],
      "segments": {
        "dev.ucp.consent.marketing.email": { "allowed": true,  "description": "Promotional emails and offers" },
        "dev.ucp.consent.marketing.sms":   { "allowed": false, "description": "Marketing text messages" },
        "com.chatapp.channel.marketing":  { "allowed": false, "description": "Marketing via ChatApp" }
      }
    },
    "dev.ucp.consent.analytics": {
      "allowed": true,
      "description": "Site analytics and performance measurement",
      "segments": {
        "com.google.analytics": { "allowed": false, "description": "Google Analytics tracking" }
      }
    },
    "dev.ucp.consent.preferences":     { "allowed": true,  "description": "Remember preferences and personalize experience" },
    "dev.ucp.consent.sale_or_sharing": { "allowed": false, "description": "Sale or sharing of personal data with third parties" }
  }
}
```

UCP-curated "well-known" purposes:
- `dev.ucp.consent.marketing`
- `dev.ucp.consent.analytics`
- `dev.ucp.consent.preferences`
- `dev.ucp.consent.sale_or_sharing` (renamed from `sale_of_data`)

Businesses are free to define own purpose buckets using rDNS convention, under which they slot optional segments. The top level contract of `buyer.consent > purpose.description > [segment.description]` allows agent to faithfully render the choice matrix without any guessing or heuristics.

@jamesandersen's to your question of "are these 4 right?". AFAIK, unfortunately there is no universally agreed enumeration we can lean on. But this is an extensible list, so our job is not to enumerate exhaustive list but to spec a well-known subset, similar to how we do in many places across the spec with well-known codes, etc. Absence of an agreed standard/enumeration should not stop us from solving the common 80%.

Thoughts, reactions?

amithanda:
Hi @igrigorik,

This nested `Consent > Purpose > Segment` design is a huge improvement! It successfully eliminates the type polymorphism from the previous draft (replacing the `boolean | object` unions with uniform objects) and provides a clear, structural roadmap for platforms to render the checkbox groupings.
I have two suggestions to build upon this nested shape that we can discuss:
### 1. Enforcing Ordered Arrays for Consent Segments (Compliance)
Should we design this as an **array of objects** for `segments` instead of a map.
*   **Why ordering is legally critical:** Under GDPR, CCPA/CPRA, and TCPA, the
    Business (merchant) carries 100% of the data controller liability. Choice
    architecture is legally scrutinized. For example, California CPRA mandates
    "equal prominence" for opt-outs, and TCPA requires SMS consent to be
    "clear and conspicuous" and positioned adjacent to its disclosure text.
    Shuffling consent options (which is common when parsing unordered JSON maps)
    can create non-compliant layouts, immediately exposing merchants to severe
    regulatory penalties and litigation.
*. **Support for Nested Hierarchy:** Inline with your argument around avoiding platforms
    to guess layout groupings, ordering etc via heuristics. Using a map *within*
    the nested segments still fails to guarantee key ordering. Standardizing
    on a nested array structure resolves both: the parent purpose defines
    the logical grouping, and the child array enforces the rendering order.
*   **Consistency:** While maps are preferred for programmatic registries, UCP
    consistently uses arrays for display-critical transactional sequences (such
    as `totals` and `line_items`). Using an array for consent segments aligns
    with this existing protocol convention.
### 2. Hierarchical Opt-In / Opt-Out Metadata (Jurisdictional Compliance)
Different channels are governed by different regulations. For example, in the
US, email marketing is legally **opt-out** (under CAN-SPAM), while SMS
marketing is strictly **opt-in** (under TCPA, where pre-checked SMS boxes
are illegal).
*   **The Solution:** We should support the `consent_model` attribute at both the
    **Purpose** and **Segment** levels, using an inheritance model. The Purpose
    defines a fallback default, and nested segments can explicitly override it
    to ensure precise, legally compliant rendering on the platform.
---
### Suggested Payload Shapes
#### Business Advertise (Response)
```jsonc
"buyer": {
  "consent": {
    "dev.ucp.consent.marketing": {
      "allowed": true, 
      "description": "Promotional communications across all channels",
      "links": [{"type": "privacy_policy", "url": "https://example.com/privacy"}],
      "consent_model": "opt_out", // Fallback default for marketing segments
      
      // Changed from a map to an array of objects to guarantee UI rendering order
      "segments": [
        {
          "id": "dev.ucp.consent.marketing.email",
          "allowed": true,
          "description": "Promotional emails and offers",
          "consent_model": "opt_out" // Inherited from parent; platform can pre-check
        },
        {
          "id": "dev.ucp.consent.marketing.sms",
          "allowed": false,
          "description": "Marketing text messages",
          "links": [{"type": "tcpa_disclosure", "url": "https://example.com/sms-terms"}],
          "consent_model": "opt_in" // Overrides parent; platform must render unchecked
        }
      ]
    },
    "dev.ucp.consent.sale_or_sharing": {
      "allowed": true,
      "description": "Do not sell or share my personal information",
      "consent_model": "opt_out" // No segments; parent-level model is authoritative
    }
  }
}
```
#### Platform Confirm (Request)
When sending decisions back, the platform omits descriptive metadata
(`description`, `links`, and `consent_model` are marked as `ucp_request: "omit"`
in the schema) and only returns the captured choices:
```jsonc
"buyer": {
  "consent": {
    "dev.ucp.consent.marketing": {
      "allowed": true, // User selected general marketing
      "segments": [
        {
          "id": "dev.ucp.consent.marketing.email",
          "allowed": true // User opted in to email
        },
        {
          "id": "dev.ucp.consent.marketing.sms",
          "allowed": false // User opted out of SMS
        }
      ]
    }
  }
}
```
### 3. Parent-Child Inheritance & Override Semantics (Prose Clarification)
We should normatively specify how parent and child values interact in the spec
prose to prevent logical conflicts:
*   **Consent State (`allowed`):**
    *   **Parent-level `false` (Master Block):** Overrides all nested segments. A
        general opt-out blocks all channels, regardless of individual segment
        states.
    *   **Parent-level `true` (Fallback):** Acts as general consent for the
        purpose; individual segment-level toggles (e.g., `sms: false`) act as
        explicit overrides for those specific channels.
*   **Consent Type (`consent_model`):**
    *   **Parent-level fallback:** Defines the default consent type (`opt_in` /
        `opt_out`) for the category.
    *   **Segment-level override:** Individual segments can explicitly define
        their own `consent_model`, overriding the parent fallback for that
        channel.

Let's discuss and take a call.

jamesandersen:
@igrigorik  / @amithanda I feel like this is close to the finish line and appreciate support from both of you!

Here's my quick view:
- [x] updated the schema to make buyer optional on create, update, _and_ complete
- [x] rationale for keeping the "purpose" layer of the structure - pressure-tested and sound reasoning for keeping it ... now with room to extend more easily as needed
- [x] Boolean / object polymorphism - It looks like the samples from both of the recent comments have modeled `allowed` as a property of the purpose level which feels like it accomplishes the "blanket" consent for purpose without the need for actual polymorphism complexity
- [x] regarding the hypothetical case of a platform needing to provide some _additional_ data to accompany the consent e.g. a mobile app push token ... `consent_segment` is an object we *could* extend later if this is deemed important but I'm content ignoring this for now

Outstanding:
* *Best representation of Business intent* - I second @amithanda proposed tweak from map to array to allow businesses to more clearly indicate the consent ordering as there could be some regulatory implications; both of you have highlighted the importance of letting the business express more clearly the *intended* UX display.  *However*, I also think we need to be careful with this ... a business ordering and grouping of segments alone doesn't imply or complete the compliance the business from its responsibilities.  In fact the split nature of requesting (business) and rendering (platform) these consents makes reasoning about compliance even more challenging IMO. Perhaps something like the following in the docs would be appropriate:
> UCP carries the business's consent configuration to the platform. Ordering of segments, descriptions, and links reflect the business's expressed intent and SHOULD be preserved by the platform in rendering. Compliance with applicable privacy and consent regulations (GDPR, TCPA, CCPA/CPRA, etc.) is the responsibility of the implementing business and platform, not of the protocol. Businesses are responsible for the consent options they advertise, including applicable default states, disclosures, and segment ordering. Platforms are responsible for faithfully presenting those options to buyers.
* *Consent Model* - @igrigorik I didn't see this in your latest comment but also think this is important to include (as in @amithanda 's last comment) for a business to control the consent default experience which can vary by jurisdiction

igrigorik:
@amithanda @jamesandersen updated the draft to capture the shape we reviewed above. PTAL.

wsbrunson:
This looks great and would still solve the problems we had when we opened this proposal: https://github.com/Universal-Commerce-Protocol/ucp/pull/407

At the time we weren't sure how we could extend the marketing consent channels options to the other types of consents like analytics. What you have here works really well and would hopefully be the last breaking change to `buyer.consent`. It should be simple for Platforms and Businesses to add new types of consents and new options for collecting that consent.

The one thing I'm unsure of: Does it make sense for Businesses to send an `allowed` value? I could see an issue where there was confusion from the Business on what value to send and they accidentally opt-in a buyer. If we had two fields, one for a Buyer's selection and one for a Business's preference, there would be no confusion. Consent would always be flowing from Buyer -> Platform -> Business.

This would add more complexity to something that may or may not be an issue. Whatever we decide here, I'm good with this proposal 👍.

docs/specification/buyer-consent.md
```diff
@@ -50,88 +80,203 @@ Businesses advertise consent support in their profile:
...
+  }
 }
 ```
 
-## Security & Privacy Considerations
+## Data dependencies
+
+Some consent states depend on additional buyer or checkout data. For example,
+SMS marketing consent may require a buyer phone number. When required data is
+missing, the business SHOULD use standard `messages[]` warnings or errors to
```
jamesandersen: Kind of nit... but do we _want_ to be opinionated about using an `error` if consent is being granted on a checkout **complete** operation without some requisite contact point?  If using a warning only there's an edge case of the transaction having been completed but marketing consent having failed that leaves the platform in a slightly strange state where it can't directly send a correction to address the warning because the transaction is already complete.

source/schemas/shopping/buyer_consent.json
```diff
@@ -61,7 +115,7 @@
               "ucp_request": {
                 "create": "optional",
                 "update": "optional",
-                "complete": "omit"
+                "complete": "optional"
```
jamesandersen, line 126: Do we need to update the [buyer](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/shopping/checkout.json#L46) node to `optional` as well?  I'm not sure if the schema builder tooling has well-defined behavior for handling an `optional` on a nested property overriding an `omit` on a parent level.

jamesandersen:
@igrigorik I'm feeling really good about where this is headed.  I just left a few smaller comments.  

The one slightly challenging area left (as per our discussion and @wsbrunson highlighting again) is how the platform is expected to handle `allowed` (in the absence of the consent model proposal from @amithanda above).  I'm trying to get comfortable with the approach you've modeled ... but just want to be *super* clear about this so bear with the following consideration of this example:
```
"dev.ucp.consent.marketing": {
  "allowed": false,
  "description": "Promotional communication...",
  "links": [...],
  "segments": {
    "dev.ucp.consent.marketing.email": { "allowed": true, "description": "...", links: [] },
    "dev.ucp.consent.marketing.sms":   { "allowed": false, "description": "...", links: [] }
  }
}
```

Just for _extreme clarity_ - is this how you expect a platform to interpret this in light of the drafted normative language requiring (i.e. `MUST`) the platform to display any segments the business provides:

1. Email segment appears with `allowed: true`   (why would a business request consent it already feels it has?)
   * By *including* the segment at all (rather than omitting it), the business is indicating it **does** want to present an explicit consent choice to the user
   * By setting `allowed: true` the business is indicating it already feels it *has* this consent (e.g. default **opt-out** regulatory jurisdiction)
   * It would be appropriate for the platform to present this as "opt-out" (e.g. **checked** checkbox) to the user

2. Sms consent is `allowed: false`
  * By *inclusion* of the segment, business is seeking explicit user consent for this segment
  * By setting `allowed: false` the business is indicating it doesn't have this consent (e.g. default **opt-in** regulatory jurisdiction)
  *  It would be appropriate for the platform to present this as "opt-in" (e.g. **unchecked** checkbox) to the user
  
For *both* cases, unless for some reason the business needs to re-validate consent, it would *omit* these segments  on subsequent cart/checkout transactions so the platform doesn't continually re-prompt for the same consent.  This could be as a result of the business...
  * Receiving affirmative consent from the user
  * Prompting X time in Y days and *not* receiving buyer consent ... thus deciding to backoff on this consent prompt for a period of time ... business side logic (not platform)

source/schemas/shopping/buyer_consent.json
igrigorik: We don't, it'll propagate up correctly.

igrigorik:
@jamesandersen @wsbrunson ty both. PTAL at the updated shape: https://github.com/Universal-Commerce-Protocol/ucp/pull/451/changes/e51a4fd661a69877cb09830e27f6faed59d9d1ea. Taking a different approach and I believe it addresses your feedback cleanly.

docs/specification/buyer-consent.md
```diff
@@ -262,21 +305,44 @@ For example:
 ]
 ```
 
+On `complete_checkout`, businesses MUST NOT transition the checkout to
+`completed` while a confirmed consent decision has unmet data dependencies.
+Missing dependencies MUST be surfaced via the standard [Error
```
jamesandersen, line 354: Excellent - avoids complete transaction + incomplete consent

docs/specification/buyer-consent.md
```diff
@@ -262,21 +305,44 @@ For example:
...
+   set `source: "platform"` only when the `checked` value reflects the buyer's
+   stated preference. For choices to which no buyer preference applies,
+   platforms MUST echo the advertised `source` unchanged. The protocol does
+   not prescribe what counts as a buyer-stated preference; that determination
+   is made by the platform under applicable policy.
+
+6. **Per-business attribution.** A consent decision captured with a specific
+   business is attributable only to that business; platforms MUST NOT
+   propagate it as the basis for `source: "platform"` in a different business's
+   request. Broader buyer preferences that are not tied to a specific business
```
jamesandersen, line 395: OK - don't use consent at one business as implied at another - makes perfect sense.  Just for clarity, it appears the latter part here is essentially saying if a platform hypothetically has a global "shopping preferences" screen of some kind on which a user could indicate "it's ok to use my email for marketing" then the platform could apply that user choice automatically.   Am I following correctly?

If so we might want to adjust the language in point (1) above ("Identifiers are opaque handles; platforms MUST NOT infer semantics from identifier paths alone.) such that platforms *could* use the UCP standard identifiers (e.g. `dev.ucp.consent.marketing.email`) semantically to map to it's hypothetical settings.

docs/specification/buyer-consent.md
```diff
@@ -262,21 +305,44 @@ For example:
...
 ```
 
+On `complete_checkout`, businesses MUST NOT transition the checkout to
+`completed` while a confirmed consent decision has unmet data dependencies.
+Missing dependencies MUST be surfaced via the standard [Error
+Handling](checkout.md#error-handling) flow.
+
 ## Normative requirements
 
 1. **Use the advertised settings.** Platforms MUST initialize and present
```
jamesandersen: The introduction of `source` / `checked` prompts two questions for me here:
1. I'm beginning to worry slightly about platforms needing to render a large chunk of intimidating consent related UX at checkout via the `MUST` guidance here e.g. our example above (albeit primarily to illustrate several _possible_ states) would require a platform to render 8 consent choices to the user.  Should we consider adding an _additional_ example where a business subsequently omits `checked=true` purposes / segments (e.g. their privacy policy allows them to retain a user choice for 30 days without re-prompting, etc) 
2. Is the `source` attribution meaningful to the user e.g. how might a platform render this to the user in clear way?  If it's not, let's drop it from the list of items the platform is suppose to use to render the consent.

docs/specification/buyer-consent.md
igrigorik: Re, global prefs: yep! Added clarification on identifiers.

docs/specification/buyer-consent.md
igrigorik: Good prompts. Clarified both, ptal.

docs/specification/buyer-consent.md
```diff
@@ -50,88 +84,270 @@ Businesses advertise consent support in their profile:
...
+The source signals how the platform should treat the current value:
+`source: "business"` invites the platform to present the choice for buyer
+engagement; `source: "platform"` indicates a recorded buyer preference,
+and platforms MAY suppress re-presentation.
+
+### Advertise example
+
+Businesses advertise available purposes and segments with the current consent
+state (either the buyer's prior decision or the business default):
+
```
jamesandersen, line 214:
```suggestion
#### Minimal example

<!-- ucp:example schema=shopping/buyer_consent def=consent op=read extract=$.buyer.consent -->

```json
{
  "ucp": {},
   "id": "checkout_456",
   "currency": "USD",
  "buyer": {
    "consent": {
      "dev.ucp.consent.marketing": {
        "checked": false,
        "source": "business",
        "description": "Promotional emails and exclusive offers",
        "segments": {
          "dev.ucp.consent.marketing.email": {
            "checked": false,
            "source": "business",
            "description": "Promotional emails and exclusive offers",
            "links": [{ "type": "privacy_policy", "url": "https://example.com/privacy" }]
          }
        }
      }
    }
  }
}

The smallest useful consent map is a single purpose with **no** segments. In the simple example above a business is advertising a specific email marketing segment within the marketing purpose, 
defaulted off.  Other consent purposes are omitted.  The platform here is invited to render only
single email consent segment to the user.  Once the user's response is submitted, the `source` would switch to `platform` and the platform may opt not to re-prompt the user for this choice. 

#### Multiple consent example
```
@igrigorik I think a minimal example like this would be useful e.g. what does it look like if the business is just trying to gather a single consent (e.g. email or SMS).

One thing implied by the example / text here is that if a purpose contains segments the platform is not expected to render a user prompt for the purpose (generally) and for each segment it contains ... segments would take precedence. Do you agree?

jamesandersen (commented):
nit only: "checked" seems a bit anchored in a particular UX representation ... would "consented" (or something similar?) work?

amithanda:
> nit only: "checked" seems a bit anchored in a particular UX representation ... would "consented" (or something similar?) work?

I think it is a good point, `checked` is heavily anchored in visual UI representations and may be semantically awkward for future non UX based scenarios e.g voice/chat yes or no etc. 
Maybe rename it to  **`consented`** as @jamesandersen suggested or  **`granted`** -  here are some reasons:
* **Modality Neutral:** "Granted" fits both visual checkboxes and verbal agent interactions ("Consent is granted").
* **Legally Consistent:** Consent is either "granted" (`true`) or "denied/withdrawn" (`false`).
* **Resolves Opt-Out Mapping Confusion:** For some tricky consent types e.g. `dev.ucp.consent.sale_or_sharing` (Sale or sharing of data), `granted: false` means consent to share data is not granted. This maps directly to the semantic state of the purpose, avoiding the logical inversion confusion that arises when trying to map a visual "Do Not Sell" checkbox state to a field named `checked`.

source/schemas/shopping/buyer_consent.json
```diff
@@ -3,55 +3,119 @@
...
-  "description": "Extends Checkout with buyer consent tracking for privacy compliance via the buyer object.",
+  "description": "Extends the buyer object with per-purpose consent. Each purpose is keyed by a reverse-DNS identifier and carries the current `checked` state, the `source` of that state (business default or platform-captured buyer decision), a `description`, optional `links`, and optional `segments` for finer-grained channel, vendor, or program decisions scoped to that purpose.",
   "$defs": {
-    "consent": {
+    "consent_purpose": {
       "type": "object",
-      "description": "User consent states for data processing",
+      "description": "A buyer's consent decision for a purpose (e.g., marketing, analytics). Carries the current binary state, its source (business default or platform-captured buyer decision), human-readable context, and optional refinements scoping the decision to specific channels, vendors, or programs.",
+      "required": ["checked", "source", "description"],
+      "additionalProperties": false,
```
amithanda: Both object schemas define `"additionalProperties": false`. This prevents future extensibility and contradicts our design rule as specified in the **[Schema Authoring Guide (Open Objects)](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/docs/documentation/schema-authoring.md#open-objects-additionalproperties)**. 

Was this intentional to make these closed?

docs/specification/buyer-consent.md
```diff
@@ -50,88 +84,270 @@ Businesses advertise consent support in their profile:
...
+  }
+]
+```
+
+On `complete_checkout`, businesses MUST NOT transition the checkout to
+`completed` while a confirmed consent decision has unmet data dependencies.
+Missing dependencies MUST be surfaced via the standard [Error
+Handling](checkout.md#error-handling) flow.
+
+## Normative requirements
```
amithanda, line 357: Should we add something to clarify the persistence and revocation semantics of consent states? 
Something like:
> **Consent Persistence & Revocation**
> 
> UCP consent states represent **persistent buyer preferences** associated with the buyer's customer record (CRM, email list, or authenticated profile) at the business. They are not ephemeral, single-transaction approvals.
> 
> *   **Persistence:** Once confirmed (e.g., `checked: true, source: "platform"`), the preference remains active across future transactions. Platforms may suppress re-presentation of these choices to prevent buyer fatigue.
> *   **Revocation:** Buyers can revoke prior consent during checkout by deactivating the option (e.g., unchecking a pre-checked box). The platform communicates this by sending `checked: false, source: "platform"`. Businesses receiving this state transition are responsible for updating their internal customer registries to reflect the revocation.

source/schemas/shopping/buyer_consent.json
igrigorik: Good catch, fixed!

docs/specification/buyer-consent.md
igrigorik: Refactored / fixed, ptal. We already had related text, opted to extend that.

igrigorik:
@amithanda @jamesandersen updated, ptal! 🤞🏻

docs/specification/buyer-consent.md
```diff
@@ -401,3 +402,6 @@ Handling](checkout.md#error-handling) flow.
    request. Broader buyer preferences that are not tied to a specific business
    are independent of per-business decisions and may be applied to each
    business's advertised choices on their own basis.
+
+7. **Buyer review and revocation.** Platforms MUST provide a way for buyers
```
jamesandersen: @igrigorik this is still just in the context of the UCP interaction right? e.g. if the business advertises some prior consent choices with source = platform as per #1 platforms decide which choices to present to users ... but here we're saying platforms must also have some other affordance for users to see/change prior choices *from the business's advertised set*.  This adds some more UX complexity but I can see the reasoning.

What's I'm hesitant about is any confusion that the platform needs to independently persist these choices and present them *outside* the context of a UCP interaction (where there's not really a way to change them).

igrigorik: @jamesandersen clarified in https://github.com/Universal-Commerce-Protocol/ucp/pull/451/commits/8ab2b043cd7e7d6e06ecf3e0c0ff85293acd3760. 

The intent was to highlight that platform persists across sessions, then it shouldn't be one a one-way thing that a buyer can't change their mind on. Merged this into single section.

jamesandersen (approved):

amithanda:
Nice work on this, the extension reads much cleaner than the prior iterations. A few things I think turned out really well:

- **Unifying on a single `{granted, source, description, links, segments}` object** keyed by reverse-DNS, and dropping the old `boolean | object` polymorphism, is the right call. It makes purposes and segments open-extensible without future schema changes, and it resolves the polymorphism concern raised earlier in the thread.
- **`source` separating authorship from state** is a clean model
- Folding persistence and reversibility and the `additionalProperties` cleanup both look good.

One side effect worth recording (which @jamesandersen's comment touched upon), not a blocker: because the `complete: "optional"` annotation sits on the whole `buyer` ref, it also reopens `first_name`, `last_name`, `email`, and `phone_number` for changes & submission at `complete`, where base checkout omits the buyer object entirely. 

Why I think this is fine: those fields are already submittable on `create` and `update`, so this opens no new field and no new data sink. The only practical change is that a client can now also send or overwrite buyer PII on the `complete` call. That seems acceptable.

Approving. Thanks for the careful iteration on this one.

amithanda (approved):

PR #11, opened 2026-01-11
feat: add intent trace extension

0xtotaylor:
# Enhancement Proposal: Intent Trace Extension

  | Status | **Provisional** |
  |--------|-----------------|
  | Author | @0xtotaylor |
  | RFC | `rfcs/intent-trace.md` |

  ## Summary

  This proposal extends `POST /checkout-sessions/{id}/cancel` to accept an optional `intent_trace` object, enabling agents to transmit structured data about why a checkout session was abandoned.

  ## Type of change

  - [x] Enhancement Proposal (new capability/extension per GOVERNANCE.md)

  ## Proposal Contents

  Per the Enhancement Proposal process, the RFC (`rfcs/intent-trace.md`) includes:

  - Summary
  - Motivation
  - Detailed Design
  - Risks
  - Test Plan
  - Graduation Criteria

  ## Implementation

  This PR also includes a complete implementation ready to merge once the proposal reaches **Implementable** status:

  - `source/schemas/shopping/intent_trace.json` — Extension schema
  - `docs/specification/intent-trace.md` — Specification documentation
  - `mkdocs.yml` — Navigation updates

  ## Checklist

  - [x] RFC document follows Enhancement Proposal template
  - [x] Implementation aligns with RFC detailed design
  - [x] Documentation complete
  - [x] Schema validation passes
  - [x] `mkdocs build --strict` passes

  ## Requested Action

  Requesting TC review to advance from **Provisional** → **Implementable**.

PR #458, opened 2026-05-20
feat: Add Payment Authentication Actions

raginpirate:
### Description                                                           
Introduces `dev.ucp.shopping.payment_authentication`, a Checkout extension defining standard browser-surface Actions for payment authentication:
   - `dev.ucp.payment.device_data_collection`                                           
   - `dev.ucp.payment.three_ds_challenge`                                               

This PR is stacked on #582, which defines the common extension-owned Actions primitive. Payment Authentication contributes the concrete Action types, config, rendering behavior, security requirements, notification protocol, deadlines, and Checkout integration needed for DDC and 3DS challenges.                              

### The approach

EMVCo defines the interaction between the issuer’s ACS and the merchant’s 3DS provider, mediated by the cardholder’s device and card-network Directory Server. Businesses and platforms do not have to be involved directly in this specification; often [Payment Providers create an abstraction layer](https://gist.github.com/raginpirate/fa4b1a0e4051e0d4ea03d7e5bf4e4b5c) around the concerns of the EMVco spec. With this world view in mind, UCP attempts to standardize the abstraction layer that businesses and payment providers truly need to offer when handling 3DS and DDC on arbitrary business OR platform surfaces; a simple signal pattern to resolve or reject the action.

## Category (Required)

## Related Issues

https://github.com/Universal-Commerce-Protocol/ucp/issues/517
https://github.com/Universal-Commerce-Protocol/ucp/pull/582

## Checklist

### DIFF
--- a/docs/specification/payment-actions/device-data-collection.md
+++ b/docs/specification/payment-actions/device-data-collection.md
@@ -23,8 +23,7 @@ the [Payment Authentication extension](../payment-authentication.md):
 dev.ucp.payment.device_data_collection
 ```
 
-It asks the Platform to run a scoped, invisible browser-capable surface associated
-with a payment instrument or attempt.
+It asks the Platform to mount an invisible payment-authentication surface to complete device data collection.
 
 ## Purpose and Scope
 
@@ -39,7 +38,7 @@ availability, method timeouts, or collection outcomes into EMVCo state such as
 `threeDSCompInd`. If a provider or Business skips collection, including when an
 ACS supplies no method URL, the Business does not emit this Action.
 
-## Emission and Gated Effect
+## Blocking Action
 
 A Business **MAY** emit this type from a Create, Update, or Complete Checkout
 response after it can associate collection with a payment instrument and
--- a/docs/specification/payment-actions/device-data-collection.md
+++ b/docs/specification/payment-actions/device-data-collection.md
@@ -23,7 +23,8 @@ the [Payment Authentication extension](../payment-authentication.md):
 dev.ucp.payment.device_data_collection
 ```
 
-It asks the Platform to mount an invisible payment-authentication surface to complete device data collection.
+It asks the Platform to mount an invisible payment-authentication surface to
+complete device data collection.
 
 ## Purpose and Scope
 
@@ -80,9 +81,9 @@ The config shape is defined inline by the
 | `payment_instrument_id` | string | ✓ | ID of the associated instrument in the containing Checkout. |
 | `url` | string | ✓ | Absolute HTTPS URL for the invisible collection surface. |
 
-The Business **SHOULD** provide a Business- or provider-operated wrapper URL that
-owns any provider-specific POST, such as posting `threeDSMethodData` to an ACS
-method URL. The Platform performs an ordinary navigation to `config.url`.
+`config.url` **SHOULD** be Business- or provider-operated and own any
+provider-specific POST, such as posting `threeDSMethodData` to an ACS method
+URL. The Platform performs an ordinary navigation to it.
 
 ## Platform Behavior
 
@@ -102,8 +103,8 @@ Mounting the surface **MUST** follow the shared
 and [Embedded Protocol security requirements](../embedded-protocol.md#security).
 
 On the web the surface is typically a hidden iframe. A native Platform may use
-an isolated webview or equivalent browser surface. It must not be visible to the
-Buyer or request Buyer interaction.
+an isolated webview or equivalent browser surface. It **MUST NOT** be visible to
+the Buyer or request Buyer interaction.
 
 The Platform distinguishes only whether its surface finished or could not
 continue. It does not determine whether device data was collected successfully,
@@ -131,10 +132,9 @@ When the Platform-facing collection step is finished, the surface **MUST** send:
 }
 ```
 
-The surface may send `action.done` when handler-owned logic determines that the
-Platform should close the surface. The notification does not reveal whether
-collection succeeded, was unavailable, was skipped, or mapped to a particular
-provider completion indicator.
+Handler-owned logic determines when that step is finished. The notification does
+not reveal whether collection succeeded, was unavailable, was skipped, or mapped
+to a particular provider completion indicator.
 
 ### Error
 
@@ -163,14 +163,13 @@ it **MAY** send:
 }
 ```
 
-Recommended surface-level codes are `action_unavailable` when initialization
-fails, `action_expired` when the surface's session expired, and `action_failed`
-for another terminal surface error. These are diagnostic codes, not EMV 3DS or
-payment outcomes. A provider-domain timeout or unavailable method that the
-handler can safely continue past should normally result in `action.done`, with
-the Business and provider recording the domain outcome server-side. After either
-notification, the Platform follows the shared Payment Authentication
-reconciliation contract.
+Surface-level codes are the shared
+[well-known `action.error` codes](../payment-authentication.md#surface-rendering-and-notifications).
+They are not EMV 3DS or payment outcomes: a provider-domain timeout or
+unavailable method that the handler can safely continue past should normally
+result in `action.done`, with the Business and provider recording the domain
+outcome server-side. After either notification, the Platform follows the shared
+Payment Authentication reconciliation contract.
 
 ## Deadline and Fallback
 
--- a/docs/specification/payment-actions/device-data-collection.md
+++ b/docs/specification/payment-actions/device-data-collection.md
@@ -1,0 +1,208 @@
+<!--
+   Copyright 2026 UCP Authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+
+# Payment Device Data Collection Action
+
+This specification defines the device data collection Action type declared by
+the [Payment Authentication extension](../payment-authentication.md):
+
+```text
+dev.ucp.payment.device_data_collection
+```
+
+It asks the Platform to run a scoped, invisible browser-capable surface associated
+with a payment instrument or attempt.
+
+## Purpose and Scope
+
+Device data collection lets a Business, payment handler, PSP, or 3DS provider
+collect device or browser data before payment processing continues. UCP does not
+carry the collected data. The loaded surface sends provider-specific data directly
+to the Business or payment provider through its own integration.
+
+For EMV 3DS, this Action represents only the Platform-facing instruction to
+mount the collection surface. UCP does not define how a Business maps ACS method
+availability, method timeouts, or collection outcomes into EMVCo state such as
+`threeDSCompInd`. If a provider or Business skips collection, including when an
+ACS supplies no method URL, the Business does not emit this Action.
+
+## Emission and Gated Effect
+
+A Business **MAY** emit this type from a Create, Update, or Complete Checkout
+response after it can associate collection with a payment instrument and
+negotiated payment handler. `config.payment_instrument_id` identifies that
+instrument.
+
+The Action gates advancement of the associated payment attempt while the
+occurrence remains outstanding in the authoritative Checkout. A terminal surface
+event starts reconciliation; it does not itself lift the gate or assert that
+collection succeeded. If collection is merely an optimization that the Business
+will not wait for, the Business **MUST NOT** emit an Action.
+
+## Runtime Shape
+
+The Action is emitted under its type key:
+
+<!-- ucp:example schema=shopping/payment_authentication def=dev.ucp.shopping.checkout extract=$.actions target=$.actions op=read -->
+```json
+{
+  "actions": {
+    "dev.ucp.payment.device_data_collection": [
+      {
+        "id": "ddc-1",
+        "config": {
+          "payment_instrument_id": "instrument_1",
+          "url": "https://payments.example.com/ucp/payment/ddc/session_123"
+        }
+      }
+    ]
+  }
+}
+```
+
+The config shape is defined inline by the
+[Payment Authentication extension schema](site:schemas/shopping/payment_authentication.json).
+
+| Field | Type | Required | Notes |
+| :---- | :--- | :------- | :---- |
+| `payment_instrument_id` | string | ✓ | ID of the associated instrument in the containing Checkout. |
+| `url` | string | ✓ | Absolute HTTPS URL for the invisible collection surface. |
+
+The Business **SHOULD** provide a Business- or provider-operated wrapper URL that
+owns any provider-specific POST, such as posting `threeDSMethodData` to an ACS
+method URL. The Platform performs an ordinary navigation to `config.url`.
+
+## Platform Behavior
+
+The Platform **MUST**:
+
+1. validate the URL according to the Payment Authentication contract, the
+   payment handler and instrument context, and Platform policy;
+2. load the URL in an isolated, invisible browser-capable surface;
+3. complete the `action.ready` handshake before allowing collection to begin;
+4. correlate embedded notifications with both this Action occurrence and the
+   mounted surface; and
+5. unmount the surface after `action.done`, `action.error`, load failure, or
+   timeout.
+
+Mounting the surface **MUST** follow the shared
+[Payment Authentication rendering contract](../payment-authentication.md#surface-rendering-and-notifications)
+and [Embedded Protocol security requirements](../embedded-protocol.md#security).
+
+On the web the surface is typically a hidden iframe. A native Platform may use
+an isolated webview or equivalent browser surface. It must not be visible to the
+Buyer or request Buyer interaction.
+
+The Platform distinguishes only whether its surface finished or could not
+continue. It does not determine whether device data was collected successfully,
+whether an ACS method timed out, or whether 3DS authentication should proceed.
+
+## Embedded Notifications
+
+After completing the shared
+[`action.ready` handshake](../payment-authentication.md#ready-handshake), the
+surface sends JSON-RPC 2.0 notifications defined by the
+[payment Action embedded contract](../payment-authentication.md#surface-rendering-and-notifications).
+
+### Done
+
+When the Platform-facing collection step is finished, the surface **MUST** send:
+
+<!-- ucp:example schema=transports/jsonrpc def=request op=read -->
+```json
+{
+  "jsonrpc": "2.0",
+  "method": "action.done",
+  "params": {
+    "id": "ddc-1"
+  }
+}
+```
+
+The surface may send `action.done` when handler-owned logic determines that the
+Platform should close the surface. The notification does not reveal whether
+collection succeeded, was unavailable, was skipped, or mapped to a particular
+provider completion indicator.
+
+### Error
+
+If the surface cannot continue from the Platform-facing transport perspective,
+it **MAY** send:
+
+<!-- ucp:example schema=transports/jsonrpc def=request op=read -->
+```json
+{
+  "jsonrpc": "2.0",
+  "method": "action.error",
+  "params": {
+    "id": "ddc-1",
+    "error": {
+      "ucp": { "version": "{{ ucp_version }}", "status": "error" },
+      "messages": [
+        {
+          "type": "error",
+          "code": "action_unavailable",
+          "content": "Device data collection could not start.",
+          "severity": "recoverable"
+        }
+      ]
+    }
+  }
+}
+```
+
+Recommended surface-level codes are `action_unavailable` when initialization
+fails, `action_expired` when the surface's session expired, and `action_failed`
+for another terminal surface error. These are diagnostic codes, not EMV 3DS or
+payment outcomes. A provider-domain timeout or unavailable method that the
+handler can safely continue past should normally result in `action.done`, with
+the Business and provider recording the domain outcome server-side. After either
+notification, the Platform follows the shared Payment Authentication
+reconciliation contract.
+
+## Deadline and Fallback
+
+[EMV 3-D Secure Protocol and Core Functions v2.3.1.1](https://www.emvco.com/specifications/emv-3-d-secure-protocol-and-core-functions-specification-6/){ target="_blank" }
+requires the 3DS Server to set `threeDSCompInd` to `N` when the 3DS Method does
+not complete within five seconds (Req 315). That provider-domain timer begins
+when the method is invoked and remains owned by the Business and provider; the
+Platform **MUST NOT** derive it from the outer surface's load event.
+
+The Platform applies a bounded outer-surface deadline defined by the handler or
+Platform policy, measured from navigation and capped by Checkout `expires_at`.
+That deadline controls only how long the Platform keeps the UCP surface mounted;
+it does not set `threeDSCompInd` or another provider outcome.
+
+If no valid `action.done` or `action.error` arrives before the outer deadline,
+the Platform **MUST** unmount the surface, treat the interaction as timed out,
+and follow the non-successful surface reconciliation path. It may unmount sooner
+after load failure or a valid terminal notification. The Platform **MUST NOT**
+remount the same Action type and `id`; an unchanged occurrence follows the
+parent Checkout Actions fallback, handoff, and cancellation rules.
+
+## Security
+
+The common [Payment Authentication security requirements](../payment-authentication.md#security-and-data-handling)
+apply. In addition:
+
+- The handler's trust policy **MUST** authorize the initial and redirect origins
+  before the Platform loads them.
+- The surface **MUST NOT** receive Platform credentials or access Platform
+  storage.
+- Provider payloads and collected device data **MUST NOT** be copied into the
+  Action, notification params, Checkout, or payment instrument.
+- `action.done` and `action.error` are advisory surface signals. A later
+  Checkout response, based on Business/provider state, remains authoritative.
--- a/docs/specification/payment-actions/three-ds-challenge.md
+++ b/docs/specification/payment-actions/three-ds-challenge.md
@@ -23,7 +23,8 @@ This specification defines the 3DS challenge Action type declared by the
 dev.ucp.payment.three_ds_challenge
 ```
 
-It asks the Platform to present a buyer-facing payment-authentication surface to complete 3ds challenges.
+It asks the Platform to present a buyer-facing payment-authentication surface to
+complete 3DS challenges.
 
 ## Purpose and Scope
 
@@ -79,11 +80,10 @@ The config shape is defined inline by the
 | `payment_instrument_id` | string | ✓ | ID of the associated instrument in the containing Checkout. |
 | `url` | string | ✓ | Absolute HTTPS URL for the buyer-facing challenge surface. |
 
-The Business **SHOULD** provide a Business- or provider-operated wrapper URL that
-owns any provider-specific POST, such as submitting CReq to an ACS challenge
-URL. The Platform performs an ordinary navigation to `config.url`; it
-**MUST NOT** construct provider requests, parse challenge payloads, or relay 3DS
-protocol data.
+`config.url` **SHOULD** be Business- or provider-operated and own any
+provider-specific POST, such as submitting CReq to an ACS challenge URL. The
+Platform performs an ordinary navigation to it; it **MUST NOT** construct
+provider requests, parse challenge payloads, or relay 3DS protocol data.
 
 ## Platform Behavior
 
@@ -139,8 +139,11 @@ send:
 
 `action.done` means only that the Platform should close the challenge surface
 and retrieve Checkout. It does not mean authentication or payment succeeded.
-After closing the surface, the Platform uses Get Checkout and does not call
-Complete Checkout again to resume the accepted operation.
+After closing the surface, the Platform should issue Get Checkout to confirm
+processing state. Get Checkout is a read-only observation and does not
+itself advance the payment attempt; the Business advances Checkout state
+from server-side provider state (typically an out-of-band provider
+push/callback), independent of when the Platform calls Get Checkout.
 
 A handler-owned surface **MAY** include provider-specific or diagnostic params
 such as correlation identifiers or SDK outcomes. Platforms **MUST** ignore
@@ -173,18 +176,17 @@ If the surface cannot continue, it **MAY** send:
 }
 ```
 
-Recommended surface-level codes are `action_unavailable` when initialization
-fails, `action_expired` when the challenge session expired, and `action_failed`
-for another terminal surface error. Buyer abandonment and provider
-authentication outcomes may be unavailable to the surface or known only by the
-provider backend; the Business and payment provider remain authoritative for
-those distinctions. After either notification, the Platform follows the shared
-Payment Authentication reconciliation contract.
+Surface-level codes are the shared
+[well-known `action.error` codes](../payment-authentication.md#surface-rendering-and-notifications).
+Buyer abandonment and provider authentication outcomes may be unavailable to the
+surface or known only by the provider backend; the Business and payment provider
+remain authoritative for those distinctions. After either notification, the
+Platform follows the shared Payment Authentication reconciliation contract.
 
 ## Combined Provider Flows
 
-A payment provider may use one wrapper URL to orchestrate device data collection
-and a challenge. The emitted Action type describes the Platform-facing behavior:
+A payment provider may serve both device data collection and a challenge from
+one `config.url`. The emitted Action type describes the Platform-facing behavior:
 
 - If the surface remains invisible and never requests Buyer interaction, the
   Business emits device data collection.
@@ -193,8 +195,8 @@ and a challenge. The emitted Action type describes the Platform-facing behavior:
 The Business must not emit an invisible DDC Action and then make that surface
 visible. If collection completes first, the preferred flow is to remove the DDC
 occurrence and emit a new challenge occurrence in a later Checkout response.
-The two occurrences have different Action types and IDs even when their wrapper
-URL origin is the same.
+The two occurrences have different Action types and IDs even when their
+`config.url` origin is the same.
 
 ## Deadline and Fallback
 
@@ -202,7 +204,7 @@ URL origin is the same.
 gives the ACS 10 minutes (600 seconds) after successfully sending each challenge
 interface to the challenge iframe (Req 226). The surface and provider own that
 per-interface timer. It does not define a single UCP deadline beginning at the
-initial wrapper navigation.
+initial navigation.
 
 The Platform applies a bounded outer-surface deadline defined by the handler or
 Platform policy, measured from navigation and capped by Checkout `expires_at`.
--- a/docs/specification/payment-actions/three-ds-challenge.md
+++ b/docs/specification/payment-actions/three-ds-challenge.md
@@ -23,8 +23,7 @@ This specification defines the 3DS challenge Action type declared by the
 dev.ucp.payment.three_ds_challenge
 ```
 
-It asks the Platform to present a scoped, buyer-facing payment-authentication
-surface.
+It asks the Platform to present a buyer-facing payment-authentication surface to complete 3ds challenges.
 
 ## Purpose and Scope
 
@@ -38,7 +37,7 @@ The Platform presents the surface and observes only when the Platform-facing
 interaction finishes or cannot continue. The Business determines the
 authentication and payment outcome from its server-side provider state.
 
-## Emission and Gated Effect
+## Blocking Action
 
 The 3DS challenge Action gates the associated payment attempt from producing a
 completed Checkout while the occurrence remains outstanding. It does not assert
--- a/docs/specification/payment-actions/three-ds-challenge.md
+++ b/docs/specification/payment-actions/three-ds-challenge.md
@@ -1,0 +1,230 @@
+<!--
+   Copyright 2026 UCP Authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+
+# Payment 3DS Challenge Action
+
+This specification defines the 3DS challenge Action type declared by the
+[Payment Authentication extension](../payment-authentication.md):
+
+```text
+dev.ucp.payment.three_ds_challenge
+```
+
+It asks the Platform to present a scoped, buyer-facing payment-authentication
+surface.
+
+## Purpose and Scope
+
+A 3DS challenge Action lets the Business and payment handler ask the Platform to
+present an issuer or provider challenge to the Buyer. UCP does not carry EMV 3DS
+payloads such as CReq/CRes, AReq/ARes, RReq/RRes, or `transStatus`. Those details
+remain within the loaded surface and the Business, payment handler, PSP, 3DS
+server, issuer, or ACS integration.
+
+The Platform presents the surface and observes only when the Platform-facing
+interaction finishes or cannot continue. The Business determines the
+authentication and payment outcome from its server-side provider state.
+
+## Emission and Gated Effect
+
+The 3DS challenge Action gates the associated payment attempt from producing a
+completed Checkout while the occurrence remains outstanding. It does not assert
+that the Complete Checkout operation was rejected, and a terminal surface event
+does not itself lift the gate.
+
+A Business **SHOULD** accept Complete Checkout, begin the payment attempt, and
+return `complete_in_progress` with the challenge Action. If the Business cannot
+accept completion before the challenge, it follows the core Checkout rule: it
+returns `incomplete` with a recoverable error Message whose `path` selects the
+challenge occurrence.
+
+## Runtime Shape
+
+The Action is emitted under its type key:
+
+<!-- ucp:example schema=shopping/payment_authentication def=dev.ucp.shopping.checkout extract=$.actions target=$.actions op=read -->
+```json
+{
+  "actions": {
+    "dev.ucp.payment.three_ds_challenge": [
+      {
+        "id": "three-ds-challenge-1",
+        "config": {
+          "payment_instrument_id": "instrument_1",
+          "url": "https://payments.example.com/ucp/payment/3ds/session_456"
+        }
+      }
+    ]
+  }
+}
+```
+
+The config shape is defined inline by the
+[Payment Authentication extension schema](site:schemas/shopping/payment_authentication.json).
+
+| Field | Type | Required | Notes |
+| :---- | :--- | :------- | :---- |
+| `payment_instrument_id` | string | ✓ | ID of the associated instrument in the containing Checkout. |
+| `url` | string | ✓ | Absolute HTTPS URL for the buyer-facing challenge surface. |
+
+The Business **SHOULD** provide a Business- or provider-operated wrapper URL that
+owns any provider-specific POST, such as submitting CReq to an ACS challenge
+URL. The Platform performs an ordinary navigation to `config.url`; it
+**MUST NOT** construct provider requests, parse challenge payloads, or relay 3DS
+protocol data.
+
+## Platform Behavior
+
+The Platform **MUST**:
+
+1. validate the URL according to the Payment Authentication contract, the
+   payment handler and instrument context, and Platform policy;
+2. load the URL in an isolated, visible browser-capable surface;
+3. complete the `action.ready` handshake before presenting the challenge;
+4. prevent interaction with the rest of Checkout while the challenge is
+   outstanding;
+5. correlate embedded notifications with both this Action occurrence and the
+   mounted surface; and
+6. close the surface after `action.done`, `action.error`, unrecoverable load
+   failure, or abandonment.
+
+Mounting the surface **MUST** follow the shared
+[Payment Authentication rendering contract](../payment-authentication.md#surface-rendering-and-notifications)
+and [Embedded Protocol security requirements](../embedded-protocol.md#security).
+
+On the web the surface may be a full-page frame, modal frame, or separate window
+when the negotiated handler defines that presentation and its security policy.
+A native Platform may use an isolated webview, browser view, or system browser.
+A top-level or system-browser handoff is permitted only when the handler defines
+a completion channel that the Platform can authenticate and correlate with the
+Action occurrence.
+
+The Platform **MUST NOT** infer an authentication outcome from frame content,
+URL changes, or notification diagnostics.
+
+## Embedded Notifications
+
+After completing the shared
+[`action.ready` handshake](../payment-authentication.md#ready-handshake), the
+surface sends JSON-RPC 2.0 notifications defined by the
+[payment Action embedded contract](../payment-authentication.md#surface-rendering-and-notifications).
+
+### Done
+
+When the Platform-facing challenge interaction finishes, the surface **MUST**
+send:
+
+<!-- ucp:example schema=transports/jsonrpc def=request op=read -->
+```json
+{
+  "jsonrpc": "2.0",
+  "method": "action.done",
+  "params": {
+    "id": "three-ds-challenge-1"
+  }
+}
+```
+
+`action.done` means only that the Platform should close the challenge surface
+and retrieve Checkout. It does not mean authentication or payment succeeded.
+After closing the surface, the Platform uses Get Checkout and does not call
+Complete Checkout again to resume the accepted operation.
+
+A handler-owned surface **MAY** include provider-specific or diagnostic params
+such as correlation identifiers or SDK outcomes. Platforms **MUST** ignore
+unknown params for UCP processing unless the negotiated handler defines them.
+They are never authoritative payment outcomes by default.
+
+### Error
+
+If the surface cannot continue, it **MAY** send:
+
+<!-- ucp:example schema=transports/jsonrpc def=request op=read -->
+```json
+{
+  "jsonrpc": "2.0",
+  "method": "action.error",
+  "params": {
+    "id": "three-ds-challenge-1",
+    "error": {
+      "ucp": { "version": "{{ ucp_version }}", "status": "error" },
+      "messages": [
+        {
+          "type": "error",
+          "code": "action_failed",
+          "content": "The challenge surface could not continue.",
+          "severity": "recoverable"
+        }
+      ]
+    }
+  }
+}
+```
+
+Recommended surface-level codes are `action_unavailable` when initialization
+fails, `action_expired` when the challenge session expired, and `action_failed`
+for another terminal surface error. Buyer abandonment and provider
+authentication outcomes may be unavailable to the surface or known only by the
+provider backend; the Business and payment provider remain authoritative for
+those distinctions. After either notification, the Platform follows the shared
+Payment Authentication reconciliation contract.
+
+## Combined Provider Flows
+
+A payment provider may use one wrapper URL to orchestrate device data collection
+and a challenge. The emitted Action type describes the Platform-facing behavior:
+
+- If the surface remains invisible and never requests Buyer interaction, the
+  Business emits device data collection.
+- If the surface may display a Buyer challenge, the Business emits 3DS challenge.
+
+The Business must not emit an invisible DDC Action and then make that surface
+visible. If collection completes first, the preferred flow is to remove the DDC
+occurrence and emit a new challenge occurrence in a later Checkout response.
+The two occurrences have different Action types and IDs even when their wrapper
+URL origin is the same.
+
+## Deadline and Fallback
+
+[EMV 3-D Secure Protocol and Core Functions v2.3.1.1](https://www.emvco.com/specifications/emv-3-d-secure-protocol-and-core-functions-specification-6/){ target="_blank" }
+gives the ACS 10 minutes (600 seconds) after successfully sending each challenge
+interface to the challenge iframe (Req 226). The surface and provider own that
+per-interface timer. It does not define a single UCP deadline beginning at the
+initial wrapper navigation.
+
+The Platform applies a bounded outer-surface deadline defined by the handler or
+Platform policy, measured from navigation and capped by Checkout `expires_at`.
+The Business **MUST** choose an expiry that accommodates the handler-defined
+window. The outer deadline governs only the Platform-owned container and does
+not replace or reset an ACS timer.
+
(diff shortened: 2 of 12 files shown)


docs/specification/checkout.md
```diff
@@ -451,6 +451,499 @@ The platform resolves the recoverable error programmatically while
...
+| `id`          | string   | ✓        | Unique, unguessable identifier for this action. Echoed in the result and used by the platform to correlate the result against the outstanding action. Treat as a capability token: origin checks are defense in depth, not the trust boundary.    |
+| `code`        | string   | ✓        | Reverse-domain identifier for the action category, owned by the capability or handler protocol that defines the action's semantics. See [Action Codes](#action-codes).                                                                              |
+| `severity`    | string   | ✓        | `optional` \| `recoverable` \| `requires_buyer_input`. See [Action Severity](#action-severity).                                                                                                                                                    |
+| `url`         | string   | ✓        | The URL to render or navigate to. **MUST** use the `https` scheme.                                                                                                                                                                                  |
```
aneeshali: The current proposal simplifies the action object by only providing a url (which implies a standard HTTP GET), and relies on the business/PSP to host a same-origin "relay" page if they need to perform a POST or carry rich payloads (like EMVCo's creq or threeDSMethodData). While this keeps the UCP platform implementation simple, it forces the business/PSP to host intermediate relay pages for common flows like 3DS. It might be more robust to allow UCP to transport the request configuration directly.

Please consider expanding the action object to support rich request payloads natively with http method, body, headers, encoding etc.

docs/specification/checkout.md
```diff
@@ -451,6 +451,499 @@ The platform resolves the recoverable error programmatically while
 rendering the allergen disclosure in proximity to the referenced line
 item.
 
+## Actions
+
+### Overview
```
aneeshali: The introduction of the actions primitive significantly alters the Checkout lifecycle and state transition flows. Instead of a single terminal complete_checkout call, we now have multi-step loops. We should explicitly document and visualize these flows where a complete can be followed by an _update_ and a subsequent _complete_. It might be better to have a dedicated state diagram to capture this variant when actions are involved.

Also wondering if we need to call out any checkout state transitions when actions are involved.

docs/specification/checkout.md
```diff
@@ -451,6 +451,499 @@ The platform resolves the recoverable error programmatically while
...
+{
+    "id": "act_dd_01HZX9TZK8Q1J5N3D7P2VQ4F0M",
+    "code": "dev.ucp.payment.device_data",
+    "severity": "optional",
+    "url": "https://psp.example.com/device-data/sess_xyz",
+    "display": { "form": "invisible" }
+}
+```
+
+**3DS challenge (modal, required).** The ACS lives on a different
```
aneeshali: I was wondering how we should be documenting the nuances of 3DS challenges so the PSPs have clarity on the exact communication for the following:
* DDC step and its optimization using updateCheckout.
* Passing additional signals to complement the DDC step.
Should any of these mentioned in the UCP and provided as a guide? OR should it be treated more of the payment handler responsibility, so they can have their own documentation?

aneeshali (commented):
Thanks a lot @raginpirate for this PR. Added a few comments based on the last discussion.

docs/specification/checkout.md
raginpirate: 👍 yeah this base shape needed work. I just pushed an update to simplify actions to their core: a rev-dns code + severity + config and opened up some 3ds as an example implementation of an action, with a defined transport mechanism and config in isolation. This is still heavily a WIP but hopefully the rough direction is making sense, and I'm happy to keep gathering this very early feedback!

docs/specification/payment-actions/device-data-collection.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Done Message
+
+When collection is complete, the loaded surface **MUST** send a `pa.done`
+notification to its parent/opener:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "method": "pa.done",
```
aneeshali: Hey @raginpirate while defining only a "pa.done" without any extra payload simplifies the work for the platform, please confirm if there are any security implications with this approach.

docs/specification/checkout.md
```diff
@@ -460,6 +451,82 @@ The platform resolves the recoverable error programmatically while
...
+checkout, the platform **MAY** satisfy escalation by loading `continue_url` in an
+embedded context.
+
+Completion is defined by the owning action specification. For some actions, the
+platform updates checkout state with a handler-defined field. For others, the
+rendered surface or invoked provider reports directly to the business, and the
+platform's next checkout call causes the business to observe that the external
+step has completed.
+
+### Examples
```
aneeshali: IIRC, there was a suggestion to also keep a non-3ds example to validate that the solution works more broadly. Student/military discounts and account verification were brought up as examples. It would be great if we could illustrate one such example as well.

aneeshali (commented):
Thanks a lot @raginpirate for a productive discussion. Left a couple minor comments.

docs/specification/payment-actions/device-data-collection.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Ignoring, Timeout, and Instrument Switching
+
+If the action is `optional` and the platform ignores it, the business **MUST NOT**
+treat the missing collection as a protocol error. The business may continue
+without the data, emit a required/blocking collection action later, request a
+challenge, decline the payment, or escalate according to the owning handler's
+rules.
+
+If the action times out or the surface fails to load, the platform follows the
```
dumitru-adyen: Have a question/comment on semantics of the DDC step. Based on EMVCo, device fingerprint (DDC) is **optional** and I cannot identify a clear instruction of the follow step in case of DDC failure (timeout after 10s for example) or even the fact that ACS nay not even provide a method URL. I would normally expect in case of failure of DDC step to proceed with authentication request (AReq) and notify that in the completion indicator _threeDSCompInd_ . Is this intentional or am I missing something ?

source/services/payment-actions/embedded.openrpc.json
```diff
@@ -0,0 +1,57 @@
...
+          "required": true,
+          "schema": {
+            "$ref": "../../schemas/common/types/reverse_domain_name.json",
+            "description": "Runtime action code for the completed action. Well-known values include dev.ucp.payment.device_data_collection and dev.ucp.payment.three_ds_challenge."
+          }
+        }
+      ]
+    },
+    {
+      "name": "pa.error",
```
dumitru-adyen: For payment service providers a bit more granularity here can be important to make a difference between a failure and a shopper drop off (shopper left the session without completion). From authentication perspective this metrics offer different insights and makes troubleshooting and correlations easier to spot.

docs/specification/checkout.md
dumitru-adyen: fully agree with @aneeshali here since it will present a problem both on DDC (which is a POST with threeDSMethodData to the ACS method URL) as well as the challenge step (a POST with the CReq to the ACS challenge URL).

docs/specification/payment-actions/device-data-collection.md
```diff
@@ -0,0 +1,195 @@
...
+
+This specification defines the standard payment device data collection action.
+It uses the common [Runtime Actions](../overview.md#runtime-actions) envelope and
+is intended for payment handlers that need the platform to run a scoped,
+invisible browser-capable collection step before payment processing continues.
+
+The action code is:
+
+```text
+dev.ucp.payment.device_data_collection
```
kmfreeman, line 23: Discussed a bit already, but in our case the ddc and challenge are wrapped under the same redirect url. My understanding is that as long as the postMessage data ("method": "pa.done") is used this clearly defines the completion as such. Would need a different payment-action defined.

docs/specification/payment-actions/three-ds-challenge.md
```diff
@@ -0,0 +1,220 @@
...
+### Done Message
+
+When the challenge interaction is complete from the platform-facing surface's
+perspective, the loaded surface **MUST** send a `pa.done` notification to its
+parent/opener:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "method": "pa.done",
```
kmfreeman: rather than having a separate postMessage design for UCP and our own direct integration or clientside SDK use we would likely add key:values in the postMessage for our own purposes. I would hope/assume these would be ignored by UCP?

these could be additional identifiers for support purposes or expanding on the detail of the outcome e.g. authenticated, rejected, unavailable....

docs/specification/payment-actions/device-data-collection.md
raginpirate: Hey @dumitru-adyen 👋                                                                                                                                        
 
 I think the key framing here is that this approach is intentionally trying to avoid re-implementing the EMVCo spec inside UCP. A few important points:       
                                                                                                                                                              
 - UCP does not specify how a business/PSP deals with receiving a DDC / 3DS Method URL from its provider.                                                     
 - The platform does not need to see or learn about DDC if it is skipped by either the provider or the business, including cases where the ACS does not provide a method URL.                                                                                                                                      
 - The platform also does not need to learn the domain outcome of the DDC step. It only needs to know that the platform-facing action surface has completed so it can re-drive the checkout operation.                                                                                                                 

You bring up a good point about timeouts though. I don’t want platforms to be responsible for the EMVCo-specific meaning of a DDC timeout or for setting/deriving provider state like threeDSCompInd. But we should define a UCP frame abandonment policy: if the invisible surface does not complete within the configured timeout, the platform can abandon/unmount the frame and re-drive the checkout operation, giving the business/PSP a chance to re-emit DDC, proceed without it, emit a challenge, or abandon/escalate based on its own provider state.

docs/specification/payment-actions/three-ds-challenge.md
raginpirate: Yup totally; additional properties is the supported default for any object passed over UCP. This can empower you to adhere to the subset standard while growing the superset and I'd advocate you can always push for more of that superset when it makes sense for that information to live with the platform and not just the business!

source/services/payment-actions/embedded.openrpc.json
raginpirate: From my POV I expected that metric to be conveyed on the backend source of truth between the business and the provider; I am not certain it makes sense for that context to be shared with the platform; an opaque "done" seems sufficient on their end, but I'm open to understanding if that data point (beyond error) is important to them.

docs/specification/payment-actions/device-data-collection.md
raginpirate: 👍 Yeah exactly! You can use the exact same endpoint for either action, as long as it adheres to the spec. And the business is the one who with the additional BE data of the payment request they start, can properly surface that URL with the right contextual action code.

docs/specification/checkout.md
raginpirate: Marking this thread as resolved but please top-level the concern again if we want to re-open how granular this spec should be for data flowing through the hands of platforms rather than just the business<->provider interaction 👀  I'm not sure if were all speaking the same thing in this thread.

docs/specification/payment-actions/device-data-collection.md
```diff
@@ -0,0 +1,254 @@
...
+  "method": "pa.error",
+  "params": {
+    "id": "ddc-1",
+    "code": "dev.ucp.payment.device_data_collection",
+    "error": {
+      "ucp": { "version": "{{ ucp_version }}", "status": "error" },
+      "messages": [
+        {
+          "type": "error",
+          "code": "action_failed",
```
aneeshali: I was wondering if we want to highlight some of the possible values for code along with action_failed, and document them as guidance.

source/services/payment-actions/embedded.openrpc.json
```diff
@@ -0,0 +1,74 @@
...
+    "title": "UCP Payment Action Embedded Protocol",
+    "description": "Embedded transport messages for standard UCP payment action surfaces. A platform loads an action config URL in an isolated browser-capable surface. The surface sends JSON-RPC 2.0 notifications to the platform via postMessage (or the equivalent native bridge) to indicate completion or fatal surface failure. Notifications MAY include additional provider-specific fields; platforms MUST ignore fields that are not defined by this protocol or by the active payment handler for UCP processing."
+  },
+  "servers": [],
+  "methods": [
+    {
+      "name": "pa.done",
+      "summary": "Payment action surface completed",
+      "description": "The loaded payment action surface notifies the platform that the scoped interaction is done from the platform-facing surface's perspective. This does not assert payment success or authentication success; the business determines the outcome when the platform resubmits the UCP operation.",
+      "params": [
```
aneeshali, line 52: Should we also set "additionalProperties": true to indicate this can include arbitrary payloads as well?

aneeshali (approved):
Thanks Daniel for driving this alignment across may parties.

docs/specification/payment-actions/three-ds-challenge.md
kmfreeman: @raginpirate sorry if this is a bit late but a few additional thoughts came up now I'm getting in to the details of feature creation:

- should `method` actually be `result` as this is the one direction [ postMessage](https://www.jsonrpc.org/specification#response_object)?
- if you make `result` (or `method` if I'm wrong) an object the design becomes more open in terms of allowing others to use the standard jsonrpc result design e.g.

```json
{
  "jsonrpc": "2.0",
  "result": {
    "event": "pa.done",  // used by UCP
    "outcome": "3dsAuthenticated",  // could be any key:value for use by the gateway e.g. used by clientside SDK
  },
  "id": 1
}
```

This is all within the context of wanting to have a single postMessage design for both direct 3ds integration (with the gateway) and extending it for UCP usage

<s>Also means that for other postMessages we plan to omit such as when a challenge is required we can just remove the `event` from the result object as UCP does not need to react to this</s>
we would probably add `pa.challenged` or `pa.prompt` instead so can be parsed reliably for the challenge case

raginpirate: Thanks @kmfreeman — using one postMessage design for both direct gateway integrations and UCP is definitely the goal!

I believe `method` is the correct JSON-RPC shape here. This is because we're expecting the frame to postMessage that event to the parent page, and result is undefined because there is no expected response to carry back to the parent page after receiving it.

Additional params such as `outcome` or provider correlation data can still support a direct gateway SDK. UCP Platforms only process the fields defined by the active Action contract and ignore other params.

docs/specification/payment-actions/three-ds-challenge.md
```diff
@@ -0,0 +1,217 @@
...
+    "id": "three-ds-challenge-1"
+  }
+}
+```
+
+`action.done` means only that the Platform should close the challenge surface
+and retrieve Checkout. It does not mean authentication or payment succeeded.
+After
+closing the surface, the Platform uses Get Checkout and does not call Complete
+Checkout again to resume the accepted operation.
```
aneeshali: With the approach to complete the flow by calling Get Checkout rather than Complete Checkout, please confirm if the expectation is that the Platform can only perform 2 things:
- Resume the flow if further actions are needed, like proceed with challenge once DDC is complete.
- Mark the flow as completed if the Get Checkout returns a response similar to a successful non-3DS complete checkout call.

I'm wondering how will some of the failure cases be handled gracefully in this scenario. For example, if 3DS fails for whatever reason, the platform will make a subsequent Get checkout call, and the merchant may repeatedly return the same action payload as before. If the platform wants to let the user choose a different payment method or similar remedies, there may not be a way to do it in this scenario. Please share your thoughts on how such cases can be handled.

docs/specification/payment-actions/three-ds-challenge.md
```diff
@@ -0,0 +1,217 @@
...
+those distinctions.
+
+## Combined Provider Flows
+
+A payment provider may use one wrapper URL to orchestrate device data collection
+and a challenge. The emitted Action type describes the Platform-facing behavior:
+
+- If the surface remains invisible and never requests Buyer interaction, the
+  Business emits device data collection.
+- If the surface may display a Buyer challenge, the Business emits 3DS challenge.
```
aneeshali, line 193: Some PSPs are adopting a slightly different approach for their merchants in the standard 3DS flows. They provide a single URL, which performs device data collection in a visible manner with some standardized loading screen. Depending on the result of DDC, they may either redirect the user to a challenge UI or just complete the flow and pass the control back to the merchant. We could either consider this as a subcase of three_ds_challenge or we could call out as a separate option. I think what the loading screen will look like is out of scope of this PR but providing clarity on that scenario will be helpful considering this is a relevant pattern.

aneeshali (commented):
Really appreciate the work on this and the related discussions, @raginpirate. Left a couple of comments to fully understand the implementation specifics.

raginpirate:
👋 I’ve force-pushed now that the base Actions primitive has landed.

The update aligns with the landed Action and Checkout lifecycle contracts:
- Removed the old occurrence-level required concept; each Action type statically defines what it gates.

Additional improvements I pushed in here:
- Signal the payment instrument id so its clear which element we're acting on.
- Added an action.ready handshake for both surfaces. The frame identifies the Action and extension version, and waits for acknowledgment before beginning work. better matching how all embedded protocols work in the spec today.
- Clarified origin validation, rendering isolation, reconciliation, fallback, and handler-authoring requirements.

docs/specification/payment-actions/three-ds-challenge.md
raginpirate: If the Business learns a terminal provider outcome, it must remove the original Action, replace it with new work under a new ID, or advance the Checkout. It must not keep returning an occurrence that can no longer be processed, and the Platform must not reopen the same occurrence.

An incomplete Checkout can still be updated to select another instrument. Once Complete Checkout has been accepted and the Checkout is complete_in_progress, Update and another Complete are intentionally unavailable. Recovery at that point follows the handler fallback, buyer handoff through continue_url, or cancellation and creation of a new checkout.

docs/specification/payment-actions/three-ds-challenge.md
raginpirate: One wrapper URL may orchestrate both provider steps. If the surface may ever become buyer-visible, the Business emits a `three_ds_challenge` Action, because honestly thats what they are doing as the action (ddc is simply a side-effect of their modal's purpose). A device_data_collection Action must remain invisible throughout its lifetime and should only be used by providers that do separate the steps.

docs/specification/payment-actions/device-data-collection.md
```diff
@@ -0,0 +1,208 @@
...
+<!-- ucp:example schema=shopping/payment_authentication def=dev.ucp.shopping.checkout extract=$.actions target=$.actions op=read -->
+```json
+{
+  "actions": {
+    "dev.ucp.payment.device_data_collection": [
+      {
+        "id": "ddc-1",
+        "config": {
+          "payment_instrument_id": "instrument_1",
+          "url": "https://payments.example.com/ucp/payment/ddc/session_123"
```
amithanda, line 68: How do we plan to support App channel which requires sdk and no url? Is that deferred to future? Even if we plan to defer it, do we have high level idea on how that will come in? Will this same extension expand to support it or will it be a sibling action e.g dev.ucp.payment.three_ds_sdk_challenge or similar?

docs/specification/payment-actions/three-ds-challenge.md
```diff
@@ -0,0 +1,230 @@
...
+[Payment Authentication extension schema](site:schemas/shopping/payment_authentication.json).
+
+| Field | Type | Required | Notes |
+| :---- | :--- | :------- | :---- |
+| `payment_instrument_id` | string | ✓ | ID of the associated instrument in the containing Checkout. |
+| `url` | string | ✓ | Absolute HTTPS URL for the buyer-facing challenge surface. |
+
+The Business **SHOULD** provide a Business- or provider-operated wrapper URL that
+owns any provider-specific POST, such as submitting CReq to an ACS challenge
+URL. The Platform performs an ordinary navigation to `config.url`; it
```
amithanda: How are we thinking about the trust anchoring (invariant 11 from #517)?

source/services/payment-actions/embedded.openrpc.json
```diff
@@ -0,0 +1,88 @@
...
+        {
+          "name": "id",
+          "required": true,
+          "schema": {
+            "type": "string",
+            "description": "ID of the outstanding Action occurrence represented by this surface."
+          }
+        },
+        {
+          "name": "error",
```
archrao, line 78: In device-data-collection.md and three-ds-challenge.md, codes like action_unavailable, action_expired, and action_failed are listed as "recommended" surface-level error codes for action.error. To ensure consistency and prevent payment providers from inventing custom error strings, can we formalize these error codes as an enum here?

docs/specification/payment-authentication.md
```diff
@@ -0,0 +1,325 @@
...
+capability-specific Embedded Protocol `ready` method or `MessageChannel` upgrade;
+`action.ready` is its scoped handshake.
+
+Each Action occurrence uses a fresh, isolated browsing context. Device data
+collection is hidden and non-interactive. A 3DS challenge is visible, blocks
+interaction with the rest of Checkout, and **SHOULD** show a loading state until
+its content renders. Web Platforms use an iframe or other handler-approved
+browser context. For messages from a native surface to the Platform, native
+Platforms **MUST** expose a JSON-string bridge through
+`window.EmbeddedActionProtocolConsumer` or
```
archrao: nit - add (Android)

docs/specification/payment-authentication.md
```diff
@@ -0,0 +1,325 @@
...
+`window.webkit.messageHandlers.EmbeddedActionProtocolConsumer`. For the ready
```
archrao: nit - add (iOS)

docs/specification/payment-authentication.md
```diff
@@ -0,0 +1,325 @@
...
+Payment Authentication uses the Action methods below. It does not use a
+capability-specific Embedded Protocol `ready` method or `MessageChannel` upgrade;
+`action.ready` is its scoped handshake.
+
+Each Action occurrence uses a fresh, isolated browsing context. Device data
+collection is hidden and non-interactive. A 3DS challenge is visible, blocks
+interaction with the rest of Checkout, and **SHOULD** show a loading state until
+its content renders. Web Platforms use an iframe or other handler-approved
+browser context. For messages from a native surface to the Platform, native
+Platforms **MUST** expose a JSON-string bridge through
```
archrao, line 251: Since Android bindings typically expect a method signature taking stringified JSON, whereas iOS WebKit expects a JavaScript object, should we add explicit JS snippet examples for both Android and iOS bridge invocations in the spec to avoid SDK interoperability issues? This will provide clarity for developers building cross-platform bridge requirements

docs/specification/payment-authentication.md
```diff
@@ -0,0 +1,325 @@
...
+from its server-side payment-provider state.
+
+## Decline, Failure, and Abandonment
+
+A Platform **MAY** decline an instance that violates its runtime policy. If it
+cannot process the instance, it **SHOULD** use a valid `continue_url` to hand the
+session to the Business when available.
+
+For an Action on an `incomplete` Checkout, the Platform may instead choose a
+different payment instrument through an ordinary Update and later submit a new
```
archrao, line 214: When a buyer switches payment instruments via Update Checkout (e.g., from Card A to Card B), should we explicitly state here that the Business must purge any outstanding Action instances associated with the previous instrument ID from the `actions` map?

docs/specification/payment-actions/three-ds-challenge.md
simonwfarrow: I am not sure a `GET` call is the correct approach here. A Platform calling `GET` may well trigger a mutating action on the checkout resource on the Business side. This should be modelled as a `PUT` or `PATCH` on the checkout resource.

It would also means the agent cannot pass on any extra data that might be required as an output of the action which took place. To be extensible, I think an action outcome executed on the Platform should allow additional properties to be passed to the Business. An action definition should define which properties are required/allowed and the checkout update call (via a PATCH/PUT/POST) can submit these as part of the action outcome.

Taking the example from PR #582 , the response to the complete checkout call contains the required actions. As this forms part of the checkout contract and follows REST principles, it would naturally follow a subsequent call to update the checkout resource  could also update the action with any extra fields required:

Response:
```json
"actions": {
    "com.example.identity.student_verification": [
      {
        "id": "verify-student-1",
        "config": {
          "verification_url": "https://business.example.com/verify/abc"
          "required_fields" : ["student_id"]
        }
      }
    ]
  },
```

Request:
```json
"actions": {
    "com.example.identity.student_verification": [
      {
        "id": "verify-student-1",
        "student_id": "123"
      }
    ]
  },
```

simonwfarrow: To add a more specific 3ds example, if an external MPI is in use , allowing the outcome of the 3ds action to carry an identifier that identifies the authentication result on the 3ds provider would be very useful.

i.e.

```json
"actions": {
    "dev.ucp.payment.three_ds_challenge": [
      {
        "id": "three_ds_challenge",
        "authentication_id": "athlf5Y0SJfc_azJ746ry8Ja"
      }
    ]
  },
```

docs/specification/payment-actions/device-data-collection.md
raginpirate: Yes we should defer, and I believe it should be independent from this extension entirely; not just a different action.

There is no one unified 3DS SDK at this time: EMVCo specifies the SDK's interfaces and approves implementations, but ships none — so every provider carries their own approved build. As a result what we'd be looking at is a unique SDK per provider that must be negotiated as an independent extension today, enabling:
- a unique rev-dns signalled contract for what SDK the platform has agreed to implement and the business has agreed to advertise.
- the business to give the native signal to their payment provider at payment time, and propagate the native signals back to the platform.
- the platform to then embed that specific provider's contract for how to use their specific SDK, and a result to finally land with the business (either in-spec or out-of-band).

Even if we imagine one SDK existing and shared across all providers, I still think this fits the bill for a fundamentally isolated extension because theres almost nothing shared between the two specs: different trust, implementation, parameters, negotiation.

docs/specification/payment-actions/three-ds-challenge.md
raginpirate: Good prompt to re-discuss 11. The handler specification names the origins its Action surfaces load from — the handler guide and publishing checklist now ask for them — and a Platform declines any occurrence whose config.url origin isn't among them, so the expectation exists before the Action is emitted and a response can't bootstrap its own trust. This differs a bit from 11, which overcorrects to a relation between providers and platforms; payment handlers can (and often should) state that businesses are allowed to advertise 3ds urls themselves, with any given business using only the subset it actually operates. Platforms have the easy escape hatch of escalation if they do not actively support the full subset as well.

source/services/payment-actions/embedded.openrpc.json
raginpirate: Good question, I've always deeply struggled with enums in UCP (or I should say, JSON schemas) to be honest 😅 

But... code is an open string everywhere in UCP — error_code.json says freeform codes are permitted — so an enum here would make this extension stricter than the core type it references, and closing an enum makes it so we have to author a breaking change if we try to extend it. It also wouldn't change any behaviour: a Platform unmounts and reconciles the same way for all three codes and for any code it doesn't recognize.

docs/specification/payment-authentication.md
raginpirate: The bridge is already a JSON string both ways: the Embedded Protocol defines postMessage(message: string) carrying a JSON-RPC message, and this spec says the same for the ready response back to the surface. WKWebView accepts a string fine, so one surface implementation works on both without branching. 

Examples are not a bad idea, that could be a good addition to bring to embedded-protocol.md so every capability gets them rather than just this one! But maybe we can descope that from this specific PR.

raginpirate:
Pushed up a small doc improvement based on last comments, please re-review!

docs/specification/payment-actions/three-ds-challenge.md
amithanda: Thanks for clarifying the mechanism for Invariant 11, @raginpirate!

Quick follow up: I want to clarify whether we should keep this strictly as **specification prose** or if we should add a **machine-readable schema property** to support automated Platform validation.
Currently, because allowed origins are only required in the handler's markdown specification and publishing checklist (`payment-handler-guide.md`), a Platform cannot dynamically validate `config.url` against an allowlist using JSON Schema validation or Discovery profile introspection. Today, Platforms implementing this spec would need to manually configure out-of-band allowlists per `handler_id` based on reading published documentation.
You also noted a great point: *"payment handlers can (and often should) state that businesses are allowed to advertise 3ds urls themselves, with any given business using only the subset it actually operates."* 
When Businesses advertise their own subset of 3DS URLs, having a machine-readable property (e.g., an optional `allowed_origins: string[]` in `source/schemas/payment_handler.json` or in the Business/Platform profile declaration) would allow a Platform to automatically:
1. Ingest the negotiated origin allowlist during discovery/profile exchange.
2. Programmatically reject any runtime `config.url` whose origin falls outside that allowlist before mounting the iframe.

**Question for alignment:**
Should we add an optional `allowed_origins` array to `payment_handler.json` in this PR so Platforms can programmatically enforce Invariant 11 out of the box, or is there a strong reason to keep this strictly as a prose/governance requirement for now and defer schema-level origin discovery to a follow-up enhancement?

docs/specification/payment-actions/device-data-collection.md
drewolson-google: Agreed on deferring. Many 3DS implementations fall back to web regardless, so this seems like a reasonable first pass.

docs/specification/payment-actions/three-ds-challenge.md
```diff
@@ -0,0 +1,230 @@
...
+# Payment 3DS Challenge Action
+
+This specification defines the 3DS challenge Action type declared by the
+[Payment Authentication extension](../payment-authentication.md):
+
+```text
+dev.ucp.payment.three_ds_challenge
+```
+
+It asks the Platform to present a scoped, buyer-facing payment-authentication
```
drewolson-google: What does "scoped" mean here?

docs/specification/payment-actions/three-ds-challenge.md
```diff
@@ -0,0 +1,230 @@
...
+present an issuer or provider challenge to the Buyer. UCP does not carry EMV 3DS
+payloads such as CReq/CRes, AReq/ARes, RReq/RRes, or `transStatus`. Those details
+remain within the loaded surface and the Business, payment handler, PSP, 3DS
+server, issuer, or ACS integration.
+
+The Platform presents the surface and observes only when the Platform-facing
+interaction finishes or cannot continue. The Business determines the
+authentication and payment outcome from its server-side provider state.
+
+## Emission and Gated Effect
```
drewolson-google: This title is kind of weird. Maybe "Blocking Action"?

docs/specification/payment-actions/three-ds-challenge.md
```diff
@@ -0,0 +1,230 @@
...
+  "params": {
+    "id": "three-ds-challenge-1",
+    "error": {
+      "ucp": { "version": "{{ ucp_version }}", "status": "error" },
+      "messages": [
+        {
+          "type": "error",
+          "code": "action_failed",
+          "content": "The challenge surface could not continue.",
+          "severity": "recoverable"
```
drewolson-google, line 171: I'm assuming this example is demonstrating some kind of transient failure -- something failed in the challenge but we can try again. Is that correct?

drewolson-google (approved):
Looks good, a few small comments but none are blocking.

aneeshali:
Hey @raginpirate also sharing a couple sequences we discussed to capture the potential e2e flows for unified URL and separate URL.

### DDC+Challenge unified

```mermaid
sequenceDiagram
    participant User
    participant Platform as Platform
    participant Business as Business
    participant PSP as PSP (or PSP + 3DS Provider)
    participant 3DS as Issuer ACS (Bank)

    User->>Platform: Express Purchase Intent
    Platform->>Business: POST /checkout-sessions/{id}/complete (token)
    Business->>PSP: Authorize
    PSP-->>Business: Authentication URL
    Business-->>Platform: Checkout (status: complete_in_progress, action: three_ds_challenge [url])

    Platform->>PSP: Render URL
    PSP-->>Platform: postMessage (action.ready)
    Platform-->>PSP: postMessage (ready response)
    Note over Platform, PSP: DDC loading screen<br/>(Served by PSP • Standardized appearance if Platform mandates)

    PSP->>3DS: 3DS Method + Authentication
    3DS-->>PSP: Frictionless, OR Challenge Required

    opt Challenge Required
        3DS-->>Platform: ACS Challenge Screen (Rendered by Issuer)
        User->>3DS: Complete Challenge (OTP/Biometric)
    end

    Note over Business, PSP: PSP shares the results with Business

    PSP-->>Platform: postMessage (action.done)
    Platform-->>PSP: postMessage (done response)

    Platform->>Business: GET /checkout-sessions/{id}
    Business-->>Platform: Checkout (status: completed)
    Platform->>Platform: Show Transaction Success
```

### DDC+Challenge separate

```mermaid
sequenceDiagram
    participant User
    participant Platform as Platform
    participant Business as Business
    participant PSP as PSP (or PSP + 3DS Provider)
    participant 3DS as Issuer ACS (Bank)

    Note over Platform, 3DS: Phase 1: Invisible Device Data Collection (DDC)
    User->>Platform: Express Purchase Intent
    Platform->>Business: POST /checkout-sessions/{id}/complete (token)
    Business->>PSP: Authorize / Initiate DDC
    PSP-->>Business: DDC URL (Invisible 3DS Method)
    Business-->>Platform: Checkout (status: complete_in_progress, action: device_data_collection [url])

    Platform->>PSP: Render Invisible Surface (Hidden Iframe / Webview)
    Note over Business, PSP: PSP shares the results with Business
    PSP-->>Platform: postMessage (action.ready)
    Platform-->>PSP: postMessage (ready response)
    PSP->>3DS: 3DS Method (Device Data Collection)
    3DS-->>PSP: 3DS Method Complete
    PSP-->>Platform: postMessage (action.done)
    Platform-->>PSP: postMessage (done response)

    Note over Platform, 3DS: Phase 2: Check Payment Outcome (Frictionless vs. Challenge)
    Platform->>Business: GET /checkout-sessions/{id}

    alt Case A: Frictionless (No Challenge Needed)
        Business-->>Platform: Checkout (status: completed)
        Platform->>Platform: Show Transaction Success
    else Case B: Challenge Required
        Business-->>Platform: Checkout (status: complete_in_progress, action: three_ds_challenge [url])
        Platform->>PSP: Render Visible Challenge Surface (config.url)
        PSP-->>Platform: postMessage (action.ready)
        Platform-->>PSP: postMessage (ready response)
        3DS-->>Platform: ACS Challenge Screen (Rendered by Issuer)
        User->>3DS: Complete Challenge (OTP/Biometric)
        Note over Business, PSP: PSP shares the results with Business
        PSP-->>Platform: postMessage (action.done)
        Platform-->>PSP: postMessage (done response)
        Platform->>Business: GET /checkout-sessions/{id}
        Business-->>Platform: Checkout (status: completed)
        Platform->>Platform: Show Transaction Success
    end
```

source/services/payment-actions/embedded.openrpc.json
archrao: The reasoning makes sense, thanks!

archrao (approved):
lgtm. Thanks for the work on the PR.

docs/specification/payment-actions/three-ds-challenge.md
raginpirate: Thanks for the input @simonwfarrow!

> A Platform calling GET may well trigger a mutating action on the checkout resource on the Business side. This should be modelled as a PUT or PATCH on the checkout resource.

On this, the get request's purpose is to get the checkout state, not to perform a mutating action; that is the result of the out-of-band communication between the acs, psp, and business which the platform was not involved in.

> It would also means the agent cannot pass on any extra data that might be required as an output of the action which took place.

This is true, and a point acknowledged in https://github.com/Universal-Commerce-Protocol/ucp/pull/582#discussion_r3611764504. Anything performed as a part of an action returned in a complete_in_progress must not require directly mutating the checkout object state at this time. If something should require directly updating the checkout state, the correct answer is for that checkout to fail and return to a ready_for_complete state at this time.

If product scenarios were to arise that are blocked by the current state machine of `complete_in_progress` we should continue those discussions on the checkout state machine! In the case you showed here, we do not want 3ds authentication urls to become a state management problem shared with platforms at this time.

docs/specification/payment-actions/three-ds-challenge.md
raginpirate: Scoped to the specific checkout state... but it was strange to mention this here. Updated the wording!

docs/specification/payment-actions/three-ds-challenge.md
raginpirate: Thanks for the deep thinking Amit. You're right about the current state: origins live only in the handler's prose and publishing checklist, so a platform does have to configure them per handler out of band, by hand. I think that's the correct cost here, and **_I'd argue against changing it now_**.

When a platform implements a payment handler there's a lot of prose and out-of-band certificate negotiation etc. that brings the platform into the handler's trust boundary, so it can use delegated payment APIs. Part of that can be a publicly or privately documented set of endpoints to allowlist.

For businesses it goes the same way, but the trust tier matters. For a normal, low-trust business the platform has no guarantees about the state of a 3DS action it would perform. When a business enters a high-trust relationship with a platform there's the same out-of-band negotiation, and that should be what yields the acknowledged trusted origins for platform-owned actions. UCP doesn't aim to classify low- vs high-trust business↔platform relationships in core — we build for low-trust.

So: for low trust, an allowed-origin list protects nobody. The business declares N, emits an N+1th, and just updates the declaration — it's a check the checked party holds the pen on. The only thing it catches is a business's own bug, and I don't want a field that reads to every platform as a security boundary when it isn't one. For high trust, the list should be maintained out of band, precisely so the trust boundary can't be renegotiated mid-flight.

docs/specification/payment-actions/three-ds-challenge.md
simonwfarrow: Thanks for coming  back to me.

On GET call after the action completes, it's not clear to me how a business would know to check the status of the action. In the case of 3ds, a lot of the time that would require the business to invoke their 3ds suppliers endpoint to retrieve the status of the authentication. Are they supposed to do this on every GET call? Can there only be a single in flight action, and does a business need to infer that if the action is in-flight any GET calls must check the status of that action?

scripts/scaffolds/shopping_payment_authentication_response.json
```diff
@@ -0,0 +1,48 @@
...
+    "instruments": [
+      {
+        "id": "instrument_1",
+        "handler_id": "card_handler_1",
+        "type": "card",
+        "selected": true
+      }
+    ]
+  },
+  "actions": {}
```
sumudul:
```suggestion
  }
```

Drive-by observation on this scaffold, as per the [overview](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/docs/specification/overview.md?plain=1#L97), `actions` key must be omitted when there are no outstanding actions. I realize scaffolds are CI infrastructure rather than rendered spec, so nothing non-conformant reaches readers. The only concern would be just that a file named `*_response.json` is a likely copy-source for future extension authors.

docs/specification/payment-actions/three-ds-challenge.md
raginpirate: Good questions! Answering each directly.

**_Do they check on every GET?_** No. There's only provider state to resolve while an occurrence is outstanding on that payment attempt, and the window is bounded: after action.done the Platform polls Get Checkout with bounded backoff. The expected path is that the provider pushes the payment-attempt state change (or forwards the ACS notification) out of band, so the GET reads already-applied state. Resolving lazily on GET while an occurrence is outstanding is also valid but not enforceable by spec, and a reasonable safety net if a push is dropped or not something the provider does.

**_Single in-flight?_** For this extension, yes — the Business emits only the next interaction the Platform can process (DDC, then a challenge under a new id in a later response).

On carrying authentication_id back through the Platform you asked earlier, I'll replay now that I grasped the concern a bit better: with an external payment attempt that identifier exists server-side before the challenge is presented, so relaying it isn't new information — and the Business must validate any Platform-supplied reference against its own attempt binding anyway, so the relay buys no trust while adding an attempt-substitution surface. Where a provider genuinely produces a value in the browser (a result JWT, for example), the required Business- or provider-operated wrapper URL is same-origin with that provider: the wrapper posts it to its own backend and the Platform stays out of the path. That's what makes an action-outcome payload unnecessary for 3DS specifically.

amithanda:
**LGTM! Approving this PR.** 🎉

Huge thanks to `@raginpirate`, `@drewolson-google`, and everyone who contributed to this design. With **PR #582** landing the common extension-owned `actions` primitive in `dev.ucp.shopping.checkout`, **PR #458** cleanly delivers the concrete browser-surface abstraction for payment authentication (`device_data_collection` and `three_ds_challenge`).

Below is a quick recap of how this stacked implementation tracks against the **14 Design Invariants from Issue #517**

### 1. Progress Against Issue #517 Design Invariants
| # | Invariant | Status | How PR #582 + PR #458 Satisfy the Requirement |
| :- | :-------- | :----- | :-------------------------------------------- |
| **1** | **In-band resumability** | **Covered** | Actions gate completion during `complete_in_progress` (or `incomplete` for DDC). After `action.done`, the Platform re-drives `GET /checkout-sessions/{id}` to resume the same checkout in-band without creating a new session. |
| **2** | **Native delegation on common path** | **Covered** | Frictionless 3DS (~90% of attempts) completes immediately without emitting an Action. Step-ups render in Platform-managed modals/frames without forcing top-level redirects away from the merchant/platform. |
| **3** | **Hidden/background execution** | **Covered** | Explicit separation between `dev.ucp.payment.device_data_collection` (which **MUST** run in an invisible browsing context) and `dev.ucp.payment.three_ds_challenge` (which **MUST** run in a visible modal). |
| **4** | **Headless / out-of-band** | **Deferred / Fallback** | Non-browser/headless channels (e.g. mobile app-to-app SDKs or decoupled push) gracefully degrade to `continue_url` or Update Checkout today. Native SDK channels are deferred to an independent future extension. |
| **5** | **Domain-generic, validated** | **Covered** | PR #582 defines the domain-generic `actions` map in core Checkout. PR #458 proves the architecture by layering two concrete payment action types without core schema changes. |
| **6** | **Zero overhead on no-action path** | **Covered** | When no authentication action is required, the `actions` map is absent or empty—adding zero payload overhead or latency. |
| **7** | **Explicit, separable lifecycle** | **Covered** | Action occurrences have a separable lifecycle gated by `id`. Surface completion (`action.done`) finishes the UI interaction but does not assert payment success; authorization declines follow standard Checkout error rules. |
| **8** | **Idempotency / replay safety** | **Covered** | Action occurrences are **single-use**. After `action.done`, `action.error`, timeout, or dismissal, Platforms **MUST NOT** reopen the same occurrence `id`. Retries require a fresh `id` from the Business. `GET` polling is replay-safe. |
| **9** | **Standardized recovery** | **Covered** | Explicit recovery paths: `GET` with bounded backoff (up to 30s) after `action.done`; single `GET` after errors/timeouts, falling back to `continue_url` handoff or instrument switching via Update Checkout. |
| **10** | **Server-side outcome authority** | **Covered** | `action.done` / `action.error` are purely **advisory** Platform signals. Platforms never inspect authentication tokens; the Business determines success exclusively from server-side provider state. |
| **11** | **Independent input trust (`config.url`)** | **Covered** | Trust is anchored to the negotiated Payment Handler's published allowlist rather than Business-asserted URLs (see deep dive below). |
| **12** | **Regulatory & PCI integrity** | **Covered** | PANs, OTPs, and banking credentials never transit UCP schemas or Platform servers. All 3DS payloads stay within the isolated iframe/webview between the browser and the ACS. |
| **13** | **Forward-compatible & vendor-agnostic** | **Covered** | Open reverse-DNS action keys and error strings. Zero vendor-specific EMVCo fields (`CReq`/`CRes`/`threeDSMethodData`) are hardcoded into UCP schemas. |
| **14** | **Graceful degradation to `continue_url`** | **Covered** | If a Platform cannot mount a surface (policy violation, unsupported environment, or unrecoverable error), it falls back to `continue_url` to hand the session to the Business. |

On Invariant #11 — A critical security requirement from #517 is that **a Platform must never trust `config.url` simply because a Business returned it in a Checkout response.** An attacker or compromised Business could otherwise inject a phishing URL (`https://evil-phishing.com/login`) into an iframe or modal.

I reviewed the latest changes again and confirmed that this is **sufficiently covered** across the core extension, action specifications, and authoring governance:
* **Core Extension (`payment-authentication.md`)**: Mandates that Platforms MUST verify `config.url` against the handler's published allowlist before rendering, MUST decline untrusted origins, and MUST enforce CSP `frame-src` restrictions to prevent cross-origin redirects.
* **Action Specifications (`device-data-collection.md` & `three-ds-challenge.md`)**: Explicitly require in both their *Platform Behavior* and *Security* sections that the handler's trust policy authorize initial and redirect origins before loading any surface.
* **Authoring Governance (`payment-handler-guide.md` & `payment-handler-template.md`)**: Enforces origin allowlist documentation as a mandatory item in the Payment Handler publishing checklist.

docs/specification/payment-actions/device-data-collection.md
```diff
@@ -0,0 +1,207 @@
...
+# Payment Device Data Collection Action
+
+This specification defines the device data collection Action type declared by
+the [Payment Authentication extension](../payment-authentication.md):
+
+```text
+dev.ucp.payment.device_data_collection
+```
+
+It asks the Platform to mount an invisible payment-authentication surface to complete device data collection.
```
igrigorik: nit: line wrap

docs/specification/payment-actions/device-data-collection.md
```diff
@@ -0,0 +1,207 @@
...
+
+The config shape is defined inline by the
+[Payment Authentication extension schema](site:schemas/shopping/payment_authentication.json).
+
+| Field | Type | Required | Notes |
+| :---- | :--- | :------- | :---- |
+| `payment_instrument_id` | string | ✓ | ID of the associated instrument in the containing Checkout. |
+| `url` | string | ✓ | Absolute HTTPS URL for the invisible collection surface. |
+
+The Business **SHOULD** provide a Business- or provider-operated wrapper URL that
```
igrigorik: What is a "wrapper URL"? Same applies to other action doc.

docs/specification/payment-actions/device-data-collection.md
```diff
@@ -0,0 +1,207 @@
...
+   mounted surface; and
+5. unmount the surface after `action.done`, `action.error`, load failure, or
+   timeout.
+
+Mounting the surface **MUST** follow the shared
+[Payment Authentication rendering contract](../payment-authentication.md#surface-rendering-and-notifications)
+and [Embedded Protocol security requirements](../embedded-protocol.md#security).
+
+On the web the surface is typically a hidden iframe. A native Platform may use
+an isolated webview or equivalent browser surface. It must not be visible to the
```
igrigorik:
```suggestion
an isolated webview or equivalent browser surface. It MUST NOT be visible to the
```

docs/specification/payment-actions/device-data-collection.md
```diff
@@ -0,0 +1,207 @@
...
+{
+  "jsonrpc": "2.0",
+  "method": "action.done",
+  "params": {
+    "id": "ddc-1"
+  }
+}
+```
+
+The surface may send `action.done` when handler-owned logic determines that the
```
igrigorik: nit: this sentence feels redundant give statement right above?

docs/specification/payment-actions/three-ds-challenge.md
```diff
@@ -0,0 +1,229 @@
...
+# Payment 3DS Challenge Action
+
+This specification defines the 3DS challenge Action type declared by the
+[Payment Authentication extension](../payment-authentication.md):
+
+```text
+dev.ucp.payment.three_ds_challenge
+```
+
+It asks the Platform to present a buyer-facing payment-authentication surface to complete 3ds challenges.
```
igrigorik: nit: line wrap

docs/specification/payment-actions/device-data-collection.md
```diff
@@ -0,0 +1,207 @@
...
+          "content": "Device data collection could not start.",
+          "severity": "recoverable"
+        }
+      ]
+    }
+  }
+}
+```
+
+Recommended surface-level codes are `action_unavailable` when initialization
```
igrigorik: Recommended: are these the "well-known" codes we are registering? If so, let's extract them into a table and state them as such, same as we do for all other well-known values in the spec.

docs/specification/payment-actions/three-ds-challenge.md
simonwfarrow: I guess this assumes all providers backing actions will post async updates back to the business. Do we want that restriction in this model? I am not clear on why we moved from POST to GET after a resulting postMessage, as it forces the async model and stops transfer of any action data. It feels like that puts quite tight restrictions on actions and the providers that are backing them.

docs/specification/payment-actions/device-data-collection.md
```diff
@@ -0,0 +1,207 @@
...
+handler can safely continue past should normally result in `action.done`, with
+the Business and provider recording the domain outcome server-side. After either
+notification, the Platform follows the shared Payment Authentication
+reconciliation contract.
+
+## Deadline and Fallback
+
+[EMV 3-D Secure Protocol and Core Functions v2.3.1.1](https://www.emvco.com/specifications/emv-3-d-secure-protocol-and-core-functions-specification-6/){ target="_blank" }
+requires the 3DS Server to set `threeDSCompInd` to `N` when the 3DS Method does
+not complete within five seconds (Req 315). That provider-domain timer begins
```
igrigorik, line 178: Running a quick check, it looks like 5-second value is introduced by SB-279 not 315 which it supersedes?

docs/specification/payment-actions/three-ds-challenge.md
```diff
@@ -0,0 +1,229 @@
...
+  "params": {
+    "id": "three-ds-challenge-1"
+  }
+}
+```
+
+`action.done` means only that the Platform should close the challenge surface
+and retrieve Checkout. It does not mean authentication or payment succeeded.
+After closing the surface, the Platform uses Get Checkout and does not call
+Complete Checkout again to resume the accepted operation.
```
igrigorik: nit: resume reads like an operation, but GET is and should be side effect free. Let's spell it out...

```suggestion
After surface is closed, the platform should issue Get Checkout to confirm 
processing state. Get Checkout is a read-only observation and does not
itself advance the payment attempt; the Business advances Checkout state 
from server-side provider state (typically an out-of-band provider 
push/callback), independent of when the Platform calls Get Checkout.
```

docs/specification/payment-authentication.md
```diff
@@ -0,0 +1,334 @@
...
+      }
+    ]
+  }
+}
+```
+
+A Business normally emits only the next interaction the Platform can process.
+If collection must precede a challenge, it emits the DDC Action first and emits
+the challenge under `dev.ucp.payment.three_ds_challenge` only in a later
+Checkout response.
```
igrigorik: Let's strengthen "normally"...

```suggestion
A Business **SHOULD** keep at most one Payment Authentication Action
outstanding per payment attempt at a time, emitting only the next interaction
the Platform can process, because an invisible device data collection surface
and a visible 3DS challenge cannot be presented coherently at the same time.
When collection must precede a challenge, the Business emits the device data
collection Action first and emits the challenge under
`dev.ucp.payment.three_ds_challenge` only in a later Checkout response.
```

source/schemas/shopping/payment_authentication.json
```diff
@@ -0,0 +1,88 @@
...
+    "dev.ucp.shopping.checkout": {
+      "allOf": [
+        {
+          "$ref": "checkout.json"
+        },
+        {
+          "type": "object",
+          "properties": {
+            "actions": {
+              "type": "object",
```
igrigorik, line 17: Actions are response-only (platform never sends them), so request visibility must be omit...

```suggestion
              "type": "object",
              "ucp_request": "omit",
```

igrigorik (changes requested):
Great work on this! A couple of minor authoring nits and this should be ready to land. 

Meta and non-blocking note: we duplicate _a lot_ of content across action docs. I would have preferred  that we consolidate the common bits under payment-authentication.md and slim the actions docs.

docs/specification/payment-authentication.md
raginpirate: Good point to make this feel less handwavy. One correction I'll make; one at a time makes sense if your doing ddc -> 3ds for one instrument; if splitting, its actually appropriate to do ddc + ddc, and then some number of linear 3ds prompts after (0, 1, or 2 but in sequence).

docs/specification/payment-actions/device-data-collection.md
raginpirate: 315 is the standard requirement; SBs are just the bulletins that announce what will become a standard requirement.

aneeshali (approved):
Thank you so much @raginpirate for driving the alignment across many stakeholders.

docs/specification/payment-authentication.md
igrigorik: Fair, lets leave this as is.

raginpirate:
👋 Before we merge, I want to acknowledge an open ask from providers to extend this spec. At the moment, as @aneeshali diagrams above, we expect providers to push their payment authentication state to businesses, such that the state machine moves forward without a reliance on a platform -> business message. An alternative path is a pull-based solution, in which the platform signals the business to check/move the state machine forward rather than the provider directly. However, when a checkout is `complete_in_progress`, you cannot update its state, preventing this class of actions from being authored.

Outside the scope of this first release, we should discuss whether or not we'd like to support these entity-modifying actions, and how we consider evolving the lifecycle of checkout to support it. cc @simonwfarrow thanks for pushing on this topic.

docs/specification/payment-actions/three-ds-challenge.md
raginpirate: Thanks for this @simonwfarrow, captured your stance top-level. Not an unfair ask at all, and we can take this as a follow up post-release of the first version 👍
https://github.com/Universal-Commerce-Protocol/ucp/pull/458#issuecomment-5185068293

igrigorik (approved):
LGTM and great work.

Push vs. pull: aligned and agree, great followup discussion that can build on the foundation defined in this PR.

amithanda (approved):

docs/specification/payment-actions/three-ds-challenge.md
simonwfarrow: Thank you. Sorry one final point . We recommend to our customers to also send in device data as a backup in case DDC fails . UCP will not support this, which may result in a higher number of challenges. See [How much data to provide](https://docs.worldpay.com/access/products/3ds/web/authentication#how-much-data-to-provide)

aneeshali: Hi @simonwfarrow this is something UCP already supports via [signals](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/shopping/types/signals.json) that can be sent as part of completeCheckout. UCP doesn't dictate the exhaustive list of signals however, that should be between the Platform and the Business. As long as the Business is able to pass through these signals to the PSP, we could assume that the PSP has the same visibility as a non-UCP 3DS flow.

simonwfarrow: Ah great. So on the first post to complete_checkout a platform will share the signals. The business must decide which signals to apply in the authentication request and create the required DDC/Challenge action. Thanks.

PR #466, opened 2026-05-21
fix!: require merchant_authorization in AP2-active checkout responses

sakinaroufid:
# Description

The spec mandates `ap2.merchant_authorization` in three places:

- `ap2-mandates.md:112` — "The business **MUST** include `ap2.merchant_authorization` in all checkout responses."
- `ap2-mandates.md:146-148` - "Businesses **MUST** embed their signature in the checkout response body under `ap2.merchant_authorization`..."
- `ap2-mandates.md:216-218` — "The checkout mandate **MUST** contain the full checkout response including the `ap2.merchant_authorization` field."

But `ap2_with_merchant_authorization` had no `required` array, so a response that omits the field validates clean. A spec-driven implementer ships a business that emits non-AP2-conformant responses, breaking the cryptographic chain at the platform's mandate-verification step — only discoverable when payment processors reject the mandate downstream.

The chain works only if every link is present:

1. Business signs the checkout body → JWS in `ap2.merchant_authorization`
2. Platform verifies, issues an SD-JWT+kb mandate that covers the full checkout *including* `merchant_authorization`
3. Business verifies the mandate at completion, expecting its own signature embedded in the signed-over content

Omitting step 1 makes step 3's verification fail by design.

Follows the established response-only-required pattern in `checkout.json` (`ucp`, `id`, `status`, `currency`, `totals`, `links` — all combine `ucp_request: "omit"` with inclusion in the parent `required` array).

## Category (Required)

## Checklist

Local verification:

- `ucp-schema lint source/` — 88 files passed
- `scripts/validate_examples.py --schema-base source/schemas/` — 253 passed, 0 failed (every existing AP2 spec example already includes `merchant_authorization`)
- `scripts/test_validate_examples.py` — 43 passed
- `ucp-schema resolve ap2_mandate.json --response --op read` confirms the resolved response schema has `required: ["merchant_authorization"]`
- `ucp-schema resolve ap2_mandate.json --request --op create` confirms request variants still strip `merchant_authorization` correctly via `ucp_request: "omit"`

## Breaking-change note

AP2-active business responses that previously validated without `merchant_authorization` will now fail schema validation. That is the intended behavior per spec. Downstream impact:

- The conformance suite (`ap2_test.py`) primarily exercises the request side, so it should be unaffected; any mock business responses elsewhere that omit `merchant_authorization` will need to add it.
- `python-sdk` regenerates Pydantic models from schemas, so the field becomes non-Optional on the next regen.


TateLyman:
Source-read note: this looks directionally right, but I think it may only close the `ap2: {}` case, not the missing-namespace case.

`ap2_with_merchant_authorization` now has `required: ["merchant_authorization"]`, so an AP2 object without the signature should fail. But in `dev.ucp.shopping.checkout`, the `ap2` property itself is still only declared under `properties` and is not listed in a parent `required` array. Under normal JSON Schema semantics, a checkout response with all base checkout required fields and no `ap2` object can still satisfy the extension `allOf` because absent optional properties are not validated.

That seems to leave the same AP2-active response gap described in the PR body: the docs say once `dev.ucp.shopping.ap2_mandate` is negotiated, the business MUST include `ap2.merchant_authorization` in all checkout responses, but a schema-driven mock/business could still omit the entire `ap2` namespace unless the resolved response schema also requires `ap2`.

Would it make sense to require `ap2` at the extension composition level for the response variant as well, while preserving the request-side `create/update: omit` behavior? If the custom `ucp-schema resolve --response --op read` already injects `required: ["ap2"]` from the extension negotiation context, then this is just worth capturing in the PR validation note; I could not see that guarantee from the source schema alone.

GarethCOliver (approved):

yanheChen (approved):

sakinaroufid:
Does this look good to merge? @GarethCOliver @yanheChen

kmcduffie:
@sakinaroufid could you sync this branch and we'd be happy to re-review.

sakinaroufid:
> @sakinaroufid could you sync this branch and we'd be happy to re-review.

@kmcduffie Synced!

damaz91:
@kmcduffie can you please take a look now that @sakinaroufid re-synced the PR? Thanks!

vishkaty:
The conflict here looks like a file move rather than anything in the change itself. #741 relocated `source/schemas/shopping/ap2_mandate.json` to `source/schemas/common/payment_ap2_mandate.json` on 2026-08-25, after this was approved.

The gap is still open on main at d3ccb55c: `$defs/ap2_with_merchant_authorization` carries no `required`, so a checkout response that omits `ap2.merchant_authorization` validates, while ap2-mandates.md:112 says the business MUST include it in all checkout responses.

I applied the one line at the new path locally to check it still behaves. With ucp-schema 1.4.1:

* response view: `required: ["merchant_authorization"]`
* complete request view: the property is absent and `required` is empty, because `merchant_authorization` is already `ucp_request: omit` there, so the request direction is unaffected
* `ucp-schema lint source/`: 124 files, all passed
* `scripts/validate_examples.py`: 343 passed, 0 failed

Happy to open a rebased branch you can pull from if that saves you time, or to leave it with you.

PR #481, opened 2026-05-26
feat(shopping): Add additional fields proposal (from Shopify)

SaifBenChahed:
## Summary

Adds an Additional Fields proposal using a single extension with optional config-based type narrowing.

Businesses sometimes need to collect checkout data that is specific to their workflow, such as PO numbers, gift options, delivery notes, requested dates, or buyer acknowledgements. Today there is no standard way for a business to describe those fields to a platform, for the platform to collect values, or for captured values to appear on orders.

This proposal standardizes that exchange:

- checkout `additional_fields` definitions and submitted values
- order read-only additional field summaries
- standard input types: `text`, `boolean`, `date`, and `choice`
- text validation hints, multiline text, date bounds, and choice options
- validation errors through existing `messages[]`

It also adds optional `config.supported_types` on the capability. Platforms and businesses may use `supported_types` to narrow the active type set; when both declare it, the active set is their intersection.

## Notes

The payload schema remains the full standard type superset. Config is treated as an optional negotiation constraint, not the primary extension model.


SaifBenChahed:
cc @igrigorik @richmolj

docs/specification/additional-fields.md
```diff
@@ -0,0 +1,472 @@
...
+## Type Support Config
+
+Platform and business profile declarations for
+`dev.ucp.shopping.additional_fields` MUST include `config.supported_types`.
+This keeps type support explicit as the extension evolves.
+
+{{ extension_schema_fields('additional_fields.json#/$defs/additional_fields_config', 'additional-fields') }}
+
+The active type set is the intersection of the platform and business
+declarations.
```
jamesandersen, line 135: What happens if the intersection is empty?   ... presumably these transactions will just terminate in `requires_escalation` but might be work noting here.

docs/specification/additional-fields.md
```diff
@@ -0,0 +1,472 @@
...
+{
+  "config": {
+    "supported_types": ["text", "boolean"]
+  }
+}
+```
+
+Businesses MUST only return fields whose `input.type` is in the active type
+set. If a required field cannot be represented using an active type, the
+business SHOULD use a supported fallback or require escalation.
```
jamesandersen: nit: maybe ref the `requires_escalation` enum [in the spec](https://github.com/Universal-Commerce-Protocol/ucp/blob/d984d4b2e0d7793b57866bee5db7c75b63e1148f/source/schemas/shopping/checkout.json#L69)

docs/specification/order-additional-fields.md
```diff
@@ -0,0 +1,148 @@
...
+      ]
+    }
+  }
+}
+```
+
+Businesses MAY advertise additional field support for checkout only, order only,
+or both. Platforms SHOULD check which resources are extended before expecting
+additional fields on order read responses.
+
```
jamesandersen, line 67: I'd suggest adding some normative language that the platform SHOULD render the inputs in the order the business sends them

source/schemas/shopping/additional_fields.json
```diff
@@ -0,0 +1,518 @@
...
+          "description": "Help text or instructions for the field.",
+          "ucp_request": "omit"
+        },
+        "input": {
+          "$ref": "#/$defs/additional_field_input",
+          "description": "Input control configuration.",
+          "ucp_request": "omit"
+        },
+        "value": {
+          "type": [
```
jamesandersen, line 302: Just a thought, for booleans, might consider making the base value accept a union, then use the same if/then pattern the PR already uses for input to pin boolean-typed fields to a real boolean:

  // additional_field.value base type:
  "value": { "type": ["string", "boolean", "null"], ... }

  // additional_field.allOf — add:
  {
    "if": {
      "properties": { "input": { "properties": { "type": { "const": "boolean" } }, "required": ["type"] } },
      "required": ["input"]
    },
    "then": { "properties": { "value": { "type": ["boolean", "null"] } } }
  }

  Request becomes { "key": "gift_wrap", "value": true }, cleared as { "key": "gift_wrap", "value": null }.


Otherwise I'd just add some language indicated the expected string formats for boolean values

jamesandersen (commented):
@SaifBenChahed thanks for submitting this!  I think this will unlock agentic commerce for a non-trivial cohort of merchants requiring some incidental extra inputs.

source/schemas/shopping/additional_fields.json
SaifBenChahed: Thanks, that makes sense. The conditional schema approach would work well for responses, since each field includes `input.type`.
For submissions, though, we only send `key` + `value`, without the field definition. The business still needs to resolve the key back to its field definition to validate type-specific rules, whether that rule is boolean format, choice membership, date bounds, or text constraints.
My slight preference is to keep `value` as `string | null` here for a uniform submission shape, so I added normative language that non-null boolean values MUST be the exact strings `"true"` or `"false"`.

jamesandersen: While there's a preference for machine resolvable schema, this seems reasonable; we'll see if there's any second opinion here ;-)

docs/specification/additional-fields.md
```diff
@@ -0,0 +1,478 @@
...
+
+```json
+{
+  "status": "incomplete",
+  "messages": [
+    {
+      "type": "error",
+      "severity": "recoverable",
+      "code": "additional_field_invalid_value",
+      "path": "$.additional_fields[?@.key=='po_number']",
```
SaifBenChahed: We could also have the path reference the additional field directly by its array position, such as `$.additional_fields[0]`, instead of by key.

jamesandersen (approved):

docs/specification/additional-fields.md
```diff
@@ -0,0 +1,478 @@
...
+[Order Additional Fields](order-additional-fields.md).
+
+## Schema
+
+When this capability is active, checkout is extended with an
+`additional_fields` array.
+
+### Checkout Extension
+
+{{ extension_fields('additional_fields', 'additional-fields') }}
```
damaz91, line 75: it should be {{ extension_fields('additional_fields', 'checkout') }}

docs/specification/additional-fields.md
```diff
@@ -0,0 +1,478 @@
...
+### Additional Field
+
+{{ extension_schema_fields('additional_fields.json#/$defs/additional_field', 'additional-fields') }}
+
+## Input Type Details
+
+Each field definition includes an `input` object with a required `type` string.
+Platforms SHOULD respect the declared type. Platforms that do not recognize a
+type SHOULD NOT render it as a different supported type, coerce it to text, or
+collect a value unless a negotiated extension defines that type.
```
jingyli, line 86: Trying to unpack this sentence here along with the `Standard Input Types` below and the schema design. Is it correct to interpret this as:
- UCP defines 4 well-known input types (essentially authored and endorsed by UCP governing body) = the standard input types
- Platforms & businesses can negotiate on other primitive supported input types (for example, "number" for numeric input) out-of-band. When the same input value is advertised and negotiated, then it can be leveraged as-is.

My confusion is around what we mean by "a negotiated extension defines that type"? Are we talking about the fact that we can have something like `com.example.number` as a type value? Can we bring some clarity in the sentence here by giving some examples? Also some naive questions on top:
- How are we thinking of mapping type value (i.e.`number`) to the actual input schema (i.e. `com.example.number`) for these custom extensions?
- How would we advertise these custom type extensions in the UCP profile?

Would we have:
```
"dev.ucp.shopping.additional_fields": [
        {
           ...
          "config": {
            "supported_types": ["text", "boolean", "com.example.number"]
          }
        }
      ],
"com.example.number": [
    {
       ....
       extends: "dev.ucp.shopping.additional_fields"
    }
]
```
If yes to the example above, then I believe this may be the first time where an extension is now extending another extension explicitly in UCP (not saying this is not okey, but just want to explicitly call it out this is the first time UCP is broadly advocating for this kind of pattern).

jingyli: Read through  `Extensibility` section below and it aligned with my understanding/assumption in the comment, though some additional clarity in language & examples would still be nice here 😃.

However, the 2 naive questions bullet still hold:1) Are we expecting the reverse-DNS to be part of `supported_types`? 2) This is the first instance of extension decorating an extension.

docs/specification/additional-fields.md
```diff
@@ -0,0 +1,478 @@
...
+All non-null values are submitted as strings regardless of input type. `null`
+represents no submitted value or a cleared value.
+
+### Additional Field Input
+
+{{ extension_schema_fields('additional_fields.json#/$defs/additional_field_input', 'additional-fields') }}
+
+### Standard Input Types
+
+This version defines the following standard types:
```
jingyli, line 97: Curious how did we land on the standard types here? Did we evaluate other primitives like numeric and/or list inputs to be a well-known value?

docs/specification/order-additional-fields.md
```diff
@@ -0,0 +1,150 @@
...
+```json
+{
+  "ucp": {
+    "version": "{{ ucp_version }}",
+    "capabilities": {
+      "dev.ucp.shopping.additional_fields": [
+        {
+          "version": "{{ ucp_version }}",
+          "extends": "dev.ucp.shopping.order",
```
jingyli, line 51: Trying to envision the scenario where `dev.ucp.shopping.additional_fields` is supported only in order but **not** checkout in UCP. In checkout, let's say platform opted to redirect/hand off user to business for collection of these values. If the extension is supported in order, does that mean platform would still get back those values through this structured `additional_fields` extension for read-only?

Or is there a strong dependency that values returned here **MUST** be collected via the extension when decorating UCP checkout?

jingyli (commented):
Thanks for the proposal! Some fundamental questions I had:

1. _[Discussed and aligned on most, would love to see if it would be good to consider adding a section in `overview.md` to clarify best practices on the use cases and scenarios further]_ Now in UCP, we will have multiple mechanisms to allow data to be collected on top of the main request & response wiring: 1) embedded transport via `continue_url`, 2) escalation redirect handoff via `continue_url`, 3) new actions construct being introduced in #582. How should implementators think about when to use what - let's take an example of needing to collect a user's age for verification, I can naively see it fall into all 4 buckets - what would be best practices and guidelines we set around them? Or is this something we don't want to get into at all and leave it to the implementors to define?
2. With this extension's vast flexibility to support various use cases, what are the principles we will now hold to decide what use cases to promote into UCP as first-class citizens vs. just re-leveraging this extension? An example can be scheduled delivery: this extension will be able to support the collection of its values, but I do see a path to also model this directly into the fulfillment extension.
3. I see the scope of this extension, so far, is only decorating `checkout` & `order` (which is 100% fine to start with a narrow scope). But I'm curious what are our thoughts on this extension decorating other capabilities like `cart` or even going beyond shopping service?
4. Are there other data (on top of the supported type) we want to negotiate up front between platforms & businesses when using this extension? An example: a platform may have UI wiring for `type = choice`, but it can only display up to 5 options. Then should that restriction be negotiated up-front?
5. Any guidelines we should set for businesses & platforms on rendering contracts? What about privacy & security considerations since now platforms will need to dynamically understand/interpret data needs from businesses and collect/disclose them from the user?

wsbrunson:
Another consequence of allowing the Business to provide additional checkout requirements outside of an extension is that the Platform would now have to do a three-step Create -> Update -> Complete checkout any time `additional_fields` were involved, instead of the ideal 2 step (Create -> Complete).

At least in this current proposal, it doesn't seem like the Business can advertise to the Platform what fields will need to be collected before a checkout_session is created. That would have an impact on how Platforms render UI and guide a buyer to checkout.

I represent PayPal and we implemented a similar feature in our first version of agentic checkout. We had a partner with thousands of merchants that all had various levels of product customization. We ended up going with something like `additional_fields` and let the merchants tell us what they needed to complete a checkout. Examples were age verification, marketing consents, and item customizations. We quickly ran into problems with Platforms trying integrate with this feature though. The number 1 piece of feedback we heard is that they want to know what fields were going to be required up-front.

UCP already has a capability for consents and seems like they will soon have something for age verification (Actions). From your list of examples, I feel like Gift Options would also be well-served as a capability because both sides would need to opt-in to use it.

If we scoped this to just item customizations, it would help answer @jingyli's second question:

> With this extension's vast flexibility to support various use cases, what are the principles we will now hold to decide what use cases to promote into UCP as first-class citizens vs. just re-leveraging this extension? An example can be scheduled delivery: this extension will be able to support the collection of its values, but I do see a path to also model this directly into the fulfillment extension.

I think item customizations would be hard to implement without an approach like `additional_fields`. If we tried to tackle it per customization, we could end up with hundreds of extensions, one for each product type.

I think there would also be synergies with the catalog and order capabilities. Items could advertise their customizations in the catalog so Platforms could see what fields would be required to checkout an item. Orders could be rejected for improper customizations that aren't caught until a human actually sees it (Inappropriate language when customizing a t-shirt for example).

ihoosain (approved):
LGTM, i think this solves for additional fields at the checkout level very well. Maybe one enhacement for later is it seems the additional fields is at the checkout level but could this be more powerful at the cart/line item level, in addition to the entire checkout?

docs/specification/order-additional-fields.md
SaifBenChahed: > is there a strong dependency that values returned here MUST be collected via the extension when decorating UCP checkout

Not necessarily. They would still receive the order’s additional fields, even if the business collected them separately.

docs/specification/additional-fields.md
```diff
@@ -0,0 +1,478 @@
...
+    `input.type` value using `allOf`, following the normal UCP schema
+    composition pattern.
+
+For vendor-defined types, use the vendor namespace in the type identifier, for
+example `com.example.file_upload`.
+
+## Operations
+
+Additional field values are submitted via standard checkout create and update
+operations.
```
jamesandersen, line 171: Aligning value submission with the buyer-consent precedent (#451) so the common case is two steps (create → complete), with an update round-trip only when a value needs correction:

```suggestion
Additional field values are submitted via standard checkout create, update, and
complete operations. Platforms MAY submit `additional_fields` values on
`complete_checkout` so the common case can
complete in two steps (create → complete) with an update round-trip only when a
value needs correction. When a required field is unmet or invalid, the business
MUST NOT transition the checkout to `completed` and MUST return the error via
`messages[]` (see [Invalid Values](#invalid-values)).
```

docs/specification/additional-fields.md
```diff
@@ -0,0 +1,478 @@
...
+   limitations under the License.
+-->
+
+# Additional Fields Extension
+
+## Overview
+
+Additional Fields extension allows businesses to indicate that they need extra
+checkout data collection fields, and specifies how field definitions and
+submitted values are shared between the platform and the business.
```
jamesandersen, line 23: Making the scope boundary explicit — this keeps late submission on `complete` safe and gives a clear promotion criterion for first-class capabilities (addresses @jingyli's Q2):

```suggestion
submitted values are shared between the platform and the business.

**Scope:** Additional fields are incidental data capture. They MUST NOT
materially affect payment processing or checkout completion mechanics — they
cannot change available payment handlers, totals, or fulfillment options, and
MUST NOT gate completion beyond a recoverable request to provide or correct a
value. Data needs that move those levers (for example consents or age
verification) belong in a purpose-built capability, not this extension.
```

jamesandersen (commented):
@wsbrunson the "forced 3-step" point is a good catch and worth pinning down precisely — thanks for bringing the PayPal experience to it.

A couple thoughts here:

**The extra round-trip is fixable and shouldn't be inherent.** The field *definitions* already come back in the create-checkout **response** (`additional_fields[]` with `value: null`), so the business does advertise what it needs — just at create-time. What forces Create → Update → Complete today is that the draft routes value submission through create/update only ("submitted via standard checkout create and update operations") and never mentions `complete`.

We already solved this exact shape for consent in #451: platforms **MAY submit captured consent on `complete_checkout`**, with asymmetric validation — `warning` on create/update, `error` at `complete` (the business MUST NOT transition to `completed` while dependencies are unmet). Applying the same pattern here collapses the happy path back to **Create → Complete**, with the Update loop only when a value fails validation. I've left a suggested change below to align `additional_fields` with that precedent explicitly.

**Pre-`create` visibility is a trickier problem** — (though not necessarily blocking IMO).  I think a separate discussion on whether some field hinting belongs at discovery/catalog would be interesting (which also connects to your catalog/line-item synergy point).  This could be taken up by the catalog domain working group we're planning to spin up (e.g. related to RFC #550 )

Both of these lean on a principle I'd like us to make explicit — and it also partially answers @jingyli's governance question (Q2) about what belongs in this extension vs. a first-class capability:

> **Additional fields are incidental data capture and MUST NOT materially affect payment processing or checkout completion mechanics** — they cannot change available payment handlers, totals, or fulfillment options, and cannot gate completion beyond a recoverable request to provide or correct a value.

That boundary is what makes late submission on `complete` safe (a missing PO number is recoverable; it can't reshape the payment step), and it's the line for promotion to a first-class capability: anything that *does* move those levers (consents, age verification, gift options that change price/fulfillment) belongs in a purpose-built capability both sides opt into, not here. I've left a second suggested change to state this in the spec.

wsbrunson:
> Additional fields are incidental data capture and MUST NOT materially affect payment processing or checkout completion mechanics — they cannot change available payment handlers, totals, or fulfillment options, and cannot gate completion beyond a recoverable request to provide or correct a value.

Okay yeah I think that would solve my major hesitation. Would the guidance then be that any fields that should be able to affect payment processing or checkout completion mechanics should be codified as a capability/extension?

Agreed that pre-create visibility is tricky but not a blocking issue.

raginpirate (changes requested):
Thanks for the proposal and the thoughtful review so far. I agree there is a real need for incidental, Business-specific data. However, the current shape recreates part of JSON Schema through text, boolean, date, choice, validation hints, stringified values, and supported_types negotiation. That leads directly to the questions raised here about native booleans, numeric/list support, custom type extensions, and renderer capability.

I’ve sketched an imperfect but much smaller alternative in this [commit: Declared Fields](https://github.com/Universal-Commerce-Protocol/ucp/commit/90d9198ca2dbce2b62639ba0746a64ca5709df82).

A Business advertises standard JSON Schemas in the active extension’s response config, keyed by reverse-DNS family (I figured this would be helpful to let external partners build segmented additional fields that N businesses can adopt 👯):

 ```json
   {
     "config": {
       "com.example.gift_message": {
         "type": "object",
         "properties": {
           "include_gift_message": {
             "type": "string",
             "title": "Write a gift message",
             "ucp_request": {
               "create": "optional",
               "update": "optional"
             }
           }
         }
       }
     }
   }
 ```

 The same family key carries values on Cart, Checkout, or Order:

 ```json
   {
     "declared_fields": {
       "com.example.gift_message": {
         "include_gift_message": "Happy Birthday!"
       }
     }
   }
 ```

Mechanically, the Platform selects config["com.example.gift_message"], resolves its UCP annotations for the operation, and validates declared_fields["com.example.gift_message"] against it. The Business repeats that validation and reports failures through existing messages[].

This gives us native JSON values and standard JSON Schema features without another UCP-specific type system:

 - text → type: "string"
 - boolean → type: "boolean"
 - date → type: "string", format: "date"
 - choice → enum or labeled oneOf
 - structured values → existing object/array schemas

It also lets us reuse pure core UCP value types through $ref. A Platform supporting Declared Fields should understand an explicit allowlist of those types and their local normative contracts—for example, referencing the shared Description type also carries its rich-content sanitization requirements. This must not become a way to bypass negotiation for effectful types such as payment credentials, Actions, consent, fulfillment, or mandates.

**_One intentional difference from the late-submission suggestion_**: Checkout marks declared_fields as omitted on Complete. Complete is a narrow payment/finalization slice, not a final generic Checkout update—just as it omits line items, context, fulfillment, and most Buyer state. Required declared values must be accepted through Create or Update before the Business returns ready_for_complete. I recognize this retains an extra Update call in some cases, but that has been a long-standing and accepted tradeoff to create a well-defined boundary between complete and update.

**_My proposed map shape also deliberately drops the current array’s ordering semantics_**. JSON object and JSON Schema properties order are not portable. #525 is the open TC proposal for ordering reverse-DNS maps; it could restore family ordering, but must also cover properties within a family before it fully replaces the current flat field order. Until then, I believe we should avoid trying to solve this ordering concern.

Would love your thoughts on this alternative, and if its simplicity and tradeoffs feel UCP-native 😄 I put this as request changes to ensure we have a solid conversation on this direction in our deep dive before shipping 🙇

vixdug:
Thanks all for the review here — and to @SaifBenChahed for driving the proposal to this point.
  
 A small group huddled today to work through the remaining open threads. We landed on closing this PR and coming back with a V2. The underlying need is real and worth solving - this isn't a rejection of the problem. We think there's a better shape available and we want to do more solutioning before bringing it back.

PR #483, opened 2026-05-27
feat: Web Bot Auth (WBA) interop 

igrigorik:
UCP specifies RFC 9421 message signatures, but had incompatible restrictions and missing guidance for interoperating with Web Bot Auth (WBA):

- **Algorithm (hard block)**: the JWK schema accepted only EC keys; Ed25519 was not a valid UCP key type, so a UCP signer literally could not produce a WBA-conformant signature.
- **Key publication (undefined role)**: UCP's canonical field is `signing_keys[]`; WBA verifiers read an RFC 7517 `keys[]` JWK Set. UCP neither forbade `keys[]` nor defined its role.
- **Header semantics (undefined)**: `UCP-Agent` points at the UCP profile; `Signature-Agent` advertises WBA keys. UCP didn't define how a verifier picks between them.

This PR resolves all three with a **single-signature, dual-audience** shape: one key, one signing operation, one signature on the wire that both a UCP verifier (via `UCP-Agent`) and a WBA verifier (via `Signature-Agent`) accept. WBA's protocol draft §4.2.4 permits additional covered components, which makes this valid under both regimes once the algorithm is supported and header roles are defined.

## What this PR does

1. **Open algorithm vocabulary.** The JWK schema validates EC and OKP via `if/then` (open `kty`/`crv`/`alg`; unknown keys are tolerated, never whole-profile-rejected). `ES256` stays the universal MUST baseline; `EdDSA` (Ed25519) is OPTIONAL for general verifiers and MUST for verifiers that support WBA.

2. **Profile as a JWK Set.** The profile gains an optional top-level `keys[]` (RFC 7517) mirroring canonical `signing_keys[]`; when both are present they MUST list the same `kid`s with semantically equivalent JWKs (identical RFC 7638 thumbprints). RFC 7517 §5 permits the extra members, so the document is simultaneously a UCP profile and a valid JWK Set — a signer MAY reuse it as its WBA key source. (`keys[]` is published as an interim step toward promoting it to canonical and deprecating `signing_keys[]` in a future UCP version — separate PR.)

3. **"Dual-audience" defined, with scoped requirements.** A dual-audience signature carries `tag="web-bot-auth"`, signs the `signature-agent` component, sets `keyid` to the key's RFC 7638 thumbprint, and includes `created`/`expires` (nonce SHOULD). For such keys the published `kid` **MUST** equal the thumbprint, so the WBA lookup (by thumbprint) and the UCP lookup (`keyid`→`kid`) resolve the same key. WBA interop is request-scoped; responses use standard UCP signatures.

4. **Discovery: three equal `Signature-Agent` variants.** `type=jwks_uri` (a JWK Set URL — point it at the profile), `type=cimd` (a Client ID Metadata Document whose `jwks_uri` may point back at the profile), `type=directory` (an origin hosting a well-known directory). UCP does **not** rank them — the right choice depends on the counterparty's verifier. The `type` parameter and its values are defined in [`draft-meunier-webbotauth-httpsig-directory-00` §4.1](https://datatracker.ietf.org/doc/draft-meunier-webbotauth-httpsig-directory/00/) (the renamed/re-adopted directory draft, formerly `draft-meunier-http-message-signatures-directory`). The directory's format, per-key self-signatures, and media type are **deferred to the same draft §5.2** — UCP does not restate them.

5. **Verifier (Identity Resolution Algorithm): capability-based, not tag-gated.** The `tag` is a hint, not a gate. A UCP verifier always resolves via `UCP-Agent` (so a UCP-only verifier never implements WBA discovery — it only needs Ed25519 + RFC 9421 §2.1.2 to verify the bytes); a WBA-aware verifier MAY additionally resolve via `Signature-Agent`. A **closed, explicit covered-component gate** is enforced in every regime, so a WBA-minimal signature can never authenticate a UCP request whose body/method/path/headers are unbound. `keyid`==thumbprint is MUST for WBA-shape signatures. Identity is the URL that supplied the verifying key; a `Signature-Agent` key is treated as the same identity as `UCP-Agent` only when the URLs match.

6. **Fetch safety.** UCP specifies it for every identity-resolution dereference (profile, `jwks_uri`, CIMD), aligned with the WBA protocol draft's SSRF guidance: reject special-use IPs (RFC 6890) with a loopback carve-out for local dev, resist DNS rebinding, don't follow in-document URLs to internal addresses, and bound response size (a deployment guard — the schema sets no size limit).

### Default UCP signature (unchanged)

```
POST /checkout-sessions HTTP/1.1
Host: merchant.example.com
UCP-Agent: profile="https://platform.example/.well-known/ucp"
Idempotency-Key: 550e8400-e29b-41d4-a716-446655440000
Content-Digest: sha-256=:X48E9q...:
Signature-Input: sig1=("@method" "@authority" "@path" "ucp-agent" "idempotency-key" "content-digest" "content-type");keyid="merchant-2026"
Signature: sig1=:base64_ecdsa_signature_value:
```

### Dual-audience opt-in signature (new)

```
POST /checkout-sessions HTTP/1.1
Host: merchant.example.com
UCP-Agent: profile="https://platform.example/.well-known/ucp"
Signature-Agent: sig1="https://platform.example/.well-known/ucp";type=jwks_uri
Idempotency-Key: 550e8400-e29b-41d4-a716-446655440000
Content-Digest: sha-256=:X48E9q...:
Signature-Input: sig1=("@method" "@authority" "@path" "signature-agent";key="sig1" "ucp-agent" "idempotency-key" "content-digest" "content-type");keyid="poqkLGiymh_W0uP6PZFw-dvez3QJT5SolqXBCW38r0U";created=1738617600;expires=1738621200;tag="web-bot-auth"
Signature: sig1=:base64_ed25519_signature_value:
```

One signature on the wire. UCP verifiers route via `UCP-Agent`; WBA verifiers route via `Signature-Agent`. Both verify the same bytes against the same key, and both headers can point at the same `/.well-known/ucp` URL (or `Signature-Agent` can point elsewhere).

## Backwards compatibility

- Default UCP signers/verifiers are unchanged; ES256 stays the baseline.
- Profiles publishing only `signing_keys[]` remain valid; `keys[]` is OPTIONAL.
- WBA opt-in is per-signature; non-opting sites emit only `UCP-Agent`.

## Future direction

A UCP+AP2+WBA merchant operates **two keys** today (Ed25519 for HTTP/WBA, ECDSA P-256 for AP2 `merchant_authorization`, per AP2 v0.2). Two upstream moves collapse this to one: (1) [AP2 #268](https://github.com/google-agentic-commerce/AP2/issues/268) relaxing the non-deterministic-signature requirement, and (2) UCP promoting `keys[]` to canonical. The wire format and verifier algorithm here already point at that future.

Routed to a follow-up tracking issue (not expanded here): promoting `keys[]`/`jwks_uri`/CIMD convergence + migration, and `tag` namespace reservation.

## Checklist
- [x] Core Protocol
- [x] Contributing Guide followed
- [x] Documentation updated
- [x] Linting/formatting pass


docs/specification/signatures.md
```diff
@@ -131,49 +179,193 @@ defined in [RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517).
...
+<!-- ucp:example schema=shopping/checkout op=create direction=request -->
+```json
+POST /checkout-sessions HTTP/1.1
+Host: merchant.example.com
+Content-Type: application/json
+UCP-Agent: profile="https://platform.example/.well-known/ucp"
+Signature-Agent: sig1="https://platform.example/.well-known/ucp"
+Idempotency-Key: 550e8400-e29b-41d4-a716-446655440000
+Content-Digest: sha-256=:X48E9qOokqqrvdts8nOJRJN3OWDUoyWxBf7kbu9DBPE=:
+Signature-Input: sig1=("@method" "@authority" "@path" "signature-agent";key="sig1" "ucp-agent" "idempotency-key" "content-digest" "content-type");keyid="NzbLsXh8uDCcd-6MNwXF4W_7noWXFZAfHkxZsRGC9Xs";created=1738617600;expires=1738621200;tag="web-bot-auth"
```
amithanda: The `kid` on the Ed25519 example is a real RFC 7638 thumbprint, but of RFC 7638's RSA example key, not of the Ed25519 key it's attached to. Computed from the published key:

```
{"crv":"Ed25519","kty":"OKP","x":"11qYAYKxCrfVS_7TyWQHOg7hcvPapiMlrwIaaPcHURo"}
-> SHA-256 thumbprint = kPrK_qmxVWaYVA9wwBF6Iuo3vVzz7TxHCTwXBygrS4k
```

(the worked example in RFC 8037 §A.3 for this exact key). The value currently used, `NzbLsXh8uDCcd-6MNwXF4W_7noWXFZAfHkxZsRGC9Xs`, is the thumbprint of the RSA key in RFC 7638 §3.1.

I know example crypto here may be illustrative, the `Signature` is a placeholder and the `Content-Digest` is RFC 9530's `{"hello": "world"}` value rather than a digest of the body below (which I would recommend correcting, if possible). The reason I'd still flag this one specifically: unlike those obviously-fake placeholders, this value reads as a correct thumbprint, and it's the field the WBA opt-in makes a MUST ("MUST set `keyid` to the JWK SHA-256 Thumbprint") and that the verifier algorithm checks. So the `tag="web-bot-auth"` request example at signatures.md currently fails its own document's MUST, and an implementer is likely to lift it as a test vector and then doubt their own thumbprint code.

Could we either swap in the real thumbprint (`kPrK_...`, which is also a recognizable published RFC value so we keep that benefit). The same value also appears in the PR description's wire and profile examples, worth updating there too.

source/schemas/profile.json
```diff
@@ -8,37 +8,91 @@
   "$defs": {
     "jwk_public_key": {
       "type": "object",
-      "description": "EC public JSON Web Key used for HTTP Message Signatures and signed webhook verification. UCP profiles publish public keys only; private key material MUST NOT appear in a profile. Additional public JWK members are permitted; consumers ignore unknown members.",
-      "required": ["kid", "kty", "crv", "x", "y"],
+      "description": "Public JSON Web Key used for HTTP Message Signatures and signed webhook verification. UCP profiles publish public keys only; private key material MUST NOT appear in a profile. UCP supports two key types: EC (ECDSA P-256, P-384) and OKP (Ed25519); OKP keys are RECOMMENDED for HTTP transport identity when Web Bot Auth (WBA) interop is desired. A single profile MAY publish keys of either or both types; consumers select keys by kid. Additional public JWK members are permitted; consumers ignore unknown members.",
+      "required": ["kid", "kty"],
+      "oneOf": [
```
amithanda: This is the one we should discuss a bit before merging. The new `oneOf` over closed `kty`/`crv`/`alg` enums makes the schema reject any key type or curve it doesn't already enumerate, and because `items` applies per element, a single unrecognized key invalidates the entire `signing_keys[]`/`keys[]` array (and so the whole profile):

- future `kty` (e.g. a post-quantum `AKP`): FAILS validation
- future EC `crv` (e.g. `P-521`): FAILS validation
- existing EC P-256 + OKP Ed25519: pass

This is notable because the pre-PR schema had `crv` as an open string, so we'd be narrowing an existing open field into a closed enum, and re-creating the exact "couldn't add Ed25519 without a schema change" problem this PR exists to fix (next algorithm = another breaking schema bump + old validators reject new profiles). A verifier that doesn't support an algorithm should skip that key during key lookup, not reject the whole document at schema-validation time.

Could we keep the recognized values as `examples` (or per-branch constraints) but let unknown key types pass validation? One forward-compatible shape that preserves the EC/OKP constraints without the closed `oneOf`:

```json
"required": ["kid", "kty"],
"properties": {
  "kty": { "type": "string", "description": "JWK key type. Well-known: EC (ECDSA P-256/P-384), OKP (Ed25519). Verifiers MUST tolerate unknown key types and skip keys they cannot use.", "examples": ["EC", "OKP"] },
  "crv": { "type": "string", "description": "Curve name. Well-known: P-256, P-384, Ed25519.", "examples": ["P-256", "P-384", "Ed25519"] },
  "alg": { "type": "string", "description": "JWA algorithm. Well-known: ES256, ES384, EdDSA.", "examples": ["ES256", "ES384", "EdDSA"] }
},
"allOf": [
  { "if": { "properties": { "kty": { "const": "EC" } }, "required": ["kty"] },
    "then": { "required": ["crv", "x", "y"] } },
  { "if": { "properties": { "kty": { "const": "OKP" } }, "required": ["kty"] },
    "then": { "required": ["crv", "x"] } },
  { "if": { "properties": { "crv": { "const": "P-256" } }, "required": ["crv"] },
    "then": { "properties": { "alg": { "const": "ES256" } } } },
  { "if": { "properties": { "crv": { "const": "P-384" } }, "required": ["crv"] },
    "then": { "properties": { "alg": { "const": "ES384" } } } }
]
```

This keeps the per-curve `alg` coupling and the EC-needs-`x,y`/OKP-needs-`x` requirements, but an unknown `kty` simply carries no extra constraints rather than failing. Would that work for what we need here?

docs/specification/overview.md
```diff
@@ -363,10 +363,29 @@ registries **MUST** be present even when empty. `ucp.capabilities` is optional
...
+field that every UCP verifier reads. This preserves backward
+compatibility with every existing UCP verifier.
+
+Profiles **MAY** additionally publish a top-level `keys[]` array per
+[RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517) (JWK Set
+Format) that **mirrors** `signing_keys[]`. RFC 7517 §5 permits
+additional members alongside `keys`, so the same document serves as
+both a UCP profile and a Web Bot Auth-compatible key directory. See
+[Deployment Patterns for WBA Interop](#deployment-patterns-for-wba-interop)
+below. When both arrays are present, they **MUST** list the same set
```
amithanda, line 469: The cross-array equivalence MUST is good, and for an added or generally out-of-sync key the failure is safe: a verifier reads only its own array, misses the key, and returns key_not_found (fail-closed). The case I'd want to nail down is removal, specifically compromise revocation. The atomicity guarantee is currently phrased around rotation ("Rotation updates apply to both arrays atomically"), but the removal steps say "Remove the old key from the profile" (rotation step 4) and "Immediately remove compromised key from profile" (compromise step 1) without naming both arrays. If an operator removes a compromised key from signing_keys[] but the keys[] update lags or is missed, WBA-shape verifiers (which read keys[]) keep accepting signatures from that key, a fail-open outcome. Could we make the atomicity rule explicitly cover removal/revocation across both arrays, e.g. "When a key is removed (rotation or compromise), it MUST be removed from both arrays atomically; until it is absent from both, the regime still reading it will continue to accept it"? Same single-source-of-truth intent, aimed at the case that actually fails open rather than at verifier-side divergence detection.

amithanda:
Hi @igrigorik, Overall this is a strong, well-scoped PR that solves a real problem: a UCP signer couldn't produce a WBA-conformant signature at all before this, and the PR closes that gap without disturbing existing integrators. A few things I think are genuinely well done: the single-signature dual-audience shape (one key, one signing operation, verified by both regimes off the same bytes) is elegant; backward compatibility holds (ES256 stays the universal baseline, keys[] and the WBA opt-in are both optional).

I left three inline comments, only one of which I'd treat as a pre-merge discussion:

1. profile.json (the oneOf over closed kty/crv/alg enums): this is the one worth resolving before merge. It narrows previously-open fields into closed enums and makes any future key type or curve fail whole-profile validation, which makes it quite rigid.

2. signatures.md (the Ed25519 example kid): a should-fix correctness nit, the kid is a real RFC 7638 thumbprint but of the wrong (RSA) key, so the example contradicts the MUST the PR introduces.

3. overview.md (revocation atomicity): a small but security-relevant clarification, the atomicity guarantee should explicitly cover key removal/revocation across both arrays, since that's the one divergence that fails open. PTAL.

None of these are objections to the direction, which is right.

source/schemas/profile.json
igrigorik: Good catch, fixing.

igrigorik:
@amithanda ty sir, great catch and feedback on the oneOf. Updated, PTAL.

docs/specification/signatures.md
```diff
@@ -368,6 +629,17 @@ UCP-Agent: profile="https://platform.example/.well-known/ucp"
 profile_url = "https://platform.example/.well-known/ucp"
 ```
 
+**Example (Signature-Agent, WBA-shape):**
+
+```text
+// Headers
+Signature-Agent: sig1="https://platform.example/.well-known/ucp"
```
dpeacock: I think from the [IETF Spec](https://datatracker.ietf.org/doc/draft-meunier-http-message-signatures-directory/) this is valid.  But it looks like there may be a gap with [Cloudflare's expectations](https://developers.cloudflare.com/bots/reference/bot-verification/web-bot-auth/#2-host-a-key-directory)   today:

> Host a key directory at /.well-known/http-message-signatures-directory (note that this is a requirement)

igrigorik:
Rereading the Web Bot Auth Architecture [1] and HTTP Message Signatures Directory [2] drafts, directory discovery is ambiguous. The two specs describe it two different ways with no stated precedence, and it affects whether a profile can double as a key directory.

- **Directory §4** [2]: `Signature-Agent` "contains a URI allowing retrieval of [the] directory" — dereference the URI as-is. **Architecture §4.4** [1] agrees ("fetch … as indicated by `Signature-Agent`").
- **Architecture §4.5** [1]: "The reference for discovery is a FQDN. It SHOULD provide a directory hosted on the well known" — take the host, use `/.well-known/http-message-signatures-directory`.

Concretely, for `Signature-Agent: sig1="https://example.com/.well-known/ucp"`:

- a *dereference-verbatim* verifier fetches `…/.well-known/ucp` ✅ 
- a *FQDN + well-known* verifier fetches `…/.well-known/http-message-signatures-directory` ❌ 

### Desired state 
One `/.well-known/ucp` document, carrying one JWK Set, that's simultaneously a UCP profile and a Web Bot Auth directory — discoverable via `Signature-Agent` — instead of maintaining a separate `/.well-known/http-message-signatures-directory`. The verbatim-deref reading makes that clean; the FQDN reading forces a second endpoint. What we need to clarify...

1. Can `Signature-Agent` point at an arbitrary URL?
2. If yes: §3 [2] requires `application/http-message-signatures-directory+json`. Can this be relaxed to ease implementation and allow fused directory / profile we're after.

---

[1] https://datatracker.ietf.org/doc/html/draft-meunier-web-bot-auth-architecture-05
[2] https://datatracker.ietf.org/doc/html/draft-meunier-http-message-signatures-directory-05

docs/specification/signatures.md
igrigorik: Yeah, there is confusion / divergence across specs on this. Let's pull this out of a comment thread: https://github.com/Universal-Commerce-Protocol/ucp/pull/483#issuecomment-4713565398

docs/specification/overview.md
```diff
@@ -363,10 +363,35 @@ registries **MUST** be present even when empty. `ucp.capabilities` is optional
...
-used for HTTP Message Signatures and signed webhooks. See
-[Message Signatures](signatures.md) for key format, algorithms, lookup, and
-rotation.
+Profiles **MAY** include public JSON Web Keys used for HTTP Message
+Signatures and signed webhooks. When a profile publishes signing keys,
+they **MUST** appear in `signing_keys[]` — the canonical UCP profile
+field that every UCP verifier reads. This preserves backward
+compatibility with every existing UCP verifier.
+
+Profiles **MAY** additionally publish a top-level `keys[]` array per
```
thibmeu, line 462: with WBA, we'd like to move away from having signing keys directly within a profile in favor of having a JWKS URI. The advantages are multiple:
1. allow rotation of the keys without modifying the profile
2. eases offloading of key management
3. aligns with CIMD that's used for MCP for instance

For UCP, this would be compatible with https://github.com/thibmeu/http-message-signatures-directory/pull/98.

Two changes though:
1. there would need to be a client_id that's the URL at which the profile is published
2. JWKS SHOULD present a signature over `@authority` (limit copy and impersonation of the key set)

docs/specification/overview.md
```diff
@@ -363,10 +363,35 @@ registries **MUST** be present even when empty. `ucp.capabilities` is optional
...
+[RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517) (JWK Set
+Format) that **mirrors** `signing_keys[]`. RFC 7517 §5 permits
+additional members alongside `keys`, so the same document serves as
+both a UCP profile and a Web Bot Auth-compatible key directory. See
+[Deployment Patterns for WBA Interop](#deployment-patterns-for-wba-interop)
+below. When both arrays are present, they **MUST** list the same set
+of `kid` values, and entries sharing a `kid` **MUST** be semantically
+equivalent — i.e., have identical RFC 7638 SHA-256 thumbprints
```
thibmeu:
```suggestion
equivalent — i.e., have identical base64url JWK SHA-256 Thumbprint as defined in [Section 3.2 of RFC 7638](https://rfc-editor.org/rfc/rfc7638#section-3.2) for RSA and EC, and in [Appendix A.3 of RFC 8037](https://rfc-editor.org/rfc/rfc8037#appendix-A.3) for ed25519.
```

adding base64url mention. pointing to RFC8037 for ed25519, even though that probably could be ignored

docs/specification/overview.md
```diff
@@ -479,6 +504,33 @@ Businesses publish their profile at `/.well-known/ucp`. An example:
...
   },
   "signing_keys": [
+    {
+      "kid": "kPrK_qmxVWaYVA9wwBF6Iuo3vVzz7TxHCTwXBygrS4k",
+      "kty": "OKP",
+      "crv": "Ed25519",
+      "x": "11qYAYKxCrfVS_7TyWQHOg7hcvPapiMlrwIaaPcHURo",
+      "use": "sig",
+      "alg": "EdDSA"
+    },
```
thibmeu, line 615: nits: it'd be good to reuse test keys from [RFC9421](https://datatracker.ietf.org/doc/html/rfc9421#name-example-ed25519-test-key) to ease detection of key material that's meant for test and not for prod

docs/specification/overview.md
```diff
@@ -493,9 +545,29 @@ Businesses publish their profile at `/.well-known/ucp`. An example:
...
+both UCP HTTP transport (with WBA interop) and AP2 mandate signing:
+
+- An **Ed25519** key (OKP) for HTTP transport identity, WBA-compatible.
+  The `kid` is the JWK SHA-256 Thumbprint per RFC 7638.
+- An **ECDSA P-256** key (EC) for AP2 mandate signing
+  (`ap2.merchant_authorization`), required to be non-deterministic
+  per AP2 v0.2 §Payment Mandate.
+
+A business that does not interact with AP2 or WBA may publish a single
+ES256 key in `signing_keys[]` only (the universal baseline). See
```
thibmeu, line 676: `ecdsa-p256-sha256` is allowed in Web Bot Auth (see [spec](https://datatracker.ietf.org/doc/html/draft-meunier-web-bot-auth-architecture-05#section-4.2-6)). it's not been deployed to minimise the required supported algorithm. That's something we could add as a test vector and on cloudflare implementation though

docs/specification/overview.md
```diff
@@ -1154,6 +1240,144 @@ validation (invalid schema, signing keys, signature mismatch),
...
+
+The body bytes are identical between the two responses; only the
+`Content-Type` header and the presence of `Signature-Input` /
+`Signature` response headers differ.
+
+Best fit for: dynamic hosting infrastructure (edge runtimes, API
+gateways, application servers) where content negotiation is
+straightforward.
+
+#### Pattern 2: Two URLs (isolated signing complexity)
```
thibmeu: I agree we should align to have only one URL as an entrypoint. I think https://github.com/thibmeu/http-message-signatures-directory/pull/98 should help address that. ucp would become a CIMD document, and the keys would be in jwks_uri

`Signature-Agent` would have one extra parameter `type` that allows for `cimd`.

Eyes welcome

docs/specification/overview.md
```diff
@@ -1154,6 +1240,144 @@ validation (invalid schema, signing keys, signature mismatch),
...
+A request MAY carry multiple signatures per
+[RFC 9421 §4.3](https://www.rfc-editor.org/rfc/rfc9421#section-4.3).
+Verifiers attempt each signature independently; the request is
+authenticated when at least one signature verifies. The algorithm
+below processes a single signature.
+
+1. **Determine the verification regime from the signature's `tag`.**
+    - `tag="web-bot-auth"` — WBA-shape. The signature **MUST** satisfy
+      the agent-signature requirements in
+      [draft-meunier-web-bot-auth-architecture](https://datatracker.ietf.org/doc/draft-meunier-web-bot-auth-architecture/)
```
thibmeu:
```suggestion
      [draft-meunier-web-bot-auth-architecture-05](https://datatracker.ietf.org/doc/draft-meunier-web-bot-auth-architecture/05/)
```

pinning the version, as that document may evolve

thibmeu (commented):
The change looks good overall. I think there is opportunity to align the two specifications

1. Add examples for `ecdsa-p256-sha256` in web bot auth specification, with an associated implementation
2. Have a CIMD shape, specifically with jwks_uri. This would allow both WBA and UCP to keep their own namespace ("ucp" and "web-bot-auth"), and have key management to be decorrelated. For WBA, the changes are being proposed https://github.com/thibmeu/http-message-signatures-directory/pull/98

docs/specification/overview.md
```diff
@@ -1154,6 +1240,144 @@ validation (invalid schema, signing keys, signature mismatch),
...
+   §5.2. Discard keys without a valid self-signature.
+5. **Match the signature's `keyid`** to a `kid` in the resolved key
+   list. **Skip** this signature if no match. For WBA-shape
+   signatures, the verifier **SHOULD** also confirm that `keyid`
+   equals the RFC 7638 SHA-256 thumbprint of the matched JWK. The
+   WBA architecture draft §4.2 requires `keyid` to be the thumbprint;
+   verifying the equality detects directory tampering and ensures
+   WBA-interop behavior. If the thumbprint check fails, skip this
+   signature.
+6. **Verify the signature** using the matched key. The signing
```
shopdave, line 1495: Insert a regime-independent coverage gate before the crypto-verify step.

```suggestion
6. **Enforce UCP's minimum covered-component set, in every regime.**
   Independent of `tag`, the signature that authenticates the request
   **MUST** cover at least:
    - `@method`, `@authority`, and `@path` (always);
    - `@query` if the request carries a query string;
    - `content-digest` and `content-type` if the request has a body;
    - `idempotency-key` for POST, PUT, PATCH, and DELETE;
    - `ucp-agent` if the `UCP-Agent` header is present.

   If any required component is absent from the signature's covered
   set, **skip** this signature; a body, method, or path that the
   authenticating signature does not cover is treated as unsigned. This
   prevents a signature satisfying only Web Bot Auth's minimal covered
   set (`@authority`, `signature-agent`) from authenticating a
   state-changing UCP request whose body, method, or path is unbound.
7. **Verify the signature** using the matched key. The signing
    algorithm is derived from the JWK's `kty`/`crv`. If the verifier
    does not support the matched key's `kty`, `crv`, or `alg`, **skip**
    this signature; it yields `algorithm_unsupported` if no other
    signature authenticates the request.
```

docs/specification/signatures.md
```diff
@@ -381,13 +653,23 @@ verify_rest_request(request):
...
+    public_key = find_key_by_kid(directory, keyid)
     if not public_key:
         return error("key_not_found")
 
+    // 2a. Skip keys whose algorithm this verifier does not support.
+    // The kty/crv/alg vocabularies are open (see Signature Algorithms);
+    // an unsupported key never invalidates the whole key set.
+    if not algorithm_supported(public_key):
+        return error("algorithm_unsupported")
+
```
shopdave, line 705: Step 3 currently checks the digest only `if "content-digest" in components`, so a body-bearing request whose signature omits `content-digest` silently skips integrity verification. (This predates WBA, but the WBA path makes minimal-coverage signatures legitimate.) Enforce presence before verification.

```suggestion

    // 2b. Enforce UCP's minimum covered-component set (all regimes).
    // A signature MUST bind the request's integrity-relevant fields
    // regardless of tag; a WBA-shape signature covering only WBA's
    // minimum (@authority, signature-agent) does not authenticate a
    // body-bearing or state-changing UCP request.
    required = ["@method", "@authority", "@path"]
    if request.query:                                required += ["@query"]
    if request.has_body:                             required += ["content-digest", "content-type"]
    if request.method in [POST, PUT, PATCH, DELETE]: required += ["idempotency-key"]
    if "UCP-Agent" in request.headers:               required += ["ucp-agent"]
    for component in required:
        if component not in components:
            return error("signature_invalid")
```

docs/specification/signatures.md
```diff
@@ -422,12 +704,19 @@ verify_rest_response(response, signer_profile_url):
...
+    public_key = find_key_by_kid(profile, keyid)
     if not public_key:
         return error("key_not_found")
 
+    // 2a. Skip keys whose algorithm this verifier does not support.
+    // The kty/crv/alg vocabularies are open (see Signature Algorithms);
+    // an unsupported key never invalidates the whole key set.
+    if not algorithm_supported(public_key):
+        return error("algorithm_unsupported")
+
```
shopdave, line 770: Mirror of the gate for responses (no method/idempotency to bind, but the body still must be covered).

```suggestion

    // 2b. Enforce minimum covered components for responses (all regimes).
    required = ["@status"]
    if response.has_body:
        required += ["content-digest", "content-type"]
    for component in required:
        if component not in components:
            return error("signature_invalid")
```

pjordan:
[This is a quick review to unblock @igrigorik, largely assisted by claude]

## Overall

This is a strong, carefully-iterated PR that turns a previously *impossible* thing — emitting a Web Bot Auth-conformant signature from a UCP signer — into a clean, backward-compatible capability. I'm in favor of the direction, and after the 06-15 fixes the substrate is in good shape. My position is **approve after a small set of correctness/interop fixes, with the contested key-publication architecture routed to a tracking issue rather than blocking here.**

A few things I want to credit specifically (extending @amithanda's top-level praise rather than repeating it):

- The **per-signature `tag`-dispatch** is exactly the right primitive. One set of signed bytes, two verification regimes selected on `tag`, "authenticated if at least one verifies," with **unknown tags skipped** — that's the correct reading of RFC 9421 §4.3 (multiple signatures) composed with WBA architecture §4.2.3 (UCP's `ucp-agent`/`idempotency-key`/`content-digest`/`content-type` ride as permitted "additional components"). It's forward-compatible by construction.
- The **`@authority` cache-key MUST** in *WBA Response Signing* ("any cache … MUST include the request authority in its cache key … otherwise a cached signature served for `Host: a.example` will fail under `Host: b.example`") is a subtle, easy-to-omit correctness rule that the directory draft itself doesn't spell out (it only says "cache the directory contents," directory-05 §5.1). It lines up exactly with the `@authority;req`-MUST-set self-signature in directory-05 §5.2 and is precisely the trap Pattern 1's "identical body bytes" design invites. Nice catch.
- **A1 landed well** (commit `44d1dba`): `$defs.jwk_public_key` is now `required: ["kid","kty"]`, the `kty` `const:"EC"` is gone, enums are `examples`, and an `allOf` of `if/then` branches preserves EC-needs-`x,y` / OKP-needs-`x` and the per-curve→`alg` coupling *while letting an unknown `kty` pass*. That's the right shape — it avoids re-creating the exact "couldn't add Ed25519 without a schema bump" trap this PR exists to fix.

### Security & key lifecycle
A3 is resolved in prose — revocation now correctly states a key "is not effectively revoked until it is absent from **both** arrays" and that add/rotate/remove "MUST update both arrays atomically." Good. The residual I'd put on the record (not re-litigating A3): that MUST is **prose-only and structurally unenforceable** — `signing_keys[]` and `keys[]` are two independent arrays in `profile.json`, JSON Schema 2020-12 can't express "`keys[]` mirrors `signing_keys[]`," and per the Identity Resolution Algorithm each signature reads exactly one array, so no verifier in the resolution path can detect drift either. A compromised key dropped from `signing_keys[]` but lingering in `keys[]` keeps verifying for every WBA-shape verifier. I wouldn't bolt a cross-array check onto verifiers (it doesn't reach the regime that fails open) — this is the sharpest standalone argument for @thibmeu's single-list direction, and I'd feed it into that discussion rather than block.

Two net-new security items worth resolving before merge: (1) **`nonce` is absent entirely** — the WBA opt-in list omits it, though WBA architecture §4.2.2 says agents SHOULD add a `nonce` (base64url, ~64 bytes, unique within the validity window) and origins MAY require one (§4.4); a UCP signer following our 6-item list to the letter is nonce-less and a strict WBA origin is entitled to re-challenge it (§4.3). UCP's `idempotency-key` is business-layer dedup, **not** a transport-bound nonce, so it isn't a substitute. (2) A **freshness contradiction**: the WBA section says RFC 9421-conformant verifiers "will enforce the freshness window," while the standing note still says `created` is OPTIONAL and "UCP handles replay protection at the business layer … not signature timestamps." These are reconcilable by scoping (WBA-shape MUSTs `created`/`expires`; default UCP doesn't), but as written a reader can't tell which governs. (Minor precision: RFC 9421 §3.2.1 makes expiry/nonce enforcement an *application-defined* requirement a verifier MUST then enforce — it isn't an automatic consequence of conformance — so "UCP verifiers MUST reject expired WBA-shape signatures" is the accurate phrasing.)

### Interop reality
This is the item I'd weight most heavily, and it sharpens @dpeacock's Cloudflare comment and @igrigorik's discovery-precedence question from open-questions into a concrete interop break. I tried to mentally run **Pattern 1** (and the worked examples that point `Signature-Agent` at `…/.well-known/ucp`) against the one WBA verifier shipping today, Cloudflare, and I don't think it verifies. Cloudflare's docs make the directory at `/.well-known/http-message-signatures-directory` a hard requirement, require `Content-Type: application/http-message-signatures-directory+json`, and require per-key response self-signatures; its signed-request example carries a **bare origin** `Signature-Agent`, which (the only reading consistent with both facts) means it takes the *authority* and appends the well-known path rather than dereferencing the URL verbatim. So Pattern 1 fails on path, media type, and self-signature simultaneously. To be fair to the PR: the verbatim-deref reading **is** spec-legal under directory-05 §4.1 — this isn't "wrong per the draft," it's "incompatible with the deployed §4.5-reading verifier." Pattern 2 already provisions a real signed directory endpoint and is the conformant path. I'd resolve this before merge by either (a) making Pattern 2 (bare-origin `Signature-Agent`, the directory media type, per-key self-sigs) the WBA-conformant default and scoping Pattern 1 as *"spec-legal per directory-05 §4.1 but not interoperable with deployed verifiers today,"* or (b) holding Pattern 1 until the directory draft adds an explicit precedence rule. Relatedly: the PR tells verifiers to "discard keys without a valid self-signature" (IRA step 4) but never names the concrete contract to do so — the directory self-signature uses **`tag="http-message-signatures-directory"`**, distinct from the request `tag="web-bot-auth"`, and that string appears 0× in the diff. Two sentences naming the tag, the `@authority;req` component, and the `+json` media type would make step 4 actionable.

### Architecture & future-proofing
The central choice — inline `keys[]` mirror vs @thibmeu's `jwks_uri`/CIMD direction (thibmeu/http-message-signatures-directory#98) — is currently being made implicitly. I don't think the PR should pivot; shipping the mirror now to unblock dual-audience is reasonable and lower-cost. But I'd like @igrigorik and @thibmeu to put the decision on the record: **is `keys[]` a true mirror we intend to deprecate in favor of a `jwks_uri`/`type=cimd` shape, or a permanent canonical field?** The migration seam matters — the PR twice promises "a future UCP version may promote `keys[]` to canonical" but the IRA step 3 hard-binds "default UCP signature → read `signing_keys[]`," so a later deprecation is a breaking change unless we write the fallback rule (e.g., "verifiers SHOULD fall back to `keys[]` when `signing_keys[]` is absent") now. I'd route convergence to a dedicated issue alongside the discovery-precedence thread. Two smaller future-proofing notes: the `tag` namespace UCP is establishing as a routing key is an un-registered free-form RFC 9421 §2.3 string (worth stating how UCP allocates the values it recognizes, and considering reserving `tag="ucp"` while still accepting untagged-as-UCP for backcompat); and the normative coupling to two **individual, non-WG-adopted** drafts (architecture = Informational, directory = Standards Track, both "Expires 3 September 2026") should be pinned to `-05` per @thibmeu's T6 — all five `draft-meunier-*` links are currently bare.

### Adoption & governance
Two cheap, high-leverage fixes: the **PR description still carries the RSA thumbprint `NzbLsXh8…` on Ed25519 keys (3×)** — the spec files are fixed to `kPrK_…`, the body just needs the same swap so the canonical example doesn't fail its own item-4 MUST (extending @amithanda's A2, which asked for the body fix too). And per @thibmeu's T3, one fully-worked vector reusing the named RFC 9421 `test-key-ed25519` (`x=JrQLj5P_89iXES9-vFgrIy29clF9CC_oPPsw3c5D0bs`, thumbprint `poqkLGiymh_W0uP6PZFw-dvez3QJT5SolqXBCW38r0U`) would make test material recognizable as non-prod. On the AP2 two-key story: it's accurate and honestly labeled interim (AP2 v0.2 §Payment Mandate does hard-MUST exclude Ed25519 for the Checkout JWT). I'd add two calibrations — that the single-key future is gated on AP2 #268, which is OPEN, author-filed, and heading to the FIDO Agentic Payments WG (a strength worth stating plainly, not concealing), and that the actual fix is a **payload-entropy** rule UCP satisfies via the Checkout `id`, not algorithm-permissiveness (plain Ed25519, including ph/ctx, contributes zero signature entropy). On `ecdsa-p256-sha256` (T4): I read architecture-05 §4.2 as permitting any IANA-registered algorithm via an open set rather than naming ES256 specifically (the draft ships vectors only for `rsa-pss-sha512` and `ed25519`), and Cloudflare accepts Ed25519 only today — so "spec-permitted via the open set, not deployed" is the precise framing, and the Ed25519 SHOULD here is right.

**Merge recommendation: approve-after-changes.** Land the baseline; gate merge on the cheap correctness/interop items (PR-body thumbprint, `nonce` SHOULD + the freshness scoping, the directory self-sig tag, the Pattern-1 conformance caveat, `-05` pinning, one worked vector). Route the discovery-precedence decision and the inline-mirror-vs-`jwks_uri` fork to a tracking issue for @igrigorik / @thibmeu / @amithanda to converge — and strongly consider splitting the uncontested Ed25519/OKP schema fix into its own merge so the ecosystem can adopt the unblock immediately. None of this is an objection to the design, which is the right one. Thanks for the careful iteration — and for filing AP2 #268 upstream rather than waiting on it. 🚀

docs/specification/signatures.md
```diff
@@ -131,49 +188,197 @@ defined in [RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517).
...
+   interval SHOULD be at most 24 hours.
+6. **MUST include `tag="web-bot-auth"`.** WBA verifiers select
+   signatures by this tag.
+
+**Component requirements preserved.** UCP's existing signed components
+(`ucp-agent`, `idempotency-key`, `content-digest`, `content-type`)
+stay in the signed list. WBA accepts them as "additional components"
+per
+[draft-meunier-web-bot-auth-architecture](https://datatracker.ietf.org/doc/draft-meunier-web-bot-auth-architecture/)
+§4.2.3.
```
shopdave: As written this reads as signer-side guidance ("stay in the signed list"), but nothing in the verifier algorithm enforces it. Tie the requirement to the (updated) Identity Resolution Algorithm and state the property plainly.

```suggestion
**Component requirements preserved — and enforced.** A WBA-shape
signature **MUST** still cover UCP's full component set (`@method`,
`@authority`, `@path`, `ucp-agent`, `idempotency-key`, `content-digest`,
`content-type`) exactly as a default UCP signature does; Web Bot Auth
accepts these as "additional components" per
[draft-meunier-web-bot-auth-architecture](https://datatracker.ietf.org/doc/draft-meunier-web-bot-auth-architecture/)
§4.2.3. This is enforced at the verifier, not left to signer discipline:
per the [Identity Resolution Algorithm](overview.md#identity-resolution-algorithm),
a UCP verifier **MUST** skip any signature — WBA-shape or default — that
omits a component required for the request (`content-digest` and
`content-type` for a body; `idempotency-key` for a state-changing
method), even when it verifies cryptographically. Opting into Web Bot
Auth therefore never widens what UCP authenticates.
```

shopdave:
Regarding @pjordan's remarks on interop reality: I do think there is something to address here. Adding something like:

_Publishing top-level keys[] is necessary but not sufficient for UCP-WBA interop. A profile or directory endpoint used for WBA key discovery MUST be served with per-key HTTP Message Signatures on the response; otherwise WBA verifiers following the directory draft’s recommended validation behavior will ignore the keys._

or something to that effect may resolve it.

igrigorik:
Ty all for thorough feedback. Pushed a major revision, updated PR description to match. Key changes:

1/ **UCP doesn't become a WBA directory.**
The profile is a valid RFC 7517 JWK Set a signer MAY reuse; WBA discovery is just `Signature-Agent`'s typed variants (`jwks_uri`/`cimd`/`directory`), presented as options, each able to point back at the profile. The in-spec directory self-signature contract is removed — format/self-sigs/media-type are deferred to the directory draft §5.2. This reverses the earlier "name the self-sig contract" direction: that mechanism belongs to the draft, not UCP.

2/ **A UCP verifier behaves the same whether or not a signature is WBA-shaped.**
It always finds the signing key via `UCP-Agent`, regardless of the signature's `tag` — so WBA support stays optional and a UCP-only verifier needs zero WBA-specific code. Separately, every signature (WBA-shaped or not) MUST cover the same required request parts (method, path, body digest, identity headers); one that covers only WBA's bare minimum can't authenticate a UCP request whose body or path it left unsigned.

3/ **"Dual-audience" is now defined and scoped.**
`kid` MUST equal the RFC 7638 thumbprint for dual-audience keys (else the same signature a WBA verifier accepts fails `UCP-Agent` resolution), the `keyid`==thumbprint check is MUST, and fetch safety (RFC 6890 special-use-IP rejection + loopback carve-out + response-size bound) is specified — aligned with the WBA architecture draft's SSRF guidance.

[PR #98](https://github.com/thibmeu/http-message-signatures-directory/pull/98) landed – hooray! @thibmeu I filed thibmeu/http-message-signatures-directory#100: would love to remove unnecessary friction here, wdyt?

Ready for next review pass — ptal 🙏

_p.s. this PR introduces `keys[]` as parallel option, which keeps this PR backward compatible. Having done a few more turns on this proposal, I'm conviced we should make `keys[]` the default but I'll stage that as separate PR that we can fast-land after we land this one -- we can backport this change and make `keys[]` default as breaking change in next (08) release._

docs/specification/overview.md
igrigorik: Ah, ty for flagging this. Updated.

docs/specification/overview.md
igrigorik: Reworked.

source/schemas/profile.json
```diff
@@ -67,7 +96,12 @@
         },
         "signing_keys": {
           "type": "array",
-          "description": "Public keys used to verify signatures from this profile owner.",
+          "description": "Canonical UCP profile field for publishing signing keys. When a profile publishes signing keys, they MUST appear here; this is where every UCP verifier reads them. A future UCP version will promote keys[] to canonical and deprecate signing_keys[]; profiles wanting WBA interop today MAY publish both arrays as mirrors.",
```
thibmeu, line 99: I feel stating that `keys` will be promoted may be to strong. Specifically because WBA got multiple comments to favor `jwks_uri` (over even `jwks`)


```suggestion
          "description": "Canonical UCP profile field for publishing signing keys. When a profile publishes signing keys, they MUST appear here; this is where every UCP verifier reads them. A future UCP version may promote keys[] to canonical and deprecate signing_keys[]. This migration is tracked separately. Profiles wanting WBA interop today MAY publish both arrays as mirrors.",
```

docs/specification/overview.md
```diff
@@ -1135,25 +1232,193 @@ request access. Strategies include:
...
+### Deployment Patterns for WBA Interop
+
+A UCP profile carrying a top-level `keys[]` array is a valid RFC 7517
+JWK Set, which a signer can optionally reuse as its Web Bot Auth key
+source. The `Signature-Agent` header's `type` parameter
+([draft-meunier-http-message-signatures-directory](https://datatracker.ietf.org/doc/draft-meunier-http-message-signatures-directory/))
+selects how a verifier resolves the advertised keys. (The `type`
+parameter and its `jwks_uri`/`cimd`/`directory` values were added via
+[directory-draft PR #98](https://github.com/thibmeu/http-message-signatures-directory/pull/98);
+they are in the editor's copy, pending publication as -06.) Each
```
thibmeu: @thibmeu that's an item on me. Should be this week

docs/specification/overview.md
```diff
@@ -1135,25 +1232,193 @@ request access. Strategies include:
...
+      dual-audience shape); the verifier resolves them via `UCP-Agent`,
+      treats `signature-agent` as an ordinary covered component, and
+      need not implement Web Bot Auth key discovery. (Verifying a
+      dual-audience signature does still require supporting the key's
+      algorithm — whichever the signer used, per
+      [Signature Algorithms](signatures.md#signature-algorithms) — and
+      RFC 9421 §2.1.2 Dictionary-member component selection to cover
+      `signature-agent;key="<label>"`.) Signatures whose `tag` denotes
+      another application's purpose are skipped (see below) — UCP does
+      not claim signatures scoped to other applications.
```
thibmeu:
```suggestion
      `signature-agent;key="<label>"`.) Signatures with tags other than `web-bot-auth` are skipped unless UCP defines or explicitly accepts that tag.
```

docs/specification/overview.md
```diff
@@ -1135,25 +1232,193 @@ request access. Strategies include:
...
+parameter and its `jwks_uri`/`cimd`/`directory` values were added via
+[directory-draft PR #98](https://github.com/thibmeu/http-message-signatures-directory/pull/98);
+they are in the editor's copy, pending publication as -06.) Each
+variant can stand alone or point back at the UCP profile:
+
+- **`type=jwks_uri`** — the member value is a JWK Set URL, fetched
+  directly. Point it at the UCP profile URL and the profile's `keys[]`
+  serves as the JWK Set: one document is both profile and key source.
+  Integrity derives from TLS to the profile origin, with no per-key
+  self-signature.
```
thibmeu: this is defined in the draft-meunier-http-message-signatures-directory draft. probably worth pointing to it directly (will be in [-06](https://thibmeu.github.io/http-message-signatures-directory/draft-meunier-http-message-signatures-directory.html#section-4.1))

thibmeu (commented):
The changes to the specification make it much closer to WBA, specifically with the optional `keys`. 

The addition of `Signature-Agent` with `directory`/`jwks_uri`/`cimd` helps a lot, and reuses discovery via signatures components.

Only question mark I'd have is regarding using `keys` rather than `jwks_uri`, but that's a recommendation that can evolve later, and that's valid today

pjordan:
[Second pass, post the 06-19 revision. Largely assisted by Claude, with the external claims checked against primary sources by Claude.]

I am approve ready.

## Claude's review

The 06-19 revision is a big step and it resolved most of what I — and @amithanda, @shopdave, @thibmeu — raised last round. Crediting the changes specifically so we don't re-litigate them: `nonce` is now a SHOULD with the WBA §4.2.2 shape (~64-byte base64url, unique within the validity window) and the "`Idempotency-Key` is not a nonce" distinction is explicit; the freshness contradiction is resolved by scoping (`created`/`expires` MUST for WBA-shape, OPTIONAL for default UCP) with the correct "enforcement is application-defined per RFC 9421 §3.2.1" framing; the AP2 two-key story moved to `ap2-mandates.md` and is accurately calibrated (entropy-vs-algorithm-class, AP2 #268, UCP's per-session Checkout `id` as the entropy source); `ecdsa-p256-sha256` is framed as permitted-via-open-registry; the architecture refs are pinned to `-05`; and the Ed25519 example `kid` is now the correct RFC 7638 thumbprint of the RFC 9421 test key (`poqkLGiymh…`, which I recomputed — it matches, and the old RSA-thumbprint value is gone from both the spec and the PR body). The capability-based Identity Resolution Algorithm (tag-as-hint, not gate), the closed covered-component gate (@shopdave's), the private-key `not`-guard in `profile.json`, the open `kty`/`crv`/`alg` vocabulary, and the new "Authenticated identity" section closing the `Signature-Agent`-vs-`UCP-Agent` identity-confusion gap are all the right calls. Net: I'm at **approve-after-changes**, and the list is now short.

One thing I checked that turned out fine: the `type=directory` semantics. Against the merged directory PR #98 (commit `40119d8`), `type=directory` does mean "member value is an **origin**, append the well-known path," while `jwks_uri`/`cimd` are verbatim-deref — so the PR's description is **correct, not a bug**. That verification surfaced the items below.

source/schemas/profile.json
igrigorik: @thibmeu I'll keep this as is, the "will promote keys[] to canonical" is not due to WBA; UCP already defines `signing_keys[]` and we really should have called it `keys[]` to begin with.

docs/specification/overview.md
igrigorik: fixed

thibmeu (approved):
looks good to me!

igrigorik:
@thibmeu ty for your help and feedback!

@pjordan clarification on `keys[]`: current PR adds this object as optional and parallel to existing `signing_keys[]`, with suggestion to mirror same keys in both -- this shape will land as backwards compatible via this PR. After this lands, I'll stage a separate PR to make `keys[]` the default and remove `signing_keys[]` -- this will eliminate duplication and make UCP profile a valid JWKS; this will go into next major release as breaking change.

That aside, I believe this is good land. Ready for final pass.

docs/specification/signatures.md
```diff
@@ -377,14 +667,38 @@ verify_rest_request(request):
...
-    // See overview.md#profile-requirements for URL validation and fetch rules
-    profile_url = get_profile_url_from_ucp_agent(request.headers["UCP-Agent"])
-    profile = fetch_profile(profile_url)  // HTTPS-only, no redirects
-    public_key = find_key_by_kid(profile.signing_keys, keyid)
+    // 2. Resolve signer's public key (capability-based; see
+    // overview.md#identity-resolution-algorithm).
+    key_set = resolve_signer_key_set(request.headers)
+    public_key = find_key_by_kid(key_set, keyid)
     if not public_key:
         return error("key_not_found")
```
shopdave: Fix for multi-signature guarantee:

```suggestion
        return skip_signature("key_not_found")
```

docs/specification/signatures.md
```diff
@@ -377,14 +667,38 @@ verify_rest_request(request):
...
 
+    // 2a. Skip keys whose algorithm this verifier does not support.
+    // The kty/crv/alg vocabularies are open (see Signature Algorithms);
+    // an unsupported key never invalidates the whole key set.
+    if not algorithm_supported(public_key):
+        return error("algorithm_unsupported")
```
shopdave: Fix for multi-signature guarantee:

```suggestion
        return skip_signature("algorithm_unsupported")
```

docs/specification/signatures.md
```diff
@@ -377,14 +667,38 @@ verify_rest_request(request):
...
+    if request.query:                        required += ["@query"]
+    if request.has_body:                     required += ["content-digest", "content-type"]
+    if "Idempotency-Key" in request.headers: required += ["idempotency-key"]
+    if "UCP-Agent" in request.headers:       required += ["ucp-agent"]
+    if "Signature-Agent" in request.headers: required += ["signature-agent"]
+    for component in required:
+        if component not in components:
+            // coverage failure, not a crypto failure; under multi-signature
+            // handling this is skip-and-try-next (see IRA step 5)
+            return error("signature_invalid")
```
shopdave: Fix for multi-signature guarantee:

```suggestion
            // coverage failure, a target/body/header the signature does not
            // cover is treated as unsigned (see IRA step 5)
            return skip_signature("coverage_insufficient")
```

docs/specification/signatures.md
```diff
@@ -399,12 +713,12 @@ verify_rest_request(request):
 
     // 5. Verify signature
     signature = parse_signature(request.headers["Signature"])
-    if not ecdsa_verify(signature_base, signature, public_key):
+    if not verify(signature_base, signature, public_key):
         return error("signature_invalid")
```
shopdave: Fix for multi-signature guarantee:


```suggestion
        return skip_signature("signature_invalid")
```

docs/specification/signatures.md
shopdave, line 681: verify_rest_request()/verify_rest_response() hard-return error(...) on every failure path, but the IRA says a request is authenticated if any signature verifies and rejected only when all fail.

The inline comment at signatures.md:698–699 even says "skip-and-try-next" directly above a return error("signature_invalid"). As written, one valid + one failing signature → wrongful rejection.

Suggest making explicit that this routine processes a single candidate and its error returns are "skip this candidate" from the IRA's perspective (or wrap it in the outer loop).

docs/specification/signatures.md
```diff
@@ -419,12 +733,26 @@ verify_rest_response(response, signer_profile_url):
     keyid = sig_input.keyid
     components = sig_input.components
 
-    // 2. Fetch signer's public key
+    // 2. Resolve signer's public key from the signer's profile.
     profile = fetch_profile(signer_profile_url)
-    public_key = find_key_by_kid(profile.signing_keys, keyid)
+    public_key = find_key_by_kid(profile, keyid)
```
shopdave: Nit:

```suggestion
    public_key = find_key_by_kid(profile.signing_keys, keyid)
```

docs/specification/signatures.md
```diff
@@ -377,14 +667,38 @@ verify_rest_request(request):
...
-    // See overview.md#profile-requirements for URL validation and fetch rules
-    profile_url = get_profile_url_from_ucp_agent(request.headers["UCP-Agent"])
-    profile = fetch_profile(profile_url)  // HTTPS-only, no redirects
-    public_key = find_key_by_kid(profile.signing_keys, keyid)
+    // 2. Resolve signer's public key (capability-based; see
+    // overview.md#identity-resolution-algorithm).
+    key_set = resolve_signer_key_set(request.headers)
+    public_key = find_key_by_kid(key_set, keyid)
     if not public_key:
         return error("key_not_found")
```
shopdave: The keyid == RFC 7638 thumbprint check is a MUST for WBA-shape signatures in IRA step 4. It's the binding that ties an advertised key's identity to its bytes. But the verify_rest_request() pseudocode never performs it; it only string-matches keyid to kid. An implementer coding from the pseudocode omits the core WBA trust check.

```suggestion

   // pre-2a. WBA-shape signatures bind key identity to key bytes:
   // keyid MUST equal the matched JWK's RFC 7638 thumbprint
   // (see IRA step 4 / WBA architecture draft §4.2).
   if sig_input.tag == "web-bot-auth":
       if keyid != rfc7638_thumbprint(public_key):
           return skip_signature("signature_invalid")
```

docs/specification/overview.md
```diff
@@ -1135,25 +1232,191 @@ request access. Strategies include:
...
+  ```text
+  Signature-Agent: sig1="https://platform.example/agent";type=cimd
+  ```
+
+- **`type=directory`** *(default)* — when `type` is omitted or set to
+  `directory`, the member value is an **origin** (not a full URL): the
+  verifier appends the registered well-known path
+  (`/.well-known/http-message-signatures-directory`) to that origin and
+  fetches a signed directory there. Its format, per-key self-signatures,
+  and media type are defined by the directory draft §5.2.
```
shopdave, line 1408: `type=directory` is the default when type is omitted, and it requires a signed directory with per-key self-signatures at /.well-known/http-message-signatures-directory .The most likely UCP deployment, a static /.well-known/ucp file with mirrored keys[], has none of these, so a conforming WBA directory verifier ignores every key and rejects. The "profile doubles as a WBA key source" framing needs this caveat; "both verifiers accept it" is false for static hosting on the default path.

.cspell/custom-words.txt
```diff
@@ -123,4 +123,28 @@ vulnz
 worktree
 yaml
 yml
+jwks
```
shopdave, line 139: Need to run `uniq` on these.

docs/specification/signatures.md
```diff
@@ -131,49 +186,225 @@ defined in [RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517).
...
+**Component requirements preserved — and enforced.** A WBA-shape
+signature **MUST** still cover the same components a default UCP
+signature does (the Required set in the
+[Signed Components](#rest-request-signing) table); WBA accepts them as
+"additional components" per
+[draft-meunier-web-bot-auth-architecture-05](https://datatracker.ietf.org/doc/draft-meunier-web-bot-auth-architecture/05/)
+§4.2.3. The verifier enforces this regardless of `tag` per the
+[Identity Resolution Algorithm](overview.md#identity-resolution-algorithm),
+so opting into Web Bot Auth never widens what UCP authenticates.
+
```
shopdave, line 354: Nit, maybe worth highlighting.

```suggestion

**Interop is directional.** This gate makes a UCP verifier a *strict
subset* of Web Bot Auth: a minimal WBA signer covering only `@authority`
(permitted by WBA architecture-05 §4.2) is **rejected** by a UCP verifier.
UCP interop targets WBA *verifiers* — which accept UCP's richer covered set
as "additional components" — not arbitrary minimal WBA *signers*. A UCP
signer satisfies a WBA verifier; a minimal WBA signer does not
automatically satisfy a UCP verifier.
```

shopdave (approved):
LGTM with minor corrections.

amithanda:
Hi @igrigorik,

+1 to addressing @shopdave's suggestions few hours ago. I wanted to flag a concern regarding the normative coupling to the unreleased `-06` version of the key directory draft.

#### The Concern: Spec Drift & Out-of-Date References
In `docs/specification/overview.md` (around L1275-1300), we define the `Signature-Agent` `type` parameters (`jwks_uri`, `cimd`, and `directory`):
* These parameters are introduced to UCP in this PR to support Pattern 1 (`type=jwks_uri` pointing to the UCP profile).
* While I see that draft references were pinned to `-05` in a recent commit to align with prior feedback, these parameters are **not** actually present in the currently published [draft-meunier-http-message-signatures-directory-05](https://datatracker.ietf.org/doc/html/draft-meunier-http-message-signatures-directory-05). As noted in the PR description, they are in the editor's copy, pending publication as `-06`.
* Currently, pinning the UCP spec text to a Datatracker URL that resolves to `-05` means a developer reading the official document will not find these parameters, leading to validation confusion. 
* More importantly, because these parameters are part of an active IETF working group (`webbotauth`) draft that is not yet finalized, there is a risk that the working group could rename, restructure, or drop these parameters before `-06` or the final RFC is published. If that happens, UCP's wire format will drift from the standard. I would like to hear your thoughts on this.

**Suggestion:** I completely appreciate the goal to keep UCP isolated from WBA directory internals. However, to protect early implementers from spec drift before the final RFC lands, could we either temporarily point the Markdown links directly to a commit-locked version of the editor's copy where they are defined, or add a brief informative footnote indicating that these parameters track the upcoming `-06` working group evolution?

####  Single-Array Migration
The security gaps @shopdave highlighted in the `verify_rest_request()` pseudocode—where the logic completely misses the `keyid == thumbprint` validation step—perfectly underscores why the parallel array structure represents an architectural risk, so I fully endorse the resolution plan (in response to @pjordan's comment) in your status note to promote `keys[]` as the default source of truth very quickly after this.

docs/specification/overview.md
```diff
@@ -1135,25 +1232,191 @@ request access. Strategies include:
...
 status code (see [Error Handling](#error-handling)).
 
+### Deployment Patterns for WBA Interop
+
+A UCP profile carrying a top-level `keys[]` array is a valid RFC 7517
+JWK Set, which a signer can optionally reuse as its Web Bot Auth key
+source. The `Signature-Agent` header's `type` parameter selects
+how a verifier resolves the advertised keys. The parameter and
+its `jwks_uri`/`cimd`/`directory` values are defined in §4.1 of
+[draft-meunier-http-message-signatures-directory](https://datatracker.ietf.org/doc/html/draft-meunier-http-message-signatures-directory).
```
thibmeu:
```suggestion
[draft-meunier-webbotauth-httpsig-directory-00](https://datatracker.ietf.org/doc/draft-meunier-webbotauth-httpsig-directory/00/).
```

docs/specification/signatures.md
```diff
@@ -131,49 +186,225 @@ defined in [RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517).
...
+but `Signature-Agent` MAY point elsewhere.
+
+The three `sig1` labels are bound together — the `Signature-Agent`
+dictionary member key, the `;key="sig1"` parameter on the signed
+`signature-agent` component, and the `Signature-Input` signature label —
+per item 3 of the opt-in list below.
+
+To opt in, a signer makes the following changes to their primary UCP
+signature. Items marked **MUST** are required by
+[draft-meunier-web-bot-auth-architecture-05](https://datatracker.ietf.org/doc/draft-meunier-web-bot-auth-architecture/05/)
```
thibmeu:
```suggestion
[draft-meunier-webbotauth-httpsig-protocol-00](https://datatracker.ietf.org/doc/draft-meunier-webbotauth-httpsig-protocol/00/)
```

docs/specification/signatures.md
```diff
@@ -131,49 +186,225 @@ defined in [RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517).
...
+   (WBA §4.3–4.4).
+7. **MUST include `tag="web-bot-auth"`.** WBA verifiers select
+   signatures by this tag.
+
+**Component requirements preserved — and enforced.** A WBA-shape
+signature **MUST** still cover the same components a default UCP
+signature does (the Required set in the
+[Signed Components](#rest-request-signing) table); WBA accepts them as
+"additional components" per
+[draft-meunier-web-bot-auth-architecture-05](https://datatracker.ietf.org/doc/draft-meunier-web-bot-auth-architecture/05/)
```
thibmeu:
```suggestion
[draft-meunier-webbotauth-httpsig-protocol-00](https://datatracker.ietf.org/doc/draft-meunier-webbotauth-httpsig-protocol/00/)
```

docs/specification/overview.md
```diff
@@ -1135,25 +1232,191 @@ request access. Strategies include:
...
+      other than `web-bot-auth` are skipped unless UCP defines
+      or explicitly accepts that tag.
+    - **`Signature-Agent` — Web Bot Auth key lookup, OPTIONAL
+      (WBA-aware verifiers).** For a signature carrying
+      `tag="web-bot-auth"`, a WBA-aware verifier **MAY** instead resolve
+      via the `Signature-Agent` member, parsed per the
+      [Signature-Agent parsing rules](signatures.md#rest-request-verification).
+      Such a signature **MUST** satisfy the WBA agent-signature
+      requirements in
+      [draft-meunier-web-bot-auth-architecture-05](https://datatracker.ietf.org/doc/draft-meunier-web-bot-auth-architecture/05/)
```
thibmeu:
```suggestion
      [draft-meunier-webbotauth-httpsig-protocol-00](https://datatracker.ietf.org/doc/draft-meunier-webbotauth-protocol/00/)
```

thibmeu (commented):
the latest editor copy of the webobotauth drafts have been published on the datatracker. there has been name changes (based on feedback from the wg), which i've added here, in addition to pinning the versions

igrigorik:
@thibmeu ty, landed updates with a few additional section pointer fixes.

@amithanda prescient comments!

The `type` parameter (`jwks_uri`/`cimd`/`directory`) is now in the published `-00` of the renamed directory draft, at **§4.1** — so a developer following our link lands on a version that actually contains it. I've re-pinned every citation to `-00`. Since datatracker versions are immutable, this is effectively your option (a): a commit-locked pointer to where the parameters are defined, with no silent drift. I've also updated the PR description.

Re, single-array migration: 👍 and I'll stage that as soon as we land this.

amithanda:
I spent some time on whether anchoring the WBA key discovery on the `draft-meunier-webbotauth-httpsig-*` drafts is a good idea, given they are still individual (pre-adoption) submissions and were renamed/renumbered even during this PR. My conclusion is that it is an acceptable, deliberately contained trade-off rather than a problem: the dependency lives entirely behind the WBA opt-in, so any further draft churn touches only that section and never UCP's base key discovery (`signing_keys[]`), and the `type` dispatch (`jwks_uri` / `cimd` / `directory`) already hedges across mechanisms rather than betting on one. Tracking a live, chartered-WG, multi-vendor standard this way is the right call for an opt-in interop profile.
One genuinely optional, question and a note. Nothing here should hold up the merge:
- Has the recommended `type=jwks_uri` profile-as-JWKS path been validated end-to-end against a verifier the UCP ecosystem will actually use? 
- Whenever the drafts are formally WG-adopted, the references will want to move from `draft-meunier-webbotauth-httpsig-*` to `draft-ietf-*`. Purely a future-cleanup note.

Nice work on this. Approving.

amithanda (approved):

igrigorik:
> Has the recommended type=jwks_uri profile-as-JWKS path been validated end-to-end against a verifier the UCP ecosystem will actually use?

We will prototype / validate this at Shopify. 

> Whenever the drafts are formally WG-adopted, the references will want to move from draft-meunier-webbotauth-httpsig-* to draft-ietf-*. Purely a future-cleanup note.

Yep, I'll track this.

pjordan:
Looks good to me!

pjordan (approved):

PR #485, opened 2026-05-27
fix: specify idempotency payload-mismatch contract

igrigorik:
Fixes an underspecified case in UCP's idempotency contract: same-key + mismatched-payload behavior. The base requirements table in `signatures.md` previously said only *"On duplicate: Return cached response, do not re-execute"*, with no qualifier for whether the duplicate-key request's payload matches the original. The `overview.md` Protocol Errors table correctly mapped HTTP `409` to *"Idempotency key reused with different payload"*, but no normative contract told businesses *when* to return it.

This was surfaced during PR #409 (split-payments) review, where the spec needed to call out "use a new idempotency key when retrying with modified payment instruments." That's not a split-payments-specific contract — it's a base UCP idempotency contract that was underspecified. 

### Compatibility

Technically this is a normative tightening: a literal pre-PR implementation of "return cached response on any duplicate key, regardless of payload" is non-conforming after this PR. In practice the impact is expected to be minimal because (a) `overview.md`'s Protocol Errors table already published HTTP 409 for the mismatch case pre-PR, (b) industry convention is to reject payload mismatch, and (c) the pre-PR contract was ambiguous, not endorsing, on the mismatch case.

I think we can+should land this as non-breaking tightening. 

---
## Checklist

### DIFF
--- a/docs/specification/signatures.md
+++ b/docs/specification/signatures.md
@@ -483,8 +483,21 @@ Signature: sig1=:MEUCIQD...:
 | **Entropy** | Minimum 128 bits (e.g., UUID v4, 22+ char alphanumeric) |
 | **Uniqueness** | Per-client, per-operation type |
 | **Server storage** | Minimum 24 hours, recommended 48 hours |
-| **On duplicate** | Return cached response, do not re-execute |
+| **On duplicate (matching payload)** | Return cached response, do not re-execute |
+| **On duplicate (mismatched payload)** | Reject with `409 Conflict` (REST) / `-32000` (MCP); do not execute |
 | **On storage failure** | Fail closed (reject request with 503) |
+
+**Payload Matching:** Businesses **MUST** detect whether the payload of a
+duplicate-key request matches the payload of the original. Because
+`Content-Digest` (RFC 9530) is a SHA-256 hash over the raw body bytes and
+is already required on any signed request with a body, businesses
+**SHOULD** persist the `Content-Digest` alongside the idempotency key and
+compare on duplicate-key requests. For unsigned requests, businesses
+**MUST** use an equivalent deterministic payload-comparison mechanism
+(e.g., hashing the raw body). Platforms therefore **MUST** generate a
+fresh idempotency key whenever they modify the request payload —
+including retries with modified payment instruments, updated shipping
+addresses, swapped line items, or any other change to the request body.
 
 **Note:** The RFC 9421 `created` parameter is **OPTIONAL**. UCP handles replay
 protection at the business layer through idempotency keys, not signature timestamps.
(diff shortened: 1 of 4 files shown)


docs/specification/signatures.md
```diff
@@ -483,9 +483,22 @@ Signature: sig1=:MEUCIQD...:
...
 | **On storage failure** | Fail closed (reject request with 503) |
 
+**Payload Matching:** Businesses **MUST** detect whether the payload of a
+duplicate-key request matches the payload of the original. Because
+`Content-Digest` (RFC 9530) is a SHA-256 hash over the raw body bytes and
+is already required on any signed request with a body, businesses
+**SHOULD** persist the `Content-Digest` alongside the idempotency key and
+compare on duplicate-key requests. For unsigned requests, businesses
+**MUST** use an equivalent deterministic payload-comparison mechanism
+(e.g., hashing the raw body). Platforms therefore **MUST** generate a
```
amithanda: Comparing the raw body bytes or raw body hash can be highly fragile for unsigned requests. Intermediate proxies, API gateways, or client-side JSON encoders often re-serialize JSON payloads, altering insignificant whitespace or key order. This can trigger false `409 Conflict` (payload mismatch) errors on legitimate retries.
Should we update the "Payload Matching" section to recommend semantic JSON comparison or server-side canonicalization:
```markdown
**Payload Matching:** Businesses **MUST** detect whether the payload of a
duplicate-key request matches the payload of the original. Because
`Content-Digest` (RFC 9530) is a SHA-256 hash over the raw body bytes and
is already required on any signed request with a body, businesses
**SHOULD** persist the `Content-Digest` alongside the idempotency key and
compare on duplicate-key requests. For unsigned requests, businesses
**MUST** use a deterministic payload-comparison mechanism. Comparing raw
body bytes or raw body hashes is discouraged as differences in whitespace
or key ordering (e.g., due to client-side library changes or intermediate
proxies) can cause false conflicts. Instead, businesses **SHOULD** parse
the JSON payloads and perform a semantic comparison, or use JSON
Canonicalization Scheme (JCS) per RFC 8785 to generate a deterministic
hash for comparison. Platforms therefore **MUST** generate a
fresh idempotency key whenever they modify the request payload...

amithanda (approved):
Added a minor suggestion, but looks good overall.

raginpirate (approved):

docs/specification/signatures.md
igrigorik: Good prompt, re-reading the copy.. It introduced an unnecessary branch. 

We already spec and require Content-Digest signing over raw body, which directly addresses the problem you're pointing to. Fixed the text to apply same logic here.

PR #340, opened 2026-04-08
feat: loyalty extension for catalog/cart/checkout capability

ziwuzhou-google:
# Summary

A new loyalty extension is proposed to address a critical trust and transparency milestone: ensuring existing loyalty members can seamlessly access their benefits during an agentic cart and checkout experience. By enabling buyers to see their specific tier and eligible rewards before finalizing a purchase, we address a foundational expectation for program members and remove friction from the checkout funnel.

# Motivation

Loyalty is a core concept in Commerce, serving as a primary driver for customer retention and long-term business growth. The value of a program is usually realized at the point of sale. UCP in its current format can provide some support for loyalty, mostly in the checkout capability, in the format of auto-applied monetary member-only benefits such as member price discount and/or cheaper and faster fulfillment. However, loyalty is far more than immediate-value monetary benefits. Fungible delayed-value rewards (e.g. points, miles, cashbacks), for example, is another key motivator, and in certain verticals the single most important motivator, in transactional activity. Buyers expect full visibility into their member-exclusive perks before committing to a transaction. Providing this information via an agent-to-business inquiry builds user trust and ensures that the convenience of an AI-driven checkout does not come at the expense of earned member benefits.

# Proposal Scope

Different than [PR #251](https://github.com/Universal-Commerce-Protocol/ucp/pull/251) that aims to provide an all-inclusive schema and support a wide range of use cases, this PR focuses specifically on a baseline version, although still general enough to allow future expansion, that can be used to decorate checkout capability to provide extra loyalty information that enhances the experience. More advanced use cases such as loyalty relationship management (e.g. sign-up, tier upgrade/downgrade) and loyalty rewards transfer/redemption are out of the scope of this proposal.

Specifically the following core use cases of benefit recognition for known members are addressed with the proposal:
 - Price-Impacting Benefits: Real-time application of member-only discounts and free shipping offers
 - Non-Price Benefits: Transparent display of rewards earned or rewards applicable to future purchases
 - Status Recognition: Verification and display of the buyers’ specific loyalty tier

# Design Details

## Core concepts and hierarchy

Four core concepts are introduced in the schema and structured in a hierarchy:

 - Tracks: Distinct enrollment pathways or program categories that a user can join independently or simultaneously
 - Tiers: Specific achievement ranks or status milestones within a track that unlock escalating value as a member progresses through activity or spend
 - Benefits: ongoing perks and privileges granted to a customer based on their current tier or membership status
[...]

PR #507, opened 2026-06-11
feat!: fulfillment methods on catalog

igrigorik:
Extends the fulfillment extension (`dev.ucp.shopping.fulfillment`) onto the **catalog** capability (`catalog.search`, `catalog.lookup`, `get_product`) so a variant can advertise *how it can be fulfilled* at discovery — the same `fulfillment.methods[]` a buyer/agent reasons about at checkout, applied to a second surface. (Also touches the shared extension: option base, open `type`, config.)

## Response — `fulfillment.methods[]` on the variant

```json
{
  "products": [
    {
      "id": "prod_kettle",
      "title": "Electric Kettle",
       ...,
      "variants": [
        {
          "id": "var_ss",
          "availability": { "available": true, "status": "in_stock" },
          "fulfillment": {
            "methods": [
              {
                "type": "shipping",
                "description": { "plain": "Ships to your address in 1–4 business days" },
                "availability": { "available": true, "status": "in_stock" },
                "options": [
                  { "id": "std", "title": "Standard", "description": { "plain": "Arrives in 4 business days" } },
                  { "id": "exp", "title": "Express",  "description": { "plain": "Next business day" } }
                ]
              },
              {
                "type": "pickup",
                "description": { "plain": "Pickup today at Downtown Store" },
                "location": "loc_downtown",   <-- optional ID key for upcoming location capability
                "availability": { "available": true, "status": "in_stock" }
              }
            ]
          }
        }
      ]
    }
  ]
}
```

- **`type`** — open string; well-known `shipping`, `pickup`, `curbside`. Additive: platforms present every method regardless of `type` (recognizing one only adds optional UX), so a new type needs no extension. A method is identified by `type` + fulfillment scope — same-scope variation (Standard vs Express) is `options`, distinct shipments are separate methods.
- **`options[]`** — optional, a *selective preview* (not exhaustive): without a destination/full cart, catalog previews boundary choices (cheapest, fastest); the full set is negotiated at cart/checkout. Each option is the shared `fulfillment_option_base` (`id`, `title`, `description`) that checkout composes with cost/timing (`totals`, carrier, times). Cost/time live on an option, never the method. A discovered `option.id` SHOULD be accepted as `selected_option_id` in cart/checkout (best-effort — a single-product option may differ in a multi-product cart).
- **`availability`** — for the **single** in-scope location (the one in `fulfills_to`, or resolved from `context`). A method's own `availability` is authoritative for that method; variant-level `availability` is the roll-up across methods. Availability *across* locations ("which stores have this") is the upcoming locations capability.
- **`location`** — for place-based methods, a stable location id. A business advertising pickup there MUST accept the same id at cart/checkout as `selected_destination_id`, so a discovered location round-trips into the order.

## Request — `context` vs `filters`

```json
{
  "query": "electric kettle",
  "context": { "address_country": "CA", "address_region": "ON" },
  "filters": {
    "fulfills_to": { "address_country": "CA", "postal_code": "M5B 2H1" },
    "methods": ["pickup"]
  }
}
```

- **`context`** — the *anchor*: coarse buyer locale (country/region/postal); a non-binding hint availability is reported against.
- **`filters.fulfills_to`** — the *target*: where the order is *fulfilled to* — a `location` id (store / saved address) **or** a coarse locality (may differ from `context`, e.g. a gift). Authoritative; supersedes `context`. Shares the `locality` type with `context`; no full street address / PII.
- **`filters.methods`** — restrict to method types (e.g. `["pickup"]`).

Single-market by design: one coarse destination keeps the response single-market (one currency / price / availability). Multi-destination or cross-market queries are separate parallel requests.

## Breaking changes

Two breaking changes land on the existing **checkout** `fulfillment` types (shared via the new `fulfillment_option_base`): `fulfillment_option.description` becomes the structured `description.json` object — a bare string no longer validates (wrap as `{ "plain": "…" }`) — and `fulfillment_method.type` drops its `["shipping", "pickup"]` enum for an open string.

## Notable design points

- **Open `type` everywhere.** `fulfillment_method.type`, `catalog_fulfillment_method.type`, `fulfillment_available_method.type`, and `expectation.method_type` are all open strings (the closed `["shipping","pickup"]` enums are gone). The old closed-enum "Adding New Methods" mechanism wasn't deliverable — `allOf` *intersects* enums (can't widen one). New types are advertisable, filterable, and checkout-negotiable with **no** extension — at the cost of not being backward-compatible for strict closed-enum consumers.
- **Shared option base.** Extracted `fulfillment_option_base` (`id`, `title`, structured `description`); checkout's `fulfillment_option` composes it via `allOf` + cost/timing, and its `description` is upgraded from a flat string to the structured `description.json` (examples updated).
- **Open, extensible config.** Business/merchant fulfillment config carries `multi_destination` (array of `{ method, …constraints }`; listing a method permits it, the object holds optional per-method constraints) and `method_combinations` (permitted method-type sets per cart). A third party appends a new method type — with constraints — with no protocol rev. Both replace the earlier closed `additionalProperties: false` maps.
- **Per-operation contracts preserved.** `get_product`/`lookup` compose fulfillment onto `detail_product`/`lookup_variant`, keeping `selected`/`options` and required `inputs`.


source/schemas/shopping/types/catalog_destination.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/catalog_destination.json",
+  "title": "Catalog Destination",
+  "description": "A buyer-visible destination for a fulfillment method, with availability.",
+  "type": "object",
+  "required": ["id"],
+  "additionalProperties": true,
+  "properties": {
```
maximenajim: The one area where I think this gets a little constrained is that the “how” of fulfillment — the method — and the “where-kind” of fulfillment — store, locker, partner counter, or the buyer’s own address — can vary independently.

One possible lightweight option is to add a single optional `destination_type` field:

```json
"destination_type": {
  "type": "string",
  "description": "Kind of place this destination is. Open enum; well-known values: `store`, `locker`, `pickup_point`, `address`. Consumers MAY ignore unknown values.",
  "enum": ["store", "locker", "pickup_point", "address"]
}
```

The nice thing about this is that simpler platforms can ignore it and continue to work with a single destination shape. At the same time, it gives us a clean way to distinguish a locker handoff from a store counter or partner pickup point, which is important for flows like Drive Up, in-store pickup, and partner pickup.

As a bonus, this could let us retire the `id == "address"` sentinel, since `destination_type: "address"` would carry the meaning currently being overloaded into the magic ID.

source/schemas/shopping/types/fulfillment_method.json
```diff
@@ -14,8 +14,7 @@
     },
     "type": {
       "type": "string",
-      "enum": ["shipping", "pickup"],
```
maximenajim, line 17: Dropping the closed `["shipping","pickup"]` enum to a bare string isn't backward-compatible for existing strict consumers: a consumer still validating against the closed checkout schema will reject any payload carrying a new `type` value a producer sends.  I'm not against opening it, but if we do it should ride a `feat!:` and a `dev.ucp.shopping.fulfillment` version bump rather than land as additive.

source/schemas/shopping/fulfillment.json
maximenajim, line 230: One concrete use case I want to make sure this design still supports is the Target.com experience today, where **Pickup** and **Drive Up** are search filters. A guest can narrow the entire result set to items they can collect this afternoon. I think there are two gaps in the current shape that make that hard to express on this surface.

First, `fulfills_to` filters by **where** — an address — but we don’t currently have a way to filter by **how**. In other words, there isn’t a server-side way to ask for “only items available for pickup.” The response does advertise fulfillment methods per variant, so a platform could theoretically over-fetch and post-filter client-side. But at search scale, that is not really a substitute for letting the business restrict and rank the result set server-side.

Second, **Drive Up** is effectively curbside fulfillment, and `curbside` is not currently part of the well-known vocabulary. Under the “consumers may ignore unknown values” rule, if we mint that value ourselves, it may disappear on platforms that don’t recognize it. That would make one of our highest-volume fulfillment methods invisible at discovery time.

I think the fix can be small and consistent with the pattern this PR already establishes: add an optional `method_types[]` array to `fulfillment_search_filters`, next to `fulfills_to`.  That, combined with promoting `curbside` to a well-known value through the RFC path, would make today’s Target.com filter rail expressible end to end. Happy to bring the exact schema text if helpful.

source/schemas/shopping/fulfillment.json
```diff
@@ -20,6 +20,135 @@
...
+      "$defs": {
+        "lookup_request": { "$ref": "#/$defs/fulfillment_lookup_request" },
+        "lookup_response": { "$ref": "#/$defs/fulfillment_lookup_response" },
+        "get_product_request": { "$ref": "#/$defs/fulfillment_get_product_request" },
+        "get_product_response": { "$ref": "#/$defs/fulfillment_get_product_response" }
+      }
+    },
     "dev.ucp.shopping.checkout": {
       "title": "Checkout with Fulfillment",
       "description": "Checkout extended with hierarchical fulfillment.",
```
maximenajim, line 209: One thing I wanted to check on the `fulfillment_get_product_response` composition: it currently composes onto `catalog_lookup.json#/$defs/get_product_response` and then overrides `product` with `fulfillment_product`.

The part that gives me pause is that `fulfillment_product` extends the plain `types/product.json`, while the base `get_product_response.product` is `#/$defs/detail_product`. My understanding is that `detail_product` is where we carry `selected` and the per-option availability signals used for interactive variant narrowing on a PDP. By swapping in a product shape that extends the bare product type, the composed fulfillment response seems to lose that richer detail-product contract.

It still validates today because `additionalProperties: true` allows those extra fields to pass through untyped, so CI would not catch the difference. But the contract becomes a little weaker than what the base `get_product_response` provides.

I noticed Loyalty (#340) avoided a similar issue by adding `loyalty` as a sibling field rather than re-overriding `product`. Fulfillment is a little different because the enrichment hangs off the variant, so we probably cannot use that exact shape. But I think the same principle could apply: compose the fulfillment enrichment onto `detail_product` rather than onto the bare `product.json`.

Concretely, maybe the get-product path gets its own product definition:

```json
"fulfillment_detail_product": {
  "allOf": [
    { "$ref": "catalog_lookup.json#/$defs/detail_product" },
    {
      "type": "object",
      "properties": {
        "variants": {
          "type": "array",
          "items": {
            "$ref": "#/$defs/fulfillment_variant"
          }
        }
      }
    }
  ]
}
```

Then `fulfillment_get_product_response.product` could point at `fulfillment_detail_product`.

Was extending the bare `product.json` here intentional, or was it mostly the convenient base to reach for while adding the fulfillment fields?

source/schemas/shopping/fulfillment.json
```diff
@@ -20,6 +20,135 @@
...
+      ]
+    },
+    "fulfillment_search_filters": {
+      "description": "Catalog filters extended with a fulfillment destination filter.",
+      "allOf": [
+        { "$ref": "types/search_filters.json" },
+        {
+          "type": "object",
+          "properties": {
+            "fulfills_to": {
```
maximenajim, line 142: Concrete use case I want to make sure survives this design: on Target.com today, **Pickup** and **Drive Up** are search filters — a buyer narrows the entire result set to what they can collect this afternoon. Two things in the current shape prevent building that rail on this surface.

First, `fulfills_to` filters by *where* (an address), but nothing filters by *how* — there's no way to ask "only items available for pickup." The response advertises methods per variant, so a platform could over-fetch and post-filter client-side, but at search scale that's not a substitute for the business restricting and ranking server-side. Second, Drive Up is curbside, and curbside isn't in the well-known vocabulary — under the consumers-ignore-unknown rule, a value we mint ourselves just disappears on platforms that don't recognize it, so one of our highest-volume fulfillment methods becomes invisible at discovery time.

The fix is small and follows the pattern this PR already established: an optional `method_types[]` array right here in `fulfillment_search_filters`, next to `fulfills_to`. We've been prototyping exactly that field (enum'd to the well-known method values) and it composes from inside the extension the same way. That, plus promoting `curbside` to a well-known value via the RFC path, makes today's Target.com filter rail expressible end to end. Happy to bring the exact schema text.

docs/specification/fulfillment.md
```diff
@@ -280,11 +289,153 @@ If the buyer chooses pickup but the platform doesn't support split
...
+    buyer's address, supplied on the request); otherwise it is the store or
+    location id.
+* `name` and `address` — the location's name and address, when it has one.
+* `availability` — whether the variant is available via this method at this
+    destination.
+
+A method's `type` tells you what kind of destination it carries. Catalog
+only advertises these options; the buyer commits at checkout. A destination
+`id` found here can be passed to cart or checkout as
+`selected_destination_id`, though checkout can also present its own
```
maximenajim: This line tells a buyer's agent it can carry a discovered destination `id` into checkout as `selected_destination_id` — but the two sides don't share a destination type. Catalog destinations are `catalog_destination.json` (the flat `{id, name?, address?, availability}` introduced here), while checkout's `fulfillment_method.destinations[]` are `fulfillment_destination.json` (the `oneOf[shipping_destination, retail_location]` we already have and enrich in #375), and `selected_destination_id` resolves against *that* set. So "pass the catalog id to checkout" is an unbacked round-trip: it works only as long as both sides happen to mint ids the same way, and nothing in the protocol says they must.

This is adjacent to the `destination_type` thread I already opened but distinct from it — that thread is about modeling where-kind; this is about whether the catalog id is the *same identifier* checkout will accept. For Target the concrete failure is an agent showing "in stock for pickup at store 1492," binding `selected_destination_id: "1492"` at checkout, and checkout not recognizing it because catalog's id space was never promised to be checkout's. I'd rather catalog's destination be a documented projection/subset of `fulfillment_destination` (shared id semantics, shared discriminator) than an independent sibling — then this sentence is true by construction. If the intent is genuinely that ids are opaque and *not* guaranteed to round-trip, the spec should say that instead, because the current wording invites integrators to rely on it.

source/schemas/shopping/fulfillment.json
maximenajim: Also, `fulfills_to` reuses the full `postal_address` type, which allows `street_address`, `extended_address`, `first_name`, `last_name`, and `phone_number` in a catalog discovery request. That is too much PII for pre-checkout product search, especially when autonomous agents may query many merchants before user selection.

source/schemas/shopping/fulfillment.json
igrigorik: good catch, fixed.

docs/specification/fulfillment.md
igrigorik: clarified

source/schemas/shopping/fulfillment.json
igrigorik: refactored 👍

igrigorik:
Updated the draft to capture feedback from TC review - see https://github.com/Universal-Commerce-Protocol/ucp/pull/507/commits/f607ce2c6e3eceaa6c150d91abb1567c99650636. Key changes...

Each catalog variant carries a `fulfillment.methods[]`. Each method states how the variant can be fulfilled (`type`), its `availability`, and — for place-based methods like pickup — the resolved `location` id. The per-method `destinations[]` array is gone: location facts (name, address, hours) and cross-store comparison will move to a separate ~locations capability, and catalog just references a location by id. This yields...

Catalog response (method-first):

```json
{
  "products": [
    {
      "id": "prod_kettle",
      "title": "Electric Kettle",
      "variants": [
        {
          "id": "var_ss",
          "title": "Stainless Steel",
          "availability": { "available": true, "status": "in_stock" },
          "fulfillment": {
            "methods": [
              {
                "type": "shipping",
                "description": { "plain": "Ships to your address" },
                "availability": { "available": true, "status": "in_stock" }
              },
              {
                "type": "pickup",
                "description": { "plain": "Pickup today at Downtown Store" },
                "location": "loc_downtown",
                "availability": { "available": true, "status": "in_stock" }
              }
            ]
          }
        }
      ]
    }
  ]
}
```

Request (filter to pickup near a place):

```json
{
  "query": "kettle",
  "context": { "postal_code": "94043" },
  "filters": {
    "fulfills_to": { "location": "loc_downtown" },
    "method": ["pickup"]
  }
}
```

---
Related context...

### Method taxonomy

We need to model fulfillment method types, and there are real sub-variations regardless of shape (curbside, locker, same-day). Three shapes were on the table: collapse to two classes ("comes to you" / "you go to it") with variants nested underneath; a flat list of peer types; or a nested class → sub-type tree.

**Discussion:** a taxonomy only earns structural nesting if a consumer acts on that structure, and the consumer actions are *filter*, *render*, and *group*. A flat `type` makes filter and render trivial (`type=pickup`, switch on type); the two-class nest's only advantage is making the come-to-me / go-to-it grouping explicit — one derived view — at the cost of routing every specific filter through two levels and pushing variant filtering down into metadata. Separately, "what a buyer must *do* to receive an order" (a code, a signature, an age check) is orthogonal to the type axis.

**Rec:** with above in mind, my intuition is to keep `type` **flat** and well-known-keyed.

### Location data in catalog

The draft embeds a `destinations[]` array on each method, each destination carrying its facts (name, address) plus availability. We discussed whether the catalog response should carry location facts at all, or just availability. To be clear: the ability to look up locations and answer "which locations offer this product" *is* important, and we agreed it can/should be addressed separately as part of a locations capability.

**Discussion:** catalog is product-centric and returns many products at once, so embedding N locations' facts per method per variant explodes a broad search into a products × stores matrix. The buyer's discovery question is narrower — "can I get this at the place I care about?" — i.e. one location. Location facts (hours, address) and cross-store comparison are a distinct, location-centric concern with their own lifecycle, better served by a dedicated capability that catalog references by id.

**Rec:** catalog answers availability for the single location the request targets (passed via `fulfills_to`, or inferred from `context`) and references it by `location` id; drop the embedded facts and the `destinations[]` array.

### Availability + status shape

The draft models availability as `{ available, status }`, which yields confusing combinations (e.g. `available: true` + `status: out_of_stock`). Thinking about this more, I think we conflated two distinct questions: "is it ready now?" and "can I purchase it?". An item can be purchasable but not immediately fulfillable (preorder, backorder).

Note that `{ available, status }` is **not introduced by this PR**. It's already a thing on Catalog variant ([variant.json](https://github.com/Universal-Commerce-Protocol/ucp/blob/b66d781e1948b70fa9a306b6feb571171da822b3/source/schemas/shopping/types/variant.json#L106-L119)); this PR only extracted it into a shared type and reused it on the fulfillment method. So this conversation surfaced a *pre-existing* ambiguity we need to work through in the core variant contract, not a new one.

source/schemas/shopping/fulfillment.json
igrigorik: Good points. Addressed in updated draft.

igrigorik:
As followup to the TC deep-dive yesterday, ran through an exercise of modeling "flat vs nested". Upfront disclaimer: the shapes below are deliberately minimal. The decision is the **nesting**, not the option's field set. An "option" is an open object (at least `id` + `description`) that can carry cost, delivery-time estimate, free-shipping threshold, etc. 

### A: nested `options[]` 
Each option carries its own cost/time/threshold fields; a method with several choices lists several options.

```json
"fulfillment": {
  "methods": [
    {
      "type": "shipping",
      "description": { "plain": "Ships in 2–4 business days" },
      "availability": { "available": true, "status": "in_stock" },
      "options": [
        { "id": "std", "description": { "plain": "Standard" } },
        { "id": "exp", "description": { "plain": "Express" } }
      ]
    },
    {
      "type": "pickup",
      "location": "loc_downtown",
      "description": { "plain": "Pickup today at Downtown" },
      "availability": { "available": true, "status": "in_stock" },
      "options": [
        { "id": "today", "description": { "plain": "Ready today" } }
      ]
    }
  ]
}
```

### B: flat methods / multiple same-`type` entries
Each entry carries its own cost/time fields directly; same-`type` choices (Standard / Express) become peer entries. Flat also needs an `id` on every method entry and a normative "platforms group same-`type` entries" rule.

```json
"fulfillment": {
  "methods": [
    { "id": "ship_std",    "type": "shipping", "description": { "plain": "Standard shipping" }, "availability": { "available": true, "status": "in_stock" } },
    { "id": "ship_exp",    "type": "shipping", "description": { "plain": "Express shipping" },  "availability": { "available": true, "status": "in_stock" } },
    { "id": "pickup_today", "type": "pickup", "location": "loc_downtown", "description": { "plain": "Pickup today" }, "availability": { "available": true, "status": "in_stock" } }
  ]
}
```

## Recommendation

### 1. nested offers simple contract that lets every client render a good+consistent presentation
Presentation structure and ordering is an important, non-trivial, and business-specific. We allow and enforce this in many places in UCP by leaning into "the server precomputes the structure, the client renders as presented". Nested structure encodes both the order and logical grouping: a client walks `methods[]`, renders each as a section (header = `description` + `availability`), and lists `options[]` beneath — getting "Shipping: Standard / Express" and "Pickup: today" for free. 

Flat delivers the same data as peer rows (Standard shipping / Express shipping / Pickup today) and makes *every* client group by `type` to render well. Further, this would require "platforms SHOULD group same-`type` entries" rule, which pushes presentation work onto every consumer, and does not allow consistency or business-preference for how things ought to be presented or grouped -- e.g. you lose section header, which client has to compute itself. This is a bad outcome / pattern for everyone: more work for client, uncertain presentation and ordering for business and user.

### 2. Checkout/cart symmetry continuity via `options`
We discussed alternative names (e.g. `projections`), but on further inspection I would recommend we lean on existing `options` and mirror structure of cart/checkout. In effect, catalog is `method → options`, which is a lean subset of checkout's `method → groups → options`. One important caveat, which we can handle as normative text and via schema descriptions is that catalog options are representative, not exhaustive -- we already have similar contract for variants in search, and we don't invent a new name for it.

### 3. `options` is optional
`options` is optional; a method may carry zero options, surfacing `type` + `description` + `availability` only.

docs/specification/fulfillment.md
```diff
@@ -313,11 +313,34 @@ method has:
...
+Each option carries an `id` and a `title` (a short label distinguishing it
+from siblings), plus an optional renderable `description` for context. These
+are a shared base: at checkout the same option is composed with cost and
+timing (`totals`, carrier, fulfillment times). The option is open, so a
+business MAY annotate it with additional fields. A method MAY also carry
+none, surfacing only `type`, `description`, and `availability`; options are
+nested directly under the method, with no group layer (unlike checkout
+`methods[].groups[].options[]`).
+
+A discovered option `id` is the unit checkout selects: a business that
```
richmolj: This wouldn't be true for us even for cart -> checkout. It's worse for catalog, since the option is more of a prediction.

igrigorik: [8c3f0e1](https://github.com/Universal-Commerce-Protocol/ucp/pull/507/commits/8c3f0e196f0c7390ec2957ec5be81ce67ac40f7c)

docs/specification/fulfillment.md
```diff
@@ -280,11 +290,180 @@ If the buyer chooses pickup but the platform doesn't support split
...
+
+`fulfillment.methods[]` lists the methods available for a variant. Each
+method has:
+
+* `type` — the fulfillment method (e.g. `shipping`, `pickup`); see
+    [Method Types](#method-types).
+* `description` — short, buyer-facing summary of how the variant is
+    fulfilled via this method (e.g. "Ships in 2–4 business days"). Directly
+    renderable; see [Rendering](#rendering).
+* `availability` — whether the variant is available via this method at the
```
maximenajim, line 312: This might not be a big issues but this PR puts a per-method `availability` next to the variant-level `availability` (`variant.json:106`), and after this round both resolve to the *same* `availability.json` type — so they're now structurally indistinguishable, with no stated precedence when they disagree. One possible normative sentence on the method `availability` bullet:

> Per-method `availability` is authoritative for that method at the resolved `location`. The variant-level `availability` is the global roll-up across methods and locations; when a method states `availability`, consumers **MUST** use it for that method and **MUST NOT** infer per-method availability from the variant-level value.

amithanda:
I stress-tested the two shapes against how this actually renders at scale 
(Walmart / Target / Kohls PDPs) and against a small set of invariants, and I agree 
that A seems cleaner. 

A few additions to the case:

**Lead reason: the transform is asymmetric.** A to B (flatten into peer rows) is
deterministic and lossless. B to A (regroup a flat list into methods) means grouping
on `type`, which is open in this PR, so a consumer seeing two same-`type` entries
with an unknown value cannot tell "two options of one method" from "two distinct
methods." You can always derive flat from nested, never reliably the reverse, so the
source of truth should be the more structured shape: every consumer can project its
own view out of A, none can rebuild A from B.

**Real-world shape.** Both Walmart and Target present a per-type rail (Shipping /
Pickup / Delivery), each card with one summary and one availability state, speed
tiers as a drill-down (Walmart even shows "Express in 19 min" as a refinement under
the selected Delivery card, not a fourth peer). That is A directly. B hands the
platform "Standard shipping" and "Express shipping" as peers and makes it dedupe them
back into one card before it can draw the rail every major retailer ships. Same with
availability: retailers state it per type ("Shipping: Not available"), which is where
A puts it; B scatters it across rows and forces an OR-rollup.

**Two different kinds of variation, and they should not be merged.** A `shipping`
method can have more than one entry for two completely different reasons:

1. **Different speeds of the same shipment:** Standard (4 days) vs Express (next
   day). One shipment, the buyer is just choosing a service level.
2. **Genuinely different shipments:** ship from the local store today vs from the
   warehouse in two days (or ship to two different addresses). These are separate
   fulfillment paths, not speeds of one path. The spec already allows multiple
   same-`type` methods for this case.

A consumer often needs to tell these apart. Option A keeps them in two different
places: `options[]` inside a method for case 1, and separate `methods[]` entries for
case 2, so the structure itself says which is which. Option B puts everything in one
flat list, so two `shipping` entries are ambiguous: the consumer cannot tell whether
they are two speeds of one shipment (which it should group under a single card) or two
separate shipments (which it should show as distinct choices).

**Small reframe on point 1.** I'd be a little cautious leaning on "business precomputes
structure so clients render consistently" as the justification. On a multi-merchant
surface the platform legitimately wants its own consistent ordering across merchants,
so the business does not really get to dictate final presentation there. The framing
that holds up: the business discloses *structure* (which option belongs to which
method, what is available how) that only it knows and the platform cannot reconstruct,
and the platform owns *presentation* (order, grouping into a rail, collapse to
"from $5", voice vs cards). A serves exactly that split, so the case for A is actually
stronger than the consistency argument suggests.

source/schemas/shopping/types/business_fulfillment_config.json
```diff
@@ -18,7 +18,7 @@
           "description": "Multiple pickup locations allowed."
         }
       },
-      "additionalProperties": false
+      "additionalProperties": true
```
maximenajim: Opening the method-type map keys is necessary, but `additionalProperties: true` also opens the values to any JSON type. A profile like `"allows_multi_destination": { "curbside": "yes" }` now validates, and platforms will diverge between rejecting it, treating it as truthy, or treating it as absent.

This is a negotiation contract, so the value type needs to stay boolean even while the keys are open. Please change this map to typed additional properties and mirror the same fix in `merchant_fulfillment_config.json`:

```json
"additionalProperties": {
  "type": "boolean",
  "description": "Whether multiple destinations are allowed for this fulfillment method type."
}
```

docs/specification/fulfillment.md
igrigorik: good catch, clarified.

igrigorik:
@amithanda @maximenajim good feedback, updated. 

@maximenajim @richmolj ptal and sanity check https://github.com/Universal-Commerce-Protocol/ucp/pull/507/changes/7c319a9cf37108d32bc880f812cb7c16e8206c6c -- I'm assuming we want to retain per-type object/flexibility. If we think it's only ever a boolean, then we can collapse `multi_destination` to an array and use same contract as `method_combinations`.

p.s. validated above shape against 3P extension, which can now target & extend this structure: https://gist.github.com/igrigorik/1e58bf1fc6ab46d92c04c9a2beb666b7

richmolj (commented):
Missed a couple spots opening the type. `method_combinations` and the config keys are now open, but two schemas still pin the type to a closed enum: `fulfillment_available_method.type` (`["shipping","pickup"]`) and `expectation.method` (`["shipping","pickup","digital"]`).

(*NB: this reminded me that an order details page for a partner still wants to put `digital` somewhere even if no `fulfillment` extension, hence `expectation.method`*)

Nit: since we're already making a breaking change, maybe rename `platform_fulfillment_config.supports_multi_group` -> `allows_multi_group` to match the `allow`-verb vocabulary.

maximenajim:
Thanks for the quick restructure — looks good.

The one thing I'd change is dropping the required `allowed` boolean. It's redundant with the rule you already wrote in the field description — "a type that is absent is not permitted." Once absence means "not permitted," presence already means "permitted," and `allowed` is a second way to say the same thing. Worse, `allowed: false` *contradicts* that rule: `{ "shipping": { "allowed": false } }` and omitting `shipping` entirely both mean "shipping multi-destination not allowed," and a consumer now has two encodings to reconcile. That two-ways-to-express-one-state ambiguity is the same class of problem as the original type-hole — it just moved from the value type to the value semantics.

So my recommendation: presence-of-key = permitted; the object carries optional per-type *constraints* (which is where your 3P fields live); `{ "shipping": {} }` = allowed with no constraints. Single source of truth, still fully extensible. Concretely:

```json
"multi_destination": {
  "type": "object",
  "description": "Per-method-type multi-destination config, keyed by fulfillment method `type`. A type that is present is permitted; a type that is absent is not. The object carries optional per-type constraints. Open — businesses MAY add any method type; well-known keys are `shipping` and `pickup`.",
  "additionalProperties": {
    "type": "object",
    "description": "Optional per-type multi-destination constraints. An empty object means permitted with no additional constraints.",
    "additionalProperties": true
  }
}
```

One bookkeeping note: this shape lives byte-identically in both `business_fulfillment_config.json` and `merchant_fulfillment_config.json`, so whatever we land has to go in twice.

igrigorik:
@maximenajim good point. Hmm, the empty object is the same ~anti-pattern we struggled with and avoided in loyalty. As an alternative, we can model it as an array: https://github.com/Universal-Commerce-Protocol/ucp/pull/507/changes/172c6521ea5e2dc0c2bf6380894690f5e779b011 — wdyt? 

@richmolj ☝🏻

richmolj (approved):

maximenajim (approved):

amithanda:
@igrigorik - I am ready to approve.

One thing I want to highlight: this PR also makes a breaking change to the **checkout** surface, 
which the title ("fulfillment methods on catalog") doesn't signal and is quite easy to miss.

Specifically:

- `fulfillment_option.description` goes from `string` to the `description.json` object,
  so existing checkout producers sending a string `description` no longer validate.
- `fulfillment_method.type` drops its closed enum to an open string on checkout too.

The `!` is correctly in the PR title, so the breaking flag is there. The gap is that the
human-readable scope says "catalog" while the actual break lands on checkout, which is
easy to miss in review and in the release notes. Could we do one of:

- **Split** the checkout-surface changes (`description`, `type`) into their own `feat!`
  so they are reviewed and announced as a checkout break in their own right, leaving this
  PR additive on catalog; or
- **Keep them here but make it explicit:** update the PR description and any other necessary heads up 
  to call out the checkout impact with before/after, so integrators on that surface aren't surprised and we 
  don't miss it in our release notes.

Either works for me. I just want the checkout break to be visible rather than implicit.

igrigorik:
@amithanda good call, updated description 👍

amithanda (approved):

PR #508, opened 2026-06-11
feat: define recipient-required gift checkout

seungy0:
## Summary

Adds `dev.ucp.shopping.gift`, a Checkout extension for purchases sent to another person.

A Platform starts a gift checkout with:

```json
{
  "gift": {
    "recipient_required": true
  }
}
```

`gift: {}` has no effect. With `recipient_required: true`, the Business must have a selected, eligible recipient before it completes checkout. The buyer chooses the recipient; the Business validates the choice.

The Business keeps that choice as state for the Checkout. UCP does not carry recipient IDs, selection tokens, friend lists, contact data, or addresses. A user-authenticated Checkout response may include a buyer-facing recipient name.

## Flow

- If recipient selection needs buyer input, the Business returns `requires_escalation` and `continue_url`.
- The buyer can sign in and select a recipient in the Business experience. Returning from that page does not prove that selection succeeded; the Platform reads the Checkout again.
- An Update must include `recipient_required: true` to keep the Gift request. Omitting it clears the Business selection.
- Complete omits `gift`. The Business revalidates the recipient before it creates a payment or order.

## Authentication

Gift does not define a login flow. The buyer may sign in to the Business during recipient selection. If the Platform needs the recipient name, it uses the existing Identity Linking flow. Opening `continue_url` does not authenticate later UCP calls by itself.

## Not included

Recipient search, reusable recipient handles, multiple recipients, messages or cards, acceptance, scheduling, and recipient data in Orders or events are out of scope.

### DIFF
--- a/docs/specification/gift.md
+++ b/docs/specification/gift.md
@@ -1,0 +1,246 @@
+<!--
+   Copyright 2026 UCP Authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+
+# Gift Extension
+
+## Overview
+
+The Gift extension represents a buyer-paid purchase for another person. The
+buyer selects and pays for the product, while the recipient is selected and
+validated by the Business.
+
+Gift purchases differ from ordinary checkout because the buyer and recipient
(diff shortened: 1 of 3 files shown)


damaz91:
Hello @seungy0, thanks for your contribution to UCP. This does not seem just a documentation PR as the title would suggest. Seems like you are trying to propose a new extension. 

This probably would entail a schema change as well. Please add it in the PR so we can move forward with the review.

Thanks.

seungy0:
Hello @damaz91,
Thanks for pointing this out. I’ve added the Gift schema and spec to the PR.
`dev.ucp.shopping.gift` extends Checkout. The Business keeps recipient selection with the Checkout, and UCP only returns a display name, not a recipient ID.
I also updated the PR description with a fuller walkthrough of the flow and what stays on the Business side.
I’d appreciate another look when you have a moment.

seungy0:
Hi @raginpirate @prasad-stripe, just checking in on this PR. I added the Gift schema and spec requested in the earlier feedback and updated the PR description with the current flow.

Could you take a look when you have a chance? Please let me know if there’s anything else I should update. Thanks!

PR #516, opened 2026-06-15
chore: validate catalog examples by op+direction, drop def=

igrigorik:
Catalog capabilities (`catalog.search`, `catalog.lookup`) carry distinct request/response shapes under `$defs/{op}_{direction}`. Example annotations named the shape twice — `op=read` **plus** `def=search_response` — which is redundant now that `ucp-schema` derives the shape from operation + direction.

## Changes

- **Annotations (22):** `op=read/create def=<shape>` → `op=search|lookup|get_product` (+ `direction`) across `docs/specification/catalog/{mcp,index,rest}.md`. `def=` dropped.
- **Harness:** the coverage check derives `$defs/{op}_{direction}` for container schemas instead of reading `def=`. Validation already flows through `ucp-schema validate` (full schema), which auto-selects the shape. `def=` stays for shapes that *aren't* an operation+direction (transport `error_response`, profile `business_schema`, sub-types).
- **Docs:** `schema-authoring.md` `$defs` section now teaches the op+direction convention, with `def=` shown for the non-operation case.

**⚠️ Depends on ucp-schema container validation**
Blocked by **Universal-Commerce-Protocol/ucp-schema#31** (https://github.com/Universal-Commerce-Protocol/ucp-schema/pull/31). The harness shells out to the installed `ucp-schema`; an older binary won't auto-select the now-`def`-less catalog shapes and would regress them to vacuous passes. **Merge only after that PR ships in a release.**

---

## Checklist

### DIFF
--- a/docs/documentation/schema-authoring.md
+++ b/docs/documentation/schema-authoring.md
@@ -651,12 +651,22 @@ shows an envelope but only a subtree is the UCP payload under validation.
 <!-- ucp:example schema=shopping/checkout op=create direction=request extract=$.params.arguments.checkout -->
 ```
 
-**Schema with `$defs`.** Some schemas (e.g. catalog) define request/response
-inside `$defs`. Use `def=` to extract and validate against the named
-definition.
+**Schema with `$defs`.** Some schemas hold several message shapes under
+`$defs`. When a capability's request and response are different objects (e.g.
+catalog: a search request is a query, a search response is a list of products),
+just name the operation and direction — the validator selects
+`$defs/{op}_{direction}` automatically:
 
 ```text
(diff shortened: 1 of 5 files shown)


igrigorik:
Marking as ready for review, but we should land this only after we release and upgrade to new version of ucp-schema. 

Note: the old ucp-schema (1.2.0) that CI installs validates the new def-less catalog examples against the body-less container root → it passes everything vacuously. So nothing fails in current build, but this is a validation regression. With new version live in CI, we bring back full validation.

cc @wry-ry

drewolson-google (approved):

wry-ry (approved):

PR #523, opened 2026-06-17
feat: shopping permalink capability

igrigorik:
Permalinks are a popular, well-established way to turn a single shareable URL into a pre-loaded cart or checkout. The link encodes the shopping intent — items, optional discounts, a destination, and some buyer prefill — directly, so a buyer taps it and lands in the right purchase flow with the right state already in place. They show up all across the funnel: email and SMS campaigns, social and influencer links, paid ads and landing pages, QR codes on print and packaging, "buy now" and share-cart buttons, bundle offers, etc. The whole point is friction reduction — replace "browse, find, add, configure" with one tap.

Most commerce platforms already offer some flavor of cart or checkout permalink, but each is platform-specific by construction: a proprietary origin, a platform-specific item-ID scheme, and a bespoke prefill vocabulary. A common shape is a cart permalink like `/cart/{item_id}:{quantity}` with a `?discount=` parameter and platform-specific prefill fields. There is no cross-merchant, agent-addressable, schema-typed equivalent — nothing an AI agent, marketplace, or campaign tool can construct once and point at any UCP business.

This capability (`dev.ucp.shopping.permalink`) generalizes the permalink pattern into the protocol. It keeps the ergonomics — compact item path, discount fields, a destination — but expresses prefill as typed UCP field paths resolved against the business's own Cart/Checkout schema, and replaces platform-specific behavior with a single redirect contract.

## Design

The capability leans on existing UCP schemas, in a future-proof way.

It defines one control parameter, `continue_to`, and zero shopping-state fields of its own. Every data-carrying query parameter is an existing UCP field path written as a JSON Pointer with the leading `/` omitted — `buyer/email`, `context/postal_code`, `discounts/codes/0`, `line_items/0/quantity` — resolved against the business's Cart or Checkout schema.

The payment-handler preference is modeled as a new core `context.payment` field — a provisional, priority-ordered buyer hint with entries such as `{ "handler": "com.adyen", "types": ["card"] }` — rather than a permalink-owned field, so it propagates to checkout, cart, and catalog through their existing `context` reference and needs no permalink-specific handling.

## Discovery

Businesses advertise `dev.ucp.shopping.permalink` with `config.endpoint`, an absolute HTTPS endpoint without userinfo, query, fragment, or trailing slash.

```json
{
  "ucp": {
    "capabilities": {
      "dev.ucp.shopping.permalink": [
        { "version": "…", "spec": "…", "schema": "…", "config": { "endpoint": "https://merchant.example/buy" } }
      ]
    }
  }
}
```

## URL shape

```
{endpoint}
{endpoint}/{items}?{query}
```

- Compact item path: comma-separated `item_id_token:quantity` pairs. A raw token is used when the identifier matches `[A-Za-z0-9._-]+`; otherwise it is `~` plus base64url-without-padding of the UTF-8 identifier, which avoids fragile percent-encoded path separators such as `%2F`.
- Query: UCP field paths, the `continue_to` control, and non-UCP parameters (preserved on the redirect by default when safe).
- No-item route: the OpenAPI `GET /` binding represents the route relative to `config.endpoint`, for example `https://merchant.example/buy/?context/postal_code=94105&continue_to=/`.

### Playground & examples: https://ucp-permalink.vercel.app

----

## Checklist


docs/specification/permalink.md
```diff
@@ -0,0 +1,702 @@
...
+If the full field path is recognized and applicable, the Business applies it to
+the initialized shopping state; if it is not recognized or not applicable, the
+Business SHOULD consume and remove it by default. A Business MAY expose selected
+field paths on the destination URL only when it explicitly selects and, where
+appropriate, rewrites them for a defined purpose, such as preserving
+attribution. A Business MUST NOT forward sensitive values.
+
+### Destination Preference (`continue_to`)
+
+`continue_to` is a destination preference, not shopping-state data. It requests
```
jamesandersen, line 264: nit: this may be personal preference / subjective ... but what about `continue_to` being a path segment?  Is the choice to make it a query param to avoid confusion about what the first additional path segment *could* be since neither `continue_to` or `items` are required?

docs/specification/permalink.md
```diff
@@ -0,0 +1,702 @@
...
+rules prevent open redirects, response-header injection, mass-assignment of
+server-owned fields, and log injection or parser confusion from control
+characters in decoded tokens and values.
+
+A permalink URL exposes the inputs it carries: the referring page's `Referer`
+may leak the URL itself, and the `303` response's `Cache-Control` and referrer
+policy (see [Redirect Resolution](#redirect-resolution)) keep it from being
+cached or forwarded to the destination.
+
+### Privacy
```
jamesandersen, line 699: 👍

source/schemas/shopping/permalink.json
```diff
@@ -0,0 +1,60 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/permalink.json",
+  "name": "dev.ucp.shopping.permalink",
+  "title": "Permalink Capability",
+  "description": "Browser-addressable shopping intent capability: defines a Business browser endpoint and redirect resolution. It defines no shopping-state fields of its own; permalink query parameters address existing UCP field paths.",
+  "$defs": {
+    "endpoint": {
+      "type": "string",
+      "pattern": "^https://[^/?#\\s\\\\@]+(?:/[^?#\\s\\\\]*[^/?#\\s\\\\])?$",
+      "description": "Absolute HTTPS browser endpoint with a non-empty authority and without userinfo, query, fragment, whitespace, backslashes, or trailing slash. Optional compact item path and query parameters are appended to this endpoint."
```
jamesandersen, line 12: I'm thinking about the case where businesses may want to direct permalinks to their apps. In combination with iOS Universal Links / Android AppLinks, the `https` requirement here should work in **most** cases.  

There are a few edge cases where a business *might* want to use a custom app scheme e.g. `merchant://`:
* *No web fallback desired* - for some reason the business only wants to open in their app
* *Embedded webviews* (Universal Links are intentionally disabled in WKWebView / SFSafariViewController on iOS)

Overall I think custom app schemes are risky (IIUC the same scheme can be claimed by multiple apps) so I wouldn't advocate supporting them.  However, we may just want to add some mention of using Universal Links / AppLinks when there is a desire to open apps (if installed)

docs/specification/permalink.md
igrigorik: Yes, we reserve path for `items`. The `{continue,redirect,...}_to` as a query string is an established and well-known pattern, which we should stick with.

source/schemas/shopping/permalink.json
igrigorik: AFAIK, handling universal links in in-app-browser is possible but — yes — requires some additional code on both Android and iOS. But even so, allowing endpoint to be an app scheme breaks fallback behavior with zero recourse for the user, and neither platform or business know if the app is available on the device. The safe behavior for a platform would be to ignore any non-HTTPS link, which is counterproductive to the very essence of what we're trying to achieve here. I would keep the `https` scheme locked.

Added note for universal links: https://github.com/Universal-Commerce-Protocol/ucp/pull/523/changes/b44f7acc3d9ba83b58d7b98f6e92affa75724855

jamesandersen (approved):
Thanks for the note on universal links

source/schemas/shopping/types/context.json
```diff
@@ -37,6 +37,14 @@
       "items": {
         "$ref": "../../common/types/reverse_domain_name.json"
       }
+    },
+    "payment_handlers": {
+      "type": "array",
+      "description": "Buyer-preferred payment handlers in priority order (most preferred first). Each provided value SHOULD map to a handler advertised in the Business profile's `ucp.payment_handlers`. The Business SHOULD use it to preselect or prioritize that handler and MAY ignore unavailable or ineligible values; unrecognized values MUST be ignored without error.",
```
raginpirate: high-level product concern here: A payment handler's objective is to spec out how to move a credential from a provider or platform owned experience OR tokenizer and into the business.

This modelling makes sense for experience strict handlers: think google pay, apple pay, an LPM like iDEAL.

Where this modelling breaks down is for a handler for a tokenizer such as [Adyen](https://docs.adyen.com/online-payments/agentic-commerce/agent-platform/agentic-payments-ucp), which supports many different instrument types for headless tokenization, where the important preference is not targeting the Adyen handler but instead targeting the Adyen handler with the card instrument, as thats the UX the platform would build. Example: Adyen handler, advertising support for cards and manual bank entry. The objective of the permalink is likely to target the manual bank entry form the platform would render, not Adyen in general.

To this end, I'd propose adding one extra optional layer to this:
`payments: [{handler: "com.adyen", instruments: ["card", "bank"]}, {handler: "com.stripe"}]`.

To reduce complexity, I am not advocating for the ability to advertise preference just by instrument type, because speaking to a preference of tokenizer -> type is always acceptable and the hierarchy should be respected. I also avoid complex expressions such as preference for Stripe over Adyen for cards but Adyen over Stripe for banks at this time; lets spec against duplicate handler keys for simplicity.

By keeping this as an array of objects (as opposed to a hash keyed by handler name) we could support the above later if we wanted to, and we allow handlers to declare their preference order without enumerating an instrument set.

igrigorik: @raginpirate addressed via https://github.com/Universal-Commerce-Protocol/ucp/pull/523/changes/61024b5a28082db0acda8dfdab493809842f7543, ptal.

source/services/shopping/permalink.openapi.json
```diff
@@ -0,0 +1,134 @@
...
+      "variables": {
+        "endpoint": {
+          "default": "https://merchant.example/buy",
+          "description": "Absolute HTTPS browser endpoint advertised in permalink capability config."
+        }
+      }
+    }
+  ],
+  "paths": {
+    "/": {
```
raginpirate, line 20: Hmm, this seems to spec {endpoint}/ as valid but we actually try disallowing any trailing slashes. Was there a strong reason to disallow those?

source/schemas/shopping/permalink.json
```diff
@@ -0,0 +1,60 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/permalink.json",
+  "name": "dev.ucp.shopping.permalink",
+  "title": "Permalink Capability",
+  "description": "Browser-addressable shopping intent capability: defines a Business browser endpoint and redirect resolution. It defines no shopping-state fields of its own; permalink query parameters address existing UCP field paths.",
+  "$defs": {
+    "endpoint": {
```
raginpirate, line 8: Should this use `"format": "uri"` like many other examples in the codebase?

raginpirate (commented):
Sorry for the slow loop. Thanks for addressing the payments concern! Just two extra little edges I spotted that are unrelated.

source/services/shopping/permalink.openapi.json
igrigorik: clarified 👍

igrigorik:
@raginpirate ty for the review, addressed your questions -- ptal.

maximenajim:
This is great. I would like a few clarifications before approval:

1. Define cart mutation/idempotency semantics. Target has a one click cart feature that adds to the active cart and increments duplicates, but the PR correctly requires public GET permalinks to be safe/idempotent under bots and prefetchers. Should the spec say how Businesses can merge, stage, or dedupe permalink-created state?
2. Add a fulfillment example equivalent to `fft=pickup&loc=1375`, preferably using the existing fulfillment extension, and clarify how fulfillment applies to compact-path line items before server-generated line item IDs exist.
3. Add attribution/affiliate examples such as `attribution/afid=...`, since Target requires AFID today and affiliate networks often wrap/append it.
4. Add age-restricted guidance: DOB-in-URL, but UCP should steer toward age-gate remediation or a dedicated profile/extension, not a casual base field.
5. Narrow the default for preserving non-UCP query parameters, especially for cross-origin redirects.
6. Add rate-limit guidance and a 429/Retry-After response option.

docs/specification/permalink.md
```diff
@@ -0,0 +1,707 @@
...
+Location: https://checkout.merchant.example/session/chk_123
+```
+
+## Security Considerations
+
+A permalink endpoint is an unauthenticated browser GET. Every component of the
+URL is externally supplied and untrusted, and browsers, prefetchers,
+link-preview bots, and security scanners load permalink URLs without buyer
+intent. A Business
+therefore MUST treat permalink resolution as safe and idempotent: repeated loads
```
maximenajim: Target has a feature today that has an explicit add-to-cart semantics: if the guest already has an active cart, we add the requested items to that cart, and if the same item is already present, quantity is incremented.

I think that merchant behavior is valid, but it is in tension with this permalink GET/idempotency requirement unless the spec defines the permitted mutation model more clearly. Agents need to know whether this URL means “set/stage these line items,” “merge into active cart,” or “initialize a checkout session".

Suggested clarification:

> A Business MAY merge permalink items into an existing buyer cart, create a separate staged cart/checkout, or replace a prior staged permalink state, but repeated loads of the same permalink URL by the same browser/session MUST NOT repeatedly increment quantities or otherwise amplify side effects without a fresh buyer action.

docs/specification/permalink.md
```diff
@@ -0,0 +1,707 @@
...
+```
+
+Possible resolution:
+
+```http
+HTTP/1.1 303 See Other
+Location: https://checkout.merchant.example/session/chk_123
+```
+
+### Campaign link with discount and continuation
```
maximenajim, line 552: Can we add a fulfillment example here? Target links support fulfillment in the URL, e.g. pickup plus a required pickup location. UCP can express this through the existing fulfillment extension when active, but the permalink spec does not show the pattern.

Suggested example:

```text
https://merchant.example/buy/98800943:1?fulfillment/methods/0/type=pickup&fulfillment/methods/0/selected_destination_id=1375&attribution/afid=test
```

ihoosain (approved):
i chat this over with a few folks internally and this looks good to us! :shipit:

igrigorik:
@maximenajim thanks, great feedback. See commit & commit msg: https://github.com/Universal-Commerce-Protocol/ucp/pull/523/changes/3d25a5357a7d665a4af6250eed5085848788f01a. A couple additional notes..

**Attribution / AFID** — already expressible and shown. `attribution` is an open map, and the examples use `attribution/utm_source` and `attribution/ref`; `afid` is just another key. Any active-extension field is already targetable — no change needed.

**Age-restricted / DOB** — agreed, shouldn't be special-cased. Added a generic line that a Business MAY require additional steps (verification, gating, authentication) before constructing state; no-PII-in-URL is already normative. 

**Cross-origin non-UCP params** — cross-origin redirects are valid use case for some accelerated payment methods and the existing "MAY drop or rewrite when unsafe or disallowed by policy" already covers sensitive-param stripping.

amithanda (approved):
Approving! This is a very well-designed capability and a genuinely hard problem (a cross-merchant, schema-typed permalink standard) solved cleanly. 
Let's wait for @maximenajim's approval before merging.

maximenajim (approved):
Looks good, thank you!

PR #511, opened 2026-06-11
fix(schema): allow hyphenated reverse-domain namespaces

juanferrub:
## Summary

Allow reverse-domain namespace segments to contain interior hyphens, so domain-derived namespaces such as `com.example-shop.shopping.checkout` validate.

Fixes #478.

## Motivation

UCP namespace authority is based on reverse-domain ownership. Many real merchant and vendor domains include hyphens, including domains such as `harley-davidson.com`, `ray-ban.com`, `t-mobile.com`, `new-balance.com`, and `coca-cola.com`.

The current regex rejects those domain-derived namespaces, which can force implementations to use workaround domains or less-authoritative namespace choices.

## Details

This updates `common/types/reverse_domain_name.json` to permit hyphens inside namespace segments while keeping the existing constraints that segments start with a letter and do not end with a hyphen. Underscores remain supported in later segments for existing names such as `identity_linking` and `loyalty_gold`.

## Compatibility / Risk

This is backward compatible for existing identifiers because it only broadens accepted values. The updated pattern continues to reject leading/trailing hyphens in a segment, digit-starting segments, and underscore-starting segments.

Downstream implementations with stricter local validators may need to align with this schema change.

## Test Plan

- `npx --yes cspell --no-progress --no-summary source/schemas/common/types/reverse_domain_name.json`
- `npx --yes ajv-cli@5.0.0 compile --spec=draft2020 -s source/schemas/common/types/reverse_domain_name.json`
- `git diff --check`
- Local Node regex checks covering valid and invalid hyphenated namespace cases

Not run locally: `ucp-schema lint source/` because `ucp-schema` is not installed in this environment. CI should run the repository schema lint.

## Contribution Notes

I read the organization contribution guide before opening this. The PR uses a conventional commit title and is intentionally scoped to issue #478. I understand the Google CLA may need to be completed before review/merge.

Issue #478, opened 2026-05-25
[Proposal]: namespace support domain with -， like ucp-aa.com


slf2code:
### Summary

Our domain name format is like this: ucp-xx.com. However, the short dash (-) is not allowed during the extension process, which results in the inability to define the extension. But standard domain names do support the short dash (-), and we hope to be able to support it as well.

### Motivation

<img width="1246" height="1264" alt="Image" src="https://github.com/user-attachments/assets/ac8eb460-236c-4084-a1bf-b4c60c3b2876" />
support "-"

### Goals

support "-"

### Detailed Design

support "-"

### Risks and Mitigations

support "-"

### Test Plan

support "-"

### Graduation Criteria

**Working Draft → Candidate:**
* [ ] Schema merged and documented (with Working Draft disclaimer).
* [ ] Unit and integration tests are passing.
* [ ] Initial documentation is written.
* [ ] TC majority vote to advance.

**Candidate → Stable:**
* [ ] Adoption feedback has been collected and addressed.
* [ ] Full documentation and migration guides are published.
* [ ] TC majority vote to advance.

### Implementation History

* [YYYY-MM-DD]: Proposal submitted.
* [YYYY-MM-DD]: TC approved "Provisional"; capability enters "Working Draft".
* [YYYY-MM-DD]: TC approved advancement to "Candidate".
* [YYYY-MM-DD]: TC approved "Implemented"; capability enters "Stable".

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #530, opened 2026-06-19
feat: define namespace authority binding + broaden RDNS grammar

igrigorik:
Two related namespace-governance changes:

1. **Authority Binding**
   * A normative definition of *how* a capability's declared `schema` URL is bound to its reverse-domain authority. Existing spec asserted "origins MUST match the namespace authority" but never said how to *compute* the authority, and its two normative statements disagreed (MUST validate vs SHOULD reject) — leaving the rule unenforceable.
2. **Reverse-domain grammar**
   * Broadened so all valid domains can be expressed (interior hyphens, digit-leading labels, IDN TLDs), and single-sourced so the grammar can't silently diverge across schemas.

Staged related enforcement PR in Universal-Commerce-Protocol/ucp-schema#34 that implements the algorithm.

## Authority Binding

Normative changes:

- **Clarify scope: provenance, not trust.** Binding proves the name is controlled by the domain owner — *whether to negotiate/support* a capability is the client's decision. It's a hostname check, **not** a fetch-safety control; dereferencing remains subject to the protocol's URL fetch-safety (SSRF / resolved-address) requirements, which are independent.
- **Only the `schema` URL is authority-bound** — it's the machine-fetched contract. The `spec` URL is human documentation, not in the trust path, so it's `https`-only with a free origin (docs may live on a subdomain or third-party host). This avoids forcing docs onto a name-aligned origin.
- **"MUST validate" reconciled to "MUST reject"** — a failing binding is treated as not present; MUST NOT be activated.
- **Enforcement is symmetric.** UCP profile exchange is bidirectional (platforms fetch `/.well-known/ucp`; businesses fetch the platform profile via `UCP-Agent`), so *any* party consuming a profile MUST validate the binding for every declared capability/service/payment handler. Business Requirements now state this explicitly, mirroring Platform Requirements.

Renames the section **"Spec URL Binding" → "Authority Binding"** (the bound thing is the `schema` URL, not the `spec` URL) and updates the core-concepts cross-link.

## Updated Reverse-Domain grammar

```
old: ^[a-z][a-z0-9]*(?:\.[a-z][a-z0-9_]*)+$
new: ^[a-z](?:[a-z0-9-]*[a-z0-9])?(?:\.[a-z0-9](?:[a-z0-9_-]*[a-z0-9_])?)+$
```

| Reverse-domain name | Old | New | Expresses the domain |
| --- | --- | --- | --- |
| `com.example.loyalty_gold` | ✅ | ✅ | `example.com` |
| `dev.ucp.common.identity_linking` | ✅ | ✅ | (underscore identifier) |
| `com.example-shop.checkout` | ❌ | ✅ | `example-shop.com` (interior hyphen) |
| `com.2example.cart` | ❌ | ✅ | `2example.com` (digit-leading SLD) |
| `xn--p1ai.example.checkout` | ❌ | ✅ | an `xn--p1ai` IDN TLD |

Segments start with a letter or digit, never start or end with a hyphen. The first segment (the reversed TLD) is letters/digits and may carry interior hyphens only for punycode (`xn--…`) TLDs. As a bonus, all inline copies now reference the canonical schema instead of duplicating it:

- `capability.json` `extends` → `$ref common/types/reverse_domain_name.json`
- `signals.json` propertyNames → `$ref` (keeps its own description)

## Compatibility / risk

- Grammar change is backward compatible: it only widens accepted values; every previously valid name still validates.
- The binding requirement is new normative behavior (MUST).

---

## Closes
- #478
- #511

## Checklist


docs/specification/overview.md
```diff
@@ -77,18 +77,90 @@ All capability and service names **MUST** use the format:
...
+| `dev.ucp.shopping.checkout`         | `shopping.ucp.dev` | `dev.ucp.shopping` | **accept** |
+| `com.example.payments.installments` | `example.com`      | `com.example`      | **accept** |
+| `com.example.pay`                   | `evil.example`     | `example.evil`     | **reject** |
+| `dev.ucp.shopping.checkout`         | `evil.example`     | `example.evil`     | **reject** |
+| `com.examplecorp.pay`               | `example.com`      | `com.example`      | **reject** |
+| `com.example.pay`                   | `cdn.example.com`  | `com.example.cdn`  | **reject** |
+
+A capability's `schema` is served from a host whose reversed labels are a prefix
+of its name. A canonical apex host (`example.com` for `com.example.*`) always
+satisfies this; a subdomain satisfies it only when its labels line up with the
```
jamesandersen, line 147: Hmm so a `com.example.pay` capability would need to host schema at `pay.example.com` or `example.com` ... but _related_ subdomains like `schema.example.com` or `api.example.com` wouldn't validate.  While this may be an acceptable constraint we could consider allowing this flexibility with something like:
1. Get the host of the schema url
2. Derive the [registrable domain](https://url.spec.whatwg.org/#host-miscellaneous) of the host (e.g. [public suffix](https://github.com/publicsuffix/list/wiki/Format#algorithm) plus one additional label)
3. Ensure the capability starts with the reversed registrable domain

igrigorik: @jamesandersen my concern with PSL is that it's a _big_ data & implementation dependency that affects every validator **and** handshake. Both parties _SHOULD_ be enforcing the describing validation rules when performing the capability negotiation (on every request), which means both sides now have to load/maintain the PSL list and appropriate logic to match — this explodes implementation & operational complexity.

I understand the value, but I would strongly advocate against PSL.

jamesandersen: I can get behind this.    It's an easy enough change should there be a compelling reason for it in the future but your point about the implementation burden is sound.

jamesandersen (approved):

source/schemas/capability.json
jingyli: Given the verbiage in `Authority Binding` in the documentation, we probably should also have the `required` annotation for `spec` and `schema` for business capability declaration (so platforms can run proper validations against it)?

Similar comment to payment_handler's definition in https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/payment_handler.json#L38

jingyli (approved):

source/schemas/capability.json
igrigorik: Good call. Updated requirements and docs. Note: schema, spec is not necessary as required.

amithanda:
Overall looks great, I have some minor suggestions to clarify some edge cases and security boundaries. PTAL and check if they make sense:

### 1. Clarifying URL Port Numbers (Host vs. Hostname Ambiguity)

The draft uses the term "host" (e.g., *"derived from the `schema` URL host"*) to describe the domain name matching component. While this is conceptually correct under the WHATWG URL Standard (where the internal `host` data structure excludes the port), most programming language APIs represent this differently, which could lead to implementation bugs:
* **JavaScript/Node.js**: `url.host` includes the port (e.g., `example.com:8443`), whereas `url.hostname` excludes it (`example.com`).
* **Go**: `url.Host` includes the port, requiring a call to `url.Hostname()` to strip it.
* **Python**: The domain without port is `url.hostname`, while `url.netloc` includes it.

If a developer directly maps the word "host" in the specification to their language's standard library `host` property, the reversed prefix will include the port (e.g. `8443:com.example`), causing legitimate capability validations to fail in sandbox or non-default port environments.

**Suggestion**: 
We could add a simple parenthetical clarification to the definition under the Derivation Algorithm to ensure developers use the hostname rather than host-plus-port and add that as an example to be super clear:
```diff
- The authority is derived from the `schema` URL host — which names the owning domain directly...
+ The authority is derived from the `schema` URL host (excluding any port number) — which names the owning domain directly...
```

---

### 2. Public Suffix List (PSL) Validation & Path-Based Hijacking

The specification requires the host to be a registered domain name of at least two labels, but does not address multi-tenant hosts where tenants are separated by URL path rather than subdomains. Without validating against the Public Suffix List (PSL) or warning against path-based shared hosts, naively reversing hostnames allows tenants on a shared domain to hijack namespaces:
* **Scenario**: Suppose a shared hosting provider serves schemas under `https://my-shared-host.com/tenant-name/schema.json`. 
* **The Match**: An attacker on `https://my-shared-host.com/attacker/schema.json` declares a capability named `com.my-shared-host.victim.checkout`. The validator parses the schema URL and extracts the host: `my-shared-host.com`. Reversing the host labels yields `com.my-shared-host`. Since `com.my-shared-host.victim.checkout` begins with `com.my-shared-host.`, the validation check passes, allowing the attacker to claim the victim's namespace.

While we should not mandate PSL validation (due to the runtime complexity and dynamic dependency overhead it would introduce for lightweight clients), it would be helpful to explicitly warn developers about this vector.

**Suggestion**:
Consider adding a warning note in the security considerations section explaining that naively treating public suffixes or shared multi-tenant hosts (like `s3.amazonaws.com` or path-based shared servers) as registered domains allows other tenants under that suffix to satisfy the prefix match, and recommending caution when designing custom capabilities on shared domains.

---

### 3. CNAME Subdomain Naming Constraints (The DNS Subdomain Trap)

Because the authority binding relies on a prefix match of the reversed host labels, merchants using third-party SaaS hosting face a DNS configuration trap if they attempt to CNAME custom domains.

Typically, when mapping a store to a SaaS provider, a merchant configures a CNAME record using a generic subdomain like `schema.merchant.com` or `saas.merchant.com`. However, this will fail validation:
* **The Mismatch**: 
  * A merchant defines a capability named `com.merchant.checkout`.
  * They host the schema on a CNAME'd subdomain: `https://schema.merchant.com/schema.json`.
  * The validation client reverses the host to: `com.merchant.schema`.
  * **The check**: Does the capability name `com.merchant.checkout` start with `com.merchant.schema.`? **No** (`checkout` does not match `schema`). The validation fails.
* **The Correct Configuration**:
  * To pass validation, the CNAME'd subdomain labels must align exactly with the capability's namespace path.
  * If the capability name is `com.merchant.checkout`, the schema URL must be served from `https://checkout.merchant.com/schema.json`.
  * The validation client reverses the host to: `com.merchant.checkout`.
  * **The check**: Does `com.merchant.checkout` start with `com.merchant.checkout.`? **Yes** (Accept).

**Suggestion**:
We should consider documenting this direct protocol consequence in the **Enforcement** section. It would be helpful to clarify that hosting schemas on a platform's shared domain (without a custom domain mapping) restricts the capability names to that platform's namespace, and that subdomains mapped via CNAME must align exactly with the prefix path of the capability namespace.

docs/documentation/core-concepts.md
```diff
@@ -317,9 +317,15 @@ registry — domain owners control their own namespace.
 | `dev.shopify.catalog` | shopify.dev | Shopify |
 | `com.example.payments.installments` | example.com | example.com |
 
-The `spec` and `schema` URLs declared in a capability must originate from the
-namespace authority domain. Platforms **MUST** validate this binding to prevent
-spoofed capabilities.
+A capability's `schema` URL must originate from its namespace authority domain.
```
raginpirate: Nit: this applies to more than capabilities, might be nice to use the word "entity" instead.

docs/specification/overview.md
```diff
@@ -77,18 +77,93 @@ All capability and service names **MUST** use the format:
...
-| Namespace       | Required Origin           |
-| --------------- | ------------------------- |
-| `dev.ucp.*`     | `https://ucp.dev/...`     |
-| `com.example.*` | `https://example.com/...` |
-
-Platform **MUST** validate this binding and **SHOULD** reject capabilities where
-the spec origin does not match the namespace authority.
+#### Authority Binding
+
+Every capability declares a `schema` URL (the machine-readable JSON Schema,
```
raginpirate: Same nit for this section

.cspell/custom-words.txt
```diff
@@ -123,4 +123,22 @@ vulnz
 worktree
 yaml
 yml
+jwks
```
raginpirate, line 127: Nit: It looks like some duplicates merged into custom-words.txt!

docs/specification/overview.md
```diff
@@ -77,18 +77,93 @@ All capability and service names **MUST** use the format:
...
+   domains in A-label / punycode form) and **reverse its labels** to form the
+   `authority_prefix` (host `ucp.dev` → `dev.ucp`).
+4. The binding is valid if and only if `name` begins with `authority_prefix`
+   followed by a `.` (a literal trailing dot). The trailing-dot boundary is
+   required so that `com.example` (from host `example.com`) cannot satisfy a
+   neighboring namespace like `com.examplecorp.*`, where it is a textual but
+   not label-aligned prefix; it also guarantees a non-empty
+   `{service}.{capability}` remainder.
+
+The `{service}.{capability}` remainder is treated as opaque by this check; it is
```
raginpirate: Nit: this isn't always `{service}.{capability}` as a tail, right? Even in the examples you proposed its actually a `{vertical}.{entity}` (`com.example.payments.installments`)

raginpirate (commented):
👍 Thanks for pushing on provenance of schemas! Totally aligned with this direction, just had a few nits I'd love your thoughts on.

igrigorik:
@amithanda @raginpirate ty, great feedback. Incorporated in last commit -- ptal.

docs/specification/overview.md
```diff
@@ -77,18 +77,108 @@ All capability and service names **MUST** use the format:
...
+the same prefix, so declare entities only under a **registrable domain** (a
+public suffix plus one label) that you exclusively control.
+
+##### Enforcement
+
+A platform **MUST** validate each business-declared `schema` URL before fetching
+it. If the URL's origin does not match the entity's namespace authority (per
+[Derivation algorithm](#derivation-algorithm)), the platform **MUST NOT** fetch
+it and **MUST** reject the entity — treated as not present and never
+activated. A `spec` URL **MUST** be a valid `https` URL.
```
amithanda: The profile-fetch rule is explicitly scoped: implementations **MUST NOT** follow redirects (`3xx`) "on profile fetches" ([overview.md:1236]). Should `schema` fetches carry the same rule? As written they're governed by Schema Resolution / Resolution Flow, which specifies what to fetch but not redirect behavior, and the Enforcement section defers to "URL fetch-safety requirements" that only guard the resolved address (SSRF, DNS rebinding), not `3xx` following. So two conformant validators could diverge on the same profile: one stops at a redirect from the authority-aligned host, the other follows it. If following is permitted, an open redirect on an authority-aligned host also sidesteps the binding, since the redirect target is never re-validated against the namespace authority. Could we state redirect handling explicitly for `schema` fetches so the behavior is deterministic?

```suggestion
activated. A `spec` URL **MUST** be a valid `https` URL. A platform **MUST NOT** follow redirects (`3xx`) when fetching a `schema` URL, consistent with profile fetches.
```

amithanda (approved):
@igrigorik - added a minor suggestion about covering redirects (just as we cover in profile fetches). PTAL before merging.

Approving this PR 🚀🚀 . Thanks for incorporating all the feedback.

Issue #515, opened 2026-06-15
Enhancement Proposal: Mandate Evaluation and Receipts (resolves #512)


arnaud-92:
> **Status:** Proposal · **Related:** #512 · **Target:** Checkout capability extension
> **Proposed by:** Arnaud Buisine — MerchantStamp · initial namespace `com.merchantstamp.mandate-evaluation`
> **Basis:** implementation work in a UCP merchant server for long-tail merchants, and the design discussion in #512

## Summary

Define the behaviour when an AP2 mandate's TTL elapses during checkout completion. The model evaluates mandate liveness **once**, at receipt of **Complete Checkout**, and records that evaluation as a verifiable receipt:

- valid at that instant → authorization is **frozen** for the session, an *authorization receipt* is recorded, and processing runs to terminal regardless of later expiry;
- already lapsed at that instant → the session never enters `complete_in_progress`, terminates `canceled`, and emits an `EXPIRED` *cancellation receipt*.

Both outcomes share one signed-receipt envelope verified through a single path. This resolves #512 (*Undefined behavior when AP2 mandate TTL expires during `complete_in_progress`*).

## Motivation

The Checkout lifecycle has a processing state, `complete_in_progress`, between a business accepting Complete Checkout and the order reaching `completed`. If the authorizing AP2 mandate's TTL elapses *while the session is in that state*, the specification does not define what happens — whether in-flight work is abandoned, runs to completion, or what a downstream party observes. #512 reported this gap.

The root cause is conflating two properties under "TTL validity":

- **Authorization** — *did the buyer permit this charge?* Settled the instant the business accepts Complete Checkout against a mandate valid at that instant. A processing delay the buyer did not cause does not unsettle it.
- **Liveness** — *is the grant inside its window right now?* Gates whether a payment **may be entered** into processing; it does not determine whether work already admitted stays authorized.

AP2 reinforces this: an Intent Mandate carries the TTL as a hard constraint and AP2 provides **no in-protocol channel to revoke a mandate before its TTL**. The TTL is therefore a pre-agreed bound on *when the grant may be exercised*, not a live withdrawal signal. Re-evaluating it mid-processing would invent a "buyer withdrew consent" event the protocol does not have.

[...]

PR #540, opened 2026-06-23
feat(checkout): mandate-evaluation extension for AP2 TTL at admission (resolves #512)

### LINKED ISSUE #512: [Bug]: Undefined behavior when AP2 mandate TTL expires during complete_in_progress
### Describe the bug

The checkout specification does not define expected behavior when an AP2 mandate (Intent or Cart) reaches its TTL **after** `Complete Checkout` has been accepted but **before** payment processing finishes, i.e. while the checkout is in `complete_in_progress`.

We are implementing a UCP merchant server for long-tail merchants (non-Shopify platforms) and hit this question while wiring AP2 payment handlers into the checkout state machine.

The `complete_in_progress` state is explicitly designed to cover business-side processing. In practice, payment processing in this window is not instantaneous:

- 3-D Secure challenges can take minutes when the buyer is interrupted (directly relevant to the 3DS2 work in #420 / #421, which will make long-running completions a first-class scenario);
- asynchronous payment methods common in the EU (SEPA Direct Debit, iDEAL, Bancontact) confirm with delays;
- PSP retry logic on soft declines can extend authorization well past the original request.

Meanwhile, AP2 mandates carry a Time-to-Live, and the Cart Mandate is the merchant's non-repudiable evidence of buyer authorization. Nothing in the current checkout spec states *when* TTL validity must be evaluated relative to the state machine.

**Open questions for the spec:**

1. Is mandate TTL validity evaluated **once**, at receipt of `Complete Checkout`, and frozen for the remainder of processing? Or must it hold continuously until `completed`?
2. If a mandate expires mid-processing, what is the expected terminal state — `canceled`, or a transition back to `requires_escalation` with a `continue_url` so the buyer can re-authorize?
3. Is a dedicated error code (e.g. `mandate_expired`) warranted, so platforms can distinguish this from generic payment failures and trigger re-authorization flows instead of abandoning the transaction?
4. Should the spec recommend a minimum remaining TTL at `Complete Checkout` time (e.g. the merchant MAY reject completion if remaining validity is below its expected processing window)?

### To Reproduce

This is a specification gap rather than a runtime bug, so "reproduction" is the scenario that exposes it:

1. Buyer authorizes a purchase; agent holds a valid AP2 Cart Mandate with a TTL.
2. Agent calls `Complete Checkout`; the checkout enters `complete_in_progress`.
3. Payment processing takes longer than the mandate's remaining TTL (e.g. an interrupted 3DS challenge, a SEPA confirmation delay, or PSP retries on a soft decline).
4. The mandate is now expired while processing is still underway. The spec gives no normative answer on whether the merchant should complete, cancel, or escalate — and implementations will diverge.

### Expected behavior

A normative definition of TTL evaluation relative to the checkout state machine. Possible directions (non-exhaustive):

- **(a)** TTL is evaluated at receipt of `Complete Checkout`; once accepted, processing completes regardless of subsequent expiry. Simple, deterministic, favors completion.
- **(b)** TTL must hold until `completed`; on expiry the checkout transitions to `requires_escalation` with a re-authorization `continue_url`. Stricter, favors buyer control.
- **(c)** Merchant-declared policy: the capability profile advertises which of (a)/(b) the merchant applies, plus an optional `min_remaining_ttl`, so platforms can act accordingly before calling `Complete Checkout`.

### Screenshots

_No response_

### Additional context

**Why it matters:**

- **Dispute integrity:** a payment authorized against a mandate that expired mid-flight weakens the cryptographic audit trail that AP2 exists to provide. Merchants need certainty about which timestamp governs.
- **Determinism across implementations:** without a normative answer, PSP handlers will make divergent choices, and identical checkouts will succeed on one merchant stack and fail on another.
- **EU payment methods:** async rails make this a common case, not an edge case, for European merchants.

Happy to draft spec language for whichever direction maintainers prefer, and to validate it against our implementation.

### Code of Conduct

arnaud-92:
# Description

Adds a vendor-namespace Checkout capability extension, `com.merchantstamp.mandate-evaluation`
(`docs/specification/mandate-evaluation.md`), defining **when** AP2 mandate liveness is evaluated
relative to the UCP Checkout state machine and **how** that admission decision is made auditable.
Resolves the `complete_in_progress` TTL ambiguity reported in #512, landing the design converged in
#512 / #515 with @aeoess, @chopmob-cloud and @GarethCOliver.

**Normative model**
- Mandate liveness is evaluated **exactly once**, at admission of Complete Checkout.
  - Valid at admission → authorization frozen; **MAY** enter `complete_in_progress`; **MUST** run to
    terminal; later expiry **MUST NOT** abort or shorten the settlement window.
  - Expired at admission → **MUST NOT** enter `complete_in_progress`; terminates `canceled` with the
    existing `mandate_expired` error.
- `mandate_expired` is **admission-time only** — it MUST NOT fire once in `complete_in_progress`
  (the normative fix for #512).

**Compose on AP2, don't fork.** No new lifecycle status, error code, receipt envelope, signature
algorithm, or key-discovery mechanism. The outcome is an AP2-aligned terminal receipt (`Success`,
or `Error` + `mandate_expired`); admission evidence rides as a namespaced extension field
(`evaluated_at`, `reference`, `checkout_id`, `mandate_exp`, `result`) on that receipt. `evaluated_at`
and `mandate_exp` are Unix epoch integer seconds, consistent with AP2 core.

**Binding by content** — recomputable tuple `{reference, checkout_id, evaluated_at}`; no forward
`receipt_id`. **Digest rules** kept in two separate cases (SHA-256 over the exact signed
serialization for the artifact handle, with compact-vs-detached JWS pinned per representation;
JCS-then-SHA-256 for unsigned recomputable content). **Signing** reuses the existing UCP/AP2
Message Signatures profile (ES256 required, ES384/512 optional, JCS, `signing_keys[]`).

**Scope.** Point-of-use admission + post-admission lifecycle only. Pre-TTL withdrawal / revocation /
status propagation — including the closed-enum `EXPIRED` cancellation receipt — is out of scope and
lives on a separate track.

## Category (Required)

## Related Issues

Resolves #512
Refs #515

## Checklist

## Notes

Working-Draft vendor extension; conformance tests to follow in
`Universal-Commerce-Protocol/conformance`. Open items tracked in §7 of the doc (receipt-addressing
conformance vectors, validity-window commitment, receipt delivery placement, FIDO push-down of
`evaluated_at`, DCQL/OpenID4VP capability-query alignment).

cc @GarethCOliver @aeoess @chopmob-cloud — opening as text to iterate on, per
https://github.com/Universal-Commerce-Protocol/ucp/issues/515#issuecomment-4741493270.


arnaud-92:
@aeoess — good catch, and it deserves to be pinned explicitly so two implementers can't diverge.

`reference` is the **AP2 mandate reference** — the hash of the closed mandate credential, computed per AP2's existing rule — **never** the checkout reference. The checkout/session identity rides in its own `checkout_id` field. They are two distinct named fields in the tuple, so `reference` is never overloaded to mean the checkout.

**Why the mandate ref is the one that binds:** the admission decision answers "was *this grant* live at the point of use," so the evidentiary anchor has to be the mandate being evaluated — `reference` carries that. `checkout_id` scopes the decision to the flow, which is what enforces mutual exclusivity per checkout and lets a verifier reconstruct the link. Both are required, but they answer different questions: `reference` = *which authorization grant*, `checkout_id` = *which flow*.

To remove the ambiguity, I'll tighten §5.1/§5.3 in `docs/specification/mandate-evaluation.md` to state:

- `reference` — the AP2 mandate reference; hash of the **closed mandate credential**, per AP2's existing definition. **MUST NOT** be the `checkout_hash` (which hashes the inner `checkout_jwt`, equal to the Payment Mandate `transaction_id`).
- `checkout_id` — the UCP checkout/session identifier.
- binding tuple = `{reference, checkout_id, evaluated_at}`, where `reference` and `checkout_id` are two independent identifiers and are never conflated.

This and the compact-vs-detached JWS pinning (§6.1) are exactly the two implementer footguns worth nailing before the fields freeze — thanks for flagging both.

aeoess:
That resolves it.

Using `reference` as the AP2 mandate reference, meaning the hash of the closed mandate credential, while keeping `checkout_id` as the separate flow identity is the right split. The admission decision binds to the grant being evaluated, and the per-checkout exclusivity scoping lives in its own field.

The §5.1 / §5.3 wording you laid out, with `reference` explicitly not being the `checkout_hash` and the binding tuple keeping the two as independent identifiers, closes the ambiguity.

Pinning compact versus detached JWS per representation in §6.1 closes the other one. Those were the two places where two conformant implementers could have diverged on the same logical receipt.

Both are pinned now. Thanks for tightening the text.

aeoess (approved):
Approving. This follows the direction the discussion on #515 converged on, and the parts most likely to create interoperability problems look right to me: the admission-time invariant in 3.3, recomputation of the content-tuple binding in 5.3, keeping reference separate from checkout_id, preserving the two distinct digest rules in 6.1, and the clear limit placed on what evaluated_at proves in 5.5.

I have three non-blocking suggestions.

1. Section 5.4 makes mandate_exp required only when the verifier cannot independently resolve the mandate: "REQUIRED whenever the verifier cannot independently resolve the mandate." The problem is that the issuer does not know the capabilities of every future verifier when it creates the signed object. That makes the requirement difficult to apply or test at issuance time. I would make mandate_exp unconditionally required. It costs one integer, makes the time-window check in 5.5 consistent for every verifier, and leaves open item 7.2 with a cleaner question: when should a verifier prefer the live mandate over the committed validity window?

2. Section 5.2 says that the terminal receipt links back to the admission attestation, but it does not define the actual pointer used for that link. Section 6.1, case 1, already provides a natural candidate: the jwt-hash handle of the admission attestation. Naming that explicitly as the back-pointer would make assertion 8 in the test plan interoperable. Otherwise, different implementations may choose different references while each believing it follows the specification.

3. result appears to repeat information already expressed by status and error on the AP2 receipt. A short consistency rule would remove the possibility of those fields contradicting one another. For example, result could be required to be valid when status is Success, and expired when error is mandate_expired.

None of these changes the architecture, and all three seem suitable for the existing open-items process. Ready to land from my side.

PR #536, opened 2026-06-22
docs: link common types to shared references section

igrigorik:
An alternative strategy to #443 drafted in response to #412.

Rendering #443 locally and reading resulting capability pages, I think the suggested nudge in #412 went too far: yes, it consolidates repeated tables, but it sacrifices readability -- most of the types end up "see X in reference" anchors which sacrifice readability. In practice, table duplication is not a big deal because we autogen the tables and always regenerate the full site on any update - i.e. they don't get out of sync. **This PR narrows the redirect predicate to schemas under `common/types/` only.**

Link sentences ("See X in the Schema Reference …") per capability page:

| Page                | #443 (full consolidation) | this PR (common-only) |
| ------------------- | ------------------------: | --------------------: |
| `cart`              |                        14 |                     5 |
| `checkout`          |                        21 |                     8 |
| `catalog/index`     |                        15 |                     5 |
| `fulfillment`       |                        12 |                     1 |
| `order`             |                         7 |                     1 |
| `embedded-checkout` |                         6 |                     1 |

The link sentences that remain are exactly the cross-vertical primitives each page actually references:

| Page                | Redirected to `reference.md`                                                                |
| ------------------- | ------------------------------------------------------------------------------------------- |
| `cart`              | Link, Message, Message Error / Info / Warning                                               |
| `checkout`          | Error Code, Error Response, Link, Message, Message Error / Info / Warning, Postal Address   |
| `catalog/search`    | Pagination Request / Response \*                                                            |
| `fulfillment`       | Postal Address                                                                              |
| `order`             | Postal Address                                                                              |
| `embedded-checkout` | Postal Address                                                                              |

\* via `extension_schema_fields`, emitted as a bare cross-reference rather than a wrapping "See X in …" sentence.

Closes #443, #412.

---
## Checklist

- [x] **Documentation**: Updates to README, or documentations regarding schema or capabilities.
- [x] **Infrastructure**: CI/CD, Linters, or build scripts. (Requires DevOps Maintainer approval)
- [x] I have followed the [Contributing Guide](https://github.com/Universal-Commerce-Protocol/.github/blob/main/CONTRIBUTING.md)
- [x] I have updated the documentation (if applicable).
- [x] My changes pass all local linting and formatting checks.
- [x] I have added tests that prove my fix is effective or that my feature works.

PR #542, opened 2026-06-24
docs(specification): per-operation Requirement column (#412)

sriakula1:
Standalone follow-up to #443 (closed) / #536 (merged): the `Visibility` column @wry-ry asked for on #412, split out so #536's narrower centralization isn't blocked or muddied.

Schema-reference tables now render `Visibility` in place of the binary `Required` column, showing `required` / `optional` / `omit` per request operation (`create` / `update` / `complete`) and per response. Derived from the existing `ucp_request` annotation plus the base `required` array — **rendering only, no schema changes**.

Examples from the rendered reference:

| Field | Visibility |
| :--- | :--- |
| `Line Item.id` | **Required** in responses; omitted on create, optional on update |
| `Line Item.totals` | **Required** in responses; omitted in requests |
| `Checkout.line_items` | **Required** in responses; required on create & update, omitted on complete |
| `Checkout.buyer` | Optional in responses; optional on create & update, omitted on complete |

Fields with no `ucp_request` annotation render as a plain **Required** / Optional, unchanged in spirit. Operation-specific request/response tables on capability pages (resolved via `ucp-schema`) continue to show the effective state for that operation — correct, not misleading.

Implementation: a new `_field_visibility()` helper in `main.py` plus a one-cell change to the row format. All row paths funnel through it, so the table layout stays consistent across `_render_table_from_ref` and `_render_embedded_table`. `main.py` only, +61 / −4.

Verified locally on Python 3.13 by driving `auto_generate_schema_reference` over every `types/` schema in `common/` and `shopping/`: 52 tables rendered, 0 old `Required` headers remaining, 28 per-op breakdown cells, 0 malformed rows. The canonical multi-version mkdocs build + link check will run in CI.

Cell wording is a judgement call — happy to adjust if you'd prefer a different shape:
- Collapse request clauses that equal the response value (e.g. `Checkout.buyer`).
- Split into separate `Create` / `Update` / `Complete` / `Response` columns instead of one prose cell.

cc @wry-ry @igrigorik


damaz91:
Thanks for your contribution. However, during local testing of the PR (which is intended to render operation-specific rules like **Required** in responses; omitted on create...), I noticed that these detailed messages are not appearing in the final rendered docs.

The issue seems to be that the ucp-schema resolve tool (used during the doc build process to resolve refs and flatten allOf) strips the custom ucp_request annotations. Because of this, details.get("ucp_request") in main.py always evaluates to None, and the tables fall back to the base response visibility.

Not sure how this is working for you, have you tried building the entire documentation locally rather than calling the table rendering macro directly?

sriakula1:
@damaz91 thanks for building locally. I reran `./scripts/build_local.sh` against this branch and the rendered `/specification/reference/` page contains 51 multi-op Visibility cells. A few grep'd from the HTML:

```
Required in responses; omitted on create, required on update
Required in responses; required on create & update
Required in responses; required on create & update, omitted on complete
```

A/B on the same machine: `upstream/main` produces 0 multi-op cells and 90 `Required` headers; this branch produces 51 cells and 90 `Visibility` headers. The annotation survives because `auto_generate_schema_reference` loads raw JSON via `_load_json_file` (main.py:1203) rather than `_resolve_schema`, and the `$ref` recursion stays raw when `context=None`. Per-operation capability pages do route through `ucp-schema resolve` and strip the annotation, but on those pages the binary Required/Optional is the correct answer for the single operation; the PR was scoped to `reference.md` only.

Possible you were testing a per-op page? Happy to share full build output, a screenshot of the rendered table, or walk through the call chain.

damaz91:
@sriakula1 right, I was looking at an outdated local build. 

The new `Visibility` column is a great improvement, but it can become quite verbose when multiple request operations share the same visibility as the response default (which is the inherited fallback).

We can make the tables much cleaner by **only rendering operations that differ from the response visibility**, and using a simpler separator (`; ` instead of ` in responses; `).

#### Examples:

| Field | Current Rendering | Proposed Collapsed Rendering |
| :--- | :--- | :--- |
| `Fulfillment Method.type` | **Required** in responses; required on create, optional on update | **Required**; optional on update  |
| `Checkout.buyer` | Optional in responses; optional on create & update, omitted on complete | Optional; omitted on complete  |
| `Line Item.totals` | **Required** in responses; omitted in requests | **Required**; omitted in requests |

#### Implementation Suggestion:

You can achieve this by updating the `_field_visibility` helper in `main.py` to filter out redundant clauses:

```python
def _field_visibility(field_name, ucp_request, required_list):
  """Render the Visibility cell for a schema field."""
  word = {"required": "required", "optional": "optional", "omit": "omitted"}
  response = "required" if field_name in required_list else "optional"
  base_disp = "**Required**" if response == "required" else "Optional"

  if ucp_request is None:
    return base_disp

  if isinstance(ucp_request, str):
    if ucp_request == response:
      return base_disp
    return (
      f"{base_disp}; "
      f"{word.get(ucp_request, ucp_request)} in requests"
    )

  if isinstance(ucp_request, dict):
    # Only keep operations that differ from response visibility (default)
    diff_request = {}
    for op in ("create", "update", "complete"):
      val = ucp_request.get(op)
      if val and val != response:
        diff_request[op] = val

    if not diff_request:
      return base_disp

    # Group adjacent request operations that share a visibility value
    groups = []  # list of (value, [ops]) preserving operation order
    for op in ("create", "update", "complete"):
      if op not in diff_request:
        continue
      val = diff_request[op]
      if groups and groups[-1][0] == val:
        groups[-1][1].append(op)
      else:
        groups.append((val, [op]))
        
    clauses = [
      f"{word.get(val, val)} on {' & '.join(ops)}" for val, ops in groups
    ]
    if not clauses:
      return base_disp
    return f"{base_disp}; {', '.join(clauses)}"

  return base_disp

damaz91:
@sriakula1 can you please take a look at the feedback? Thanks in advance!

sriakula1:
Good call, that reads a lot cleaner. Pushed it in 7910009.

The cell now drops any request op that just matches the response default and uses `;`, so your three examples come out as:

- `Fulfillment Method.type` → **Required**; optional on update
- `Checkout.buyer` → Optional; omitted on complete
- `Line Item.totals` → **Required**; omitted in requests

Kept the adjacent-op grouping so a run like `Required; omitted on create, optional on update` still collapses where it should.

And on the annotation thing from earlier — you were right that I should've been building the whole site instead of poking at the macro in isolation. Ran a full `./scripts/build_local.sh` this time and the `ucp_request` annotations survive the resolve step fine, tables render end to end. Thanks for staying on it.

damaz91:
@sriakula1 Last feedback from my side: I suggest we change the column name from "Visibility" to "Requirement".

Once you do this, I'll approve. Thanks!

sriakula1:
@damaz91 Done — renamed the column to **Requirement** in b6b6b26. Thanks for the reviews on this one!

damaz91:
Thanks! Don't want to be annoying, but also all the method names/comments in the code keep referencing to "visibility". Can you change those as well?

sriakula1:
@damaz91 Apologies, that should have been caught the first time. Swept the rest in 6f558c7 — helper is now `_field_requirement`, the local var and all the docstrings/comments follow suit. No `visibility` left anywhere in the file.

damaz91 (approved):

damaz91:
Last nit: can you make sure the code passes the lint test? There are a few lines over 80 char now @sriakula1 . Thanks again!

sriakula1:
Good catch — fixed in `71a4083` ✅

The rename bumped a handful of docstring/comment lines to 81 chars. Wrapped them back under 80 (no logic touched); ruff E501 is clean locally now, zero lines over 80. Should be green once the workflow run gets the go-ahead. Thanks for the thorough pass on this one! 🙏

carolinerg1 (approved):

PR #547, opened 2026-06-25
docs: add server timing response header

karangoel16:
## Summary
- Add optional `Server-Timing` response header to UCP Shopping REST responses
- Document the `ucp;dur=<milliseconds>` metric for checkout, cart, catalog, and order REST bindings
- Link to the MDN `Server-Timing` reference and clarify that platforms should still measure client-observed latency
- Update the docs header macro so response headers render for non-200 responses such as `201 Created`

## Related
- Catalog REST's missing HTTP Headers section is split into #548 so that documentation structure fix can land separately first.

## Validation
- `python3 -m json.tool source/services/shopping/rest.openapi.json >/dev/null`
- `python3 -m py_compile main.py`
- `python3 scripts/validate_examples.py --schema-base source/schemas/ --file docs/specification/checkout-rest.md docs/specification/cart-rest.md docs/specification/order-rest.md docs/specification/catalog/rest.md`
- `python3 scripts/test_validate_examples.py`
- `uv run mkdocs build --site-dir /tmp/ucp-mkdocs-check`

Note: `uv run mkdocs build --strict` is blocked locally by missing native Cairo libraries required by CairoSVG image conversion.

### DIFF
--- a/docs/specification/cart-rest.md
+++ b/docs/specification/cart-rest.md
@@ -80,6 +80,16 @@ All request and response bodies **MUST** be valid JSON as specified in
 
 All REST endpoints **MUST** be served over HTTPS with minimum TLS version 1.3.
 
+### Response Timing
+
+Business responses **MAY** include the
+[`Server-Timing`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Server-Timing){ target="_blank" }
+header to report business-side processing duration. When present, UCP responses
+**SHOULD** use the `ucp` metric with `dur` in milliseconds, for example
+`Server-Timing: ucp;dur=42.3`. This value excludes network time and is
+informational; platforms **SHOULD** measure client-observed latency
+independently.
+
 ## Operations
 
 | Operation | Method | Endpoint | Description |
@@ -132,6 +142,7 @@ All REST endpoints **MUST** be served over HTTPS with minimum TLS version 1.3.
     ```json
     HTTP/1.1 201 Created
     Content-Type: application/json
+    Server-Timing: ucp;dur=42.3
 
     {
       "ucp": {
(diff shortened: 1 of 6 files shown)


igrigorik (changes requested):
Big fan of Sever-Timing and definitely encourage wider adoption of its use, but I don't think we need to or should normatively define it in UCP spec. It's a feature that can be separately and independently adopted by any HTTP service; there is noting UCP-specific or special in how it needs to integrate with Server-Timing.

karangoel16:
@igrigorik , 

One reason I still think this is worth documenting somewhere: without a shared convention, every platform would need to work with every provider independently to learn whether server-side timing is available, what metric name to parse, and what unit/scope it uses.

That said, I agree with the review concern that `Server-Timing` probably should not become part of the core protocol contract. It is generic HTTP observability, not required for UCP correctness or interoperability.

A possible compromise would be to move this out of the per-operation normative REST/OpenAPI surface and into a short non-normative implementation guidance note, something like:

```md
UCP REST implementations MAY use the standard HTTP `Server-Timing` header for diagnostics. When reporting total business-side UCP request processing time, implementations SHOULD use the `ucp` metric with `dur` in milliseconds, for example `Server-Timing: ucp;dur=42.3`.

Platforms MUST treat this value as diagnostic only and MUST NOT use it as a substitute for client-observed latency measurement.
```

That gives providers a common convention to adopt, while avoiding making timing telemetry a required or semantically critical part of the UCP protocol.

igrigorik:
> One reason I still think this is worth documenting somewhere: without a shared convention, every platform would need to work with every provider independently to learn whether server-side timing is available, what metric name to parse, and what unit/scope it uses.

Server-Timing is self-discoverable: it's presence is the signal, and there is a well-defined algorithm for how to parse all the labels, values, etc. Further, I don't think we should codify any promises here because this is diagnostics data that is not meant for public consumption; ST is typically provided for internal tooling and RUM measurement which itself can be sampled instead of being provided on every response. 

I appreciate the push and agree that Server-Timing is a useful and relevant tool for surfacing perf-data for UCP implementers, but I would still argue that we shouldn't and don't need to make any special callouts for it in UCP spec itself -- the Server-Timing contract is sufficient an agnostic to the application / application-layer protocol.

damaz91:
Thanks @karangoel16 and @igrigorik for the discussion; closing this for now based on the conversation.

PR #561, opened 2026-07-03
docs(fulfillment): correct options[].total row to match the schema (totals array)

### LINKED ISSUE #559: fulfillment.md documents groups[].options[].total (Required, nullable) but the schema has no total — it requires a totals array whose amounts cannot be null
In `docs/specification/fulfillment.md` (main), the fulfillment-option field table documents:

| `groups[].options[]` | `total` | Yes | Price in minor units (may be null if not yet available) |

The schema disagrees on three points:

- **No `total` field.** `source/schemas/shopping/types/fulfillment_option.json` composes (`allOf`) `fulfillment_option_base.json` — which defines only `id`, `title`, `description` — with an object that **requires `totals`** (an *array* of `total.json`). There is no `total` property anywhere.
- **Not nullable.** Each `totals[]` entry requires `amount` (`types/total.json`), typed as a bare `integer` via `common/types/signed_amount.json`, so `null` is valid in no position.

So both "Required: Yes" (for a field that does not exist) and "may be null" (impossible under the schema) are schema-false. The row predates 2026-01-23 and is present in every released version.

**Suggested fix:** rewrite the row to describe `totals` (a required array of `{type, amount, display_text}`). If a "price not yet available" state is intended, it would need a mechanism the schema actually supports (omitting the option, or a documented sentinel) rather than a null amount — happy to PR whichever you prefer.

_Found while building an unofficial UCP conformance suite (https://spck.dev); happy to PR a fix if useful._

vishkaty:
Corrects the `groups[].options[].total` row in `fulfillment.md` to match the schema.

`source/schemas/shopping/types/fulfillment_option.json` requires **`totals`** — an array of `total.json` — and defines no `total` field. `total.json`'s `amount` is a bare `integer` (`common/types/signed_amount.json`), so the documented "may be null" is impossible in any position.

Closes #559. Found while building an unofficial UCP conformance suite (https://spck.dev). Happy to adjust the wording — in particular, if a "price not yet available" state is intended, it needs a schema-supported mechanism, which I left out rather than guess at.

### DIFF
--- a/docs/specification/fulfillment.md
+++ b/docs/specification/fulfillment.md
@@ -184,7 +184,7 @@ human-readable fields that platforms render directly.
 | --------------------- | ------------- | -------- | ------------------------------------------------------- |
 | `groups[].options[]`  | `title`       | Yes      | Primary label that distinguishes from siblings          |
 | `groups[].options[]`  | `description` | No       | Supplementary context for the title                     |
-| `groups[].options[]`  | `totals`      | Yes      | Cost breakdown for this option: an array of `total` objects (`type`, `amount`, `display_text`) |
+| `groups[].options[]`  | `totals`      | Yes      | Cost breakdown: an array of `total` objects             |
 | `available_methods[]` | `description` | No       | Standalone explanation of alternative availability      |
 
 ### Business Responsibilities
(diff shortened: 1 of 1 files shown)


vishkaty:
@googlebot I signed it!

vishkaty:
Thanks for the re-run — that attempt was against the previous commit. The MD060 table-alignment error is fixed in 2cb81fa (verified against .github/linters/.markdownlint.json locally, 0 errors); the workflow run for the new commit is pending approval.

carolinerg1 (approved):

cusell-google (approved):

PR #566, opened 2026-07-06
feat!: promote `keys[]` to canonical profile signing key field

igrigorik:
Lands the breaking change migration of top-level `keys[]` JWK Set to the **single canonical** key-publication field in a UCP profile and removes `signing_keys[]`. A profile is now natively a valid RFC 7517 JWK Set with no dual publication.

---

#483 shipped `keys[]` and `signing_keys[]` as parallel arrays that **MUST** for backwards compatibility. Collapsing to a **single `keys[]`** (which we agreed to in 483) removes drift and complexity of mirroring across arrays, eliminates the duplication, and makes the profile a valid JWKS with no dual-publication gymnastics. 

On the wire, before (#483) — dual, mirrored arrays:

```json
{ "ucp": { "...": "..." }, "signing_keys": [ "..." ], "keys": [ "..." ] }
```

After — single canonical JWK Set:

```json
{ "ucp": { "...": "..." }, "keys": [ "..." ] }
```

**This is breaking change**: verifiers at this version read `keys[]`. The transition is publisher-controlled: a business can publish and return appropriate profile based on provided / negotiated UCP version during handshake.

## Checklist


docs/specification/overview.md
```diff
@@ -455,38 +455,23 @@ least one capability.
...
+
+Adding, rotating, or removing a key updates this single array. Removal
+is the security-critical case: a revoked or compromised key is not
+effectively revoked until it is absent from `keys[]`.
+
+Profiles migrating from an earlier major version **MAY** additionally
+publish a legacy `signing_keys[]` array for interop with verifiers built
+before `keys[]` became canonical; those verifiers read `signing_keys[]`,
+while UCP verifiers at this version read `keys[]` and ignore the legacy
+array.
```
amithanda: nit: Since a migrating profile can publish both arrays, should we restore the both-arrays removal rule this section used to carry? As written, removing a revoked key from `keys[]` alone still leaves it verifying for anyone reading the legacy `signing_keys[]`.
```suggestion
Profiles migrating from an earlier major version **MAY** additionally
publish a legacy `signing_keys[]` array for interop with verifiers built
before `keys[]` became canonical; those verifiers read `signing_keys[]`,
while UCP verifiers at this version read `keys[]` and ignore the legacy
array. A profile that publishes both arrays **MUST** remove a revoked or
compromised key from `signing_keys[]` as well as `keys[]`; a key that
remains in either array continues to verify until it is absent from both.

**Comment by amithanda** [comment]
Hi @igrigorik,

Pushed a small follow-through commit (de3166c) that brings `docs/documentation/core-concepts.md` in line with the `keys[]` promotion. That file was not otherwise in the diff, so it would have shipped still describing the removed field (including a key-lookup step telling implementers to match `kid` in `signing_keys[]`). 

The commit updates:
- Profile Structure prose: `keys`, described as a valid RFC 7517 JWK Set reusable as a Web Bot Auth key source (mirroring the overview wording)
- the business-profile example: `signing_keys` to `keys`
- the key-lookup step: match `kid` in `keys[]`
It is three lines of doc consistency, so I put it here for convenience, hope it is ok.

Separately, I added an inline suggestion on the Profile Structure section of `overview.md` about revocation during migration. Since a migrating profile MAY publish both `keys[]` and a legacy `signing_keys[]`, I suggest we restore the both-arrays removal rule that #483 carried. As written, removing a revoked or compromised key from `keys[]` alone still leaves it verifying for anyone reading the legacy `signing_keys[]`. The inline block proposes the scoped MUST.

PTAL and let me know if these look ok.

File: docs/specification/overview.md (hunk 1 above)
Comment (igrigorik) [reply docs/specification/overview.md]: My preference would be to leave this out, as it's a footgun and error-prone surface to keep in sync -- the very thing we want to get away from with this update. The business has the agents' version on the request and can return / negotiate the appropriate response for legacy clients vs. clients on version where we ship this update.

**Comment by igrigorik** [comment]
@amithanda ty, pushed update looks good. Left a comment on the signing_keys[] suggestion.

File: docs/specification/overview.md (hunk 1 above)
Comment (jamesandersen) [inline docs/specification/overview.md]: same nit, different proposal: The current language does leave room for some confusion as @amithanda notes.   @igrigorik what about just removing this suggestion of including *both* and instead prescribing the one or the other approach you suggest below:
```suggestion
Profiles migrating from an earlier major version **MAY** publish a
legacy `signing_keys[]` array for interop with verifiers operating at
earlier UCP versions before `keys[]` became canonical; those verifiers
read `signing_keys[]`, while UCP verifiers at this version read `keys[]`.
Profiles **SHOULD** include only one of these arrays to avoid 
indeterminate results should their contents differ.
```

jamesandersen (approved):
Overall LGTM - thanks @igrigorik

docs/specification/overview.md
amithanda: I read the last statement in the interop requirement - **those verifiers read `signing_keys[]`,
while UCP verifiers at this version read `keys[]` and ignore the legacy
array."**  as handling two cases:
1) signing_keys -> keys
2) signing_keys + keys -> keys

I was more looking at clarifying the case 2) and @igrigorik - I think you are right to point that out that it causes issues when we look at 1).

I think @jamesandersen's change modifies that statement and removes the ambiguity so this should work.

jamesandersen:
Incidentally, while reviewing existing docs related to this PR I found a couple minor docs issues and opened #576

igrigorik:
@jamesandersen @amithanda good catch, this was error of omission and I propose we simply remove that section. PTAL at https://github.com/Universal-Commerce-Protocol/ucp/pull/566/changes/ba1638d77a4a0d9bafcef64294105022426d62db and commit msg for explanation.

vishkaty:
With `signing_keys[]` gone, `keys[]` is a plain RFC 7517 JWK Set — which may legitimately carry non-signature keys (`use: "enc"`, or `key_ops` without `sign`/`verify`). The verification text resolves the key by matching `keyid` to a published `kid`, with no `use`/`key_ops` filter, so I tested what happens between two reasonable implementations: a verifier that resolves by `kid` alone accepts a signature from a key published with `use: "enc"`, while a verifier that pre-filters the set JWKS-style (`use == "sig"`) rejects the same request with `key_not_found`. Same wire bytes, opposite verdicts.

Since RFC 7517 leaves `use` OPTIONAL (§4.2) and distinct `kid` values within a set are only a SHOULD (§4.5), would it be worth one sentence pinning verifier behavior — e.g. verifiers consider only signature-capable keys (`use` absent or `"sig"`; `key_ops` absent or including `verify`) when resolving `keyid`?

igrigorik:
@vishkaty sharp eyes, ty for catching this. Updated: https://github.com/Universal-Commerce-Protocol/ucp/pull/566/commits/0c6a19d689b198778e58adb7c3be628e3847ba2b.

vishkaty:
Confirmed — 0c6a19d resolves it: with the resolution-time filter, both verifier styles now agree on the same request. I'll align the samples#122 verifier to this rule once it lands.

amithanda (approved):
Great work! Approving the PR.

PR #568, opened 2026-07-07
docs(order): webhook example must sign ucp-agent (identity otherwise unbound)

vishkaty:
The "Example Webhook Request" in `docs/specification/order.md` carries a `UCP-Agent` header but its `Signature-Input` signs only `("@method" "@authority" "@path" "content-digest" "content-type")` — omitting `ucp-agent`.

This contradicts the normative signed-component rules in `signatures.md`:

- The request signed-components table marks `ucp-agent` **"Required if `UCP-Agent` header is present"** (the `**` footnote), and there is no separate webhook component table.
- `signatures.md` explicitly derives webhook identity from the `UCP-Agent` header ("Business → Platform webhooks: Profile URL from `UCP-Agent` header") and requires webhooks to be signed.

Because the example doesn't sign `ucp-agent`, the sender's identity (the component's documented purpose — "binds identity") is not cryptographically bound in the illustrated signature, and a platform copying the example produces a signature that a table-following verifier would reject.

One-line fix: add `ucp-agent` to the example's signed components, matching the request example in `signatures.md`.

Found while building an unofficial UCP conformance test suite (https://spck.dev); verified against main @ 7e5fc42.

---

**Update (class sweep):** re-checked every RFC 9421 example in `docs/` for this same pattern (UCP-Agent header present but not among the signed components). One more instance: `checkout-rest.md`'s "Example Signed Request" — fixed in the second commit with the same one-line change, matching the component order of the canonical request example in `signatures.md`. All other examples either already sign `ucp-agent`, omit the header entirely, or are response/elided examples where the component doesn't apply.

### DIFF
--- a/docs/specification/checkout-rest.md
+++ b/docs/specification/checkout-rest.md
@@ -1434,7 +1434,7 @@ Content-Type: application/json
 UCP-Agent: profile="https://platform.example/.well-known/ucp"
 Idempotency-Key: 550e8400-e29b-41d4-a716-446655440000
 Content-Digest: sha-256=:X48E9q...:
-Signature-Input: sig1=("@method" "@authority" "@path" "idempotency-key" "content-digest" "content-type");keyid="platform-2025"
+Signature-Input: sig1=("@method" "@authority" "@path" "ucp-agent" "idempotency-key" "content-digest" "content-type");keyid="platform-2025"
 Signature: sig1=:6G4i8TS6oUkGrx8KnCFUpsSPwd74...:
 
 {"line_items":[{"item":{"id":"item_123"},"quantity":2}]}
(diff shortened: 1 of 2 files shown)


vishkaty:
Superseding this with #659.

This PR's purpose is signing `ucp-agent` in the order webhook example. #659 now covers that example more completely: it adds `ucp-agent` (binding signer identity, the point of this PR) alongside `webhook-id` and `webhook-timestamp`, and it adds the normative rule requiring those to be covered plus deduplication on `Webhook-Id`. Rather than leave two PRs editing the same example line, I am closing this in favor of #659.

This PR also added `ucp-agent` to a signed request example in checkout-rest.md, which #659 does not touch. That is a small, separate improvement; I can refile it on current main if it is wanted. Thank you.

PR #576, opened 2026-07-09
fix: avoid duplicated description for scalar schema $defs in docs macro

jamesandersen:
## Summary

The `extension_schema_fields` docs macro renders a schema `$def`'s
description **twice** for scalar `$defs`. On the [AP2 Mandates page](https://ucp.dev/draft/specification/ap2-mandates/)
this shows up as a duplicated paragraph under **Business Authorization**,
**Checkout Mandate**, and **Error Codes**. The same duplication appears on
the auto-generated **Schema Reference**, which renders extension `$defs`
through the same code path.

## Root cause

In `main.py`, `_read_schema_from_defs()` (behind the `extension_schema_fields`
macro) renders the schema table and then prepends the `$def`'s `description`:

```python
table = _render_table_from_schema(embedded_schema_data, ...)
desc = embedded_schema_data.get("description", "")
if desc and need_header:
    return f"{desc}\n\n{table}"   # prepends description
return table
```

- For **structured** schemas (`properties`/`allOf`), the rendered `table` is
  a property grid that does *not* contain the description, so prepending it is
  correct.
- For **scalar** schemas (a `string`/`enum` with a `pattern` or `enum` and no
  `properties`), `_render_table_from_schema`'s scalar fallback *already*
  embeds the description alongside the pattern/enum. Prepending it again
  produces the duplicated paragraph.

The three affected AP2 `$defs` (`merchant_authorization`, `checkout_mandate`,
`error_code`) are all scalar strings/enums.

## Fix

Only prepend the description for structured schemas; scalar tables already
contain it:

```python
is_structured = any(
    key in embedded_schema_data
    for key in ("properties", "allOf", "oneOf", "$ref")
)
if desc and need_header and is_structured:
    return f"{desc}\n\n{table}"
return table
```

Scalar `$defs` now render the description once (followed by pattern/enum);
structured `$defs` are unchanged (description + property table).

## Testing

The full docs build is validated by CI (`docs.yml`), which runs the real
`mkdocs` build with `ucp-schema`. Locally I reproduced the exact render path
for the affected `$defs` — since they contain no `$ref`, `ucp-schema`
bundling is a no-op for them, so the raw-schema render matches the resolved
output. Before/after for `checkout_mandate`:

**Before** (description repeated):
```
SD-JWT+kb credential in `ap2.checkout_mandate`. Proving user authorization...

SD-JWT+kb credential in `ap2.checkout_mandate`. Proving user authorization...

**Pattern:** `^[A-Za-z0-9_-]+\.[A-Za-z0-9_-]*\.[A-Za-z0-9_-]+(~[A-Za-z0-9_-]+)*$`
```

**After** (description once):
```
SD-JWT+kb credential in `ap2.checkout_mandate`. Proving user authorization...

**Pattern:** `^[A-Za-z0-9_-]+\.[A-Za-z0-9_-]*\.[A-Za-z0-9_-]+(~[A-Za-z0-9_-]+)*$`
```

Structured `$defs` (e.g. `ap2_with_merchant_authorization`) still render
their description followed by the property table — no regression.

## Type of change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [ ] I have added tests that prove my fix is effective (macro rendering is
      validated end-to-end by the docs build in CI)

---

**Also included:** the `error_code` enum in
`source/schemas/shopping/ap2_mandate.json` defines
`merchant_authorization_missing`, but the hand-written error-code table in
`docs/specification/ap2-mandates.md` omitted that row. Added it so the table
matches the schema.


carolinerg1 (approved):

damaz91 (approved):
Thanks for the fix!

cusell-google (approved):

PR #578, opened 2026-07-10
feat(discount): add gift_line_item_id to applied discount

prateek-ct:
# Description

Some promotions grant a free or discounted line item (buy-one-get-one, free-gift-with-purchase) rather than reducing the price of an item the buyer already has. Today `applied_discount` can only describe an amount-based effect — a business granting a gift line item has no structured way to link it back to the discount that produced it, so a platform/agent sees an unexplained zero-price item in `line_items[]`.

This adds an optional `gift_line_item_id` field to `applied_discount`, pointing at the `line_items[]` entry the discount granted. It's omitted for ordinary amount-based discounts, so this is backward compatible.

Motivated by implementing the discount extension against a commerce backend whose "gift line item" cart discount type is a common, first-class promotion pattern — this gap surfaced immediately when mapping that feature onto the current schema.

## Category (Required)

## Related Issues

None filed yet — happy to open one first if maintainers would rather see this go through a fuller proposal before schema changes.

## Checklist

- [x] I have followed the Contributing Guide (including Conventional Commits title requirements and `!` for breaking changes).
- [ ] My changes pass all local linting and formatting checks.

N/A — schema/documentation change only.

### DIFF
--- a/docs/specification/discount.md
+++ b/docs/specification/discount.md
@@ -201,8 +201,10 @@ free-gift-with-purchase promotion).
 When a discount grants line items this way:
 
 - Each granted item **MUST** appear as a normal entry in `line_items[]`,
-  with its price reflecting the discount (typically `0`, but a reduced
-  non-zero price is also valid for "buy X get Y at $Z" patterns).
+  priced at its regular value — the same as any other line item — with the
+  discount reflected in that line item's own `totals[]`
+  (`items_discount`/`total`), exactly like a non-gift discount. Do not zero
+  out `item.price` itself.
 - The corresponding `applied_discount` entry **MUST** set
   `gift_line_item_ids` to those line items' `id`s, so platforms can
   attribute the items to the promotion rather than treating them as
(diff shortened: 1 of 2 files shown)


source/schemas/shopping/discount.json
```diff
@@ -76,6 +76,10 @@
             "$ref": "#/$defs/allocation"
           },
           "description": "Breakdown of where this discount was allocated. Sum of allocation amounts equals total amount."
+        },
+        "gift_line_item_id": {
```
sdedeo2025: Some ecommerce systems allow for multiple gifts or bonus products being triggered from a single promotion.  i.e. buy x, get gift y.  Buy a 2nd x, get a 2nd gift y.  

Could we make this an array?

docs/specification/discount.md
```diff
@@ -191,6 +191,112 @@ segment, or promotional rules:
...
+          ]
+        }
+      ],
+      "discounts": {
+        "codes": ["FREEGIFT"],
+        "applied": [
+          {
+            "code": "FREEGIFT",
+            "title": "Free Socks with Shoe Purchase",
+            "amount": 1200,
```
damaz91, line 302: maybe naive question, but what's the rationale of keeping the amount here? shouldn't it be automatically computed based on "gift_line_item_id" to avoid discrepancies?

source/schemas/shopping/discount.json
prateek-ct: Good catch - done. Changed gift_line_item_id to gift_line_item_ids (array of strings) so promotions that grant multiple gift units from a single discount (e.g. "buy 2, get 2 free" scaling with quantity) can be represented. Added a second example showing the multi-gift case with the latest commit.

docs/specification/discount.md
prateek-ct: Not naive at all , your question made me rethink this properly and it turned up a real mistake in my example- the gift line item's price was zeroed out. Fixed now , it shows its actual price like any other line item with the discount as its own items_discount entry in that line item's totals, so the total nets to 0 (or less, for partial-price gifts). Pushed in a recent commit, happy to have you take another look.

amount on the discount itself is still required though, even with the line item priced correctly. Nothing stops a second, unrelated discount from also touching that same line item , its totals[] can carry more than one items_discount entry with no way to tell which came from which discount. amount is what lets you attribute the value to this specific discount, same as for every other discount type.

PR #580, opened 2026-07-10
feat: factor available-instrument constraints into distinct axes

igrigorik:
Alternative shape for the constraints work in #424 — stacked on this branch as a concrete variant, not a merge-over. `ucp-schema lint` + doc-example validation pass; net smaller than the current shape.

The core proposal isn't a vocabulary change — it's **factoring `constraints` by axis.** The current single `constraints` bag mixes several genuinely different things, which is what makes it hard to reason about. Pulling them apart resolves that and, as a bonus, makes each part a clean, honest mechanism.

## The insight: "constraints" is four different axes

| Axis | What it does | Right mechanism | Key |
| :--- | :--- | :--- | :--- |
| **Presence** | "this field must be present" | `required` (+ `properties` to nest) | `constraints` |
| **Allowed values** (submitted field) | "this field's value ∈ set" | `enum` under `properties.<field>` | `constraints` |
| **Derived menus** | "I support these options" (attribute is derived, not submitted — e.g. card network) | a uniform menu map | `accepts` |
| **Discriminated** | "if subtype T, then rules-for-T" | typed list, applied by **data lookup** (never `oneOf`) | `credentials` |

Mixing all four under one `constraints` key is the confusion. Factoring them by key makes each honest.

## Shape

```json
{
  "type": "card",

  "constraints": {
    "required": ["billing_address"],
    "properties": {
      "billing_address": {
        "required": ["postal_code", "address_country"],
        "properties": { "address_country": { "enum": ["US", "CA"] } }
      }
    }
  },

  "accepts": { "brand": ["visa", "mastercard"] },

  "credentials": [
    { "type": "pan",           "constraints": { "required": ["cvc"] } },
    { "type": "network_token", "constraints": { "required": ["cryptogram"] } }
  ]
}
```

## Why factoring helps (not just tidier)

- **`constraints` is now a genuine bounded JSON Schema** over the instrument's own fields (`required`/`properties`/`enum`). In the mixed bag it wasn't — a validator run over it silently ignores `credentials` (an unknown keyword), giving false confidence. Moving the non-schema axes out fixes that.
- **`brands` becomes honest.** A card's network is *derived* from the credential, not a submitted field, so "accepted brands" is a menu you read to *offer* options — not a field-value constraint. `accepts` says that plainly; a uniform map (`{ attr: [values] }`) also stops accepted-value menus from proliferating as bespoke keys.
- **Discrimination stays a data list, never `oneOf`.** A base `oneOf` can't be widened by an extension `allOf` — it re-closes the type vocabulary exactly like an enum (verified). A typed list resolved by lookup keeps extension credential types open with no dangerous keywords.

### DIFF
--- a/docs/documentation/schema-authoring.md
+++ b/docs/documentation/schema-authoring.md
@@ -443,6 +443,15 @@ All new constraint objects **SHOULD** extend
   no `type` and stays open: it narrows named properties and never forbids unknown
   ones (no `additionalProperties:false`, `oneOf`, or `if`/`then`).
 
+`constraints` covers only field-level requirements over the *submitted* object.
+Two related but distinct axes use their own keys, **not** `constraints`:
+**accepted-value menus** for derived or selected attributes (e.g.
+`accepts: { "brand": ["visa", "mastercard"] }` — a card's network is derived, not
+a submitted field, so it is a menu, not a field constraint), and **discriminated
+per-subtype requirements** (e.g. `credentials`, below). Keeping them separate is
+what lets `constraints` stay a genuine bounded JSON Schema over the object's own
+fields.
+
 Prefer field-level constraints over ad-hoc booleans. For example, prefer
 `required: ["billing_address"]` plus `properties.billing_address.required`
 over new booleans such as `requires_billing_address` or
@@ -453,14 +462,18 @@ over new booleans such as `requires_billing_address` or
 {
   "type": "card",
   "constraints": {
-    "brands": ["visa", "mastercard"],
     "required": ["billing_address"],
     "properties": {
       "billing_address": {
         "required": ["postal_code", "address_country"]
       }
     }
-  }
+  },
+  "accepts": { "brand": ["visa", "mastercard"] },
+  "credentials": [
+    { "type": "pan",           "constraints": { "required": ["cvc"] } },
+    { "type": "network_token", "constraints": { "required": ["cryptogram"] } }
+  ]
 }
 ```
 
@@ -476,10 +489,10 @@ UCP-defined branches and handler/extension branches coexist without a schema
 ```json
 {
   "credentials": [
-    { "type": "pan", "required": ["cvc"] },
+    { "type": "pan", "constraints": { "required": ["cvc"] } },
     {
       "type": "com.example.wallet_token",
-      "required": ["assurance_level"]
+      "constraints": { "required": ["assurance_level"] }
     }
   ]
 }
--- a/docs/documentation/schema-authoring.md
+++ b/docs/documentation/schema-authoring.md
@@ -429,22 +429,23 @@ runtime requirement attached to a concrete context, not a global schema change.
 All new constraint objects **SHOULD** extend
 [`constraint.json`](site:{{ ucp_version }}/schemas/shopping/types/constraint.json):
 
-- `required_fields` names fields on the local constrained object that must be
-  present in an acceptable instance for this context.
-- The schema property that embeds the constraint defines the target object. For
-  example, `available_instruments[].constraints` targets the payment instrument,
-  while `constraints.billing_address` targets that instrument's billing address.
-- Domain-specific constraint keys remain sibling properties. For example, the
-  base payment instrument can require `billing_address` and add a nested
-  `billing_address` constraint describing which address fields are needed.
-- Reusable base schemas **SHOULD NOT** close `required_fields` with enums when
-  the target object can be extended. Concrete resolved schemas MAY narrow
-  `required_fields` when they own the complete target vocabulary, while leaving
-  the object open to future domain-specific keys.
+- `required` names properties of the constrained object that must be present in
+  an acceptable instance for this context.
+- `properties` maps a property name to a nested `Constraint` applied to that
+  property (recursive), e.g. `billing_address` narrowing which address fields are
+  required. The schema property that embeds the constraint defines the target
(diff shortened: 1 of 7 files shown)


raginpirate:
Thanks for the deep thinking here Ilya. I think defaulting to using JSON schema fragments is a great observation, and it sent me down a rabbit hole of experimenting!

While trying out these shapes, I ran into two rough edges I wanted to highlight and see if you have opinions on what "good" looks like, especially compared to how #424 solved for it.

 First, let’s look at billing address.                                                                                    
                                                                                                                          
 For pure field requirements, the JSON Schema-shaped version works well:                                                  
                                                                                                                          
 ```json                                                                                                                  
   {                                                                                                                      
     "type": "card",                                                                                                      
     "constraints": {                                                                                                     
       "required": ["billing_address"],                                                                                
       "properties": {                                                                                                    
         "billing_address": {                                                                                             
           "required": ["postal_code", "address_country"]                                                                 
         }                                                                                                                
       }                                                                                                                  
     }                                                                                                                    
   }                                                                                                                      
 ```                                                                                                                      
                                                                                                                          
 Where I get stuck is when the billing-address concern is still local to billing address, but is not itself a JSON Schema 
 assertion. For example, suppose we want to advertise that a provider-supplied address is acceptable (just an imaginary example of field toggling optionality rather than constraining fields):                     
                                                                                                                          
 ```json                                                                                                                  
   {                                                                                                                      
     "type": "card",                                                                                                      
     "constraints": {                                                                                                     
       "required": ["billing_address"],                                                                                   
       "properties": {                                                                                                    
         "billing_address": {                                                                                             
           "required": ["postal_code"],                                                                                   
           "allows_provider_address": true                                                                                
         }                                                                                                                
       }                                                                                                                  
     }                                                                                                                    
   }                                                                                                                      
 ```                                                                                                                      
                                                                                                                          
 That keeps locality, but now constraints is no longer an honest JSON Schema fragment. The alternative seems to be        
 pulling that out into another axis:                                                                                      
                                                                                                                          
 ```json                                                                                                                  
   {                                                                                                                      
     "type": "card",                                                                                                      
     "constraints": {                                                                                                     
       "required": ["billing_address"],                                                                                   
       "properties": {                                                                                                    
         "billing_address": {                                                                                             
           "required": ["postal_code"]                                                                                    
         }                                                                                                                
       }                                                                                                                  
     },                                                                                                                   
     "accepts": {                                                                                                         
       "billing_address_source": ["buyer", "provider"]                                                                 
     }                                                                                                                    
   }                                                                                                                      
 ```                                                                                                                      
                                                                                                                          
 That keeps constraints clean, but loses the locality that #424 had by letting billing-address-related constraints live   
 under the billing-address constraint object.                                                                            
                                                                                                                          
 Next, accepts vs credentials.                                                                                            
                                                                                                                          
 I understand why brand moved out of constraints: it is an accepted/derived option, not a submitted field constraint. But 
 then I’m not sure why credentials is a separate sibling rather than an accepts family.                                   
                                                                                                                          
 This shape splits two “accepted option” concepts:                                                                        
                                                                                                                          
 ```json                                                                                                                  
   {                                                                                                                      
     "type": "card",                                                                                                      
     "accepts": {                                                                                                         
       "brand": ["visa", "mastercard"]                                                                                 
     },                                                                                                                   
     "credentials": [                                                                                                     
       {                                                                                                                  
         "type": "pan",                                                                                                   
         "constraints": { "required": ["cvc"] }                                                                            
       },                                                                                                                 
       {                                                                                                                  
         "type": "network_token",                                                                                         
         "constraints": { "required": ["cryptogram"] }                                                                    
       }                                                                                                                  
     ]                                                                                                                    
   }                                                                                                                      
 ```                                                                                                                      
                                                                                                                          
 Conceptually, I’m tempted by:                                                                                            
                                                                                                                          
 ```json                                                                                                                  
   {                                                                                                                      
     "type": "card",                                                                                                      
     "accepts": {                                                                                                         
       "brand": ["visa", "mastercard"],                                                                                
       "credentials": [                                                                                                   
         {                                                                                                                
           "type": "pan",                                                                                                 
           "constraints": { "required": ["cvc"] }                                                                         
         },                                                                                                               
         {                                                                                                                
           "type": "network_token",                                                                                       
           "constraints": { "required": ["cryptogram"] }                                                                  
         }                                                                                                                
       ]                                                                                                                  
     }                                                                                                                    
   }                                                                                                                      
 ```                                                                                                                      
                                                                                                                          
 I realize credentials are richer than brands: brand is a scalar menu, while credentials are typed branches with          
 branch-local requirements. But they still feel like the same negotiation class: “these are the options this handler      
 accepts.” I'm not sure why we conceptually need to separate these buckets, but I might be missing something; an
 example could help!

Overall, I have an opinion that because this approach still requires implementers to apply UCP-specific interpretation   
 anyway — applying overlays, resolving credential entries by type, merging platform/business constraints, etc. — I’m not  
 yet convinced we should optimize primarily for JSON Schema fragments if doing so makes larger/nested constraints harder to express locally. But I'd love to hear your thoughts on these examples, and see if the shape in #424 may continue to have legs to stand on.

igrigorik:
Agree on the conclusion and landed in similar spot after iterating through this. I'll close this, but I think there are some takeaways and patterns here that we should consider for #424 -- I'll leave it to your discretion though on if and which are worth bringing over. Tag me for final review once you feel #424 is ready.

PR #582, opened 2026-07-12
feat: extension-defined Actions primitive

igrigorik:
This PR introduces **actions**, a shared response-only shape for outstanding extension-defined work. Cart, Checkout, and Catalog adopt the shape without adding an Actions capability, registry, service, or lifecycle.

Each reverse-domain map key identifies a concrete Action type declared by an active negotiated extension. Its value is a non-empty array of instances with `id` and optional extension-owned `config`. One extension may declare multiple Action types. Negotiating its version activates the complete contract, so related types can be negotiated and versioned together through existing extension machinery.

## Example

This Checkout response composes a line, provisional Discount, Student Verification Action, and correlated Message:

```json
{
  "ucp": {
    "version": "2026-01-01",
    "status": "success",
    "payment_handlers": {}
  },
  "id": "chk_student_1",
  "status": "incomplete",
  "currency": "USD",
  "context": {
    "eligibility": ["org.example.student"]
  },
  "line_items": [...],
  "discounts": {
    "applied": [
      {
        "title": "Student 10% Off",
        "amount": 500,
        "automatic": true,
        "provisional": true,
        "eligibility": "org.example.student",
        "allocations": [
          {"path": "$.line_items[0]", "amount": 500}
        ]
      }
    ]
  },
  "totals": [
    {"type": "subtotal", "amount": 5000},
    {"type": "items_discount", "amount": -500},
    {"type": "total", "amount": 4500}
  ],
  "links": [
    {"type": "terms_of_service", "url": "https://business.example.com/terms"}
  ],
  "actions": {
    "com.example.identity.student_verification": [
      {
        "id": "verify-student-1",
        "config": {
          "verification_url": "https://business.example.com/verify/abc"
        }
      }
    ]
  },
  "messages": [
    {
      "type": "info",
      "code": "eligibility_accepted",
      "content": "Student discount applied provisionally. Verify your student status to keep it.",
      "path": "$.actions['com.example.identity.student_verification'][0]"
    }
  ]
}
```

The active "Student Verification" extension defines the Action type, validates `config`, and specifies how the Platform processes it. The Message explains the current response and points to the exact Action occurrence; it does not define processing or lifecycle.

## Common behavior

A newly processed successful response includes every outstanding Action and omits `actions` when none are outstanding. While the same work persists, it keeps the same key and `id`; a replacement receives a new `id`.

An Action's presence means it gates the effect defined by its type. It does not say that an operation failed, determine the parent lifecycle, or define how many Catalog products are returned. Messages correlate outcomes with Actions using positional RFC 9535 paths such as:

```text
$.actions['com.example.identity.student_verification'][0]
```

A recoverable error Message pointing to an Action says the requested effect was not applied because of it; info and warning Messages may point to an Action without reporting failure. 

---

## Category
- [x] **Core Protocol**
- [x] **Capability**
- [x] **Documentation**


raginpirate:
Thanks for the deep thinking on actions Ilya! I like how you're approaching this, but I've got a few questions to see if theres still some rough edges we need to polish.
                                                                                                                          
 1. Is each Action key a single concrete Action type?                                                                     

 My read is that:                                                                                                         
                                                                                                                          
 ```json                                                                                                                  
   "actions": {                                                                                                           
     "com.example.identity.student_verification": [...]                                                                   
   }                                                                                                                      
 ```                                                                                                                      

 means com.example.identity.student_verification is not just the owner namespace, it is the concrete Action               
 contract/type. In other words: multiple instances under that key are multiple instances of the same Action type.
                                                                                                                          
 If that is not the intention, I think we’re missing a common type/code discriminator on the instance itself. Example:    
 one payment handler may need both device data collection and a 3DS challenge. Are those intended to be:                  

 ```json                                                                                                                  
   "actions": {                                                                                                           
     "com.example.payments.device_data_collection": [...],                                                                
     "com.example.payments.three_ds_challenge": [...]                                                                     
   }                                                                                                                      
 ```                                                                                                                      
                                                                                                                          
 as separately negotiated extensions, or one payment extension with multiple action types inside it? If it’s the latter,  
 I don’t see how the platform routes generically without inspecting opaque config.                                        

 If it’s the former, I think we should explicitly call out the tradeoff. There was TC concern around bloating negotiation 
 for Actions, and this feels adjacent to why delegated identity providers are not modelled as independently negotiated     
 services. Maybe that tradeoff is still right here, but I want to make sure we’re choosing it intentionally.              

 2. Transition-blocking state feels less precise than #553.                                                               
                                                                                                                          
 I like that this keeps Actions from becoming a separate lifecycle, but I’m trying to understand the edge where an Action 
 blocks the last attempted transition, not just eventual checkout completion.                                             
                                                                                                                          
 Example: Platform sends Update Checkout to set field X. Business says: “to accept that field update, first complete      
 CAPTCHA.” In #553, severity: blocking gave a generic machine signal: the attempted transition is blocked; resolve this   
 Action, then retry/change/escalate that transition.                                                                      
                                                                                                                          
 In this PR, I see required: true, plus resource status/messages. But how does the platform know this Action is           
 specifically required to resolve the last attempted state transition, versus generally required before completion? Is    
 the expectation that the extension defines that entirely? Should the platform retry the same Update after processing the 
 Action, or Get Checkout and infer state? If messages are supposed to carry that nuance, I worry we’re moving machine     
 control-flow semantics into explanatory text.                                          
                                                                                                                          
 3. Security/trust baseline seems to have lost some useful granularity from #553.                                         
                                                                                                                          
 I agree that the concrete extension should own its trust model, but I think core should still keep a few baseline MUSTs. 
 #553 had useful guardrails like:                                                                                         
                                                                                                                          
 - arbitrary URLs in config are not loadable just because they are URLs;                                                  
 - the Action spec must define origin/trust validation;                                                                   
 - platforms must not interpolate opaque config into executable contexts;                                                 
 - completion signals are not authoritative unless independently verifiable;                                              
 - unsupported/abandoned/expired/failed behavior must be defined.                                                         
                                                                                                                          
Wondering if it still makes sense to include these, especially the one on unsupported runtime negotiations. I imagine this will come up when URLs are rejected due to an unsupported 3rd party delegate; I think we need language on that similar to how we handle window.open requests for EC.

Appreciate your impact here! 🚀

docs/specification/checkout.md
```diff
@@ -103,16 +105,67 @@ platform receives messages indicating what's needed to progress.
...
+
+A `required: true` instance gates the effect specified for its Action type
+within the containing status:
+
+* While `incomplete`, a required Action identifies work the Business needs
+    completed before it can return `ready_for_complete`. After processing the
+    Action according to its Action type's contract, the Platform **SHOULD** use
+    [Get Checkout](#get-checkout) or a subsequent
+    [Update Checkout](#update-checkout) response to obtain the latest Checkout.
+* When Complete Checkout returns `complete_in_progress`, the Platform processes
```
amithanda: The PR adds a case where `complete_in_progress` means the Business is waiting on
the Platform rather than working internally. The accept-time rejection is
covered ("the Business MUST return the current Checkout with `status: incomplete`
and a recoverable error Message"), but I could not find the path for a required
Action that is declined, abandoned, or fails *after* `complete_in_progress` was
entered.

Reading the rest of the spec, the exits look like Cancel Checkout (which SHOULD
succeed from any non-terminal state) or `expires_at` (default 6h TTL, and
`canceled` can occur from any state). Both work.

Should this section state that explicitly? Something like: if the Platform
declines or cannot complete a required Action in `complete_in_progress`, it
SHOULD Cancel Checkout rather than continue polling, and the session otherwise
expires per `expires_at`. Without it, a Platform's only documented move in that
state is Get Checkout, which never terminates on its own.

gforgab:
Reviewing this as a concrete non-3DS / async consumer — we're modeling **Pix (Brazil, via Mercado Pago)** and it drops into this primitive with zero schema change, which is a good signal for the shape.

At `complete_in_progress` the PSP hands back a render artifact (QR image + copy-paste "copia e cola" string + short expiry). It fits as a pure-render Action: `config` is optional and extension-owned, nothing to load/submit in-band, and it resolves exactly through the path this PR already prescribes — Platform renders, then polls [Get Checkout](checkout.md#get-checkout) until `completed`, never re-driving Complete. No `mode`/`execution`/executor needed, which lines up with the "no generic machinery" stance here.

Two things worth confirming:

**1. Trust boundary for render fields.** The *Trust and Execution Boundaries* section resolves a concern we'd raised earlier (render fields sitting in opaque `config`): since "the active Action-type contract defines which `config` fields a Platform processes and what they mean," the declaring extension is the right home for `image` / `code` / `instructions_url`, and the contract must explicitly mark which are renderable vs. loadable. Reading that as intended? For a display Action specifically that means the extension pins `image` to `data:`/`https:` shown as an image only (never HTML/script), `code` as display text, and `instructions_url` to an origin/scheme allowlist — exactly the per-type machinery you're pushing down to the extension.

**2. Work that outlives the session.** This primitive ties Actions to the checkout lifecycle — `ready_for_complete`, `completed`, and `canceled` all state "no required Action remains outstanding," and resolution is polling Get Checkout. That fits seconds-scale Pix cleanly. It intentionally does **not** fit the same family that settles in hours-to-days (boleto, cash vouchers), which outlives the checkout session and has no in-band re-drive. We read that as out of scope here by design, and are scoping it to a separate order-level EP resolved through the existing Order Event Webhook (`dev.ucp.shopping.order`) rather than stretching `actions`. Flagging so the boundary is explicit — does that match your intent?

(Also relevant to your own open point on transition-blocking precision: for a pure-render Action the `required: true` + `recoverable` error Message with `path` is what disambiguates "gated on out-of-band payment" from "generally required before completion" — a second, non-3DS data point that the Message/`path` mechanism carries that gating fine without a dedicated field.)

docs/specification/checkout.md
igrigorik: Good prompt. Took a run at addressing this in: https://github.com/Universal-Commerce-Protocol/ucp/pull/582/changes/7f786a143326e8d336aabb83fa0e22a08d60e129

igrigorik:
@gforgab nice, this is great example and application of action primitive!

Your trust reading is right: the Pix contract can require its QR/code fields and define which values are inert text versus loadable content, along with schemes, origins, expiry, and fallback. But there is also a meta question here around presentation: in theory the whole experience can be a URL that platform can load, or it can be a more structured contract that platform can render in 1P surface. Actions, as a primitive, can support both.

> This primitive ties Actions to the checkout lifecycle — ready_for_complete, completed, and canceled all state "no required Action remains outstanding," and resolution is polling Get Checkout. That fits seconds-scale Pix cleanly. It intentionally does not fit the same family that settles in hours-to-days (boleto, cash vouchers), which outlives the checkout session and has no in-band re-drive. 

Correct.

docs/documentation/schema-authoring.md
```diff
@@ -433,6 +433,48 @@ object fields:
...
+  every safe-retry condition;
+- the fallback when processing fails, is declined, or is abandoned, when the
+  Action is unsupported or expires, or when the Business does not update the
+  parent response;
+- the Action-specific conditions under which work is resolved, superseded, or
+  replaced, plus within-type or cross-type ordering semantics when needed.
+
+Extension authors should define only controls relevant to the concrete
+processing model. For example, an Action with no loadable URL needs no URL
+scheme or origin rules. Extension authors should keep Action-specific data
```
raginpirate, line 468: Non-blocking: I found this to be confusing when reading it without examples. I think we can ship this without actioning, and in #458 we can link to the payment actions example in the tokenizer guide.

docs/specification/catalog/index.md
```diff
@@ -150,6 +150,54 @@ as the first element. Platforms SHOULD treat the first element as featured.
...
+products the Business returns or how the Platform handles them. `required` means
+the Action is required for the effect its type defines.
+
+For Search and batch Lookup, the Business decides whether to return zero, some,
+or all otherwise relevant products under the Action-type contract and its own
+policy. A Message can point to the Action to explain the response. Successful
+Get Product still includes `product`; its existing error response is unchanged.
+
+After processing an Action, the Platform performs a fresh Catalog operation and
+the later Business response is authoritative. Actions add no Catalog lifecycle,
```
amithanda: The Overview defines an Action `id` twice in terms of a containing resource with
cross-response continuity:

> `id` — a non-empty identifier, unique for the lifetime of the containing resource.

> While the same piece of outstanding work persists across responses, it keeps the same key and `id`. [...] an `id` is never reused within the containing resource.

For Checkout that reads cleanly, because a Checkout is a resource the Platform
re-fetches. The stable `id` is what lets a later response be recognized as the
same work, and the new *Accepted completion* rules depend on exactly that:

```json
// Get Checkout at T0 and again at T1 — same resource, same id => same work
{ "id": "chk_1", "status": "incomplete",
  "actions": { "com.example.identity.student_verification":
    [ { "id": "verify-student-1", "required": true } ] } }
```

Catalog is explicit that there is no such continuity: "the Platform performs a
fresh Catalog operation and the later Business response is authoritative. Actions
add no Catalog lifecycle, polling, or resume behavior." A Search or batch Lookup
response is not a resource with a lifetime, so I am not sure how to apply either
sentence to it. Concretely, two age-gated searches in one session:

```json
// Search A: {"query": "whiskey"}
{ "actions": { "com.example.identity.age_verification":
    [ { "id": "age-check-1", "required": true } ] } }
```
```json
// Search B: {"query": "cigars"}, moments later — what id here?
{ "actions": { "com.example.identity.age_verification":
    [ { "id": "???", "required": true } ] } }
```

Is the Business free to reuse `age-check-1` (each response its own scope), or
does "never reused within the containing resource" reach across independent
searches? 

Would one line in this section settle it? Something like: for stateless Catalog
responses, an Action `id` is unique within the response and carries no
cross-response identity, and the Overview's cross-response persistence rule does
not apply to Catalog. That keeps the Overview rule intact where Checkout and Cart
rely on it, without asking the Catalog model to meet a guarantee it can't.

source/schemas/common/types/actions.json
```diff
@@ -0,0 +1,48 @@
...
+          "type": "string",
+          "minLength": 1,
+          "description": "Identifier for this Action instance, unique for the lifetime of the containing resource.",
+          "ucp_request": "omit"
+        },
+        "required": {
+          "type": "boolean",
+          "description": "Whether this Action instance gates the parent-specific effect specified for its Action type.",
+          "ucp_request": "omit"
+        },
```
raginpirate: While reviewing the docs, I started struggling with the need for `required`, as it seems like it can be inferred in all of its used contexts now:

- If an Action prevented a state-changing operation, the Business returns the current resource with a correlated error or info Message depending on if the action must be resolved or not.
- If Complete Checkout was accepted and is waiting on Platform work, the Business returns complete_in_progress with the Action. Its clear the action should be resolved before the next Get request.                                                                                            
- If the resource has no lifecycle, its impossible for it to be "required" for the result, such as catalog; you still get a response, it just might be less than you wanted.

Since its definition is very hand-wavy at the moment, I'd suggest we remove this for first release and bring it back when we run into edges it will clearly help with.

docs/specification/checkout.md
```diff
@@ -103,16 +105,92 @@ platform receives messages indicating what's needed to progress.
...
+`complete_in_progress` means Complete Checkout was accepted. The Checkout state
+reported by the Business is authoritative.
+
+When a `complete_in_progress` Checkout includes a required Action, the Platform
+processes it according to its contract. After the Action reports completion,
+the Business might need time to receive and apply the result. The Action
+completion signal is not authoritative for Checkout state. The Platform
+**MUST** use [Get Checkout](#get-checkout) to confirm that the Business
+reflected the result, for example by removing the Action or changing the
+Checkout status. When a `complete_in_progress` Checkout contains no required
```
raginpirate: Open question: are you allowed to update a checkout when its in `complete_in_progress`? This seems to imply no, only get or cancel, but if you can't use update I imagine that rules out a large class of actions which want to have an output submitted through the associated resource. We should spell this out with a clear state machine.

docs/specification/catalog/index.md
```diff
@@ -150,6 +150,54 @@ as the first element. Platforms SHOULD treat the first element as featured.
...
+polling, or resume behavior. The common shape and rules are defined in
```
raginpirate: Nit: I believe your statement to be true about the base actions primitive. However, the extensions that introduce actions may add specific lifecycle, polling, and resume behaviour for when an action is used, can't they?

docs/specification/checkout.md
```diff
@@ -340,44 +419,99 @@ access to restricted products.
...
+    ]
+  },
   "messages": [
     {
-      "type": "error",
-      "code": "eligibility_invalid",
-      "severity": "recoverable",
-      "content": "Payment credential does not match the claimed store card benefit.",
-      "path": "$.context.eligibility[0]"
+      "type": "info",
```
raginpirate, line 494: Using this as an example for how we might be able to remove `required`: we could use `type: error` over `type: info` for the action with `eligibility_required`. If we did that, its apparent the action must be taken care of before submitting.

raginpirate (commented):
Thanks for the hard work aligning here @igrigorik
While giving this a deep think I ran into two edges I want to make sure we're on the same page about before merging this:
- the operations allowed against a checkout in complete_in_progress, which influences how much actions can really do.
- the need for `required`; if I might be missing the complexity reduction that this bool brings us now that we have a great approach leveraging messages when possible.

gforgab:
Thanks @igrigorik! On the presentation meta-question — for Pix we'd deliberately pick the **structured render contract** over a load-a-URL experience, and I think it's a useful data point for why Actions supporting both matters:

- The Pix artifact is inherently non-web: a QR image + a copy-paste "copia e cola" string the buyer takes to their bank app. The optimal UX is the platform rendering it natively (image + a copy button), which works identically on a web surface, a native app, or a conversational/voice agent that can't load a frame at all.
- A load-a-URL variant would re-introduce the loadable-content trust surface the contract is meant to constrain, and it degrades or breaks on headless/1P surfaces — the exact case where the structured contract shines.

Concretely, the Pix `config` would define inert, platform-rendered fields (`image` as `data:`/`https:` shown as an image only, `code` as display text, `expires_at`) with no HTML/script/frame to load and nothing executable; the only loadable field would be an origin-allowlisted `instructions_url` fallback. Resolution itself stays the out-of-band poll on Get Checkout.

The existing worked example (Student Verification's `verification_url`) already covers the URL-in-config shape, so Pix's additive value is being a **pure-render, no-loadable-URL** example — happy to contribute it as one if useful. (I won't presume how 3DS/DDC lands, since that shape is still being worked in #517/#458.) I'm also following the thread on whether `required` stays for the first release, since it touches how a render Action signals gating.

drewolson-google (approved):
LGTM after our discussing in the Tech Council. Looks like there are some failing checks, but otherwise I'm aligned with these changes.

docs/specification/catalog/index.md
igrigorik: Good catch! Addressed in https://github.com/Universal-Commerce-Protocol/ucp/pull/582/commits/afe5e0401aaadf5033e10a9340a10b61cca85110

docs/specification/checkout.md
igrigorik: Clarified in https://github.com/Universal-Commerce-Protocol/ucp/pull/582/commits/60f593ff25deddb101959a554d0084cdbf9b9e04

docs/specification/catalog/index.md
igrigorik: Clarified.

docs/specification/checkout.md
igrigorik: 👍  adopted in https://github.com/Universal-Commerce-Protocol/ucp/pull/582/commits/60f593ff25deddb101959a554d0084cdbf9b9e04

docs/documentation/schema-authoring.md
igrigorik: 👍

amithanda (approved):
Thanks for all the great work on this. Approving! Over to the next PR to start defining actions using this core primitive 👍

raginpirate (approved):
Thanks a ton for taking this across the finish line Ilya ❤️

PR #584, opened 2026-07-13
feat: add payment instruments to the order capability

richmolj:
# Description

Order confirmation pages typically want to show how the buyer paid — card vs gift card, last four, billing address — but the Order capability has no payment field today. This adds `order.payment`, mirroring the checkout `payment` object but reduced to the subset of fields relevant at order time (dropping collection-time details like handler reference, credential, and selection state).

`order_payment_instrument` follows the existing order-subset pattern (`order_line_item`, `order_confirmation`): a standalone read-only sibling of the checkout type carrying only `id`, `type`, `amount`, `billing_address`, and an open `display` block. `type` is an open string and `display` is left open so businesses snapshot whatever they present to the buyer.

## Category (Required)

## Checklist

Rendered `Order Payment Instrument` schema table:

| Name | Type | Required | Description |
| :--- | :--- | :--- | :--- |
| id | string | **Yes** | Identifier for this instrument instance. |
| type | string | **Yes** | The broad category of the instrument (e.g., 'card', 'gift_card'). |
| amount | Amount | No | The amount charged to this instrument, in the order currency's minor units (ISO 4217). |
| billing_address | Postal Address | No | The billing address associated with this payment method. |
| display | object | No | Display information for this payment instrument (e.g., card brand and last four digits). |


vixdug:
cc @jingyli @amithanda Now that we will land split payments (https://ucp.dev/draft/specification/split-payments/#gift-card-credit-card) in 08 this follow up allows order confirmation screens to properly display the split. This came up during implementations as there is no clear way to show the buyer what portion of the payment is captured on each method. LGTM.

raginpirate (changes requested):
Hey @richmolj! Thanks for identifying and filling this gap. I agree with the core problem: Order needs a durable, buyer-facing record of how the order was paid, especially so split payments can show each instrument’s contribution.

I forked the PR to explore a different schema boundary rather than editing your branch directly: https://github.com/Universal-Commerce-Protocol/ucp/pull/638

The goal is to preserve your end-to-end behaviour while avoiding a second payment-instrument hierarchy that the current handler prose does not extend, which we should treat as a blocking issue with this proposal. LMK what you think!

source/schemas/shopping/types/order_payment_instrument.json
dkoch74, line 1: Most of these fields would be reused any time we need to display a payment instrument summary. Instead of a having a detailed object for just this use case, it would be ideal to create something like payment_instrument_display.json and have order_payment_instrument.json inherit from that. Billing address might be overly specific for the former but could be included in the order_payment_instrument.json object.

raginpirate (approved):
Sorry I never saw you merged in the PR; thanks this looks great!

paulcarey-stripe:
Hi from Stripe. We’re supportive of adding payment information to the Order, but are wondering if a slight shift in semantics might provide a better foundation. 

We’re preparing a broad auth-and-capture proposal based on production patterns we see today including delayed and multi-capture and variable final amounts. Our proposal builds on the foundation in [#638](https://github.com/Universal-Commerce-Protocol/ucp/pull/638) with a buyer-authorised max amount and a pending/authorized/paid Order state. 

Our proposal intersects with #584 because, AIUI, its `instruments[].amount` commits the protocol to “amount charged” semantics before auth or confirmed payment. Addressing that later might require duplicative fields, or breaking semantics. Concretely, renaming to e.g. `paid_amount` might resolve the concern, but we’re still designing. We expect to publish our PR for review in the next week – would you be open to coordinating or deferring this PR until the designs can be evaluated together.

paulcarey-stripe:
I created a branch off this branch, and added a single commit that I think paves the way for separate auth and capture. Specifically, it replaces the `amount` on Order payment instrument with both `authorized_amount` and `captured_amount`. 

The commit is here: https://github.com/Universal-Commerce-Protocol/ucp/commit/4871d8b555bd74a4fd7c76567b9145ccd57a9b19

I'll raise two PRs today / tomorrow - one with complementary core changes, and a separate extension for auth and capture.

docs/specification/order.md
```diff
@@ -87,6 +87,24 @@ Expectations can be split, merged, or adjusted post-order. For example:
   (common examples: `processing`, `shipped`, `in_transit`, `delivered`,
   `failed_attempt`, `canceled`, `undeliverable`, `returned_to_sender`)
 
+### Payment
+
+**Payment** describes how the order was paid, suitable for rendering on an
+order confirmation. Like the rest of the order it reflects current state, so
+it may change post-purchase (e.g. an order edit that adds or changes a tender).
```
igrigorik: test

paulcarey-stripe:
Enhancement proposal for the core change, distinguishing between auth and capture is here.
https://github.com/Universal-Commerce-Protocol/ucp/issues/815

archrao:
With PR #584 and PR #816, 'order.payment' serves both AI agents and buyer-facing receipts. Today, buyers never see terms like 'authorized_amount' or 'captured_amount' on a receipt, those are PSP/banking concepts that confuse consumers. Therefore, the platform/UI layer must translate those fields into the amount allocated to the transaction.

1. Standard Delayed Capture (authorized = 150, captured = 0 at checkout)
At checkout, the UI/receipt reflects the authorized amount ($150) with fine print such as "Your card will be charged when items ship." In 2–3 days, once fulfilled and captured (authorized = 0, captured = 150), the receipt updates to, "Card charged $150."

2. Partial Shipment or Out-of-Stock (authorized = 150, captured = 0 at checkout)
In 2–3 days, suppose $120 of items ship (captured = 120)
- If the remaining $30 item is backordered (authorized = 30, captured = 120), the UI can display something like, "Charged $120 ($30 pending shipment)."
- If the $30 item is out-of-stock and canceled (authorized = 0, captured = 120, with a $30 cancellation in order.adjustments[]), the receipt could display, "Card charged $120 ($30 item canceled / hold released)."

In short, to avoid confusing the user, the UI layer simply needs presentation logic that masks the raw authorized_amount and captured_amount fields. Given this:
1. Is it fair to say that a separate amount field is not needed, and we can support all receipt and agent use cases using authorized_amount and captured_amount (alongside adjustments)?
2. Since PR #816 builds on PR #584's handler_id decoupling and introduces these two amount fields, should we look to merge/consolidate the two PRs? 

Also,  In PR #816, 'maximum_amount' is currently marked 'required' on 'order.payment'. However, for offline/in-store POS orders ingested via UCP Order webhooks, there is no prior online UCP Checkout session (and post-swipe tips allow final capture to exceed initial tap auth). Should 'maximum_amount' be made 'optional' on Order which did not originate from an online checkout session?

westonale:
On question 1: the two fields plus adjustments cover the receipt, because presentation logic can absorb an ambiguity a reader never sees. For the agent case I think they come up short in two places.

**Pending.** Example 2 reads "$30 pending shipment" off `authorized_amount` = 30, which works only because that example captures once. #815 does not let you treat `authorized_amount` as the amount still capturable in general: it defines the field as "authorization currently effective", keeps "remaining capture capacity" as a separate concept, and requires platforms to "accept `captured_amount` greater than `authorized_amount`" because "authorization reports current state, while capture reports total activity". Once there are partial or multiple captures, `authorized_amount` is no longer what can still be charged. That quantity, how much more can move on the order, is capacity, which #815 says Core does not represent and platforms "must not infer by subtracting the reported fields"; #820 adds it as `capturable_amount` for exactly this reason. A receipt survives the gap because a person reconciles from the merchant's messages. An agent deciding whether to wait, re-order or dispute needs that number, and Core does not carry it.

**Net.** Neither example covers a refund after capture. Take example 1 after delivery, with a $30 refund: `captured_amount` stays at 150, because it is defined before refunds, and the refund lives in `order.adjustments[]`. What the buyer is actually out is 120, and no field carries it. Getting there means subtracting refund adjustments from capture across two objects, and neither #815 nor #816 says whether a platform may do that or should expect the Business to report it. A receipt can do that subtraction silently. An agent deciding whether to re-order, escalate or open a dispute is acting on exactly that number.

So I would keep the two fields and add one normative sentence: either a platform may derive net collected from `captured_amount` and refund adjustments, or the Business reports it and the platform must not derive it. Either answer works for agents. Today neither is stated.

PR #585, opened 2026-07-14
docs: clarify business-populated response values for identity-linked sessions

jamesandersen:
# docs: clarify business-populated response values for identity-linked sessions

## Motivation

This is a focused, docs-only alternate to #519. That thread converged on a
clear conclusion (thanks @karangoel16 for raising it, and @raginpirate /
@amithanda for the framing): surfacing business-owned user state — saved payment
instruments, loyalty membership, buyer profile — does **not** need a new
extension or a new schema field. It is already possible through existing base
entities. The real gap is **discoverability**: today the mechanic is implicit,
spread across the `ucp_request` annotations, the loyalty extension, and identity
linking, and an implementer is unlikely to connect those on their own.

Rather than add a standalone concept page (which risks becoming a second source
of truth that drifts from the annotations that actually govern behavior — see
@amithanda's point in
https://github.com/Universal-Commerce-Protocol/ucp/pull/519#issuecomment-4867256323),
this PR lands the same intent as small edits anchored to the specs that already
own each concept.

## What this PR changes

- **`identity-linking.md`** — adds a `### Business-Populated Response Values`
  subsection right after Access Levels (where the auth model already lives). It
  names the pattern: fields annotated `ucp_request: "omit"` are business-populated
  by definition; fields the platform MAY also supply (`optional`/unannotated) MAY
  additionally be populated by the business from stored user state on a response.
  Identity linking is what makes those values user-specific.
- **`checkout.md`** — adds a pointer in the Payments section explaining that
  `payment.instruments` may carry display-safe, business-scoped saved instruments
  (opaque `id` + `display`, no raw credential), links to the existing
  create-checkout REST example that already returns this shape, and to Identity
  Linking for the access levels/scopes that gate it.
- **`payment_instrument.json`** — corrects the `id` description. It previously
  said "assigned by the platform," which is wrong for an identity-linked saved
  instrument: there the business assigns the `id` and the platform MUST treat it
  as an opaque, business-scoped reference.
- **`loyalty.md`** — one-line cross-link from the existing "business populates
  loyalty from authenticated identity" paragraph to the new subsection, so the
  loyalty case is explicitly framed as an instance of the general pattern.

## What this PR deliberately does not do

- **No standalone page.** No `business-owned-response-values.md`; no `mkdocs.yml`
  nav entry.
- **No new schema fields.** No `payment.saved_instruments[]`,
  no `credential_state` enum, no generic tokenized-node abstraction.
- **No unencrypted-credential example.** The response example in the original
  draft that transmitted a raw credential is omitted, consistent with the
  [payment handler security guidance](https://ucp.dev/latest/specification/payment-handler-guide/#security_1)
  (and the review comments from @jamesandersen / @dkoch74).
- **Saved shipping addresses are out of scope.** Populating
  `fulfillment` from stored user state is trickier and belongs in its own PR.

## Follow-ups (intentionally separate)

These surfaced on the thread and are worth their own enhancement proposals given
the PII implications:

1. A generic reference/tokenized-node abstraction to mark any response node as an
   opaque business-owned reference (@raginpirate / @rosew).
2. A machine-readable way to distinguish a ready-to-resolve saved reference from
   an instrument that still needs credential collection — credential absence is
   ambiguous between the two, and `id` naming is the signal we agreed not to lean
   on (@karangoel16's `credential_state` idea).

## Category (Required)

- [x] **Documentation:** Updates to README, or documentation regarding schema or
  capabilities.

## Checklist

- [x] I have followed the Contributing Guide (Conventional Commits title, no
  breaking changes).
- [x] My changes pass local linting (cspell, markdownlint). *The full mkdocs
  build could not be run locally; relying on CI (`docs.yml`) as authoritative.*
- [ ] I have added tests. *Docs + one schema-description clarification; no
  behavior change.*
- [x] (For Core/Capability) The one schema change is a `description`-only
  clarification on `payment_instrument.id`; no structural or contract change.

## Verified claims

The annotation statements in the identity-linking subsection were checked
against `source/schemas/shopping/`:

- `totals`, `messages`, `order` are `ucp_request: "omit"` in `checkout.json`.
- `buyer` is `ucp_request: {create: optional, update: optional, complete: omit}`.
- `payment.instruments` carries no `ucp_request` annotation in `payment.json`
  (bidirectional by default).


dkoch74 (approved):

source/schemas/shopping/types/payment_instrument.json
```diff
@@ -12,7 +12,7 @@
   "properties": {
     "id": {
       "type": "string",
-      "description": "A unique identifier for this instrument instance, assigned by the platform."
+      "description": "A unique identifier for this instrument instance. Typically assigned by the platform for instruments it collects. For a business-owned saved instrument returned on an identity-linked response, this identifier is assigned by the business; the platform MUST treat it as an opaque, business-scoped reference and resolve it server-side."
```
amithanda: Should we reassign the actor on "resolve it server-side"? As written, the platform is the subject of both verbs: it "MUST treat it as an opaque, business-scoped reference and resolve it server-side." But resolving the reference is precisely the one thing the platform cannot do here: the reference is opaque to it by construction, and the whole design intent is that only the business can turn id back into a real instrument. So the sentence asks the platform to do the impossible in the same breath it tells it the value is opaque.

Could we split the two responsibilities so each lands on the right party?


```suggestion
      "description": "A unique identifier for this instrument instance. Typically assigned by the platform for instruments it collects. For a business-owned saved instrument returned on an identity-linked response, this identifier is assigned by the business; the platform MUST treat it as an opaque, business-scoped reference, and the business resolves it server-side when the buyer selects it."
```

That keeps the platform's obligation (treat as opaque) and the business's obligation (resolve) distinct.

amithanda (approved):

docs/specification/identity-linking.md
```diff
@@ -72,6 +72,31 @@ either way. If their profile *also* lists identity linking with
...
+Subject to the scopes that gate the operation, a business **MAY** return the
+user's loyalty membership in `loyalty` (see [Loyalty](loyalty.md)), saved
+payment instruments in `payment.instruments[]` (see [Checkout](checkout.md)), or
+buyer profile data in `buyer`.
+
+Businesses **MUST** only return values that belong to the authenticated user and
+are appropriate for the current transaction. Platforms **MUST** treat returned
+business-owned identifiers as opaque and business-scoped, and **MUST NOT** infer
+raw credentials from a display value. Businesses **SHOULD** omit this state when
+the request lacks the required user authorization.
```
raginpirate: nit: small tweak to this section, as right now the MUST feels like it overreaches: we do not mean to only surface authenticated user values, but instead to omit them if you are NOT authenticated.

```suggestion
Businesses **MUST NOT** return stored user-specific state unless the request is
user-authenticated and authorized for the operation. They **MUST** only return
values that belong to the authenticated user and are appropriate for the
current transaction, and **SHOULD** limit them to the data needed for that
transaction. Platforms **MUST** treat returned business-owned identifiers as
opaque and business-scoped, and **MUST NOT** infer raw credentials from a
display value.
```

docs/specification/identity-linking.md
```diff
@@ -72,6 +72,31 @@ either way. If their profile *also* lists identity linking with
...
+values the platform supplied. Two kinds of fields support this:
+
+* Fields annotated `ucp_request: "omit"` are business-populated by definition:
+    the platform does not send them and the business returns them (for example
+    `totals`, `messages`, and `order`).
+* Fields the platform **MAY** also supply (annotated `optional`, or unannotated)
+    **MAY** additionally be populated by the business from stored user state on
+    the response (for example `buyer` and `payment.instruments[]`).
+
+Identity linking is what makes these values user-specific rather than generic.
```
raginpirate: nit: small refactor here to focus on semantic ownership, for example I found the double-MAY that was written here hard to interpret.

```suggestion
values the platform supplied.

Request annotations determine whether the platform may send a field; they do
not assign semantic ownership of its value. A field annotated
`ucp_request: "omit"` is excluded from requests and, when permitted and present
on a response, is populated by the business (for example, `totals`, `messages`,
and `order`). Fields annotated `optional` for an operation, or left unannotated,
may also be supplied by the platform in applicable requests. Where the response
schema and the field's semantic contract permit business-owned user state, the
business **MAY** populate it from stored state on the response (for example,
`buyer` and `payment.instruments[]`).

Identity linking provides the authenticated user context for these values.
```

raginpirate (approved):
Thanks for defining this missing prose @jamesandersen! I had a couple non-blocking doc nits.

Issue #131, opened 2026-01-28
[Feat]: internal specification structural improvement (openapi refactor)


oericdacosta:
### Is your feature request related to a problem? Please describe.

Yes, the current [openapi.json](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/services/shopping/openapi.json) specification suffers from significant code duplication and incomplete documentation.

I'm finding it the same list of ~10 header parameters (like `Authorization`, `X-API-Key`, `Request-ID`) is repeated for every single endpoint. This creates unnecessary noise and increases the risk of inconsistencies.

Additionally, the API documentation is missing formal schema definitions for error responses (4xx, 5xx), which leaves consumers guessing about the error payload structure.

### Describe the solution you'd like

I would like to refactor [openapi.json](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/services/shopping/openapi.json) to:
1.  **Deduplicate Parameters:** Extract the common header parameters into shared components (`#/components/parameters`) or parameter groups, and reference them in each endpoint. This will reduce hundreds of lines of redundant code.
2.  **Standardize Error Responses:** Create a reusable `Error` schema in `components/schemas` (based on the existing [message_error.json](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/shopping/types/message_error.json) and explicitly define response mappings for standard HTTP errors (400, 401, 403, 500) across all operations.

### Describe alternatives you've considered

I considered leaving the specification as is, but this ensures that technical debt will grow linearly with every new endpoint added, making the codebase increasingly hard to manage.

### Additional context

This refactoring focuses on the [source/services/shopping/openapi.json](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/services/shopping/openapi.json) file and leverages the existing [source/schemas/shopping/types/message_error.json](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/shopping/types/message_error.json) to define the error structure without introducing breaking changes to the protocol logic.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #589, opened 2026-07-15
feat!: Introduce Location Search + Lookup capabilities

jingyli:
## Description

A mirror copy of #545  that supersedes it.

Defines standard interfaces for discovering, searching, and retrieving physical locations (e.g., retail stores, restaurants, warehouses, lodging properties). 

It introduces two new capabilities under the `dev.ucp.common` namespace (consistent with Shopping's Catalog capability design):
1. **Location Search (`dev.ucp.common.location.search`)**: Discovery-focused endpoint combining free-text queries, explicit spatial relations (`distance`, `serves`), and structured filters (`hours`, `amenities`, `items`).
2. **Location Lookup (`dev.ucp.common.location.lookup`)**: Resolution-focused endpoint supporting single & batch lookups with per-identifier input correlation, optionally refined by the same relations and filters as Search.

Some key commerce flows it will be able to unlock:
- **Local Pickup Discovery**: Finding nearby Locations such as retail stores or restaurant branches
    that support Buyer pickup, checking their operating hours, and discovering where
    selected items are currently available before selection.
- **Fulfillment Area Verification**: Checking if a specific location (e.g., utility depot, restaurant,
    or local service provider) can serve a Buyer's address (the `serves` relation).

## Category (Required)

## Related Issues

This is related to RFC #375's section 10.

## Checklist

location_ucp_doc


igrigorik (changes requested):
Overall, the shape and proposal make sense—ty for scaffolding this. A few design questions I'd like to work through...

### 1. Should `common` be the service, or the namespace containing a Location service?

The capability names are scoped to Location:

```text
dev.ucp.common.location.search
dev.ucp.common.location.lookup
```

The profile examples, however, advertise the transport as:

```json
{
  "services": {
    "dev.ucp.common": [{ "...": "..." }]
  }
}
```

I agree with `dev.ucp.common.*` as the namespace for cross-vertical primitives. However, I'm not sure about making `dev.ucp.common` one catch-all service: every future shared capability would then share an endpoint, transport schema, etc.

The alternative split is:

```text
namespace:    dev.ucp.common.*
service:      dev.ucp.common.location
capabilities: dev.ucp.common.location.search
              dev.ucp.common.location.lookup
```

Do we think a catch-all `.common` service is preferred to scoped services?

### 2. Is `geofence_radius` sufficient to model the domain?

The current `geo` object combines a physical point with a circular `geofence_radius`, while `geofence_point` asks whether a requested point falls inside that circle. This covers a useful simple case, but in my experience service areas are often irregular polygons, disjoint regions, postal-code unions, etc.

Do we need to model (thin, hopefully) RFC 7946-based service-area shape supporting `Polygon` / `MultiPolygon`?

### 3. Can we adopt a Schema.org-inspired shape instead of inventing a bespoke one?

The latest changes make the custom model deterministic by defining `24:00`, precedence between `is_closed` / `is_24_hours` / intervals, and a two-entry convention for overnight hours. That is an improvement, but it also highlights how much new protocol syntax and interpretation logic we are defining.

Schema.org's `OpeningHoursSpecification` already provides the core semantics we need:

- `dayOfWeek`, `opens`, and `closes`;
- multiple entries for split shifts;
- `closes < opens` means the interval spans the next day;
- no `opens` means closed;
- `validFrom` / `validThrough` cover date-specific exceptions.

A Schema.org-inspired UCP shape could look like:

```json
{
  "timezone": "America/New_York",
  "hours": [
    { "day_of_week": "monday", "opens": "09:00", "closes": "17:00" },
    { "day_of_week": "monday", "opens": "18:00", "closes": "22:00" },
    { "day_of_week": "friday", "opens": "22:00", "closes": "02:00" }
  ],
  "exception_hours": [
    {
      "valid_from": "2026-12-24",
      "valid_through": "2026-12-24",
      "opens": "09:00",
      "closes": "14:00"
    },
    {
      "valid_from": "2026-12-25",
      "valid_through": "2026-12-25"
    }
  ]
}
```

The contract is compact: each entry describes one open interval; repeated entries for the same day represent split shifts; a `closes` value earlier than `opens` spans midnight; and a date-bounded exception without `opens` means closed for that period. `timezone` supplies the interpretation context for local times. The example adapts Schema.org's semantics to UCP's snake_case naming and `HH:MM` convention rather than copying its JSON-LD representation verbatim.

Is there a concrete requirement that `OpeningHoursSpecification` cannot model? If not, I would align to it and remove the custom closed/24-hour flags, precedence rules, and overnight splitting convention.

### 4. Is `offerings.inventory.quantity` an inventory-disclosure query?

The store-finder use case—"which nearby location can fulfill the item I need?"—is valuable. A hard minimum quantity filter, however, lets a caller repeatedly probe thresholds and approximate a store's stock level even when the response never returns a count.

I suggest framing this as Buyer demand and coarse availability, not inventory disclosure:

- reuse UCP's existing availability semantics rather than parallel inventory concept;
- treat requested quantity as intent that a Business may coarsen;
- change language to not imply that exact on-hand counts are returned;

### 5. What does `offerings` mean? 😅 

Modelled request can filter on:

```json
{
  "offerings": {
    "amenities": ["..."],
    "inventory": [{ "id": "..." }]
  }
}
```

The standardized Location entity, however, exposes neither amenities nor offerings. Although the filter contract implies that each returned Location satisfies the requested values, a Platform cannot render the matched facts, inspect additional amenities, or present structured item availability from the response without relying on vendor-defined fields. There is also a category question: amenities are relatively static Location characteristics, while item availability is dynamic and tied to another Catalog or menu identity. Grouping both under `offerings` may be convenient structurally, but does not necessarily make them one semantic concept.

My inclination would be to put static `amenities` on Location, model dynamic availability separately using the existing UCP availability vocabulary, and flatten the filter unless `offerings` gains a clearer cross-vertical contract.

### 6. What amenity vocabulary is interoperable across Businesses and verticals?

An open string is the right wire type for UCP, but the current proposal defines no well-known values. Should UCP publish a small open vocabulary for interoperable matching, and which initial values are important enough to standardize?

### 7. What are the capability-specific security and privacy requirements?

Location handles Buyer location hints, enumerates physical sites, describes service coverage, and can be used to probe per-location item availability. The generic Signals reference and current privacy note do not cover those capability-specific risks.

I think the specification needs explicit Security and Privacy Considerations covering:

- coarse-by-default Buyer location and progressive disclosure;
- purpose limitation and retention of location inputs;
- store/site enumeration and rate limiting;
- inventory and availability probing;
- disclosure of internal, private, or non-buyer-visible locations;
- disclosure of precise service-area geometry.

### 8. What is the bounded Location projection for Catalog and Checkout?

This PR points Checkout’s pickup destination directly at the rich common Location entity. That couples Checkout to the complete Location discovery model: `geo`, hours, service area, amenities, and every future Location extension would automatically accrete into the transactional response. I don't think that's right.

In Catalog Fulfillment (#507), we deliberately introduced a thin Location projection: a stable `location` ID plus method `description`. This avoided embedding an N-store matrix in product results and deferred richer Location facts to a separately negotiated capability. That boundary still makes sense, but I think there is a middle ground: split Location into a bounded base and an extended discovery entity.

```text
Base Location
  id
  name
  address?
      │
      └── allOf → Extended Location
                    geo
                    hours
                    exception_hours
                    timezone
                    amenities
                    service_area
```

Catalog and Checkout would use the bounded, buyer-renderable base. Location Search/Lookup would return the extended entity when the service is negotiated. Platform requests would select a Location by stable `id` rather than asserting Business-owned name/address facts. This preserves the bounded Catalog model from #507, keeps Catalog and Checkout independently renderable, and prevents the full Location discovery model from accreting into every Checkout response.

amithanda:
+1 on scoping the service to `dev.ucp.common.location`. The thing that pushes me here is what the profile can still express once `common` has more than one domain in it, since the service key is what determines how many endpoints and how many version trains a business gets.

**The setup.** Suppose `common` grows the way the name invites, say Location plus Reviews plus Identity linking. Under a catch-all service the profile can only say:

```json
"services": {
  "dev.ucp.common": [{
    "version": "2026-04-08",
    "transport": "rest",
    "schema": ".../services/common/rest.openapi.json",
    "endpoint": "https://business.example.com/ucp"
  }]
}
```

One endpoint, one transport document, and one version covering store lookup, review retrieval, and identity linking. That seems to make three fairly ordinary deployments unexpressible:

1. **Routing.** Store locator data is usually served by the merchant's own site infrastructure, while reviews are very often served by a third-party review platform on that vendor's domain, and identity linking sits behind the IdP. With one `endpoint` per transport per service there is no way to say that. The workaround would be a per-capability endpoint override, which is scoped services under a different name.
2. **Versioning.** The service entry carries its own `version`. A Reviews field change bumps the shared service version and emits a churn signal to every Location implementer even though nothing about Location moved. Domains that are operated by different parties rarely share a release cadence.
3. **Operational policy.** A public store locator is anonymous, cacheable, and high volume. Identity linking handles tokens and wants stricter auth and tighter rate limits. One base URL for both forces a single shared policy, or path-based policy that the spec does not describe.

The scoped form expresses all three directly, including the case where two domains happen to share a host:

```json
"services": {
  "dev.ucp.common.location": [{
    "version": "2026-04-08", "transport": "rest",
    "schema": ".../services/common/location/rest.openapi.json",
    "endpoint": "https://stores.business.example.com/ucp"
  }],
  "dev.ucp.common.reviews": [{
    "version": "2026-01-11", "transport": "rest",
    "schema": ".../services/common/reviews/rest.openapi.json",
    "endpoint": "https://ucp.reviews-vendor.example.com/business-123"
  }],
  "dev.ucp.common.identity": [{
    "version": "2026-04-08", "transport": "rest",
    "schema": ".../services/common/identity/rest.openapi.json",
    "endpoint": "https://business.example.com/ucp"
  }]
}
```


**The asymmetry I think settles it:** scoped services can emulate the catch-all, but the catch-all cannot emulate scoped. A business serving everything from one gateway just points every scoped entry at the same `endpoint`, as Identity does above, and pays a few redundant lines of generated JSON. A business that needs reviews served by a vendor and locations served in-house has no move available under the catch-all until the spec itself changes. Where one option is a strict superset of the other in expressive power and the only cost is profile verbosity, is verbosity not the cheaper thing to pay? Profile JSON is generated and fetched once, while endpoint topology is a deployment constraint that cannot be refactored away.

**Would it be worth writing the rule down explicitly, something like: a service is the unit of deployment, not the unit of naming?** A new service is warranted whenever a domain could plausibly be operated by a different party, on different infrastructure, or on a different release cadence than its siblings.

I like that framing because it explains the existing corpus rather than contradicting it. `dev.ucp.shopping` stays one service correctly, since cart, catalog, checkout, and orders share a transactional session and one commerce backend, so the cohesion is real. `dev.ucp.common` fails that test close to by construction, since "common" is defined as whatever is cross-vertical, which is nearly the opposite of a deployment boundary. On that reading the Catalog precedent this PR follows is not being overturned, just scoped: Catalog rides the shopping service because shopping is cohesive, not because "one service per namespace" is the rule. `common` looks like the first namespace where naming and deployment come apart, which is why it seems worth deciding deliberately rather than inheriting the Catalog shape by default.

amithanda:
### For suggestion 2. above could we just drop the circle instead of upgrading it to polygons?

`geofence_radius` only appears in one place that matters, the Location response. Both filter inputs (`distance.center` and `geofence_point`) send just a point, as the `rest.md` examples show. And `geofence_point` is a predicate the Business evaluates server side, so it behaves identically whether coverage is a circle, a polygon, a postal code list, or a drive-time isochrone. The circle is only needed to publish the shape, never to ask the question.

That matters because publishing it as a circle is currently self-contradictory. `search.md:96` binds the filter to the circle ("locations whose circular service area (defined by `geofence_radius`) contains the specified point"), so a Business whose real area is a polygon has to either approximate it (wrong answers near the boundary, across a river or a highway) or evaluate its true shape (in which case the published radius disagrees with the filter result). Which one is authoritative is undefined, as is the case where a location publishes no radius but a `geofence_point` filter arrives.

**Proposal: drop `geofence_radius` and let coverage stay a server-evaluated predicate.**

* `geo.json` becomes a pure point, which is all three call sites actually need;
* the filter takes an intent-based name such as `serves` or `coverage`, since the question is "which locations serve this?" rather than "whose circle contains this?";
* the spec says Businesses **MAY** evaluate coverage with any internal representation and **MUST** treat their own evaluation as authoritative, so presence in the result set is the answer and no geometry is returned;
* would it be worth accepting a postal address here too? `index.md:27` frames the use case as coverage "for a buyer's address," but the filter takes only latitude and longitude, so postal-code-based Businesses have to reverse geocode back into the ZIP they already reason in.

**Why this looks better than adding `Polygon`/`MultiPolygon`:** it removes the contradiction rather than encoding it more precisely, it keeps the one representation every Business can implement, and it avoids publishing geometry that most would decline to share anyway.

jingyli:
Thanks @igrigorik and @amithanda for the extensive feedback! A lot of great points that further helped to tune the schema shape.

### 1. Should `common` be the service, or the namespace containing a Location service?

This is where I have some opposite thoughts and personally prefer a catch-all `.common` service:
- **Consistency with Existing UCP Services**: UCP currently groups capabilities into broad, monolithic service endpoints per vertical/domain. While it's a great point @amithanda brought up around specialized SaaS integrations (i.e. IdP for identity linking, 3p vendor platform for reviews, and the business's core infra for locations), I personally feel like this is not a new problem we are facing - it is essentially the same as Shopping's cart/checkout capability running on deployment A but order (part of OMS) running on a different deployment - and enters into a very interesting problem space on how can a business "delegate" out different UCP capabilities to different deployment endpoints. Some things I'm immediately thinking about on this includes:
    - Are there security guidelines UCP should be prescribing on what is acceptable and what is not (similar to the ones we have around reverse-DNS)?
    - What kind of effects will bring to signing_keys?
    - If we allow service advertisement at the capability level, then do we need a separate service representation for transport binding or can we just fold transport binding into capability negotiation as well?
- **Protocol Simplicity**: The client connects to one host and negotiates which features are supported. Especially given we are simplifying versioning strategy in #614 where transport bindings no longer has their own versioning, there would be no additional churn introduced from a catch-all service.

The argument around **operation policy** should still be supported regardless of the service advertisement pattern we adopt since business ultimately controls whether or not a request should be honoured. If identity linking operations enforce a much stricter auth & rate limiting than location operations, then business can still enforce it at runtime.

_Also a small meta-point_: If in the future we prefer the capability-level `.common` service, it's not a "structural change" for implementators (vs. a guideline/policy one) so hopefully the churn/lift will not be high.

### 2. Is `geofence_radius` sufficient to model the domain?

A great prompt and I'm aligned with @amithanda's feedback in https://github.com/Universal-Commerce-Protocol/ucp/pull/589#issuecomment-5018918903. In my mind:
- **In scope**: Modelling should answer questions like "What are the locations that can serve me as a user?"
- **Out-of-scope**: Business declaring their service area upfront for model interpretation.

PTAL at https://github.com/Universal-Commerce-Protocol/ucp/pull/589/changes/e5a0cddea4c17e6bf1755c4175f54194b1649340 for this update.

_Side plug_: I also added a quick sketch on how I think we can potentially follow-up with a non-breaking geojson modelling in the future:
```
{
  "properties": {
    "geo": { "$ref": "geo.json" }, // what we have today
    "service_area": {
      "type": "object",
      "required": ["type"],
      "properties": {
        "type": { "type": "string" },
        "circle": {
          "type": "object",
          "required": ["radius"],
          "properties": { "radius": { "type": "number" } }
        },
        "polygon": {
          "type": "object",
          "required": ["type", "coordinates"],
          "properties": {
            "type": { "const": "Polygon" },
            "coordinates": {
              "type": "array",
              "items": {
                "type": "array",
                "items": { "type": "array", "minItems": 2, "items": { "type": "number" } }
              }
            }
          }
        }
      }
    }
  }
}
```

### 3. Can we adopt a Schema.org-inspired shape instead of inventing a bespoke one?

This was a miss on my end, PTAL at https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/c4087e2de1c963cb7c42e27cafeb672f77de73e6.

### 4. Is `offerings.inventory.quantity` an inventory-disclosure query?

PTAL https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/e4d21896b5bd2227b02a85bcc3175269f1440275 - I do agree with the consistent availability vocabulary - with the exception that I've cut out some well-known values that are irrelevant as a search filter (i.e. `out_of_stock`, `discontinued`, etc.).

### 5. What does `offerings` mean?
### 6. What amenity vocabulary is interoperable across Businesses and verticals?

PTAL https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/918c1cb61e6532930d6625a616088eb92b834935 - the current well-known types are focused on fulfillment/delivery service offerings based on known-existing use case. We can continue growing the vocabulary moving forward.

### 7. What are the capability-specific security and privacy requirements?

PTAL https://github.com/Universal-Commerce-Protocol/ucp/pull/589/changes/1bd596e7574f4ddb89a49d1ed33f03d81b75d89d.

### 8. What is the bounded Location projection for Catalog and Checkout?

PTAL https://github.com/Universal-Commerce-Protocol/ucp/pull/589/changes/1ff81f3a858d6e260c148e37af61086eddca0110.

source/services/common/rest.openapi.json
```diff
@@ -0,0 +1,259 @@
...
+            }
+          }
+        }
+      }
+    }
+  },
+  "components": {
+    "parameters": {
+      "authorization": {
+        "name": "Authorization",
```
maximenajim, line 112: `Authorization` cannot be modeled as an ordinary header parameter in OpenAPI 3.1; compliant tooling is required to ignore a header parameter with that name. Because this document also has no `securitySchemes` or root/operation `security` requirements, generated clients and gateways will interpret these operations as unauthenticated. `UCP-Agent` identifies a claimed profile but does not prove who sent the request. Please model the supported authentication alternatives through `securitySchemes` and operation-level security, explicitly represent anonymous public search if that is supported, and state that authentication and authorization occur before location-ID resolution or filtering.

source/services/common/rest.openapi.json
```diff
@@ -0,0 +1,259 @@
...
+        "requestBody": {
+          "required": true,
+          "content": {
+            "application/json": {
+              "schema": { "$ref": "#/components/schemas/location_search_request" }
+            }
+          }
+        },
+        "responses": {
+          "200": {
```
maximenajim, line 48: Both operations currently document only a `200` response even though the specification defines invalid-request, authentication, rate-limit, and server-error behavior. Autonomous clients need machine-readable distinctions between terminal errors, caller-correctable errors, and retryable failures; otherwise they may retry unauthorized requests, silently abandon valid searches, or amplify an outage. Please add the shared UCP error envelope for at least 400, 401, 403, 429, and 5xx, require Retry-After where appropriate, and align the REST and MCP behavior for an oversized lookup batch. The current documentation alternates between implementation-defined limits and a 400; please choose one normative status and error code.

source/services/common/mcp.openrpc.json
```diff
@@ -0,0 +1,97 @@
...
+                "description": "URL to the platform's UCP profile document."
+              }
+            }
+          },
+          "idempotency-key": {
+            "type": "string",
+            "format": "uuid",
+            "description": "Unique key for retry safety. Maps to HTTP Idempotency-Key header (optional for read-only operations)."
+          },
+          "signature": {
```
maximenajim: This introduces a detached JWS in the MCP JSON metadata, while UCP specifies RFC 9421 HTTP Message Signatures for HTTP transports, including MCP Streamable HTTP. These mechanisms are not wire-compatible and the new field leaves canonicalization, covered fields, key resolution, and replay parameters underspecified. Please remove the JSON-level detached JWS and use `Signature-Input`, `Signature`, and `Content-Digest` at the HTTP transport layer, consistent with the other UCP services. If a non-HTTP MCP transport is introduced later, its signing model should be specified independently rather than overloading this metadata field.

source/schemas/common/location_search.json
```diff
@@ -0,0 +1,60 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/common/location_search.json",
+  "name": "dev.ucp.common.location.search",
+  "title": "Location Search",
+  "description": "Location search capability. Supports natural language queries, structured filtering, and pagination.",
+  "type": "object",
+  "$defs": {
+    "search_request": {
```
maximenajim, line 9: The schema currently accepts {}, a pagination-only request, a context-only request, an empty query, and filters: {} even though the normative text says a request must contain a query or filters. Different servers will therefore accept, reject, or reinterpret the same schema-valid request. Please enforce the intended rule in JSON Schema—for example, an anyOf requiring either a non-whitespace query or a non-empty, valid filter—and add corresponding negative conformance cases. If context-only “near me” browsing is intended, that should instead be made an explicit normative operation with deterministic behavior rather than left to implementation-specific validation.

source/schemas/common/types/location_filter.json
```diff
@@ -0,0 +1,79 @@
...
+  "$id": "https://ucp.dev/schemas/common/types/location_filter.json",
+  "title": "Location Filter",
+  "description": "Filter criteria to narrow location search/lookup results. All specified filters combine with AND logic.",
+  "type": "object",
+  "properties": {
+    "geo": {
+      "type": "object",
+      "description": "Filter locations by geographic proximity or geofence coverage.",
+      "properties": {
+        "distance": {
```
maximenajim, line 12: These predicates currently permit unusable or ambiguous targets. distance may omit center and rely on context even when the request contains no usable location context, while serves accepts {} or both point and address. Please make target selection deterministic: require either an explicit distance center or a specifically defined, authenticated context source; model serves.point versus serves.address with oneOf; and define a stable error when an address cannot be geocoded. An absent or invalid target must not silently broaden the search because that could disclose or select locations outside the user’s intended geography.

source/schemas/common/types/location_filter.json
```diff
@@ -0,0 +1,79 @@
...
+              "$ref": "locality.json",
+              "description": "Coarse postal address reference of the target destination (e.g., for ZIP-code based routing)."
+            }
+          },
+          "additionalProperties": false
+        }
+      },
+      "additionalProperties": false
+    },
+    "hours": {
```
maximenajim, line 47: The containing filter says predicates are combined with logical AND, but this nested object says the behavior is implementation-defined when open_now and open_at are both present, suggesting precedence or OR behavior.

source/schemas/common/types/exception_hour.json
```diff
@@ -0,0 +1,30 @@
...
+    { "$ref": "time_interval.json" },
+    {
+      "type": "object",
+      "properties": {
+        "from": {
+          "type": "string",
+          "format": "date",
+          "description": "An ISO 8601 date representing the start of the date range."
+        },
+        "through": {
```
maximenajim: The current model has three conflicting interpretations: this description says omitting an interval means open 24 hours, the referenced time_interval schema requires open and close, and the location schema uses 00:00–00:00 as a closed-day sentinel. In addition, through is described as exclusive while the REST example uses the same date for from and through, which produces an empty range.

source/schemas/common/types/location.json
```diff
@@ -0,0 +1,51 @@
...
+        },
+        "exception_hours": {
+          "type": "array",
+          "description": "Exception hours for specific dates (holidays, closures, etc.). Support split hours through multiple entries on the same day and overnight hours (open > close). For full day closures, MUST explicitly set open and close to 00:00.",
+          "items": {
+            "$ref": "exception_hour.json"
+          },
+          "ucp_request": "omit"
+        },
+        "timezone": {
```
maximenajim, line 42: The description says timezone is required when hours or exception_hours is present, but the schema does not enforce that requirement.

source/schemas/shopping/types/fulfillment_destination.json
```diff
@@ -4,13 +4,13 @@
...
   "description": "A destination for fulfillment.",
   "type": "object",
-  "ucp_shared_request": true,
   "oneOf": [
     {
       "$ref": "shipping_destination.json"
     },
     {
-      "$ref": "retail_location.json"
+      "$ref": "../../common/types/location_base.json",
```
maximenajim: This replaces the published shopping/types/retail_location.json reference and removes that schema path. Existing implementations may have compiled clients, cached schemas, validation allowlists, or $ref resolution based on the published $id; those integrations will fail even if the new wire shape is similar.

source/services/common/rest.openapi.json
jingyli: That's a great find and I think one that is also affecting our existing openapi services - e.g., https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/services/shopping/rest.openapi.json#L808! To keep the scope of the PR focused on the new capability, I'm going to keep this as-is here but have filed https://github.com/Universal-Commerce-Protocol/ucp/issues/668 to track a quick follow-up fix that would universally apply to all openapi service definitions we have.

source/schemas/shopping/types/fulfillment_destination.json
jingyli: That's a fair callout given it's changing the semantic `$ref` link here - I've labelled the PR as breaking given it also contains a refactoring of `context.json` and `signals.json` which updated the `$id` ref.

source/schemas/common/types/location_filter.json
jingyli: Yep, and it's intended:

The logical AND is being applied at the outermost level - i.e. if `hours`, `geo`, and `inventory` are all specified, then it's an AND combination. But within `hours`, if both sub-fields are specified, then it follows the other description where behaviour is implementation-defined.

Added the keyword `outermost` to make this distinction clear - PTAL [675ec06](https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/675ec064637b5013a279e5a68f2ee8207d222595).

source/schemas/common/types/exception_hour.json
jingyli: Good catch! I actually already had a queued up fix on this in the morning but never got around to publishing it to fix the REST example. PTAL [80145dd](https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/80145ddf226960397a4bc906a6c7e731a4ed9ec6).

source/services/common/mcp.openrpc.json
jingyli: PTAL [17dac60](https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/17dac60cf5896b81b07a33e3f9deac1fecb76598).

source/schemas/common/types/location.json
jingyli: PTAL [5dfcbf7](https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/5dfcbf7304c95dedc6a3c9283b461eec02bf1611).

Also just a small callout: If this is the pattern we want to introduce in UCP schemas, then we should probably follow-up with a repo-wide vetting so we can add schema rules to enforce similar requirements (e.g., https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/shopping/checkout.json#L126).

source/services/common/rest.openapi.json
jingyli: Another great callout as I also recall seeing feedback from the community - https://github.com/Universal-Commerce-Protocol/ucp/issues/131 (section in `Standardize Error Responses` on explicitly mapping out protocol-level errors.

I'd like to keep this PR focused on the capability design and propose a follow-up to enhance all openapi definitions (e.g., https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/services/shopping/rest.openapi.json).

source/schemas/common/location_search.json
jingyli: PTAL [07f2051](https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/07f2051e25b06004081a66e4c240adcad831d953) and [41a47e6](https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/41a47e697945e492dd52cdb689995ce08b892244), I actually realize that the original language is too strict as the following use cases are being left out:
- Valid extension-based inputs: if platform & business negotiate for a vendor extension that extends this capability, then there can be other types of search inputs present in the request.
- User's contextual hints: if platform provides the user's contextual location hints, then it can be an indication that this is an "open-browse" based on the user's location.

Due to the possibility of vendor extensions, it's probably not ideal to specify this as a schema requirement. Instead, I added some implementation guidance to handle `context`-based requests.

source/schemas/common/types/location_filter.json
jingyli: PTAL [054a763](https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/054a7636f6e0ce1ed37e656902a1105227a7c8a0).

jingyli:
Thanks @maximenajim for the feedback! Responses added and requesting for another passthrough.

source/schemas/shopping/types/fulfillment_destination.json
maximenajim: Every `location_base` instance has `id` and `name`, while `shipping_destination` requires only `id` and permits unknown properties through its composed postal-address schema. A normal BOPIS destination such as `{"id":"store-1","name":"Downtown","address":{"postal_code":"55403"}}` therefore matches both branches and is rejected by `oneOf`; the current schema makes the bounded pickup projection unusable in Checkout.

Please select the destination schema from the containing fulfillment method—for example, `type: "shipping"` constrains `destinations.items` to `shipping_destination` and `type: "pickup"` constrains them to `location_base`—or add an explicit discriminator that makes the branches mutually exclusive. Add a conformance case with `id`, `name`, and `address`.

source/services/common/mcp.openrpc.json
maximenajim: `17dac60` corrected the terminology but not the protocol layer: `signature`, `signature_input`, and `content-digest` remain JSON members of `meta`. RFC 9421 defines `Signature` and `Signature-Input` as HTTP fields, and UCP's existing MCP signing guidance places those fields plus `Content-Digest` in HTTP headers and explicitly avoids a duplicate JSON representation. Implementations following this OpenRPC document would sign a different wire contract than implementations following the core signing specification.

source/schemas/common/types/location.json
```diff
@@ -0,0 +1,60 @@
...
+        },
+        "amenities": {
+          "type": "array",
+          "description": "Static features of the location (e.g., 'free_wifi', 'wheelchair_accessible', 'parking', 'curbside_pickup').",
+          "items": {
+            "$ref": "amenity_type.json"
+          },
+          "ucp_request": "omit"
+        },
+        "hours": {
```
maximenajim, line 26: The revised interval behavior is inspired by schema.org, but the wire representation still replaces its deployed vocabulary with `hours`, `exception_hours`, `day`, `open`, `close`, `from`, and `through`. Existing merchant JSON-LD uses `openingHoursSpecification`, `specialOpeningHoursSpecification`, `dayOfWeek`, `opens`, `closes`, `validFrom`, and `validThrough`; schema.org-aware generators and validators will not recognize the proposed fields, and `through` also uses a different exclusive-date convention.

maximenajim:
I completed another pass. The proposal is looking good. I added a few more follow-up comments.

igrigorik (changes requested):
Thanks Jing — we're headed in the right direction. Lots of good feedback already, I'll focus on a couple of big questions/flags I'd like us to align and iterate on...

### Let's reuse the existing Catalog and Shopping contracts

Location should inherit the established search and identifier-resolution semantics wherever they apply. Unqualified `serves(target)` should mean that any currently available method can serve the target. `inventory[].id` should remain opaque and Business-resolved, while its status predicate reuses Shopping's availability vocabulary. We should explicitly define omitted and conflicting inventory-status predicates. One forward-compatibility caveat: `serves` is currently closed and probably shouldn't be; adding a `method` qualifier later would require a schema/version change unless we preserve that extension point now.

### Q: should amenities be a reverse-domain-name keyed map?

We currently model each amenity as a string, which is simple if the contract is only “this Location has this amenity.” Did we consider—or do we want—to model amenities as a reverse-domain-name keyed map instead?

I'd like to stress-test whether we'll eventually want properties attached to an amenity, such as a subtype, restrictions, availability, or other metadata. If amenities are deliberately Boolean membership identifiers, the current string array is fine. If we expect them to become extensible assertions, an RNDS map would give us namespace ownership and room for additional properties without a later array-to-map wire break.

### Remaining hours schema and semantic issues together

The latest updates fixed timezone enforcement and some earlier exception-hour inconsistencies, but the current contract still has unresolved issues:

- `open_now` and `open_at` may both be supplied, with implementation-defined precedence or OR behavior;
- `exception_hours` uses `00:00`–`00:00` for full-day closure, leaving no unambiguous 24-hour-open representation and no defined equal-time behavior for regular hours;
- exclusive `through` differs from Schema.org's inclusive `validThrough` without an explicit conversion rule;
- regular versus exception precedence, overlapping exceptions, split and overnight intervals, and closure behavior are not defined as one evaluation model;
- the renamed UCP fields do not yet have a mechanical mapping to the corresponding Schema.org vocabulary.

Building on Maxime's existing thread, I think these should be resolved as one atomic schema, prose, and examples change rather than adjusting the pieces independently—potentially as a standalone stacked PR.

### Separate the existing Fulfillment destination debt from the Location capability

This is a big one, because it points to existing spec debt. Maxime's `oneOf` comment exposes an existing Fulfillment problem: the shipping and Business Location destination branches overlap, the request projection does not reduce Location selection to a stable ID reference, and the surrounding contract does not guarantee that a discovered Location is accepted by the applicable Fulfillment method.

Replacing `retail_location.json` with `location_base.json` does not by itself solve those issues, and it also changes an existing published path. I recommend handling this as a separately reviewed breaking Fulfillment refactor—discriminator, bounded response projection, ID-based request selection, identity guarantee, and compatibility plan—that #589 can then build on.

igrigorik:
@jingyli opened https://github.com/Universal-Commerce-Protocol/ucp/pull/687 as a branch PR to address the hours feedback I flagged above, ptal. cc @gsmith85

source/schemas/common/types/location.json
jingyli: Thanks for the callout - this is being worked on as part of #687.

source/services/common/mcp.openrpc.json
jingyli: Good callout! For the purpose of this PR, I will be removing these fields from the `meta` object. But while looking at this comment a bit more closely, I'm curious if there is a potential area for improvement for the following:
1) **Duplicated data presence in MCP**: UCP's existing MCP signing doc (https://ucp.dev/draft/specification/signatures/#mcp-transport) references `UCP-Agent` and `Idempotency-Key` as standard HTTP headers, but we also have this data communicated as part of the transport-agnostic `meta` runtime context. Any reason why we need to have these data only in both places?
2) **Lack of codification on supported HTTP headers in MCP**: Unless a developer went through the documentation in details, it's very easy for an implementor to miss out on what is already available (making the same mistake as me..)

Throwing out some thoughts/ideas for consideration:
- _[Preferred?]_ Specification extension to callout `headers` directly in the `servers` section: Gateway/client proxies can choose to parse for them (via custom implementation) OR ignores them.
- Revisit contract and consolidate everything in the runtime `meta` context.
- Have a clear supported `Supported Headers` section for MCP transport in ucp.dev (similar to what we have in REST)

jingyli:
Thanks @maximenajim and @igrigorik for this latest round of feedback!

### Let's reuse the existing Catalog and Shopping contracts

PTAL at [544413f](https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/544413f8e5e012fe068a38b4056abde463b38bbc) and [3ebd1b2](https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/3ebd1b2d39c0b5db053f615bc225b4afc392b076).

### Q: should amenities be a reverse-domain-name keyed map?

PTAL [8178018](https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/8178018c74f173d521dec4f79904c8b3cd11de0b). I agree with the reverse-DNS approach, especially given that the same amenity name may actually mean very different things depending on the industry is in (namespace clashes). Some notable examples:

- `drive_through`: in food ordering it means handing off the order from kitchen as the order comes in, but in automative services, it generally means the convey or tunnel for a car wash.
- `fitting_room`: in retail it means the actual stall for trying on apparel clothing but in logistics it means something entirely different.

However, I think we are too early to think about modelling for a more structured/complex representation. The starting use case in my mind right now is to understand whether the amenity is available/offered at a location and nothing more granular. If the need arrives in the future, we can explore defining the schema to model these amenities (and a potential non-breaking way to add it into our existing modelling can be to add a dedicated `amenities_detail` reverse-DNS map).

### Remaining hours schema and semantic issues together

Thanks for helping to draft #687, now landed and should address this concern here!

### Separate the existing Fulfillment destination debt from the Location capability

Both have pointed this out and I agree with the gap here. My proposal is in alignment to the recommendation of `handling this as a separately reviewed breaking Fulfillment refactor`. What my proposal here is to decouple the scope so this PR purely looks at introducing the new location capability. PTAL at [f687a65](https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/f687a657ec070118a32ebcaf7c7af2643fc35fab) and [1fdd791](https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/1fdd7910ff1fef738c55001eaaf8b186c5c08c28).

_tl;dr_:
- I have mirrored the definition of #688's `common/types/location_summary.json` in `common/types/location_base.json`, which is then referenced by `common/types/location.json`.
- I reverted all edits to `shopping/types/fulfillment_destination`'s schema so it can be addressed properly in #688.
- Acknowledged that temporarily UCP will hold multiple definitions of a minimalistic location representation (`shopping/types/retail_location.json` and `common/types/location_base.json`).

source/schemas/shopping/types/fulfillment_destination.json
jingyli: Acknowledged and responded in last section of https://github.com/Universal-Commerce-Protocol/ucp/pull/589#issuecomment-5289053872 - will resolve this comment so we can take the conversation there and thanks!

maximenajim (approved):
The design is materially stronger than the initial draft. Thanks for the updates.

vishkaty:
Three small consistency items at 1fdd791. The Location section in fulfillment.md now renders common/types/location.json while fulfillment_destination.json still validates shipping_destination or retail_location, so the published Fulfillment doc and the wire schema describe different entities; reverting that leftover hunk also avoids a textual conflict with #688, which edits the same lines. Smaller: index.md still describes geofence_point coverage checks that no schema or binding defines (the shipped filter is serves), and serves accepts both {} and point plus address together, which search.md leaves undefined; anyOf requiring point or address plus one precedence sentence would close it.

igrigorik (changes requested):
Great work and progress! Reviewing again with fresh eyes, a few things...

##  Let's reconcile Location with the landed fulfillment contract

#688 landed after this branch was last updated, so #589 should consume that contract rather than retain its temporary parallel representation:

- compose rich `common/types/location.json` from `common/types/location_summary.json` and remove `location_base.json`;
- take #688's Fulfillment schemas and documentation, including `location_destination`, `location_summary`, and the source-agnostic selection lifecycle;
- retain the existing `ucp_shared_request` annotations instead of removing them as unrelated cleanup;
- clarify that a Location ID returned by Search or Lookup may be submitted as `selected_destination_id` on an applicable method, while the method's `type` identifies the fulfillment mode and the Business revalidates the selection.

This leaves #688 as the single authority for the Checkout projection and selection lifecycle.

## make `serves` a first-class Location Search query?

Spicy: I think we're abusing/overusing filters. `serves` asks the core retrieval question “which Locations serve this target?” Filters should be used to narrow candidate Locations by properties such as hours, amenities, and inventory. I think we should hoist `serves` out of `filters.geo` and making it a sibling of `filters`.

```json
{
  "serves": {
    "point": { "latitude": 37.422, "longitude": -122.084 }
  },
  "filters": {
    "hours": { "open_at": "2026-08-18T18:00:00Z" },
    "amenities": ["dev.ucp.amenity.shopping.curbside_pickup"]
  },
  "pagination": { "limit": 10 }
}
```

`serves` is independently sufficient to make this request meaningful.

As a non-blocking design question: do we want the base Location capability to offer free-text `query` at all, or should natural-language retrieval be extension-defined? Either answer can coexist with `serves`; I do not think that question needs to block fixing the structured query shape.

If method-specific narrowing is introduced later, it belongs in `filters` as a sibling attribute filter. When present with `serves`, it narrows the method set used to evaluate serviceability; without it, `serves` means that at least one currently available method can serve the target.

**As a related bug/issue:** current schema instead accepts an empty target, an empty address, or contradictory targets such as a Seattle point paired with a New York address:

```json
{
  "serves": {
    "point": { "latitude": 47.6062, "longitude": -122.3321 },
    "address": { "address_country": "US", "postal_code": "10001" }
  }
}
```

One Business could evaluate the point, another the address, and a third require both. We should define what we want the behavior to be. I believe the intended contract here is ~oneOf but we also don't want to use that to close extensibility. If so, one idea to explore is to model it as single-entry map: `minProperties: 1` and `maxProperties: 1`, with `point` and `address` properties. This enforces exactly one target without a closed `oneOf` or an additional `target` wrapper.

## Clarify Location browse and pagination

What happens when serves is omitted or is an empty object? This is the "browse" flow.  My expectation is that the Business returns up to the requested or default page size, ordered by its relevance policy after applying filters. It may use request `context`, observed signals such as coarse IP-derived locality, or a Business-defined default set. These hints do not establish a distance boundary; a Platform supplies an explicit `distance` query when it needs one.

Location Search already supports `pagination.limit`, opaque request/response cursors, and `has_next_page`, so browse is paginated rather than a fixed or unbounded “all locations” result.

## Availability vocabulary

`inventory[].availability_status` currently documents the positive subset `in_stock`, `backorder`, and `preorder`, while Shopping Availability also documents `out_of_stock` and `discontinued`. Is Location intentionally defining a subset, or should Shopping identifiers use the same well-known values and meanings as Shopping Availability?

## Amenity vocabulary and documentation pass

The reverse-DNS string array answers the wire-shape question. Lets add a small normative table defining the initial `dev.ucp.amenity.*` identifiers and fix `location.json`'s bare examples such as `free_wifi` and `parking`, which do not satisfy the schema. Platforms should tolerate unknown amenity identifiers in responses; an unrecognized identifier used as a request filter must not be silently treated as satisfied.

Also, a few stale references to fix:

- `geofence_point` → `serves`;
- `offerings.inventory` → `filters.inventory`;
- Lookup's “offerings/inventory” → `inventory`;
- bare amenity examples → valid reverse-DNS identifiers.

## Punt on Lookup aliases?

Lookup says multiple identifiers may resolve to one Location, returns that Location once, and then says clients correlate solely by matching returned `location.id` to requested IDs. Those rules cannot all hold when an input is an alias.

Catalog needs multiple identifier forms and therefore returns required `inputs[]` correlation metadata. Does Location need that complexity now? The smaller v1 contract is to accept canonical, stable Business `Location.id` values only, deduplicate exact duplicate strings, and require every returned `location.id` to equal a requested ID. If aliases or secondary identifiers are required, Location needs lookup-only correlation equivalent to Catalog's `inputs[]`; returning duplicate Location objects does not solve the mapping.

jingyli:
Hopefully we are getting close to getting the right Locations shape in UCP! Responses to your feedback above @igrigorik :

## Let's reconcile Location with the landed fulfillment contract
Addressed in #753 (thanks for the edits!) and [8621ea3](https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/8621ea3180207a119883476a9a3ddf2d4b95b333) - we should now have achieved all the bullets you have outlined with the exception on "retain the existing `ucp_shared_request` annotations instead of removing them as unrelated cleanup;".

I still feel strongly about cleaning up as part of this PR given this is long overdue debt we have introduced on fulfillment (including destination types) related schemas. The UCP annotation is not used anywhere (neither in `ucp-schema` pre-processor nor anywhere beyond fulfillment).

## make serves a first-class Location Search query?
Indeed a spicy topic, ultimately addressed in #753. We landed on the principle that if something is fundamental to the core resource that is being searched/looked up (e.g., geo coordinates wrt to Location OR related similar items wrt to Product Variant), then it should be outside of `filters`. `filters` only contains concepts for further refinement only.

The only concern I had is semantically it feels like how serviceability is being modelled (e.g., the shape in search requests) differs between Location & Catalog (`filters.fulfills_to`) capabilities. This inconsistency is not a problem as of now when evaluating against the what we gain from enforcing the principle above, but still want to call it out explicitly.

## Clarify Location browse and pagination
Addressed in #753. One thing worthy of callout is that we aligned to not enforce a default pagination limit consistently if it's omitted in platform issued requests. The limit will need to be specified by the business.

## Availability vocabulary
The motivation is to intentionally have Location capability define a subset of values due to:
- In my mind, there is no use case why a Buyer would issue an explicit location search query explicitly refined to look for not available items (`out_of_stock` and `discontinued` just doesn't make sense in this B2C context).
- Businesses can still advertise `out_of_stock` products when Buyers are browsing their catalog.

## Amenity vocabulary and documentation pass

Addressed in [8621ea3](https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/8621ea3180207a119883476a9a3ddf2d4b95b333).

## Punt on Lookup aliases?
Addressed in #753.

gsmith85 (commented):
LGTM. Conceptually and architecturally, there are no blockers here and the design is in very solid shape.

Left inline comments on the specific files for discussion alignment and a couple minor mechanical/editorial items. (including the `mkdocs.yml` path fix for strict doc builds.

source/services/common/mcp.openrpc.json
gsmith85: Removing the signature members from JSON `meta` in `82c6527` resolved the immediate wire contract issue for this PR. 

The broader question of formalizing HTTP headers in OpenRPC and data mirroring between HTTP headers and runtime `meta` applies across all UCP services, so let's move that to a dedicated issue/discussion to avoid blocking this PR.

source/schemas/common/types/inventory_filter.json
```diff
@@ -0,0 +1,19 @@
...
+  "type": "object",
+  "required": ["id"],
+  "properties": {
+    "id": {
+      "type": "string",
+      "description": "The stable, opaque identifier of the item that MUST be Business-resolved (e.g., product or variant ID in shopping, or dish ID in food ordering)."
+    },
+    "availability_status": {
+      "type": "string",
+      "description": "Availability status of the inventory requested at the location. Well-known values: `in_stock`, `backorder`, `preorder`."
```
gsmith85: +1 with @jingyli on scoping `availability_status` here to the positive discovery subset (`in_stock`, `backorder`, `preorder`). 

Location Search serves buyer-facing discovery (*"Where can I acquire this item?"*), so querying for negative states like `out_of_stock` or `discontinued` does not represent realistic buyer discovery intent.

mkdocs.yml
```diff
@@ -360,6 +367,24 @@ plugins:
               User authentication via the Identity Linking Capability,
               specifying direct B2C/B2B OAuth 2.0 flows (with PKCE and
               RFC 9207) to unlock personalized commerce data.
+        Location Capability:
+          - specification/location/index.md: >-
```
gsmith85: Under `plugins.llmstxt.sections.Location Capability`, the file paths need `common/` added to match the new file locations:
```yaml
Location Capability:
  - specification/common/location/index.md: >-
      Core Location Capability, detailing high-level physical location
      discovery models, geographic and geofencing context, operating
      hours, and offerings such as amenities and inventory.
  - specification/common/location/search.md: >-
      Location discovery via the Location Search capability, specifying
      text queries and filters such as operating hours and offerings.
  - specification/common/location/lookup.md: >-
      Direct location retrieval via the Location Lookup capability,
      specifying identifiers and optional filters.
  - specification/common/location/rest.md: >-
      HTTP REST transport binding for the Location Capability,
      detailing search and lookup endpoints with JSON payload examples.
  - specification/common/location/mcp.md: >-
      Model Context Protocol (MCP) transport binding for the Location
      Capability, mapping discovery operations to JSON-RPC tools with
      metadata validation rules.
```
Fixing these 5 lines will allow `DOCS_MODE=spec mkdocs build --strict` to pass without warnings and ensure Location docs are indexed in `llms.txt`.

docs/specification/common/location/mcp.md
```diff
@@ -0,0 +1,277 @@
...
+
+## Entities
+
+### Amenity Type
+
+{{ schema_fields('types/amenity_type', 'common/location/mcp') }}
+
+### Lookup Location
+
+{{ extension_schema_fields('location_lookup.json#/$defs/lookup_location', 'common/location/rest') }}
```
gsmith85: The macro context here should be `'common/location/mcp'` instead of `'common/location/rest'`:
```markdown
{{ extension_schema_fields('location_lookup.json#/$defs/lookup_location', 'common/location/mcp') }}
```

docs/specification/common/location/index.md
```diff
@@ -0,0 +1,279 @@
...
+      Platform-supplied target; the Business is authoritative for that answer, and UCP does not
+      model or expose the underlying coverage geometry — a Location does not implicitly carry a
+      geometric boundary around its coordinates. See [Spatial Relations](search.md#spatial-relations).
+* **Operating Hours**: Regular weekly schedules (`hours`) and date-specific
+    exceptions (`exception_hours`), interpreted in the Location's `timezone`.
+    See [Operating Hours](#operating-hours).
+
+### Relationship to Other Capabilities
+
+The Location capability provides the foundation for localized commerce by integrating tightly
```
gsmith85: Editorial nit: missing "with" (`...integrating tightly with other capabilities...`)

docs/specification/common/location/search.md
```diff
@@ -0,0 +1,429 @@
...
+Separates static location characteristics from dynamic availability:
+
+* **`amenities`** (Array of Strings): Static features or services of the
+    location. All specified amenities **MUST** be supported by the location (AND semantic).
+    See [Amenity Vocabulary](#amenity-vocabulary) for well-known values.
+* **`inventory`** (Array of Objects): Real-time availability of items/goods at
+    the location. Some industry specific use cases include:
+    * *Shopping*: Checking stock availability for specific products or variants.
+    * *Food Ordering*: Checking offering availability of specific dishes or menu items.
+    Each inventory filter requires an stable, opaque `id` (e.g., product/dish ID) and
```
gsmith85: Editorial nit: `...requires a stable, opaque id...`

source/schemas/common/types/location.json
```diff
@@ -0,0 +1,59 @@
...
+      "type": "object",
+      "properties": {
+        "geo": {
+          "$ref": "geo.json",
+          "description": "Geographic coordinates for the location.",
+          "ucp_request": "omit"
+        },
+        "amenities": {
+          "type": "array",
+          "description": "Static features of the location represented as a reverse-DNS string. (e.g., 'dev.ucp.amenity.wi_fi', 'dev.ucp.amenity.shopping.in_store_pickup', 'com.example.amenity.in_person_installation').",
```
gsmith85: Editorial nit: remove spurious period before parenthesis (`reverse-DNS string (e.g., ...`).

gsmith85 (approved):

docs/specification/common/location/mcp.md
jingyli: Good catch! Copy-pasted too quickly when fixing the last broken link.

amithanda (approved):

igrigorik (approved):
Kudos everyone, great work on this! LGTM.

PR #590, opened 2026-07-16
docs(signatures): make example signature values raw r||s as the spec requires

### LINKED ISSUE #569: signatures.md: example Signature values are ASN.1/DER, but the spec requires raw r||s encoding
`signatures.md` states under Signature Value Encoding:

> **ECDSA** signatures **MUST** use fixed-width raw `r||s` encoding per RFC 9421 §3.3.1, **not** ASN.1/DER.

But the `Signature:` example values in the same document are DER-encoded. Decoding the base64 prefixes:

| Example value | First bytes | Meaning |
|---|---|---|
| `sig1=:MEUCIQD…:` | `30 45 02 21` | ASN.1 `SEQUENCE { INTEGER … }` |
| `sig1=:MEQCIB…:` | `30 44 02 20` | ASN.1 `SEQUENCE { INTEGER … }` |
| `sig1=:MFQCIH…:` | `30 54 02 20` | ASN.1 `SEQUENCE { INTEGER … }` |

`30 … 02` is exactly the ASN.1/DER form the MUST prohibits — a raw `r||s` P-256 signature is 64 bytes and never begins `30 45 02`. Several examples across the REST request, response, and webhook sections share this shape.

Impact: an implementer who mirrors the examples' encoding (easy to do, since OpenSSL/Java/.NET default to DER — as the spec itself warns) emits signatures that a spec-following verifier rejects.

Suggested fix: regenerate the example `Signature` values as fixed-width raw `r||s` (base64) with your example-generation tooling — I didn't want to fabricate signature bytes in a PR, since real-looking examples presumably come from your signing pipeline.

Found while building an unofficial UCP conformance test suite (https://spck.dev); verified against main @ 7e5fc42 by decoding the example base64. Happy to help further if useful.

vishkaty:
Fixes #569: signatures.md mandates fixed-width **raw `r||s`** encoding for ECDSA signature values, but the worked examples show values beginning `MEUC…`/`MEQC…` — the base64 of a DER `SEQUENCE`, the exact encoding the normative text forbids (§ Signature value encoding). An implementer copying the examples' shape produces signatures a conformant verifier rejects.

## What changed

Replaces the **seven** example signature values across `signatures.md` (4), `checkout-mcp.md` (1), `checkout-rest.md` (1), and `order-rest.md` (1) with a **genuine ES256 raw `r||s` value** (a real P-256 signature: 64 bytes → 88 base64 chars, decodes to no ASN.1 structure), preserving each instance's truncation style (`…`).

The eighth instance, in `order.md`, is deliberately **not** touched here — #568 edits the same example block (adding `ucp-agent` to its covered components), and overlapping diffs would conflict; it can pick up the value fix on whichever lands second.

Example values only; no normative text changes.


damaz91:
Thanks for this documentation fix @vishkaty !

I see there are others examples fix worthy, like

* https://github.com/Universal-Commerce-Protocol/ucp/blob/main/docs/specification/checkout-mcp.md?plain=1#L785
* https://github.com/Universal-Commerce-Protocol/ucp/blob/main/docs/specification/checkout-rest.md?plain=1#L1449
* https://github.com/Universal-Commerce-Protocol/ucp/blob/main/docs/specification/signatures.md?plain=1#L564

Can you take a look and add them in this PR?

vishkaty:
Good catch — thank you! Both fixed, plus one more the first pass missed: those response-signing examples use a second DER shape (`MFQCIH…`), so I re-swept by **decoding every example signature value's leading byte** (DER always starts `0x30`) instead of pattern-matching, and fixed signatures.md's response-signing example too. Also aligned the stale `keyid="merchant-2025"` in checkout-rest.md with the `merchant-2026` its sibling examples use.

The only DER-shaped value now remaining anywhere in docs/ is order.md's — left deliberately since #568 edits that same example block; it can pick up the value fix on whichever lands second.

damaz91 (approved):

jingyli (approved):

yanheChen (approved):

vishkaty:
Thanks for the reviews here @damaz91 @jingyli 🙏

This one's approved by both of you and every check is green — the only thing left
is the `Governance / Approvals` gate, which still reports *"Pending required
approvals."* Since it's a small docs fix (and it closes #569), is there a specific
domain/area approver the governance rule is still waiting on? Happy to get it in
front of the right person if that helps move it along — just let me know.

PR #602, opened 2026-07-17
feat: Add Payment Terms with Well-Defined Schedules

raginpirate:
Introduces `dev.ucp.shopping.payment_terms`, an extension that lets a Business offer different payment terms for the same checkout — pay in full today, put down a deposit and settle the balance later, split into installments — and lets the Buyer choose one.

**Payment term is negotiated:** the Business publishes the terms it will accept, the Buyer selects one, and the Business renegotiates rest of checkout. The selected term determines the checkout total, the disclosure duties, which instruments are eligible, etc. 

## What this PR adds

Two fields on the Checkout's `payment` object and one on the Order's:

| field | where | direction |
| :-- | :-- | :-- |
| `payment.terms[]` | Checkout | response only (`ucp_request: omit`) |
| `payment.selected_term_id` | Checkout | settable on Update; omitted on Create and Complete |
| `payment.accepted_term` | Order | response only, singular |

A term is a `title`, a `description`, and one or more **schedules**. A schedule is a `type` (`immediate` or anything else the Business names), a human `description`, an optional absolute `due_at`, and its own `amount`. There is no recurrence engine and no anchor arithmetic: every date on the wire is an absolute RFC 3339 timestamp the Business computed.

All three fields are optional. A Business offers terms when a Checkout has something to say about payment timing — a choice between terms, or a single term whose schedules must be disclosed. Where payment is due in full at completion, `terms` and `selected_term_id` are omitted entirely and the Checkout `total` is the amount due, exactly as core Checkout already states. Adopting the capability does not change the shape of a checkout that has nothing to disclose.

## Why selection rather than description

UCP already models "a list of priced options with exactly one selection." `fulfillment_group` is that shape, and this PR adopts similar shape:

| | fulfillment | payment terms |
| :-- | :-- | :-- |
| list | `options[]` | `terms[]` |
| selection | `selected_option_id` | `selected_term_id` |
| list direction | response-only | response-only (`ucp_request: omit`) |
| timestamps | absolute (`earliest_fulfillment_time`) | absolute (`due_at`) |
| render floor | `title` + `description` + amount | `title` + `description` + amount |
| type vocabulary | open, recognition optional | open, recognition optional |

The one place the shapes diverge is the amount. A fulfillment option carries a `total.json` array; a schedule carries a scalar `amount`. See the design decisions below.

---

## Example: lodging

A $1,200 stay. Pay in full today for $1,150, or pay the first night now and the balance at check-in. **The two terms have different payable amounts** — which is the whole reason the selection has to be a protocol field rather than a UI detail. The Buyer has not chosen yet, so the Business defaults to "pay now" on initial create_checkout:

```json
{
  "selected_term_id": "pt_pay_now",
  "terms": [
    {
      "id": "pt_pay_now",
      "title": "Pay now",
      "description": { "plain": "Save $50 by paying for your stay today." },
      "schedules": [
        {
          "id": "sched_full",
          "type": "immediate",
          "description": { "plain": "Due today when you book." },
          "amount": 115000
        }
      ]
    },
    {
      "id": "pt_deposit_balance",
      "title": "First night now, balance at check-in",
      "description": { "plain": "Hold your room with one night's rate." },
      "schedules": [
        {
          "id": "sched_first_night",
          "type": "immediate",
          "description": { "plain": "Due today when you book." },
          "amount": 30000
        },
        {
          "id": "sched_balance",
          "type": "deferred",
          "description": { "plain": "Due at check-in on September 1, 2026 at 3:00 PM PDT." },
          "due_at": "2026-09-01T15:00:00-07:00",
          "amount": 90000
        }
      ]
    }
  ]
}
```

The Buyer selects the deposit term with an Update:

```json
{ "selected_term_id": "pt_deposit_balance" }
```

The Business renegotiates and its response is authoritative: the checkout total is now $1,200, and the selected term's schedules sum to it. On completion the accepted term travels to the Order, so the Buyer retains a protocol record that $900 is still owed at check-in:

```json
{
  "payment": {
    "accepted_term": {
      "id": "pt_deposit_balance",
      "title": "First night now, balance at check-in",
      "description": { "plain": "Hold your room with one night's rate." },
      "schedules": [
        {
          "id": "sched_first_night",
          "type": "immediate",
          "description": { "plain": "Due today when you book." },
          "amount": 30000
        },
        {
          "id": "sched_balance",
          "type": "deferred",
          "description": { "plain": "Due at check-in on September 1, 2026 at 3:00 PM PDT." },
          "due_at": "2026-09-01T15:00:00-07:00",
          "amount": 90000
        }
      ]
    }
  }
}
```

The doc carries one further worked example: **installments**, four payments with the uneven remainder placed on a single schedule.

## Key design decisions

* **Only one well-known `type` value: `immediate`.** It means the payment is *due* when the checkout completes. Whether it is authorized, captured, or settled at that moment is payment-handler behavior and outside this extension. 

* **A schedule states an `amount`, not a breakdown.** The purchase is priced once, at the Checkout. The breakdown type that already exists, `totals.json`, requires a `subtotal` — and a schedule has no subtotal to give, because it prices nothing; it moves an already-priced amount through time. Rather than hand-roll an array that resembles `totals.json` while obeying different rules, a schedule carries a single non-negative `amount`, and `description` remains available for anything a Business needs to say in prose. Where the selected term changes what the purchase *costs* — a discount for paying today, a finance charge for paying over time — that difference appears in `checkout.totals`, where prices live.

* **`selected_term_id` is omitted on Create and on Complete.** On Create no terms exist yet, so there is no ID to reference. On Complete, accepting a pricing field there would let the total change after the instrument was authorized against the prior total. 

* **Consumers that do not understand a term still render it.** `title` + `description` + amount is a sufficient floor, mirroring the fulfillment rule. The type vocabulary is open and recognition is optional, so an unknown term type degrades to a readable choice rather than a blocked checkout.

* **Split Payments composes with a single-schedule term, and not otherwise.** An instrument `amount` states how much, never when, so allocating several instruments across several dated schedules would let a rounding residual silently change *when* a Buyer is charged. Per-schedule funding needs an explicit instrument-to-schedule binding, which is deferred to a follow-up; adding it later is additive and reverses this restriction without a breaking change.

* **Per-schedule payment status.** What was *agreed* lives here; what was *charged* lives in #584's `payment.instruments[].amount`. #584 adds `payment` to core `order.json` and this composes with it via `allOf` — additive, not competing, but they should land in a known order.

---

## Checklist


docs/specification/payment-terms.md
```diff
@@ -0,0 +1,489 @@
...
+| `deferred` | Payment is due after checkout completion, according to `due_at`. |
+
+Businesses MAY use additional values. Platforms MUST tolerate unknown schedule
+types and either apply business-specific support or surface clear validation
+messages when a selected payment instrument cannot satisfy them.
+
+An `immediate` schedule **MUST** omit `due_at`; a `deferred` schedule **MUST**
+include a non-empty `due_at`. The shared Schedule fields are:
+
+* `anchor_date` — RFC 3339 reference point. If omitted, the anchor is the
```
drewolson-google: `the anchor is the instant successful checkout completion creates the order.`

This reads weirdly.

docs/specification/payment-terms.md
```diff
@@ -0,0 +1,489 @@
...
+  instant successful checkout completion creates the order.
+* `offset` — ISO 8601 duration from the anchor to the first occurrence.
```
drewolson-google: I'd suggest consistently using `anchor_date` if that's the name of the field.

drewolson-google (commented):
A few small notes. Leaving a larger comment on #603.

docs/specification/payment-terms.md
```diff
@@ -0,0 +1,489 @@
...
+* `interval` — ISO 8601 duration between recurring occurrences.
+* `occurrences` — finite number of occurrences; **MUST** be present when
+  `interval` is present.
+
+Each payment schedule has `totals[]`. For a recurring schedule, those totals
+apply to each occurrence. Expanding every occurrence, the sum of schedule
+`total` entries **MUST** equal the checkout `total` exactly once.
```
archrao: In case of uneven totals due to 'occurrences' ie 100 split into 3. Should the merchant send multiple schedules so the totals adds up?

source/schemas/shopping/payment_terms.json
```diff
@@ -0,0 +1,134 @@
...
+      "description": "Order extended with the payment terms accepted during checkout.",
+      "allOf": [
+        { "$ref": "order.json" },
+        {
+          "type": "object",
+          "properties": {
+            "payment": {
+              "type": "object",
+              "properties": {
+                "terms": {
```
archrao: Should we add minItems: 1 maxItem:1 as we mention order.payment.terms[] containing exactly one complete snapshot

docs/specification/payment-terms.md
archrao: so anchor = creation_date?

source/schemas/shopping/types/payment_term.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/types/payment_term.json",
+  "title": "Payment Term",
```
archrao, line 4: I believe this is not being referenced

docs/specification/payment-terms.md
```diff
@@ -0,0 +1,701 @@
...
+      "id": "sched_cycle_1",
+      "type": "immediate",
+      "description": { "plain": "Due today. Covers your first month." },
+      "totals": [{ "type": "total", "amount": 2999 }]
+    }
+  ]
+}
+```
+
+**Policy — the durable terms.** The recurrence concerns the subscribed item
```
archrao: Just to confirm, the subscription-policy-disclosure is to inform the user of this recurring payment but this payment term is for one schedule and one payment only.
We will need the PR #629 to establish a recurring subscription_id.

igrigorik:
@archrao replying here, so this doesn't get lost in stale review comments later... 

>Just to confirm, the subscription-policy-disclosure is to inform the user of this recurring payment but this payment term is for one schedule and one payment only. We will need the PR https://github.com/Universal-Commerce-Protocol/ucp/issues/629 to establish a recurring subscription_id.

Yes, the term carries one immediate schedule for first charge, and the recurrence is disclosed through a policy and its paired disclosure rather than "scheduled" primitive.

The line this PR draws is **owed vs. not-yet-owed**. Both halves of owed are modeled: due now is a schedule with `type: immediate`, due later is a schedule with `type: deferred` and an absolute `due_at`. Both are obligations of this checkout, and the checkout `total` is their sum — the $900 due at check-in in the lodging example is owed the moment the Order is created. A renewal is not "owed" and thus not covered, but it is disclosed and agreed upon via agreement to the subscription policy. The Buyer can then decline later/async it by cancelling, and execution stays with the Business either way: this extension does not define credential storage, off-session authorization, or capture.

The subscription example in the PR models this: a policy targeting the line item, paired with a disclosure warning that compels display — which is what most jurisdictions require for a renewing charge. That example uses the simplest form of the mechanism, where the policy carries only text:

```json
{
  "type": "com.example.policy.subscription",
  "description": {
    "markdown": "Renews at **$29.99/month** on the 14th until cancelled. Cancel anytime at example.com/account."
  },
  "applies_to": ["$.line_items[0]"],
  "url": "https://example.com/subscription-terms"
}
```

That is a Business-defined type, and it works today: any Platform can present it, and the paired warning compels display. What it cannot do is let an agent compare this against an annual plan without parsing the sentence. However, nothing stops us from modeling a more structured version. Defining type-specific fields on a well-known policy type is exactly what #634 is doing for `dev.ucp.shopping.policy.return`, and the same pattern applies here:

```json
{
  "type": "dev.ucp.shopping.policy.subscription",
  "description": {
    "markdown": "Renews at **$29.99/month** on the 14th until cancelled. Cancel anytime at example.com/account."
  },
  "applies_to": ["$.line_items[0]"],
  "url": "https://example.com/subscription-terms",
  "cadence": { "interval": "month", "interval_count": 1 },
  "price": { "amount": 2999, "currency": "USD" },
  "first_charge_at": "2026-09-14T00:00:00Z"
}
```

Same container, same targeting, same disclosure pairing. `type` and `description` stay the floor, so a Platform that does not model the type presents it exactly as before, and one that does can read the structured fields. Disclaimer: the fields above are illustrative rather than a proposal -- the point is that the mechanism is already there. I do think we _should_ define a subscription policy, but that's a clean and additive PR on top of the infrastructure we're defining here.

docs/specification/payment-terms.md
igrigorik: https://github.com/Universal-Commerce-Protocol/ucp/pull/602#issuecomment-5254020548

raginpirate:
Thanks for the review and feedback everyone. We've accepted #603 and #692 into this base branch which will be the main proposal for payment terms.

As I reviewed #692's final simplification i do have a few extra thoughts on continuing to extract concepts from this base branch so we land a cohesive small core. PTAL: https://github.com/Universal-Commerce-Protocol/ucp/pull/712 <-- its also opinionated about the subscription conversation above.

archrao:
> @archrao replying here, so this doesn't get lost in stale review comments later...
> 
> > Just to confirm, the subscription-policy-disclosure is to inform the user of this recurring payment but this payment term is for one schedule and one payment only. We will need the PR #629 to establish a recurring subscription_id.
> 
> Yes, the term carries one immediate schedule for first charge, and the recurrence is disclosed through a policy and its paired disclosure rather than "scheduled" primitive.
> 
> The line this PR draws is **owed vs. not-yet-owed**. Both halves of owed are modeled: due now is a schedule with `type: immediate`, due later is a schedule with `type: deferred` and an absolute `due_at`. Both are obligations of this checkout, and the checkout `total` is their sum — the $900 due at check-in in the lodging example is owed the moment the Order is created. A renewal is not "owed" and thus not covered, but it is disclosed and agreed upon via agreement to the subscription policy. The Buyer can then decline later/async it by cancelling, and execution stays with the Business either way: this extension does not define credential storage, off-session authorization, or capture.
> 
> The subscription example in the PR models this: a policy targeting the line item, paired with a disclosure warning that compels display — which is what most jurisdictions require for a renewing charge. That example uses the simplest form of the mechanism, where the policy carries only text:
> 
> ```json
> {
>   "type": "com.example.policy.subscription",
>   "description": {
>     "markdown": "Renews at **$29.99/month** on the 14th until cancelled. Cancel anytime at example.com/account."
>   },
>   "applies_to": ["$.line_items[0]"],
>   "url": "https://example.com/subscription-terms"
> }
> ```
> 
> That is a Business-defined type, and it works today: any Platform can present it, and the paired warning compels display. What it cannot do is let an agent compare this against an annual plan without parsing the sentence. However, nothing stops us from modeling a more structured version. Defining type-specific fields on a well-known policy type is exactly what #634 is doing for `dev.ucp.shopping.policy.return`, and the same pattern applies here:
> 
> ```json
> {
>   "type": "dev.ucp.shopping.policy.subscription",
>   "description": {
>     "markdown": "Renews at **$29.99/month** on the 14th until cancelled. Cancel anytime at example.com/account."
>   },
>   "applies_to": ["$.line_items[0]"],
>   "url": "https://example.com/subscription-terms",
>   "cadence": { "interval": "month", "interval_count": 1 },
>   "price": { "amount": 2999, "currency": "USD" },
>   "first_charge_at": "2026-09-14T00:00:00Z"
> }
> ```
> 
> Same container, same targeting, same disclosure pairing. `type` and `description` stay the floor, so a Platform that does not model the type presents it exactly as before, and one that does can read the structured fields. Disclaimer: the fields above are illustrative rather than a proposal -- the point is that the mechanism is already there. I do think we _should_ define a subscription policy, but that's a clean and additive PR on top of the infrastructure we're defining here.


Thanks for the clear explanation on the "owed vs. not-yet-owed" distinction as it is clear to the clients/platform to render and also keeps purchase terms clear for the buyer. Agree to continuing the 'subscription' discussion and using the 'policy' pattern as suggested.

drewolson-google (approved):

drewolson-google:
Approved. May be worth updating the title and top description to reflect the fact that we've merged #603 into this PR, and therefore are no longer going with the "well defined" approach, if I understand correctly.

rondanino:
Shape looks right, and reusing the fulfillment_group pattern makes it easy to reason about. Some notes from Shop Pay Installments, where a term is a loan offer rather than a payment split.

The main thing is that installment terms come in two phases, and 602 only has the second one. Producing a real offer costs a credit decision, so we can't publish real terms to everyone who loads a checkout. What we ship today is an indicative set of terms derived from shop-level financing plans and the cart total, with no buyer involved at all, and the per-payment figures are a straight division for display. Only once the buyer chooses to pay this way do we underwrite and get real schedules back.

That maps onto the negotiation model here quite naturally: publish an indicative term, the Buyer selects it, and the Business returns the binding one, since the PR already says the post-selection response is authoritative. What's missing is small but load bearing:

- No way to mark a term as indicative rather than binding, so a Platform can't tell whether it's displaying an estimate or a commitment, and may cache or present it as final.
- No defined behaviour for `terms[]` changing after a selection. The PR covers the total changing, but for us the schedules themselves change, and the term the Buyer picked can be replaced by its underwritten version.
- Between selecting and getting a binding term, the Business may still need something from the Buyer, for example the last four digits of an SSN, or a debit card rather than a credit card.
- The answer can be no. A loan can be declined, which is a normal outcome rather than an error.

Today `update` can only succeed or fail, so the last two both collapse into a failure with nowhere for the Platform to go. Some way for a selection to return outstanding requirements, pointing at whoever can collect them, would cover both. That could also be a home for terms whose details the Business would rather not put on the wire, an opaque term id plus a handler reference.

Smaller, and only worth doing once the above is settled, fields that have no home: the interest amount, a down payment distinct from the first scheduled payment, and links to disclosure documents the Buyer must be shown before agreeing. Those currently vanish or get flattened into `description`, so a Platform can render them but can't reason about them, and the disclosures are a legal requirement for us rather than a nicety. Worth noting APR belongs only on a binding term. Pre-decision we can only show a representative rate, so putting it on an indicative term would invite Platforms to present it as the buyer's actual rate.

None of this needs to block landing the base, I'd rather see it go in thin. Mainly want the indicative question settled early, since it changes what a Platform is allowed to assume about `terms[]`.

archrao (approved):
Thanks for the discussion and evolution of this specification across #602, #603, #692, and #712. LGTM!

igrigorik:
@archrao fyi, backtracking half of my explainer on subscriptions: the "owed vs not-yet-owed" holds, but the subscription via policy runs into a set of problems that @raginpirate caught and flagged, which led us to drop the subscription language from #602 entirely. 

That said, good news: we should be able to layer it cleanly on top of this infrastructure in this PR. A sketch...

---

| what a subscription needs | where it goes | does #602 change? |
| :-- | :-- | :-- |
| first cycle charged | an ordinary Checkout total | no |
| what recurs, disclosed | a typed policy | no |
| choosing whether an item recurs | a purchase option on the line item | no |

`payment_terms.json` never references a line item, so the mechanism carrying the choice sits beside it rather than through it. Terms are optional, so a zero-dollar trial needs none. And a subscription selection invalidating an already-selected term is covered by `payment_term_changed`, already registered.

## The model that fails

Payment terms let a Business offer several ways to pay for the same checkout and let the Buyer choose one. It is tempting to reach (as I did) for the same mechanism to express a subscription: offer a "one-time" term and a "subscribe monthly" term, and let the selection decide whether the Buyer is enrolled. That model breaks in two independent ways.

**The two terms are payment-identical.** Both are a single immediate payment. They differ only in amount, and the amount differs because *the products differ*, not because the payment timing does. A payment term is supposed to describe when money moves. Here it is describing a product configuration and carrying an enrollment side effect that has nothing to do with payment. Where two terms have the same payment structure, the term is not the thing being chosen.

**Payment terms are checkout-wide; subscriptions are per-item.** `payment.terms` sits on the Checkout, and a schedule carries `id`, `type`, `description`, `due_at`, and `amount` — there is no association with a line item anywhere in the structure. So a mixed cart cannot be expressed. Given a one-time widget and a subscribable coffee, the terms list has to enumerate the combinations:

| term | schedule | means |
| :-- | :-- | :-- |
| A | immediate $84.99 | both one-time |
| B | immediate $79.99 | coffee subscribed |

Two subscribable items require four terms. N items require 2^N. This is the same combinatorial explosion that led us to drop `entity_ref` from fulfillment, and it arrives sooner here. And the case a Buyer would actually ask for — a deposit on the widget while the coffee is subscribed — is not expressible at any number of terms, because schedules have no way to say which items they cover.

## Three axes

The confusion came from asking the payment axis to answer an item question. Separating them dissolves it.

| concern | question | mechanism |
| :-- | :-- | :-- |
| **item** | what am I buying, at what price | line item + purchase option |
| **payment** | when does money move | payment terms (#602) |
| **disclosure** | what recurs after this checkout | typed policy |

A subscription touches all three, and each part belongs to a different mechanism. The first cycle is an ordinary payment. The recurrence is a durable term about an item. The choice to subscribe is a property of the item, not of how the checkout is paid for.

## The missing piece

Two of the three axes exist today. Payment terms handle the first cycle. A typed policy carries the recurrence disclosure — cadence, amount, first charge — using the pattern established in #634, where a well-known `type` gates a structured body while unknown types still degrade to `description`.

The unowned piece is **selection**. A policy describes; it never *offers a choice*, and it should not — a selectable policy would make policies mutable checkout state. What is missing is a purchase option on the line item. Something like:

```json
"line_items": [{
  "id": "li_coffee",
  "purchase_options": [
    { "id": "po_onetime", "title": "One-time" },
    { "id": "po_monthly", "title": "Subscribe monthly" }
  ],
  "selected_purchase_option_id": "po_monthly"
}]
```

This is `options[] + selected_*_id`, the third instance of a pattern already proven twice in the protocol: `fulfillment_group.selected_option_id` and `payment.selected_term_id`. It is not a new concept, only the same one applied on the axis where the choice actually lives. Because the selection is per-item, N subscribable items produce N independent choices rather than 2^N terms, and a mixed cart is expressible by construction.

## Why the foundation composes

Three checks, each resolved by machinery that already exists.

**Selecting a subscription changes the total.** It changes the line item's price; the Business recalculates and returns authoritative state; `checkout.totals` moves and the selected term's schedules move with it. This is identical to selecting a fulfillment option, and #602 already establishes that the Business response is authoritative for all derived state.

**A changed total can invalidate the selected term.** A Business may offer deposits on one-time purchases but not on subscriptions. The `payment_term_changed` warning code is already registered for exactly this class of invalidation: the Business re-selects and warns. The mechanism was built for a caller that did not exist yet and serves this one without modification.

**The payment term does not need to know a subscription exists.** The first cycle is an ordinary immediate obligation inside `checkout.total`. The recurrence rides the policy. The Order's `accepted_term` stays honest and frozen, because nothing recurring was ever written into it.

Both shapes fall out without a new payment-terms concept. A $50 setup fee plus $29.99/month is a checkout total of $79.99 with one immediate schedule and a policy disclosing the ongoing charge. A free trial is a checkout total of zero with no terms at all, and a policy carrying the first charge date.

igrigorik:
FYI, another heads up from merged https://github.com/Universal-Commerce-Protocol/ucp/pull/712.

```jsonc
// before
"totals": [{ "type": "total", "amount": 30000 }]
// after
"amount": 30000
```

I initially copied `fulfillment_option.totals[]` for symmetry, but a schedule was only ever allowed one entry — which took `contains` + `minContains` + `maxContains`, plus an `if/then` gate to force non-negativity, because `total.json.amount` is `signed_amount` and permits negatives. 

@raginpirate appropriately flagged it, and we debated back and forth...

Reusing `totals.json` wholesale isn't an option: it mandates exactly one `subtotal`, and the schema already said the opposite — "no `subtotal` is required, because the purchase is priced at the checkout rather than once per payment." That's why the array was hand-rolled rather than `$ref`'d in the first place. But the reason given there isn't specific to subtotals; it rules out any per-schedule breakdown, since tax, discount and fee are all allocations of a number computed once at the checkout. Existing prose had already argued the container away and #712 finished the thought.

**A relaxed array**: keep `totals[]`, drop `maxContains`, require a `total` normatively. Sheds one keyword of four, keeps the `if/then` gate. This is a tempting shape for the symmetry and the optionality it yields, but it also drags in many gotchas — e.g. we now need to spec a similar sum algorithm as we did for checkout, enforce that a total must be present, and so on.

**What we merged instead**: a single non-negative `amount`, which is materially simpler and, coupled with a description, can still capture additional context in unstructured form (e.g. this installment plan has a total of X and, via description, carries text explaining the finance charge). Once the buyer selects a plan, the checkout renegotiates and they see the exact breakdown. This also has a performance benefit, because a granular breakdown may be expensive to compute for every proposed plan.

igrigorik (approved):
@drewolson-google @archrao ty for thorough reviews. You already LGTM'ed but it wasn't clear if that applied to 602+712 merge or just the shape we had in 602. I added two explainer comments for the subscription + `amount` updates. PTAL.

Assuming everyone else is on board with current shape, lgtm 👍

raginpirate:
@rondanino Thanks for the stress test with real instalments modelling. Quick pass on each one, I think we already have all of it.

**Indicative vs binding.** Only the selected term has to sum to the checkout total, so every unselected term is already indicative by construction. If the buyer needs to actually see "estimate, subject to approval," that's a presentation: "disclosure" warning with path: $.payment.terms[N], which is MUST display, MUST keep adjacent, MUST NOT dismiss.

**Requirements after selection** (SSN4). That's core Actions: return the update with an outstanding com.payments.installments.underwriting action, which blocks ready_for_complete until underwriting resolves. Its config is an open object so the loan id and collection surface are possible but not explicitly modelled.

**Terms changing after selection**. Rewrite the selected term's schedules in place on approval and let the response carry it, since the post-selection response is already authoritative for all derived state.

**Decline**. A decline is a 200: drop the action, drop the term, reselect the default, emit payment_term_changed. Same shape as the provisional discount plus student verification flow in checkout.md, where failing verification just removes the discount.

Payment terms is the glue between policies, actions, and payment handlers which when combined make these complex use cases possible.

docs/specification/payment-terms.md
```diff
@@ -0,0 +1,667 @@
...
+In every response that carries `terms`, one of them is selected. Where the Buyer
+has made no choice, the Business selects a default.
+
+A Platform changes the selection through Update Checkout, setting
+`selected_term_id` to an `id` from the latest `terms[]`. A Platform **MUST**
+omit it on create, because term IDs are scoped to a Checkout and no options
+exist yet — the create response establishes the options, if any, and the default
+among them. A Platform **MUST** omit it on complete, because the term is settled
+before a Checkout can reach `ready_for_complete`, and the instrument is
+authorized against the checkout total. A Business receiving a selection that no
```
amithanda: Should this sentence be reworded? 

First, "authorized" is a word this extension disclaims. payment_schedule.type says "whether a due payment is authorized, captured, or settled at that moment is payment-handler behavior and outside this extension," and Out of scope repeats it. This is the one place the document makes an authorization claim, and it is the one claim it disclaims everywhere else. It is also not true of the lodging example on this page, where the checkout total is 120000 and the amount taken at completion is 30000.


```suggestion
among them. A Platform **MUST** omit it on complete, because the term is already
agreed before a Checkout can reach `ready_for_complete`, and the amount due at
completion follows from it. A Business receiving a selection that no
```

docs/specification/payment-terms.md
```diff
@@ -0,0 +1,667 @@
...
+longer resolves — because the options changed — **MUST NOT** silently substitute
+a term, and **MUST** report the change as a `payment_term_changed` warning in
+`messages[]`.
+
+Selecting a term can invalidate a selection previously accepted elsewhere in the
+Checkout — a deferred term may not be available with a same-day fulfillment
+option. The Business resolves the conflict, returns the authoritative state, and
+**MUST** report the change as a `payment_term_changed` warning in `messages[]`.
```
amithanda, line 163: Both rules that emit `payment_term_changed` fire on the same thing: which term is selected. One covers a `selected_term_id` that no longer resolves, the other covers the Business swapping the selection to fix a conflict elsewhere. Line 164 gives the reason for having a code at all, that "a Platform can therefore detect a changed selection from the code alone, rather than by comparing responses."

Neither fires when the selected term keeps its `id` and its contents change. That is the underwriting case from the comment above:

- The Platform displays `pt_pay_in_4` as four payments of $25.
- The Buyer selects it, and underwriting runs.
- The Business returns `pt_pay_in_4` with three payments and a larger first amount.

Same id, so no warning is required, and the Platform can only notice by diffing responses. A swapped selection is at least visible as a different id. A rewritten schedule changes what the Buyer will be charged and when, while looking unchanged at the id level, so the case that needs the signal most is the one case that does not get it.

Should the rule key on the term rather than on the selection?

```suggestion
option. The Business resolves the conflict, returns the authoritative state, and
**MUST** report the change as a `payment_term_changed` warning in `messages[]`.
A Business **MUST** also report that warning where it changes the selected term
in place, by altering its schedules, due dates, or amounts, without naming a
different term.
```

amithanda (approved):
Approving. Left 2 minor documentation suggestions (non blocking).

Worth calling out how much narrower this got 👍 . The simplifications we discussed all landed, and the result is materially smaller than where this started.

PR #603, opened 2026-07-17
feat: Add Payment Terms with Simple Schedules

raginpirate:
A simplification of #602 by dropping due_at in favour of a simple description.

Open question for the audience: would we be happy removing further invariants from the solution to reduce past this point?


drewolson-google (commented):
To me, this feels better than #602. It also leaves the door open to move to something more rigid in the future, if we feel that is necessary.

I suspect that `deferred` being a description will solve most of the use cases for payment terms.

PR #608, opened 2026-07-17
docs: add cart session persistence guidance

### LINKED ISSUE #344: Cart sessions need persistence guidance
## Problem

The cart capability defines `expires_at` but says nothing about how cart sessions should persist, what TTLs are reasonable, or how sessions bind to buyers across reconnections.

Every implementer ends up inventing their own approach. Some use server-side durable objects, some use database-backed sessions, some use signed cookies. The spec is silent on all of it.

## What's missing

1. Recommended TTL ranges (minutes for food ordering? hours for retail? days for B2B?)
2. Session binding -- how does a returning buyer resume a cart? By cart ID alone? By buyer identity? Both?
3. Reconnection semantics -- if a cart expires mid-conversation with a voice agent, what should happen? Silent recreation? Error?
4. Cross-transport persistence -- a cart created over MCP should be retrievable over REST (and vice versa). The spec doesn't say whether this is expected.

## Suggestion

Add a non-normative "Implementation guidance" section to the cart capability covering:

- Minimum recommended TTL by use case (quick-service vs retail vs B2B)
- Cart ID as the canonical resumption handle, independent of transport
- Business SHOULD return `expires_at` and platform SHOULD warn the buyer before expiry
- Expired carts SHOULD return a specific error code (not just generic 404) so the platform can offer to recreate

This doesn't need to be prescriptive. Even a "here's how we expect this to work" section would save implementers from guessing.

hemanth:
Closes #344

## Summary

Adds a non-normative implementation guidance document for cart session persistence, covering TTL recommendations, session resumption patterns, expiry handling, and agent reconnection scenarios.

## Changes

- `docs/guides/cart-session-guidance.md`: New guidance document

## Key guidance points

- Recommended TTL ranges by use case (food ordering: 15-30min, retail: 1-24hr, B2B: 1-7 days)
- Cart ID as transport-agnostic resumption handle
- `cart_expired` error code instead of generic 404
- Agent-specific reconnection patterns

This is intentionally non-normative — it captures recommended patterns without mandating implementation details.

### DIFF
--- a/docs/specification/cart-rest.md
+++ b/docs/specification/cart-rest.md
@@ -542,6 +542,72 @@ HTTP 200 and the UCP envelope containing `messages`:
 }
 ```
 
+## Session Persistence
+
+!!! note "Non-normative"
+    This section provides implementation guidance and is not part of the normative specification.
+
+The cart capability defines `expires_at` but intentionally leaves session
+persistence details to implementers. The guidance below captures recommended
+patterns observed across early implementations.
+
+### Session Resumption
+
+The cart `id` is the canonical resumption handle. A buyer (or agent acting on
+their behalf) resumes a cart by providing its `id` to `GET /carts/{id}` or
+`PUT /carts/{id}`, regardless of how the cart was originally created.
+
+- **Cross-transport persistence**: A cart created over MCP SHOULD be retrievable
+  over REST, and vice versa. The cart `id` is transport-agnostic.
+- **Buyer binding** is a business decision. Some businesses scope carts to a
+  buyer identity; others (especially guest-friendly merchants) allow any session
+  with the cart `id` to resume.
+
+### Expiry Handling
+
+When a cart has expired, businesses SHOULD return a `cart_expired` error code
+rather than a generic `not_found`, so platforms can distinguish "this cart
+existed but expired" from "this cart ID was never valid":
+
+<!-- ucp:example schema=common/types/error_response op=read -->
+```json
+{
+  "ucp": {
+    "version": "{{ ucp_version }}",
+    "status": "error",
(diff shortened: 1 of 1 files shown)


damaz91:
Hello @hemanth ! My 2cs on the cart persistence guidance is that these should pertain to each individual platform supporting the cart capability; I don't think UCP should prescribe any TTL.

We already have a section in https://ucp.dev/latest/specification/cart-rest/ talking about how to handle cart expiration (https://ucp.dev/latest/specification/cart-rest/#business-outcomes) and the "expires_at" field is already part of the output schema (https://ucp.dev/latest/specification/cart-rest/#output-schema_1).

Overall I don't see why these guidances should live in a different page rather then https://ucp.dev/latest/specification/cart-rest/

hemanth:
@damaz91 fair points all around. You are right that UCP should not prescribe TTLs and that the existing `expires_at` field already covers the basics.

I have reworked the branch. The changes are:

1. **Dropped the separate guide file.** The content now lives inline in `cart-rest.md` as a "Session Persistence" section, right before Security Considerations.
2. **Dropped the TTL table entirely.** That is a platform-level decision, not something UCP should recommend.
3. **Kept the parts that add value beyond what is already documented:**
   - `cart_expired` as a distinct error code from `not_found` so platforms can tell the difference between "expired" and "never existed"
   - Cross-transport resumption (cart created via MCP should be retrievable via REST)
   - Agent reconnection patterns for voice/chat scenarios

The whole section is marked non-normative.

Let me know if this works better or if you think any of the remaining guidance is still too prescriptive.

sruthi-vandhana:
Can you please resolve the merge conflict @hemanth

PR #609, opened 2026-07-17
feat: add discoverable available discounts to discount extension

### LINKED ISSUE #346: No way to discover available discounts before applying codes
## Problem

The discount extension lets platforms submit codes and receive applied/rejected results. But there's no way to ask "what discounts are available right now?" before the buyer types in a code.

An agent can't tell the buyer "there's a 15% off sale running" unless the agent already knows the code. The business knows about its active promotions, but the protocol has no channel for surfacing them.

## What's missing

- A query operation to list active, applicable discounts for a given cart/checkout context
- Distinction between public promotions (anyone can see) and private codes (buyer must know)
- Automatic discount surfacing -- if the business has a "free shipping over $50" rule, the agent should know about it when the cart total crosses $50

## Suggestion

Add an optional `available_discounts` array to cart and checkout responses:

```json
{
  "available_discounts": [
    {
      "id": "disc_summer",
      "title": "Summer sale -- 15% off",
      "type": "automatic",
      "conditions": "Applied automatically to orders over $25",
      "applied": true
    },
    {
      "id": "disc_freeship",
      "title": "Free shipping",
      "type": "automatic",
      "conditions": "Orders over $50",
      "applied": false,
      "remaining_amount": 1200
    }
  ]
}
```

- `type: automatic` -- business applies it when conditions are met, no code needed
- `type: code` -- buyer needs to submit a code (the code itself is NOT exposed, just the existence of the promotion)
- `remaining_amount` -- how much more the buyer needs to spend to qualify (useful for "add $12 more for free shipping" prompts)

This is business-optional. Merchants that don't want to expose promotions just omit the field.

hemanth:
Closes #346

## Summary

Adds an `available` array to the discount extension's `discounts_object`, allowing businesses to surface discoverable promotions before the buyer applies a code.

## Changes

- `source/schemas/shopping/discount.json`:
  - Added `available_discount` definition with `id`, `title`, `type` (automatic/code), `conditions`, `applied`, and `remaining_amount`
  - Added `available` array to `discounts_object` (response-only via `ucp_request: omit`)

## Use cases

- Agent tells buyer: "There's a 15% off summer sale running"
- Agent prompts: "Add $12 more for free shipping"
- Automatic discounts surfaced so the buyer knows they're getting a deal
- Code-based offers hint at existence without exposing the code

## Design decisions

- Business-optional: merchants that don't want to expose promotions just omit the field
- Response-only (`ucp_request: omit`) — this is purely informational from business to platform
- `type: code` exposes the **existence** of a promotion but NOT the code itself
- `remaining_amount` enables threshold-based prompts ("add $X more for...")

### DIFF
--- a/source/schemas/shopping/discount.json
+++ b/source/schemas/shopping/discount.json
@@ -79,6 +79,41 @@
         }
       }
     },
+    "available_discount": {
+      "type": "object",
+      "description": "A discount available for discovery in the current context.",
+      "required": [
+        "title",
+        "type"
+      ],
+      "properties": {
+        "id": {
+          "type": "string",
+          "description": "Stable discount identifier."
+        },
+        "title": {
+          "type": "string",
+          "description": "Human-readable promotion name (e.g., 'Summer Sale — 15% off')."
+        },
+        "type": {
+          "type": "string",
+          "enum": ["automatic", "code"],
+          "description": "'automatic' — business applies when conditions are met, no code needed. 'code' — buyer must submit a code (the code itself is NOT exposed)."
+        },
+        "conditions": {
+          "type": "string",
+          "description": "Human-readable summary of conditions to qualify (e.g., 'Orders over $50')."
+        },
+        "applied": {
+          "type": "boolean",
+          "description": "Whether this discount is already applied to the current cart/checkout."
+        },
+        "remaining_amount": {
+          "$ref": "../common/types/amount.json",
+          "description": "How much more the buyer needs to spend to qualify, in ISO 4217 minor units. Useful for 'add $12 more for free shipping' prompts."
+        }
+      }
+    },
     "discounts_object": {
       "type": "object",
       "description": "Discount codes input and applied discounts output.",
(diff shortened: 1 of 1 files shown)


ShuoRen-TT (commented):
Thanks for putting this together. I reviewed the current schema change against an active TikTok commerce integration that needs to surface discoverable promotions before the buyer enters checkout.

The core shape maps well to the implementation need:

- distinguishing `automatic` from `code` offers lets a Platform explain whether buyer action is required;
- not exposing the actual code preserves Business control while still making the offer discoverable;
- `remaining_amount` is useful once cart or checkout context exists for threshold prompts.

To keep this PR focused, I do not suggest adding Catalog support here. I will open a separate Enhancement Proposal that depends on #609 and only composes the existing `discounts.available` model onto the `get_product` response, without redefining the discount type.

One contract detail would be useful to make explicit in the specification: when an entry has `available[].applied: true`, it should correspond to the same promotion in `discounts.applied`, with `discounts.applied` remaining authoritative for the calculated amount and allocations. A schema-validated example in `discount.md` covering an automatic offer, a discoverable code-based offer, and a threshold `remaining_amount` would document that invariant and provide regression coverage.

From the integration side, the proposed response-only and Business-optional behavior looks right. @sinhanurag @raginpirate, when convenient, could you please review the core `available_discount` shape? It would unblock a concrete integration need while keeping the Catalog lifecycle discussion separate.

ShuoRen-TT:
The focused Catalog follow-up is now tracked in #707. It explicitly depends on #609, reuses the same available_discount definition, and keeps this PR limited to the Cart/Checkout model and its documentation contract.

PR #610, opened 2026-07-18
fix(schema): update checkout_mandate pattern to support dSD-JWT chain (#599)

### LINKED ISSUE #599: ap2_mandate.json: checkout_mandate pattern rejects the serialization the AP2 reference SDK emits — which wire form is intended?
ap2-mandates.md describes `ap2.checkout_mandate` as an SD-JWT credential with Key Binding, and `ap2_mandate.json` pins the wire shape as:

```
```

The AP2 reference implementation ([google-agentic-commerce/AP2](https://github.com/google-agentic-commerce/AP2), `ap2` SDK) serializes a mandate presentation as a **delegate-SD-JWT chain**: hops joined by `~~`, per its dSD-JWT modules. That wire fails the pattern — `(~[A-Za-z0-9_-]+)*` requires every `~` to be followed by at least one character, so the empty segment in `~~` can never match. The plain RFC 9901 no-KB form (compact SD-JWT ending in a trailing `~`) is rejected for the same reason.

Repro (with the reference SDK's own builders):

```python
import re
# chain = MandateClient().present(...)  # per the ap2 SDK's checkout-mandate flow
pattern = r'^[A-Za-z0-9_-]+\.[A-Za-z0-9_-]*\.[A-Za-z0-9_-]+(~[A-Za-z0-9_-]+)*$'
re.fullmatch(pattern, chain)   # -> None ('~~' between hops can't match)
```

So a platform that follows the reference implementation exactly would emit an `ap2.checkout_mandate` that fails UCP's own schema validation — easy to trip on while the AP2 side is still moving.

Which form is intended here?

1. The pattern is deliberately scoped to a single SD-JWT+KB (no chain) — in which case a note in ap2-mandates.md about the divergence from the reference's serialization would save implementers some confusion; or
2. The pattern should also admit the chain serialization (and the trailing-tilde no-KB form) — happy to send the schema PR if so; or
3. The reference's wire format is ahead of this binding and the schema will follow once it settles.

Happy to help with whichever direction is right.

Harman-2:
# Description

Updates the regex validation pattern for the `checkout_mandate` property inside the `ap2_mandate.json` schema. The previous regex constraint strictly required characters immediately following a tilde (`~`), which caused schema validation to fail when encountering the trailing tilde of compact SD-JWT strings or multi-hop `~~` delegate-SD-JWT chains emitted by the reference implementation. Loosening the pattern group modifier from `+` to `*` allows both empty segments and trailing tildes while maintaining the required character set validation.

## Category (Required)

## Related Issues

Fixes #599

## Checklist

git

Verified regex behavior variation via python environment simulation:

```text
# Before Fix Pattern Verification (Fails on dSD-JWT wire format strings)
Before fix match result: None

# After Fix Pattern Verification (Successfully parses trailing tildes and multi-hop '~~' chains)
After fix match result: <re.Match object; span=(0, 62), match='eyJhbGciOiJFUzI1NiJ9.payload.sig~~disclosure1~disclosure2~'>

### DIFF
--- a/scripts/test_ap2_mandate_pattern.py
+++ b/scripts/test_ap2_mandate_pattern.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+"""Unit tests for checkout_mandate pattern in ap2_mandate.json.
+
+The pattern must accept RFC 9901 compact SD-JWT serializations, including
+trailing tildes and empty disclosure segments, and delegate chains joined
+with ``~~`` as emitted by the AP2 reference implementation.
+
+Run: python3 scripts/test_ap2_mandate_pattern.py
+Exit: 0 on all pass, 1 on any failure.
+"""
+
+import json
+import re
+import sys
+from pathlib import Path
+
+_REPO_ROOT = Path(__file__).parent.parent
+_SCHEMA_PATH = _REPO_ROOT / "source" / "schemas" / "shopping" / "ap2_mandate.json"
+
+_RESULTS: list[tuple[str, bool, str]] = []
+
+
+def _check(name: str, condition: bool, detail: str = "") -> None:
+  _RESULTS.append((name, condition, detail))
+
+
+def _report() -> int:
+  passed = sum(1 for _, ok, _ in _RESULTS if ok)
+  failed = [(n, d) for n, ok, d in _RESULTS if not ok]
+  for name, ok, detail in _RESULTS:
+    status = "PASS" if ok else "FAIL"
+    suffix = f" — {detail}" if detail and not ok else ""
+    print(f"  {status}  {name}{suffix}")
+  print(f"\n{passed} passed, {len(failed)} failed")
+  return 0 if not failed else 1
+
+
+def _load_checkout_mandate_pattern() -> re.Pattern[str]:
+  schema = json.loads(_SCHEMA_PATH.read_text(encoding="utf-8"))
+  raw = schema["$defs"]["checkout_mandate"]["pattern"]
+  return re.compile(raw)
+
+
+def test_checkout_mandate_pattern() -> None:
+  """Pattern accepts expected SD-JWT wire forms and rejects invalid strings."""
+  pattern = _load_checkout_mandate_pattern()
+  matches = pattern.fullmatch
+
+  valid = [
+    ("compact_jwt_only", "eyJhbGciOiJFUzI1NiJ9.eyJpc3MiOiJ1c3AifQ.signature"),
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by damaz91** [comment]
Hello @Harman-2, can you please sign the CLA (https://cla.developers.google.com/ )? It is a required step to contribute to UCP? Thanks in advance!

**Comment by Harman-2** [comment]
@google-cla recheck

**Comment by Harman-2** [comment]
@damaz91 Thanks for heads up! The CLA has been signed.

**Comment by dkoch74** [comment]
@Harman-2 can you please add a unit test to demonstrate the RegEx works for expected patterns? Thanks

File: scripts/test_ap2_mandate_pattern.py [hunk 1]
```diff
@@ -0,0 +1,91 @@
...
+  valid = [
+    ("compact_jwt_only", "eyJhbGciOiJFUzI1NiJ9.eyJpc3MiOiJ1c3AifQ.signature"),
+    ("trailing_tilde_no_kb", "eyJhbGciOiJFUzI1NiJ9.payload.sig~"),
+    (
+      "disclosures_and_trailing_tilde",
+      "eyJhbGciOiJFUzI1NiJ9.payload.sig~disclosure1~disclosure2~",
+    ),
+    (
+      "delegate_chain_reference_shape",
+      "eyJhbGciOiJFUzI1NiJ9.payload.sig~~disclosure1~disclosure2~",
```
GarethCOliver, line 58: This is not quite the correct shape. Delegate SD-JWT (https://datatracker.ietf.org/doc/draft-gco-oauth-delegate-sd-jwt/)  defines it as: 

Issuerjwt~issuerd1~issuerd2~~kbjwt~delegated1~delegated2~

so your test case would be: 

"eyJhbGciOiJFUzI1NiJ9.payload.sig~issuerd1~~header.payload.sig~delegated1"

Other test cases I would recommend: 

normal SD-JWT+KB: 0 disclosure, 1 disclosure
delegate SD-JWT: 1 delegation, 2 delegation, KB-JWT termination, 1 disclosure, 2 disclosure. 

I think I mentioned in another issue that validating this is probably just a bad idea, given that we'll likely end up having to change this (again) if someone decides to use other cryptographic formats, and I don't think that on-the-wire validation is really very interesting given you need to do proper verification anyways, but if you do the above it's at least correct for now.

vishkaty:
The reference SDK emits two serializations this pattern rejects: a single hop compact SD-JWT with a trailing tilde, and a multi hop chain joined by a double tilde. Both are captured as executed vectors in #599 if they are useful as the unit test requested above.

PR #418, opened 2026-05-07
docs: explain protocol vs capability versioning

ptiper:
# Description

- Update versioning.md with high-level two-tiered versioning concepts.
- Link to detailed specifications in overview.md to avoid duplication.
- Add protocol compatibility alert and negotiation links in overview.md.
- Fix markdown formatting to pass lint checks.

## Category (Required)

- [ ] **Core Protocol**: Changes to the base communication layer, global context, or breaking refactors. (Requires Technical Council approval)
- [ ] **Governance/Contributing**: Updates to GOVERNANCE.md, CONTRIBUTING.md, or CODEOWNERS. (Requires Governance Council approval)
- [ ] **Capability**: New schemas (Discovery, Cart, etc.) or extensions. (Requires Maintainer approval)
- [X] **Documentation**: Updates to README, or documentations regarding schema or capabilities. (Requires Maintainer approval)
- [ ] **Infrastructure**: CI/CD, Linters, or build scripts. (Requires DevOps Maintainer approval)
- [ ] **Maintenance**: Version bumps, lockfile updates, or minor bug fixes. (Requires DevOps Maintainer approval)
- [ ] **SDK**: Language-specific SDK updates and releases. (Requires DevOps Maintainer approval)
- [ ] **Samples / Conformance**: Maintaining samples and the conformance suite. (Requires Maintainer approval)
- [ ] **UCP Schema**: Changes to the `ucp-schema` tool (resolver, linter, validator). (Requires Maintainer approval)
- [ ] **Community Health (.github)**: Updates to templates, workflows, or org-level configs. (Requires DevOps Maintainer approval)

## Checklist

- [X] I have followed the [Contributing Guide](https://github.com/Universal-Commerce-Protocol/.github/blob/main/CONTRIBUTING.md).
- [X] I have updated the documentation (if applicable).
- [X] My changes pass all local linting and formatting checks.
- [ ] (For Core/Capability) I have included/updated the relevant JSON schemas.
- [ ] I have regenerated Python Pydantic models by running generate_models.sh under python_sdk.

## Screenshots / Logs (if applicable)

Screenshot 2026-05-07 at 10 13 31

image

PR #337, opened 2026-04-05
feat!: replace identity linking with full identity management capability

sinhanurag:
# Description

This PR introduces the `dev.ucp.common.identity_management` capability to the Universal Commerce Protocol (UCP) specification, replacing the legacy `dev.ucp.common.identity_linking` capability. This change expands the protocol to support a full identity lifecycle, including creation, retrieval, updates, and deletion, in addition to the existing OAuth-based identity linking flow.

#### Documentation:
 - Created `identity-management.md` detailing the new capability and operations.
 - Created `identity-management-rest.md` and `identity-management-mcp.md` for transport-specific bindings.
 - Removed `identity-linking.md` (content absorbed and expanded in `identity-management.md`).
 - Updated `overview.md` to reference the new capability identifier.

#### Schemas:
 - Added `source/schemas/common/types/identity.json` defining the base user identity profile.

#### API Definitions:
 - Added `/identities` and `/identities/{id}` endpoints to `source/services/common/rest.openapi.json`.
 - Added `create_identity`, `get_identity`, `update_identity`, and `delete_identity` operations to    `source/services/common/mcp.openrpc.json`.

#### Motivation and Context: 
The previous specification only supported linking existing identities. To support a complete commerce experience, platforms and businesses need to manage the full lifecycle of a user's identity. This PR establishes that foundation.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] **Breaking change** (fix or feature that would cause existing
      functionality to not work as expected, **including removal of schema files
      or fields**)
- [ ] Documentation update

---

### Is this a Breaking Change or Removal?

If you checked "Breaking change" above, or if you are removing **any** schema
files or fields:

- [x] **I have added `!` to my PR title** (e.g., `feat!: remove field`).
- [x] **I have added justification below.**

## Breaking Changes / Removal Justification

The removal of `identity-linking.md` and the replacement of the `dev.ucp.common.identity_linking` capability with `dev.ucp.common.identity_management` constitutes a breaking change for the specification. While the core OAuth flow is preserved under the new capability, the change in the capability identifier and the reorganization of the documentation files require this to be flagged as a breaking change to ensure implementers update their references.

---

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
[...]

PR #614, opened 2026-07-20
docs: clarify UCP core release versioning contract

### LINKED ISSUE #499: [RFC]: Refactor of codebase to support capability versioning independent of protocol version
## Summary

This is a proposal to refactor the GitHub repository and CI/CD pipeline to support independent versioning for UCP capabilities, extensions, and types and also prepare us for the future of UCP with multiple verticals. By decoupling these from the protocol version, the design enables granular maintenance of the specification. We will cut a new version of the protocol to support these changes.

## Motivation

Currently, the UCP GitHub structure and published JSON schemas only support versioning at the protocol level, which creates significant gaps. Specifically, shared types lack individual versions, and built JSON files are restricted to protocol-wide releases. This refactor addresses these limitations to allow capabilities and shared types (like totals.json) to evolve independently of the core protocol.

## Goals

UCP defines a versioning scheme that supports versions at the protocol level as well as the capability level. However, the structure of the GitHub, documentation pages, and published JSON schema currently only work for versioning at the protocol level. The updated design should support:

1. Protocol versions
2. Capability/extension versions
    1. Capability version to protocol version support mapping can be many-to-many
    2. Types will need versions as specific capability versions may need specific type definitions (e.g. [total.json](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/shopping/types/total.json) change between v2026-01-23 to v2026-04-08)

While refactoring the GitHub repo to achieve this goal, we need to optimize across a few conflicting considerations

1. **Authoring simplicity** - how easy is it for contributors to who are augmenting the protocol to change the documentation
2. **Reader experience** - how easy is it for the reader to navigate, search, and understand the documentation structure
3. **Maintenance burden for implementers** - how easy is it for platforms and businesses to build and update their implementations of the protocol as it evolves
4. **Infrastructure resilience** - how dependent is it on complex setups using external plugins, and how easy is it for devops maintainers to manage the overall documentation and schema build pipeline
5. **Future robustness** - how easy is it to accommodate new verticals, backport new changes

## Current gaps

1. Current GitHub structure handles versions using branches, with no versioning below
2. Spec pages only have a version selector at the protocol level
3. Built JSON files only support versions at the protocol level (e.g. https://ucp.dev/2026-04-08/schemas/shopping/checkout.json)
4. Shared types don’t have versions

## Proposed changes

Currently the capability version is always the same as the UCP version, but this proposal will decouple them. The version property in the JSON schemas for capability will represent the version of the capability.

### Capability version assumptions

We will assume that a capability is only compatible with the most recent spec version equal to or less than the capability version. This is because any change to the spec version would imply a change that affects all capabilities, and so all capability versions will be bumped at the same time.

Example, with UCP-capability version compatibility represented by having the same color boxes:

Image

UCP version bump (2026-05-01 to 2026-07-01):

Image

### GitHub source structure

1. (unchanged) Maintain versioned release branches for the spec version (e.g. release/2026-04-08)
2. Add version to vertical/types (e.g. shopping/types/2026-04-08)
3. Represent capability and types versions as subdirs

Proposed structure would look like:
```
    ├── site/
    │   └── … (everything on ucp.dev not in Specification tab)
    ├── specification/
    │   ├── overview.md
    │   ├── playground.md
    │   ├── … (other core UCP spec documentation)
    │   ├── shopping/
    │   │   ├── checkout/
    │   │   │   ├── 2026-01-23/
    │   │   │   │   ├── checkout.md
    │   │   │   │   ├── checkout-rest.md
    │   │   │   │   └── …
    │   │   │   ├── 2026-04-08/
    │   │   │   │   ├── checkout.md
    │   │   │   │   ├── checkout-rest.md
    │   │   │   │   └── …
    │   │   │   ├── draft/
    │   │   │   └── … (other checkout versions)
    │   │   ├── cart/
    │   │   └── … (other capabilities, extensions)
    │   └── … (other verticals)
    └── source/
        ├── discovery/
        ├── handlers/
        ├── services/
        │   ├── shopping/
        │   │   ├── 2026-01-23/
        │   │   ├── 2026-04-08/
        │   │   └── … (other shopping services versions)
        │   └── … (other verticals)
        ├── schemas/
        │   ├── shopping/
        │   │   ├── checkout/
        │   │   │   ├── 2026-01-23/
        │   │   │   │   ├── checkout.json
        │   │   │   │   └── …
        │   │   │   ├── 2026-04-08/
        │   │   │   │   ├── checkout.json
        │   │   │   │   └── …
        │   │   │   ├── draft/
        │   │   │   └── … (other checkout versions)
        │   │   ├── cart/
        │   │   ├── … (other capabilities)
        │   │   └── types/
        │   │       ├── 2026-01-23/
        │   │       │   ├── totals.json
        │   │       │   └── …
        │   │       ├── 2026-04-08/
        │   │       └── … (other shopping types versions)
        │   └── … (other verticals)
        └── … (other core UCP spec schemas)
```

### CI/CD

Update publishing hooks to handle the new structure above. Ultimately we should have published URLs like:
1. Documentation
    1. Unversioned structure for pages that are not specifically tied to the spec, e.g. ucp.dev/documentation/core-concepts/
    2. Protocol versioned structure for pages that are tied only to spec version, e.g. ucp.dev/2026-04-08/specification/overview/
    3. Capability and protocol versioned structure for capability documentation, e.g. ucp.dev/2026-04-08/specification/checkout/2026-04-08/overview/
2. Schemas
    1. We will change the location of the version in the path to better reflect that the version is referencing capability/service/types version instead of UCP version, e.g.:
        1. Capabilities - ucp.dev/schemas/shopping/2026-06-01/checkout.json
        1. Services - ucp.dev/services/shopping/2026-06-01/rest.openapi.json
        1. Types - ucp.dev/schemas/shopping/types/2026-06-01/totals.json

### Files

1. Add version property to all types
2. Add version to type references (e.g. "$ref": "types/totals.json" → "$ref": "types/2026-04-08/totals.json")

### Website

1. (unchanged) Specification tab will have dropdown selector to choose UCP protocol version
2. Update Specification navigation to include the versions of each capability and extension in the nav tree that are supported in the selected version of the protocol:
```
	Specification (2026-04-01 dropdown selected)
	Overview (overview.md)
	Shopping
		Checkout Capability
			Version 2026-05-01 (checkout/2026-05-01/checkout.md)
				Overview
				…
			Version 2026-04-01 (checkout/2026-04-01/checkout.md)
			Draft (checkout/draft/checkout.md)
		Cart Capability
		Catalog Capability
		…
		Reference
		Playground
	Common
		Identity Linking Capability
	Payment Handlers
	Signatures
```
### Backporting changes

Backporting changes will not be allowed across UCP versions, but will be allowed for earlier versions of capabilities within the same UCP version. To do this, you would add the change to the same PR because capability versions will be contained within the same GitHub branch.

## Open questions

1. Services currently have references back to specific capabilities, e.g. in https://ucp.dev/2026-04-08/services/shopping/rest.openapi.json:

```
"checkout": {
  "$ref": "https://ucp.dev/2026-04-08/schemas/shopping/checkout.json"
},
"checkout_response": {
  "oneOf": [
    {
      "$ref": "https://ucp.dev/2026-04-08/schemas/shopping/checkout.json"
    },
    {
      "$ref": "https://ucp.dev/2026-04-08/schemas/shopping/types/error_response.json"
    }
  ]
},
```

Suggestion: we can change the specific capability objects to oneOf an array of objects, so the above example could become:

```
"checkout": {
  "oneOf": [
    {
      "$ref": "https://ucp.dev/2026-04-08/schemas/shopping/checkout.json"
    },
    {
      "$ref": "https://ucp.dev/2026-06-01/schemas/shopping/checkout.json"
    }
  ]
},
"checkout_response": {
  "oneOf": [
    {
      "$ref": "https://ucp.dev/2026-04-08/schemas/shopping/checkout.json"
    },
    {
      "$ref": "https://ucp.dev/2026-06-01/schemas/shopping/checkout.json"
    },
    {
      "$ref": "https://ucp.dev/2026-04-08/schemas/shopping/types/error_response.json"
    },
    {
      "$ref": "https://ucp.dev/2026-06-01/schemas/shopping/types/error_response.json"
    }
  ]
},
```

2. https://ucp.dev/latest/specification/overview/#version-requirements states that extensions should declare a requires object that specify protocol and capability versions required for correct operation. However, none of our currently defined extensions declare this. Should we add it?

igrigorik:
UCP intentionally gave services, capabilities, and extensions their own version fields so they could evolve independently from `ucp.version`. The specification described that flexibility as part of the protocol model.

In practice, UCP's publishing and certification process settled on an atomic release contract: each dated core release ships its services and transport bindings, capabilities, extensions, and shared schemas as one tested snapshot. This PR codifies that current contract rather than continuing to imply independent release schedules that UCP does not currently exercise.

## The release contract

For a selected release `D`:

1. The Platform resolves one exact Business profile for `D`, either the current profile or a `supported_versions` leaf whose `ucp.version` matches its map key.
2. `ucp.version = D` selects the UCP core snapshot published under `D`.
3. Every UCP-authored service, capability, and extension declaration in that profile carries `D`.
4. Exact capability intersection selects the active capabilities and extensions from the chosen snapshot.
5. `requires.protocol` and `requires.capabilities` verify the selected versions; they do not select replacements.

```text
Business profiles
  ├── current profile
  └── supported_versions[D]
                │
                ▼
       exact profile for D
                │
                ▼
       UCP core snapshot D
                │
                ▼
   exact capability intersection
                │
                ▼
       verify requires ranges
```

Transport descriptions remain versioned release artifacts. OpenAPI and OpenRPC `info.version` identifies the artifact published under `D`; it is not another profile negotiation axis. This does not prevent transport specifications from evolving—it defines how UCP currently bundles them into a core release.

The individual version fields remain in place. They preserve the existing wire shape and leave room for UCP to define independently versioned core components in a future contract without requiring a profile migration.

## Independent contracts

Third-party extensions and payment handlers retain their author-controlled versions and release cadence. The build hook now stamps the release version only onto UCP-authored `dev.ucp.*` schemas, making that ownership boundary explicit without changing current generated output. The PR also clarifies that `dev.ucp.*` capabilities are governed by the UCP Tech Council responsible for their domain.

## Documentation (re)structure

To avoid repeating subtly different rules across the documentation:
- **Protocol Version** owns exact profile discovery and `supported_versions`.
- **Capability Versions** owns exact capability intersection.
- **Component Versioning and Release Snapshots** owns release-`D` alignment and version ownership.
- **Versioning** owns the release-branch and backport workflow.
- Core Concepts and Schema Authoring summarize the relevant rules and link to their normative definitions.

## Compatibility

This is a contract clarification, not a wire migration.

- No profile shape changes.
- No fields added or removed.
- No required-field changes.
- No payload-validation changes.
- No change to current generated artifacts.
- Schema changes are descriptions only.
- Existing illustrative names, URLs, versions, and placeholders are unchanged.

## Related Issues

Closes #499. Supersedes #418.

---

## Checklist
- [x] **Infrastructure**: CI/CD, Linters, or build scripts.


docs/versioning.md
```diff
@@ -1,27 +1,36 @@
...
 
-New development occurs on the `main` branch. We will maintain long-lived
-branches for all supported releases of the spec.
+New development occurs on the `main` branch. We maintain long-lived branches for
+all supported releases of the spec so that a published snapshot `D` stays
+available for reference and maintenance. Approved backward-compatible changes —
+features as well as security, correctness, and interoperability fixes — may be
+backported to a maintained `release/D` branch and the artifacts generated from
+it, and UCP re-certifies the snapshot before publishing the updated artifacts.
+Breaking changes are never backported; they enter the next release.
```
jingyli: Trying to pressure test this statement here: I do agree with it for almost all of the cases we will face but do we want to give us some break-glass mechanism to override this (subject to Governance Council's discretion on a case-by-case evaluation) when there is a sensitive breaking fix that needs to be backported into a previous release?

Example from the `v2026-04-08` release: https://github.com/Universal-Commerce-Protocol/ucp/discussions/441

docs/specification/overview.md
```diff
@@ -2398,26 +2430,59 @@ The following changes **MUST NOT** be introduced without a new version:
...
+`D` on every `dev.ucp.*` service, capability, and extension entry in its
+profile. An older release is selected only through a separate
+`supported_versions` leaf profile whose own `ucp.version` is that older release
+date (see [Protocol Version](#protocol-version)). Payment-handler versions remain
+independently controlled by their authors and are not constrained
+to `D`. Declaring `version` `D` does not change negotiation: `dev.ucp.*`
+capabilities and extensions are still selected by exact-version intersection (see
+[Capability Versions](#capability-versions)).
+
+#### Third-party extensions (`com.{vendor}.*`, `org.{org}.*`)
```
jingyli, line 2944: With this new language, we are now limiting the scope of third-party/vendor's participation to just extensions (vs. before it was generalized to capabilities). Is this reduction in vendor scope intentional?

In my mind, the main pros and cons will be:
- (+): Tightening the usage so that we ensure UCP prescribes the main building blocks (base capabilities) and anything vendors introduce is only supplemental to these main building blocks
- (-): Removing the flexibility for implementors to "mix-and-match" - i.e. businesses & platforms cannot mutually agree to negotiate for a vendor catalog capability and UCP's checkout capability

My gut feeling is that the con above will eliminate some interesting use cases - especially for base capabilities that UCP has not yet defined (i.e. if a vendor wants to temporarily lean on their own identity management capability before some flavour of #337 lands in UCP).

docs/specification/overview.md
```diff
@@ -2360,12 +2374,17 @@ notice.
...
 #### Capability Versions
 
-Capability versions are negotiated independently of the protocol
-version. Each capability in the profile is an array. Multiple entries
-for the same capability, each with a different `version`, advertise
-support for multiple versions of that capability. The capability
-intersection algorithm considers only capability versions supported
-by both parties.
+Capability compatibility is established only by exact version equality, not by
+inferring compatibility from date order. Each capability in the profile is an
```
jingyli, line 2830: Trying to visualize how this works in practice, especially cross-referencing against our new guidance on UCP-authored releases (please correct me if anything deviates from how you are thinking about it):

Scenario 1 - valid ✅ 
```
{
  "ucp": {
    "version": "2026-04-08",
    "supported_versions": {
      "2026-01-23": "https://business.example.com/.well-known/ucp/2026-01-23"
    }
    "services": {...},
    "capabilities": {
      "dev.ucp.shopping.checkout": [
        {
          "version": "2026-04-08",
          "spec": "https://ucp.dev/2026-04-08/specification/checkout",
          "schema": "https://ucp.dev/2026-04-08/schemas/shopping/checkout.json"
        }
      ],
    },
   ....
}
```

Scenario 2 - valid ✅ 
```
{
  "ucp": {
    "version": "2026-04-08",
    "supported_versions": {
      "2026-01-23": "https://business.example.com/.well-known/ucp/2026-01-23"
    }
    "services": {...},
    "capabilities": {
      "dev.ucp.shopping.checkout": [
        {
          "version": "2026-04-08",
          "spec": "https://ucp.dev/2026-04-08/specification/checkout",
          "schema": "https://ucp.dev/2026-04-08/schemas/shopping/checkout.json"
        },
        {
          "version": "2026-01-23",
          "spec": "https://ucp.dev/2026-01-23/specification/checkout",
          "schema": "https://ucp.dev/2026-01-23/schemas/shopping/checkout.json"
        }
      ],
    },
   ....
}
```

Scenario 3 - invalid ❌ 
```
{
  "ucp": {
    "version": "2026-04-08",
    "supported_versions": {
      "2026-01-23": "https://business.example.com/.well-known/ucp/2026-01-23"
    }
    "services": {...},
    "capabilities": {
      "dev.ucp.shopping.cart": [
        {
          "version": "2026-04-08",
          "spec": "https://ucp.dev/2026-04-08/specification/cart",
          "schema": "https://ucp.dev/2026-04-08/schemas/shopping/cart.json"
        }
      ],
      "dev.ucp.shopping.checkout": [
        {
          "version": "2026-01-23",
          "spec": "https://ucp.dev/2026-01-23/specification/checkout",
          "schema": "https://ucp.dev/2026-01-23/schemas/shopping/checkout.json"
        }
      ],
    },
   ....
}
```

Or is it the case that **Scenario 3** would also be valid - i.e. the sentence `UCP-authored
dev.ucp.* entries declare version D in release D` only applies to the release, and not on how the profile can be built?

docs/specification/overview.md
```diff
@@ -2274,17 +2285,20 @@ and signing keys. When `supported_versions` is omitted, only
...
-2. If the platform's protocol version matches `version`: use this
-    profile directly. Proceed to capability negotiation.
-3. If the platform's protocol version is a key in
-    `supported_versions`: fetch the profile at the mapped URI. This
-    profile describes the capabilities available at that protocol
-    version. Proceed to capability negotiation.
-4. Otherwise: the business does not support the platform's protocol
+2. If the Platform's protocol version matches `version`, it uses this
+    profile directly and proceeds to capability negotiation.
+3. If the Platform's protocol version is a key in
```
jingyli: Trying to poke this sentence a bit more, especially approaching it from the angle where platforms may also have multiple supported protocol versions: what this is indicating is that negotiation's scope is limited to that mapped URI being fetched (and there should ever be only one profile being "selected" for negotiation) and there should be no cross-profile "merging" happening. <- is this correct?

This would mean let's say if a business advertises cart capability **only** in their `v2026-04-08` profile (which has `supported_versions` containing `v2026-01-23`) and checkout capability in their `v2026-01-23` profile, then from platform's perspective (if both versions are supported), only cart **or** checkout capability can be negotiated (and not both via running intersection algorithms against both profiles)?

The only way to achieve negotiating both capabilities would be to have both capabilities (cart on `2026-04-08` version and checkout on `2026-01-23` version) advertised in the same `v2026-04-08` profile?

jingyli (commented):
Thanks for proposing the clarification! Given this PR, it would also significantly simplify version's correlation to UCP's releases (related RFC - #499), especially if the releases are going to be done at a fixed/regular cadence.

But curious if we have some early thoughts around: How will the first release for new services look like? Let's hypothetically say a new service (i.e. food ordering) has an approved spec tomorrow, will they need to wait until the next official release of UCP to stamp a version `D` on their capabilities? Or will these be considered as backwards-compatible changes (since they are net new capabilities) and be backported into a previous stable release?

docs/specification/overview.md
igrigorik: Addressed in https://github.com/Universal-Commerce-Protocol/ucp/pull/614/changes/7d8b56bdf2e03b65aedf55eee471fcc86cd4d34e

docs/specification/overview.md
igrigorik: Clarified in https://github.com/Universal-Commerce-Protocol/ucp/pull/614/changes/7d8b56bdf2e03b65aedf55eee471fcc86cd4d34e.

docs/specification/overview.md
igrigorik: Addressed in https://github.com/Universal-Commerce-Protocol/ucp/pull/614/changes/7d8b56bdf2e03b65aedf55eee471fcc86cd4d34e

igrigorik:
@jingyli great questions & flags, clarified in https://github.com/Universal-Commerce-Protocol/ucp/pull/614/changes/7d8b56bdf2e03b65aedf55eee471fcc86cd4d34e — ptal. 

> But curious if we have some early thoughts around: How will the first release for new services look like? Let's hypothetically say a new service (i.e. food ordering) has an approved spec tomorrow, will they need to wait until the next official release of UCP to stamp a version D on their capabilities? Or will these be considered as backwards-compatible changes (since they are net new capabilities) and be backported into a previous stable release?

Both paths are possible under our contract. If a net-new service is a backwards-compatible addition, then a newly approved vertical either:

1. ships in the next release `D2` and declares `D2` (default), or
2. if approved for backport, lands on the maintained `release/D` branch, declares `D`, and the snapshot is re-certified.

Practically, I would push us to synchronize releases and snap to common dates rather than cut per-vertical releases — the shared date is what guarantees a common core and cross-service interop. Today that cadence is quarterly (targeting ~12 weeks), so in your example food ordering slots into the next release train rather than waiting indefinitely. The cadence itself is release-process policy, not codified in the spec -- the GC can choose to adjust the schedule.

docs/specification/overview.md
```diff
@@ -2714,23 +2725,31 @@ and signing keys. When `supported_versions` is omitted, only
...
+1. The Platform fetches `/.well-known/ucp` — this is the current version
     profile.
-2. If the platform's protocol version matches `version`: use this
-    profile directly. Proceed to capability negotiation.
-3. If the platform's protocol version is a key in
-    `supported_versions`: fetch the profile at the mapped URI. This
-    profile describes the capabilities available at that protocol
-    version. Proceed to capability negotiation.
-4. Otherwise: the business does not support the platform's protocol
+2. If the Platform's protocol version matches `version`, it uses this
```
jingyli: Would it possible to reword this bullet here:
```
2. If the Platform's protocol version matches `version`, it **MAY** use this profile
    directly and proceed to capability negotiation. If the Platform supports multiple
    versions, it reserves the right to select a different mutually supported profile by
    proceeding to step 3 instead.
```

We'd like to propose loosening up the language here so platforms are not locked in on always selecting the "current version" if it also supports multiple UCP versions - platforms should have the flexibility to choose which business profile they'd like to negotiate against.

jingyli (commented):
Thanks @igrigorik for the updates in https://github.com/Universal-Commerce-Protocol/ucp/commit/7d8b56bdf2e03b65aedf55eee471fcc86cd4d34e!

We added one proposal (https://github.com/Universal-Commerce-Protocol/ucp/pull/614/changes#r3694051647) to provide platforms, who may support multiple versions, flexibility to select which business profile they would like to negotiate against and not always be forced into choosing the "current" version under the current wording - @sinhanurag @gsmith85 FYI

Also ack on https://github.com/Universal-Commerce-Protocol/ucp/pull/614#issuecomment-5096222268, personally I prefer the backporting route to minimize having to wait for `D2`, especially in the rare instances where a vertical may have just missed the release train.

docs/specification/overview.md
igrigorik: Makes sense, landed in https://github.com/Universal-Commerce-Protocol/ucp/pull/614/changes/6620c63665b94a4eba2226456c59e8fbdb3e8d5e

igrigorik:
@jingyli landed, ready for final pass!

gsmith85 (approved):
LGTM! Gave it one last read through, it looks great.  Thanks @igrigorik for formalizing the core release snapshot contract. This provides clear alignment between `ucp.version`, profile discovery, and component artifacts.

docs/documentation/core-concepts.md
```diff
@@ -127,16 +127,17 @@ interact.
...
 Each capability is identified by a reverse-domain name (e.g.,
 `dev.ucp.shopping.checkout`) and carries a date-based version. Capabilities
-are declared in the business's UCP profile at `/.well-known/ucp` and
-negotiated and confirmed in every response so that the platform always knows
-the active feature set for a given interaction.
+are declared in the Business's UCP profile at `/.well-known/ucp`, negotiated
+by exact version, and confirmed in every response so that the Platform always
+knows the active feature set for a given interaction. UCP-authored `dev.ucp.*`
+capabilities and extensions declare version `D` in each UCP release `D`.
+Third-party extensions publish versions on their own cadence.
```
raginpirate:
```suggestion
knows the active feature set for a given interaction. UCP-authored `dev.ucp.*`
capabilities and extensions are versioned in lockstep with the specification:
the version is the date of the UCP release they ship in. Third-party extensions
publish versions on their own cadence.
```

docs/documentation/core-concepts.md
```diff
@@ -204,12 +205,15 @@ A single service can be accessed via multiple transport bindings:
...
-A business declares which transport bindings it supports within each service;
-platforms pick whichever fits their context — an AI agent may prefer MCP, a
-traditional web app may use REST.
+A Business declares which transport bindings it supports within each service;
+Platforms pick whichever fits their context — an AI agent may prefer MCP, a
+traditional web app may use REST. Every UCP-defined service declares an explicit
+`version` equal to the selected `ucp.version`; because each entry pairs the
+service with one transport binding, that service `version` repeats on each entry.
+Transport bindings have no separate version. Multiple UCP-defined vertical
+services can coexist under one `ucp.version`, each declaring that version.
```
raginpirate:
```suggestion
traditional web app may use REST. Every UCP-defined service declares an explicit
`version` equal to the selected `ucp.version`, repeated on every entry since they
differ only by transport binding. Transport bindings have no separate version.
```

docs/documentation/core-concepts.md
```diff
@@ -438,20 +442,19 @@ Businesses publish their OAuth 2.0 server metadata at
...
-the checkout capability it extends.
-
-See [Versioning](../versioning.md) for the release branch process and the
-full breaking-change procedure.
+UCP uses date-based version identifiers (`YYYY-MM-DD`). A UCP release `D` is a
+snapshot of the core protocol — its services and transport bindings,
+capabilities, extensions, and shared schemas — published and certified together
+as internally compatible. The selected `ucp.version` selects that snapshot;
+selection is exact, so an older date is available only when the Business
+advertises it in `supported_versions`.
```
raginpirate:
```suggestion
capabilities, extensions, and shared schemas — published together as one                                                                                                                                                                                                            
internally compatible set. A profile's `ucp.version` selects that snapshot, and                                                                                                                                                                                                     
the match is exact: an older date is available only when the Business                                                                                                                                                                                                               
advertises it in `supported_versions`. 
```

docs/documentation/core-concepts.md
```diff
@@ -438,20 +442,19 @@ Businesses publish their OAuth 2.0 server metadata at
...
+
+Every UCP-defined service, capability, and extension in release `D` declares
+version `D`. Capabilities and extensions are selected by exact-version
+intersection. Third-party extensions and payment handlers version independently
+on their authors' own schedules.
```
raginpirate:
```suggestion
version `D`. Third-party extensions and payment handlers are versioned by their                                                                                                                                                                                                     
authors, independently of UCP releases. 
```

docs/documentation/schema-authoring.md
```diff
@@ -99,7 +106,14 @@ Examples: `checkout.json`, `fulfillment.json`, `discount.json`, `order.json`
 Define transport bindings that appear in `ucp.services{}` registries. Each transport
 (REST, MCP, A2A, Embedded) is a separate entry.
 
-- **Top-level fields**: `$schema`, `$id`, `title`, `description`, `name`, `version`
+- **Registry identity**: Reverse-domain service name used as the registry key
+- **Entry fields**: `version`, `spec`, `schema`, `config`, and transport-specific fields
+- **Versioning**: Every service entry declares an explicit `version`; in release
+  `D` it is `D`. Because each entry pairs the service with one transport binding,
```
raginpirate, line 112:
```suggestion
- **Versioning**: Every service entry declares an explicit `version`; in UCP release
  `D` it is `D`. Because each entry pairs the service with one transport binding,
```

docs/specification/overview.md
```diff
@@ -414,6 +407,15 @@ standard formats:
 - **A2A**: Agent Card Specification
 - **EP(embedded)**: OpenRPC (JSON format)
 
+A service is identified by its reverse-domain registry key (e.g.,
+`dev.ucp.shopping`). In a profile, services are keyed by that name, and each
+entry in `services[name][]` pairs the service with one transport binding and
+declares the service `version`: in release `D` that version is `D`. This is the
```
raginpirate, line 413:
```suggestion
declares the service `version`: in UCP release `D` that version is `D`. This is the
```

docs/specification/overview.md
```diff
@@ -568,16 +570,22 @@ This convention ensures:
...
 ##### Version Requirements
 
-Extension schemas **SHOULD** declare a `requires` object (alongside
-`name`, `title`, `description`) to indicate the protocol and
-capability versions required for correct operation:
+Extension authors **SHOULD** declare a `requires` object in the extension
+schema (alongside its `name`, `title`, and `description`) stating the versions
+the extension depends on. A third-party extension schema declares its own
+author-controlled `version`, which advances independently of `ucp.version`;
+a UCP-authored `dev.ucp.*` extension declares version `D` in release `D`.
```
raginpirate, line 577:
```suggestion
a UCP-authored `dev.ucp.*` extension declares version `D` in UCP release `D`.
```

docs/specification/overview.md
```diff
@@ -2800,12 +2823,22 @@ notice.
...
-intersection algorithm considers only capability versions supported
-by both parties.
+Capability compatibility is established only by exact version equality, not by
+inferring compatibility from date order. Each capability in the profile is an
+array; multiple entries with different `version` values advertise support for
+multiple versions where the applicable publication policy permits it. When the
+exact shared set contains more than one version, the intersection algorithm
+orders those shared dates to select the latest one. Supported third-party
+extension versions advance independently of `ucp.version`. UCP-authored
+`dev.ucp.*` entries declare version `D` in release `D`: a profile for
```
raginpirate, line 2836:
```suggestion
`dev.ucp.*` entries declare version `D` in UCP release `D`: a profile for
```

docs/specification/overview.md
```diff
@@ -2838,26 +2881,62 @@ The following changes **MUST NOT** be introduced without a new version:
...
-#### Vendor Capabilities (`com.{vendor}.*`)
-
-Capabilities outside the `dev.ucp.*` namespace version fully independently.
-Vendors control their own release schedules and versioning strategy.
+### Component Versioning and Release Snapshots
+
+A UCP release `D` is a snapshot of the core protocol — its services and transport
+bindings, capabilities, extensions, and shared schemas — published and certified
+together as internally compatible. Selecting `ucp.version` `D` selects that
+snapshot. For a release `D`:
```
raginpirate, line 2892:
```suggestion
snapshot. For a UCP release `D`:
```

docs/specification/overview.md
```diff
@@ -2838,26 +2881,62 @@ The following changes **MUST NOT** be introduced without a new version:
...
+breaking-change notice process. UCP **MUST** re-certify the snapshot before
+publishing any updated artifacts.
+
+A Business or Platform that selects `ucp.version` `D` **MUST** declare version
+`D` on every `dev.ucp.*` service, capability, and extension entry in its
+profile. An older release is selected only through a separate
+`supported_versions` leaf profile whose own `ucp.version` is that older release
+date (see [Protocol Version](#protocol-version)). Payment-handler versions remain
+independently controlled by their authors and are not constrained
+to `D`. Declaring `version` `D` does not change negotiation: `dev.ucp.*`
```
raginpirate: Why do we have a special clause for payment handlers? If we were to ever introduce a core payment handler to UCP, we don't want to version it together?

docs/specification/overview.md
```diff
@@ -2838,26 +2881,62 @@ The following changes **MUST NOT** be introduced without a new version:
...
+5. UCP certifies the snapshot — components, bindings, and supported
+    compositions — together before publishing it.
+
+UCP **MAY** backport an approved backward-compatible change — a feature, or a
+security, correctness, or interoperability fix — to a supported release and its
+generated `D` artifacts. Breaking changes enter the next release and are not
+backported by default; in exceptional cases, the Governance Council **MAY**
+approve backporting a breaking change to a supported release, following the
+breaking-change notice process. UCP **MUST** re-certify the snapshot before
+publishing any updated artifacts.
```
raginpirate: This seems to imply that forever, any release of UCP is unstable and may require dynamic changes to transact with. Should there be a maximum time-window for these exceptional cases to be resolved within if its to correct immediate breaking bugs? Like maybe a maximum-one-back breaking release window for back-porting? 🤔

That can turn this clause from unstable for any implementor reading the fine-print, to a reasonable control we have to keep our protocol functioning.

raginpirate (commented):
Almost all nits, one concern about our backporting stance for breaking changes. (https://github.com/Universal-Commerce-Protocol/ucp/pull/614#discussion_r3723985614)

docs/documentation/schema-authoring.md
igrigorik: UCP bit is implicit, I'll leave these out.

docs/specification/overview.md
igrigorik: clarified

docs/specification/overview.md
igrigorik: clarified

igrigorik:
@raginpirate good nits! Updated: https://github.com/Universal-Commerce-Protocol/ucp/pull/614/changes/dac71fb12f3a1868f41017cc174126f7f977f61d.

raginpirate (approved):
🚀

tdjin-stripe (approved):

docs/documentation/core-concepts.md
```diff
@@ -438,20 +441,18 @@ Businesses publish their OAuth 2.0 server metadata at
...
-the checkout capability it extends.
-
-See [Versioning](../versioning.md) for the release branch process and the
-full breaking-change procedure.
+UCP uses date-based version identifiers (`YYYY-MM-DD`). A UCP release `D` is a
+snapshot of the core protocol — its services and transport bindings,
+capabilities, extensions, and shared schemas — published together as one
+internally compatible set. A profile's `ucp.version` selects that snapshot, and
+the match is exact: an older date is available only when the Business
+advertises it in `supported_versions`.
```
amithanda: `main` currently carries the bullet "**Non-breaking additions** do not increment
the date" at `core-concepts.md:444`, and this PR removes it. The replacement prose
describes `D` as a "snapshot", "published together as one internally compatible
set", and in `schema-authoring.md:86` as "the entity's complete, published schema
closure". Those all read as frozen.

The mutability is still real, it just now lives only in the backport clause at
`overview.md:2904`: backwards-compatible changes may be backported into "a
supported release and its generated `D` artifacts", and that is the default path,
not the exceptional one. So two parties can both declare
`dev.ucp.shopping.checkout` at `D`, pass the exact-equality test that
[Capability Versions](../specification/overview.md#capability-versions) makes the
sole compatibility check, and still hold different schema closures if one cached
before a backport and the other after. Nothing in the profile distinguishes them.

Should the snapshot framing say plainly that a published `D` can be amended? One
sentence keeps it discoverable:

```suggestion
advertises it in `supported_versions`. A published snapshot is not frozen:
approved backwards-compatible changes may be backported into a released `D`
and its artifacts republished, so parties that cached at different times may
hold slightly different copies of the same `D`. Both copies are valid and
both parties remain conforming: backports are restricted to the
backwards-compatible class, so an earlier copy is missing later additions,
never in conflict with them. A party that meets a field or value its copy
does not define should ignore it rather than reject the payload, and should
not depend on a backported addition being understood by a peer that cached
earlier. Only an exceptional breaking backport can invalidate an earlier
copy, which is why that path requires Governance Council approval.
```

If the intent is that additive backports are always safe for a stale cache,
saying that explicitly would be better still, since it tells implementers the skew
is bounded rather than leaving them to infer it.

docs/specification/overview.md
```diff
@@ -2838,26 +2881,64 @@ The following changes **MUST NOT** be introduced without a new version:
...
+UCP **MAY** backport an approved backward-compatible change — a feature, or a
+security, correctness, or interoperability fix — to a supported release and its
+generated `D` artifacts. Breaking changes enter the next release and are not
+backported by default; in exceptional cases, the Governance Council **MAY**
+approve backporting a breaking change to a supported release, limited to
+defects that compromise the security, correctness, or interoperability of the
+release as published, and following the breaking-change notice process. UCP
+**MUST** re-certify the snapshot before publishing any updated artifacts.
+
+A Business or Platform that selects `ucp.version` `D` **MUST** declare version
```
amithanda, line 2930: > A Business or Platform that selects `ucp.version` `D` **MUST** declare version
> `D` on every `dev.ucp.*` service, capability, and extension entry in its
> profile.

This tells a publisher what not to emit. I could not find anywhere that tells a
Platform what to do when it fetches such a profile anyway, so two conforming
Platforms can behave differently on the same document.

Concretely, the profile in question is one where `ucp.version` and an entry
disagree inside a single document:

```jsonc
{ "ucp": { "version": "2026-04-08",
  "capabilities": {
    "dev.ucp.shopping.cart":     [ { "version": "2026-04-08" } ],
    "dev.ucp.shopping.checkout": [ { "version": "2026-01-23" } ]
  } } }
```

Worth separating this from the case the spec does cover. `overview.md:2751`
addresses a Platform holding **two** fetched profiles and wanting capabilities
from both, and it resolves that permissively: parallel negotiations are a
**MAY**, combining within one negotiation is a **SHOULD NOT**. The profile above
is a different shape. It is one document, so nothing is being combined across
profiles, and that guidance does not reach it.

The schema cannot catch it either. `source/schemas/ucp.json#/$defs/version` only enforces
the `YYYY-MM-DD` pattern, and nothing cross-checks a registry entry against
`ucp.version`, so such a profile is well formed and schema valid.
`profile_malformed` at `overview.md:1099` is scoped to "not valid JSON or violates
schema", which would not reach it. Build tooling does not help either, since this
governs profiles published by other parties at runtime.

Without a stated outcome the failure is indirect, and in one case silent. A
compliant Platform advertising `dev.ucp.shopping.checkout` at `D` finds no mutual
version with the mis-declared entry, and
[Intersection Algorithm](#intersection-algorithm) step 2 **excludes** it. If the
capability is required by the operation, the Business returns
`capabilities_incompatible`, which points the implementer at negotiation when the
fault is in the profile. If it is optional, the capability disappears from the
active set with no signal at all.

The document already defines outcomes for two comparable failures, so there is a
house pattern to follow. At `overview.md:2738`, a `supported_versions` profile
whose `ucp.version` disagrees with its map key gets "on mismatch the Platform
**MUST NOT** use that profile", rather than being left to fall through to
`version_unsupported`. At `overview.md:370`, an entity whose `schema` URL origin
fails the namespace-authority check is rejected, "treated as not present and
never activated". The second is the closer fit: it is entity-scoped rather than
profile-scoped, so one bad entry does not cost the Platform the whole profile.

```suggestion
A Business or Platform that selects `ucp.version` `D` **MUST** declare version
`D` on every `dev.ucp.*` service, capability, and extension entry in its
profile. A Platform that encounters a `dev.ucp.*` entry whose `version`
differs from the profile's `ucp.version` **MUST** reject that entry, treated
as not present and never activated, and **MAY** continue with the remaining
entries. An older release is selected only through a separate
```

The permissive direction works equally well: state that a Platform uses the
entries whose `version` matches `ucp.version` and ignores the rest. Either
answer is fine. The part I would want closed is that a Platform author today has
a hard publisher-side **MUST** and no consumer behavior to pair with it, so the
same malformed profile can be rejected by one Platform and silently
half-negotiated by another.

igrigorik:
@amithanda good catch on both comments — working through them surfaced a few related gotchas (notice-process ordering for security fixes, hosting policy not covering schema artifacts), landed updated wording in [1e1dea06](https://github.com/Universal-Commerce-Protocol/ucp/pull/614/commits/1e1dea06e3d20f99488964af84c8d02af48c406e).

amithanda (approved):
`f5581c8` covers both of my comments well. The skew paragraph states the part I
was after, that `D` is amended in place and earlier copies stay valid, and
grounding refresh behavior in standard HTTP caching plus the new
`ETag`/`Last-Modified` validator is a better answer than the one I proposed, it
makes skew self-correcting for runtime fetchers rather than merely documented.
Extending Hosting to cover schema and transport-description artifacts closes the
gap that made skew undetectable in the first place.

The consumer-side rejection rule reads right, and keeping it entity-scoped means
one malformed entry does not invalidate the full profile.

Approving!

PR #618, opened 2026-07-21
docs(ap2): clarify ES512 signature scope

### LINKED ISSUE #571: ap2_mandate requires ES512, but signatures.md's algorithm table only defines ES256/ES384 — should ES512 be added?
`ap2_mandate.json` requires the mandate JWS header `alg` to be one of `ES256`/`ES384`/`ES512`, and `ap2-mandates.md` lists all three — citing `signatures.md` as the algorithm authority ("uses the cryptographic primitives defined in the Message Signatures specification").

But `signatures.md`'s "Signature Algorithms" table enumerates only ECDSA P-256 (`ES256`), ECDSA P-384 (`ES384`), and EdDSA — `ES512`/P-521 appears nowhere in the document, and the implementation requirements only discuss the ES256 baseline and optional ES384.

So an implementation built to `signatures.md` wouldn't know to support verifying `ES512` and would reject a schema-valid AP2 mandate.

Question: should `ES512` be added to signatures.md's algorithm table, or is the AP2 extension's ES512 support intentional and scoped to mandates only (in which case a note in ap2-mandates.md clarifying the divergence would help)?

Found while building an unofficial UCP conformance test suite (https://spck.dev); verified against main @ 7e5fc42.

ShuoRen-TT:
## Summary

- separate AP2 JWS algorithm requirements from UCP HTTP Message Signatures
- document the `ES256`/P-256, `ES384`/P-384, and `ES512`/P-521 pairings
- require AP2 signing JWKs to include `alg`, and require verifiers to bind the
  protected JWS `alg` to the selected key's curve and JWK `alg`
- add a schema-annotated P-521/`ES512` profile key example
- clarify that AP2's accepted algorithms do not expand the universal HTTP
  signature baseline

## Motivation

AP2 currently accepts `ES512` for its JWS objects, while the shared HTTP
Message Signatures table stops at P-384/`ES384`. That can be read as either
an omitted global requirement or an AP2-specific extension.

This change takes the AP2-scoped interpretation already implied by the current
specification: AP2 owns its JWS algorithm policy, while reusing UCP's JWK
publication, discovery, and rotation mechanisms.

## Scope

This is a documentation clarification only. It does not:

- add P-521/`ES512` to UCP HTTP Message Signatures
- change the universal `ES256` implementation baseline
- modify `profile.json` or any generated schema

The existing profile JWK vocabulary is open, so the annotated P-521 example is
schema-compatible without a core schema change.

## Validation

- `git diff --check`
- focused check that the new example parses and is admitted by the current
  open JWK vocabulary
- `scripts/test_validate_examples.py`: 34 contract checks passed; its single
  integration check could not run locally because the `ucp-schema` binary is
  unavailable
- strict MkDocs build was attempted and reached the AP2 page, then stopped for
  the same missing local `ucp-schema` dependency

The PR Docs workflow will run schema example validation and the strict site
build with the repository's complete CI toolchain.

Closes #571


vishkaty:
Really like the alg↔curve binding you added here — that's the right safeguard. One thing I noticed while checking it against the schema: `profile.json` already machine-enforces that binding with `if/then` rules for P-256, P-384, and Ed25519, but there's no P-521 rule, so a P-521 key published with a mismatched `alg` currently passes validation (I confirmed with `ucp-schema`; the same mistake on a P-256 key is rejected).

I opened #625 with the one-line `P-521 pairs with ES512` rule so the binding your PR documents is also enforced at the schema layer — different file, no conflict, and I checked that your P-521 example JWK validates against it. Purely complementary; happy to fold it in here instead if you'd prefer to keep it in one place.

ShuoRen-TT:
Thanks for digging into this and for validating the example against #625 — I
really appreciate it. I agree that the schema asymmetry is real, and that the
algorithm-to-curve binding should be machine-enforced wherever P-521 is
recognized.

For now, I would prefer to keep #625 separate rather than fold it into this
draft. #618 is intentionally AP2-scoped and documentation-only while
maintainers confirm the intended layering. #625 also adds P-521/ES512 to the
profile's well-known values, examples, and descriptions, so it carries a
broader schema-level decision that is useful to triage independently.

One small consistency point for that review: the top-level
`jwk_public_key` description still lists EC well-known keys as P-256 and
P-384. If maintainers choose to make P-521/ES512 a UCP-wide well-known pairing,
that description should be aligned as well; if the pairing remains
AP2-specific, it may be clearer to qualify it that way.

The two changes remain technically compatible, and your confirmation that the
example here validates against #625 is very helpful. Thanks again for catching
the gap and preparing the focused validation.

docs/specification/ap2-mandates.md
```diff
@@ -129,18 +129,53 @@ If a public key cannot be resolved, or if the signature is invalid, the business
...
+| `ES384` | `EC` / `P-384` | SHA-384 |
+| `ES512` | `EC` / `P-521` | SHA-512 |
+
+This algorithm set applies only to AP2 JWS objects. In particular, `ES512`
+support here does not add P-521 to the UCP HTTP Message Signatures algorithm
+table or change its universal `ES256` implementation baseline.
 
 * **Algorithm:** per AP2's Checkout JWT signing rule — AP2 v0.2 requires
-  ECDSA (`ES256`/`ES384`/`ES512`); see the note below.
+  ECDSA (`ES256`/`ES384`/`ES512`); see the note below and
```
GarethCOliver, line 147: worth noting that we are likely to remove this restriction in FIDO in 0.3 (and just mandate randomness in the Checkout. But this seems fine for now.

ShuoRen-TT:
A concise review handoff:

- the scope remains AP2-specific and documentation-only;
- it does not modify `profile.json`, the universal `ES256` baseline, or the UCP HTTP Message Signatures algorithm requirements;
- #625 remains a separate schema-level/TC decision, and the two changes are technically compatible but independently reviewable;
- the PR is currently mergeable, and the CLA and `check-changes` checks pass;
- the external-contributor Docs/Lint/Spelling/Conventional Commits workflows are still awaiting maintainer approval to run.

I will respond promptly to any scope or wording feedback.

docs/specification/ap2-mandates.md
```diff
@@ -129,18 +129,53 @@ If a public key cannot be resolved, or if the signature is invalid, the business
...
 * **Key Format:** JWK ([RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517))
 * **Key Discovery:** `keys[]` in `/.well-known/ucp` (see
   [Key Discovery](overview.md#key-discovery))
 
 See [Message Signatures](signatures.md) for key format and rotation.
 
+When a verifier selects a key, it **MUST** ensure that the protected JWS
+`alg` is compatible with the key's curve according to the table above. This
+binding follows [RFC 8725 Section 3.1](https://datatracker.ietf.org/doc/html/rfc8725#section-3.1).
+If the JWK includes its optional `alg` member, that value **MUST** also match
```
GarethCOliver: I'd just mandate this be set, it's best practices.

GarethCOliver (approved):

docs/specification/ap2-mandates.md
ShuoRen-TT: Thanks for the suggestion - agreed. I updated this so keys used with the AP2
extension MUST include the JWK `alg`, and a missing or mismatched value makes
the AP2 signature invalid. The requirement remains AP2-scoped and does not
change the global profile JWK schema.

ShuoRen-TT:
Thanks again @GarethCOliver for the review. I pushed `119ffd4` to adopt the
`alg` suggestion:

- AP2 signing JWKs now **MUST** include `alg`;
- a missing or mismatched value makes the AP2 signature invalid; and
- the requirement remains AP2-scoped without changing the global profile JWK
  schema.

The focused cspell, markdownlint, and `git diff --check` validations pass. The
example validator remains unavailable locally because `ucp-schema` is not
installed; the repository CI provides that full validation once the
external-contributor workflows are approved.

@amithanda, could you please help approve the pending workflow runs and route
this documentation change to two eligible maintainers for review when
convenient? Thank you!

docs/specification/ap2-mandates.md
ShuoRen-TT: Thanks for the heads-up, Gareth. That makes sense. This PR keeps the
requirement explicitly scoped to AP2 v0.2 and retains the note pointing to
google-agentic-commerce/AP2#268. Once FIDO 0.3 publishes the normative Checkout
payload entropy requirement, we can follow up to align UCP with it rather than
anticipating the final wording here.

ShuoRen-TT:
Hi @amithanda, @jingyli, and @DanielFalconGuedes — I hope you're doing well.
I wanted to leave a gentle follow-up whenever you have a moment.

This PR remains explicitly scoped to AP2 v0.2 and is documentation-only.
Gareth's two non-blocking review comments have both been addressed:

- AP2 signing JWKs now **MUST** include `alg`, with the protected JWS `alg`
  bound to the JWK `alg` and curve.
- The possible FIDO 0.3 entropy-based change remains documented as future
  work, without anticipating normative wording that has not yet been
  published.

I have resolved both review conversations now that the feedback is incorporated
and documented. The branch remains cleanly mergeable, and the CLA and
`check-changes` checks pass. #625 remains a separate TC-level schema decision
and is not a dependency of this documentation clarification.

@amithanda, could you please approve the pending external-contributor workflow
runs when convenient?

@jingyli and @DanielFalconGuedes, I would greatly appreciate your review when
time permits.

Thank you very much for your time and help.

vishkaty (approved):
I filed #571, so I wanted to give this a real verification pass rather than just a thumbs-up. I fetched the branch (`119ffd4`), merged it onto current `main` (`5a5d9e1`) locally, and checked it mechanically:

- **Composes cleanly with current main**, including the #657 addition to `ap2-mandates.md` — it lands in the Canonicalization subsection, well below the section-intro and pseudocode hunks this PR edits, so nothing overlaps. The merged section intro (table → scoping note → binding requirement → example → algorithm note) reads coherently.
- **Example validation passes.** `scripts/validate_examples.py --schema-base source/schemas --docs docs` on the merged tree: 303 passed, 0 failed — versus 302 on main alone, so the new P-521 example is genuinely exercised and admitted by the open profile JWK vocabulary, as the PR describes. (This covers the validation that couldn't run locally without `ucp-schema`.)
- **The example key is real.** The `x`/`y` coordinates decode to 66-byte values forming a valid point on P-521 — worth stating because invalid example points are a classic spec pitfall that copy-paste propagates.
- **Tree-wide consistency.** I grepped the merged docs+schemas for `ES256`/`ES384`/`ES512`/`EdDSA`/`P-521`. Every site is consistent with the AP2-scoped layering: the HTTP Message Signatures table is unchanged with the new explicit non-extension sentence, `ap2_mandate.json`'s description already listed all three algorithms, the pseudocode's new `key_supports_jws_algorithm` assert matches the prose MUST, and signatures.md's "excludes Ed25519" cross-reference matches the ECDSA-only AP2 table.

The alg↔curve binding (RFC 8725 §3.1) plus the mandatory JWK `alg` is the right verifier guardrail for a multi-algorithm key set. And as a data point from the #571 thread: the AP2 reference stack signs and verifies ES256 only today, so this clarification breaks nothing deployed while making an ES512-signed mandate verifiable by implementations built from the spec.

The one residual — the well-known EC vocabulary in signatures.md's key-format section and profile.json still stops at P-384, so the P-521↔ES512 pairing documented here isn't yet machine-enforced — is exactly the scope split already discussed above, tracked in #625, which remains compatible with this PR (the example JWK here validates against that rule too). Non-blocking for this documentation change.

This resolves #571 as filed: the divergence is now explicitly documented on both sides of the boundary. LGTM — thanks for taking the AP2-scoped interpretation to a clean landing.

ShuoRen-TT:
Hi @mmohades and @yanheChen — a quick follow-up on #618 when convenient.

As of 2026-08-20, this PR is still mergeable, has two approvals, and all review conversations are resolved. I am not seeing any remaining technical feedback to address on my side.

If there is still a blocker, could you please point me to the specific change needed? Otherwise, I would appreciate a review or approval when time permits.

If maintainer action is still needed for the remaining governance or workflow gating, I would also appreciate help routing or clearing that final step. Thank you.

ShuoRen-TT:
Hi @damaz91, thank you for helping route reviews on this PR.

The merge conflicts with the latest main have now been resolved, and the PR is mergeable again. The change checks and CLA pass; the remaining blocker is the required maintainer approval. @mmohades and @yanheChen have been requested for some time, but I understand they may be busy.

If they are unavailable, could you please help assign or route this to another eligible maintainer for review when convenient? I would be happy to address any feedback promptly. Thank you.

docs/specification/payment/extensions/ap2-mandates.md
```diff
@@ -330,8 +367,9 @@ verify_merchant_authorization(checkout, merchant_profile):
     // Reconstruct signing input (header + payload)
     signing_input = encoded_header + "." + base64url_encode(canonical_bytes)
 
-    // Get business's public key and verify
+    // Get business's public key, bind the algorithm to its curve, and verify
     public_key = get_key_by_kid(merchant_profile.keys, header.kid)
+    assert key_supports_jws_algorithm(public_key, header.alg)
```
karangoel16: verify_alg_key_binding(public_key, header.alg) or assert_alg_bound_to_key  should we do rename in here ?

docs/specification/payment/extensions/ap2-mandates.md
```diff
@@ -129,18 +129,55 @@ If a public key cannot be resolved, or if the signature is invalid, the business
...
 * **Canonicalization:** JCS ([RFC 8785](https://datatracker.ietf.org/doc/html/rfc8785))
 * **Key Format:** JWK ([RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517))
 * **Key Discovery:** `keys[]` in `/.well-known/ucp` (see
   [Key Discovery](../../overview/index.md#key-discovery))
 
 See [Message Signatures](../../signatures.md) for key format and rotation.
 
+When a verifier selects a key, it **MUST** ensure that the protected JWS
+`alg` is compatible with the key's curve according to the table above. This
+binding follows [RFC 8725 Section 3.1](https://datatracker.ietf.org/doc/html/rfc8725#section-3.1).
```
karangoel16: any deployed keys this would break? since we are making ```alg``` required ? and in n the JWK standard (RFC 7517) it's optional

docs/specification/payment/extensions/ap2-mandates.md
```diff
@@ -129,18 +129,55 @@ If a public key cannot be resolved, or if the signature is invalid, the business
...
+AP2 defines the signed object and its accepted JWS algorithms separately:
+
+| JWS `alg` | JWK `kty` / `crv` | Hash |
+| :-------- | :---------------- | :--- |
+| `ES256` | `EC` / `P-256` | SHA-256 |
+| `ES384` | `EC` / `P-384` | SHA-384 |
+| `ES512` | `EC` / `P-521` | SHA-512 |
+
+This algorithm set applies only to AP2 JWS objects. In particular, `ES512`
+support here does not add P-521 to the UCP HTTP Message Signatures algorithm
```
karangoel16: Since the P-521 key lives in the shared keys[] array, what should a verifier do if it's referenced by an HTTP Message Signature — reject, ignore? Worth one sentence.

docs/specification/payment/extensions/ap2-mandates.md
```diff
@@ -129,18 +129,55 @@ If a public key cannot be resolved, or if the signature is invalid, the business
...
+mechanisms defined in the
+[Message Signatures](../../signatures.md) specification.
+AP2 defines the signed object and its accepted JWS algorithms separately:
+
+| JWS `alg` | JWK `kty` / `crv` | Hash |
+| :-------- | :---------------- | :--- |
+| `ES256` | `EC` / `P-256` | SHA-256 |
+| `ES384` | `EC` / `P-384` | SHA-384 |
+| `ES512` | `EC` / `P-521` | SHA-512 |
+
```
karangoel16, line 143: The table lists ES256/ES384/ES512 as AP2's accepted algorithms, but it's unclear what this requires of a verifier. 

Must a conformant AP2 verifier support all three, or is this just the set a signer may choose from? Since the universal baseline stays ES256, an ES256-only verifier's conformance status is ambiguous — and it would reject valid ES512-signed mandates. 

Worth one sentence clarifying whether the table is a verifier requirement or just the accepted set.

karangoel16 (commented):
@ShuoRen-TT PR looks good overall, but I have few questions on the same, passing back to your queue with questions

docs/specification/payment/extensions/ap2-mandates.md
ShuoRen-TT: Thank you for the suggestion! I have renamed the helper to
`assert_alg_bound_to_key(public_key, header.alg)` and included its definition,
so it explicitly checks the EC key type, curve-to-algorithm binding, and the
JWK `alg` when present. A failed assertion terminates verification.

Both the platform verification example and the business's embedded-checkout
verification now use the same helper, with the accepted-algorithm and local
verifier-policy checks before cryptographic verification. This should make the
intended safeguard clearer and keep the two examples consistent.

docs/specification/payment/extensions/ap2-mandates.md
ShuoRen-TT: Thank you for raising this compatibility question. I do not have a complete
inventory of deployed AP2 signing keys, so I cannot substantiate a claim that
requiring `alg` would affect no existing deployments.

You are right that RFC 7517 makes the JWK member optional, and the current UCP
profile schema also permits its omission. An application profile can require
it, but doing so here would introduce a stricter acceptance rule, not just
clarify the existing binding.

The original `MUST` followed @GarethCOliver's helpful recommendation to make
the key's intended algorithm explicit. I have pushed a compatibility-oriented
alternative for review:

- Businesses **SHOULD** publish `alg` for Checkout signing keys.
- If present, it **MUST** match the protected JWS header's `alg`.
- If absent, the verifier still enforces the EC `kty`/`crv` pairing and its
  independently configured algorithm policy. Absence alone does not invalidate
  the signature; the protected JWS header's `alg` remains required.

This preserves strict algorithm/key binding without requiring existing keys
to add the member. @GarethCOliver and @karangoel16, would this balance work for
you? I am leaving the discussion open for confirmation. Thank you both for
helping make the security and compatibility expectations precise.

docs/specification/payment/extensions/ap2-mandates.md
ShuoRen-TT: Thank you, agreed that the shared `keys[]` array makes this worth spelling out.
I have added a sentence linking to the existing HTTP signature rules: if the
HTTP verifier does not support the referenced P-521 key, that candidate
signature fails with `algorithm_unsupported`, without invalidating the entire
key set. In a multi-signature request, verification proceeds to other
candidates; the unsupported signature does not authenticate the request.

This does not prohibit publishing the AP2 key alongside HTTP signing keys,
nor does it add P-521 support to the HTTP implementation requirements.

docs/specification/payment/extensions/ap2-mandates.md
ShuoRen-TT: Thank you for pointing out this distinction. The intended scope is the
permitted algorithm/key pairs for the merchant-signed Checkout
(`ap2.merchant_authorization`), not a new requirement for every AP2 verifier to
implement all three algorithms. I have made that scope explicit and added
fail-closed verification against an independently configured set of supported
and permitted algorithms.

After rechecking AP2 v0.2, I have not found an explicit minimum verifier
implementation set that would justify either mandating all three here or
transferring the HTTP ES256 baseline into AP2. The revised text therefore does
not claim that an ES256-only implementation is necessarily fully AP2-conformant.

Could you and the maintainers please confirm that documenting the permitted
Checkout set, without establishing a new mandatory-to-implement set in this
PR, is the right boundary? If an explicit AP2-wide conformance baseline is
needed, I suggest treating that as a separate protocol decision rather than
implying it through this table. I am leaving this thread open for that
confirmation. Thank you again for catching the ambiguity.

ShuoRen-TT:
Thank you @karangoel16 for the careful review, and @GarethCOliver for the earlier
security guidance. I have pushed `4cfa37f2` and replied to each of the four
threads.

The revision keeps the change scoped to Checkout JWS verification, makes the
binding helper explicit in both verification paths, and clarifies unsupported
HTTP-key handling. The proposed JWK `alg` compatibility wording and the
permitted-algorithms versus implementation-baseline distinction remain open
for confirmation; I have not resolved the review threads.

Local validation passed: all pre-commit hooks, 344 documentation examples
(50 existing annotated skips), 52 validator tests, and a strict specification
build. Focused local binding/policy and cryptographic smoke checks also passed;
details and the separate dependency-audit caveat are in the updated PR body.

@archrao, I would appreciate your maintainer review of the revised scope and
compatibility wording when convenient. Thank you all for your time and help.

karangoel16 (approved):

PR #620, opened 2026-07-21
fix(shopping): add missing error declarations to MCP methods for REST parity

prasad-stripe:
## Summary

The MCP service definition is missing error declarations that REST already surfaces. Update methods (update_checkout, update_cart) lack idempotency_conflict (REST returns HTTP 409 on PUT). Read methods (get_checkout, get_cart, get_order) lack invalid_params for malformed meta.

This PR adds:
- idempotency_conflict to update_checkout and update_cart
- invalid_params to get_checkout, get_cart, and get_order

## Why this matters

MCP error declarations are the contract agent frameworks use to understand which failures are possible on each method. When declarations are incomplete, agents treat known error conditions as unexpected crashes rather than actionable signals they can recover from (fix params, retry with new idempotency key). REST already surfaces these errors; the MCP spec should declare the same contract.

## Test plan

- [x] Valid OpenRPC 1.3.2 JSON after changes
- [x] Error refs point to existing components/errors entries (invalid_params and idempotency_conflict both defined)
- [ ] Declaration-only change; no behavioral impact

### DIFF
--- a/source/services/shopping/mcp.openrpc.json
+++ b/source/services/shopping/mcp.openrpc.json
@@ -68,6 +68,28 @@
           }
         }
       }
+    },
+    "errors": {
+      "authentication_failed": {
+        "code": -32002,
+        "message": "Authentication failed",
+        "data": { "$ref": "../../schemas/common/types/error_response.json" }
+      },
+      "invalid_params": {
+        "code": -32602,
+        "message": "Invalid params",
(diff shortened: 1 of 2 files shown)


carolinerg1:
Hi @prasad-stripe, thank you for submitting this PR. It looks like the Lint Code Base check is currently failing - could you please take a look.

PR #622, opened 2026-07-21
feat(shopping): add rate_limit_exceeded error for agent retry signaling

prasad-stripe:
## Summary

Add rate_limit_exceeded error (code -32004) to the MCP Shopping service and declare it on all 13 methods. Any method can be rate-limited; agents need a structured signal to distinguish "slow down" from "your request is broken."

## Why this matters

UCP is agent-native. Agents operating in automated loops will hit rate limits from merchant implementations. Without a declared error code, agents cannot distinguish retriable throttling from permanent failures, leading to either blind retries (hammering the merchant) or premature abandonment (degraded UX).

Code -32004 follows MCP conventions: implementation-defined range (-32000 to -32019) for server-specific conditions that clients must act on distinctly (ref: MCP spec schema.ts allocates -32020+ for protocol-defined codes, leaving -32000 to -32019 for implementations like UCP).

The data payload uses existing error_response.json, consistent with all other MCP errors. The content field can carry retry-after hints in human-readable form; a structured retry_after field can be proposed separately if the TC wants machine-readable backoff.

## Test plan

- [x] Valid OpenRPC 1.3.2 JSON
- [x] Code -32004 does not conflict with existing codes (-32002, -32003)
- [x] All 13 methods reference the new error
- [ ] Declaration-only; no behavioral change

### DIFF
--- a/source/services/shopping/mcp.openrpc.json
+++ b/source/services/shopping/mcp.openrpc.json
@@ -68,6 +68,28 @@
           }
         }
       }
+    },
+    "errors": {
+      "authentication_failed": {
+        "code": -32002,
+        "message": "Authentication failed",
+        "data": { "$ref": "../../schemas/common/types/error_response.json" }
+      },
+      "invalid_params": {
+        "code": -32602,
+        "message": "Invalid params",
+        "data": { "$ref": "../../schemas/common/types/error_response.json" }
+      },
+      "not_found": {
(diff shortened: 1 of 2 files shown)


carolinerg1:
Hi @prasad-stripe, thank you for submitting this PR. It looks like the Lint Code Base check is currently failing - could you please take a look.

PR #626, opened 2026-07-22
split instrument availability into `constraints` and `options` axes

igrigorik:
Current #424 shape put everything an instrument declares — presence, per-field
   constraints, accepted card brands, and credential families — into one
   `constraints` bag. Prototyping a resolver against that shape showed the cost:
   to enforce field requirements a consumer must first classify every key by
   resolving its concrete schema (is this a field constraint? a value menu? a
   typed family?). That classification is the whole problem — it needs the target
   schema, defeats "compile from data alone," and produced real bugs in the
   prototype (per-key allOf composition, and a `required` name collision).

   Partition the bag into two sibling axes:

   - `constraints` — field requirements over the instrument's OWN fields (an
     Object Constraint: `required` plus one direct key per constrained field).
     Compiles to a JSON Schema overlay (`required` + `properties`) a standard
     validator runs.
   - `options` — accepted value menus (`brands`) and typed families
     (`credentials`), resolved by data lookup and never part of the overlay.

   With the menus and families moved to `options`, `constraints` holds nothing but
   `required` and field constraints — and `object_constraint` now enforces exactly
   that: every non-`required` key is a nested Object Constraint or a Value
   Constraint (recursing via `$ref "#"`). Compiling the overlay is therefore a
   pure data walk — no schema resolution, no classification. That is the win: the
   resolver is clean by construction, not by careful bookkeeping. Proven in the
   ucp-schema resolver prototype (~20-line compile; the compiled overlay enforces
   presence and allowed values against real instances).

   Trade-offs:

   - `constraints` is deliberately not a literal JSON Schema: fields are direct
     keys, so a consumer compiles it (take `required`, lift field keys under
     `properties`). A `properties`-keyed sub-map would make it literal but collide
     the JSON Schema `properties` keyword with the data key when a schema narrows a
     nested field. The compile is trivial, and a resolver runs regardless.
   - `required` is reserved inside a constraint object, so a field literally named
     `required` can be listed as present but cannot carry a nested constraint.
     Rare, and documented.
   - The shape still does not enforce name resolution (does the named field exist
     on the target?). Unchanged from before; a lint pass can add it later — the
     partition makes it tractable, since every non-`required` key is unambiguously
     a field name.


source/schemas/shopping/types/card_payment_instrument.json
```diff
@@ -6,31 +6,30 @@
...
                   "items": {
                     "if": {
                       "type": "object",
                       "required": ["type"],
-                      "properties": {
-                        "type": { "const": "token" }
-                      }
+                      "properties": { "type": { "const": "token" } }
                     },
                     "then": { "$ref": "token_credential.json#/$defs/constraint" }
```
jamesandersen, line 34: Not introduced here ... but unless I'm missing something, I don't see a `token_credential.json#/$defs/constraint`?

docs/documentation/schema-authoring.md
```diff
@@ -422,58 +422,59 @@ typos in core metadata like the `ucp` block).
...
+| [`ValueConstraint`](site:{{ ucp_version }}/schemas/shopping/types/value_constraint.json) | A property's value | `enum`, `const` |
+| [`TypeConstraint`](site:{{ ucp_version }}/schemas/shopping/types/type_constraint.json) | A typed branch | `type`, optional `constraints` |
+
+Inside an Object Constraint, `required` is reserved (the presence list); every
+other key names a target field and carries that field's nested constraint — an
+Object Constraint (recurse) or a Value Constraint. The two are distinguishable
+**from the data alone** — `enum`/`const` is a Value Constraint, otherwise it's an
+Object Constraint — so a consumer compiles the overlay without resolving the
+concrete schema. Field constraints are **open by default**: they live in the wire
+data and validate against the open Object Constraint, so most handlers only narrow
```
jamesandersen: nit: inclusion in schema-authoring.md implies use across UCP broadly so "handlers" could be slightly confusing here.

docs/documentation/schema-authoring.md
```diff
@@ -422,58 +422,59 @@ typos in core metadata like the `ucp` block).
...
-| Primitive | Target | Built-in vocabulary |
-| :-------- | :----- | :------------------ |
-| [`ObjectConstraint`](site:{{ ucp_version }}/schemas/shopping/types/object_constraint.json) | Object | `required` |
-| [`ValueConstraint`](site:{{ ucp_version }}/schemas/shopping/types/value_constraint.json) | Property value | `enum`, `const` |
-| [`TypeConstraint`](site:{{ ucp_version }}/schemas/shopping/types/type_constraint.json) | Typed branch | `type`, optional `constraints` |
-
-A concrete Object Constraint **MUST** extend `object_constraint.json` and define
-every supported key. The key's schema determines its meaning:
+- **`constraints`** — field requirements over the target's OWN fields, as an
+  Object Constraint: `required` (presence) plus one key per constrained field.
```
jamesandersen, line 430: Should we clarify whether the `required` should include the keys from `options`?  They are presumably required (with more than just presence)

igrigorik:
@jamesandersen good flags, addressed in https://github.com/Universal-Commerce-Protocol/ucp/pull/626/changes/387eb336acff653215457f2f529a9a26683a6e87:

- **`token_credential.json#/$defs/constraint`**: it does resolve; it's defined at `$defs/constraint` which is pre-existing code in proto branch -- this PR is not a replacement, it's a set of updates to apply on top.
- **"handlers"**: agreed, this doc is cross-domain — generalized to "most schemas", and marked the card example explicitly as "a payment handler".
- **`required` vs `options` keys**: no,`required` names fields the buyer *submits*; `options` attributes like `brands` name values the seller *accepts*. A set like a card's `brand` is derived from the PAN post-submission, so there's no submittable field to overlay, which is exactly why it's `options`, not `constraints`.

jamesandersen (approved):
Thanks for the quick clarifications @igrigorik 

 I think this culmination of the various explorations and feedback from @raginpirate and @alexpark20 is a good resolution to the problem that prompted the original  PR (#288 ) and nicely scales to other use cases.  I appreciate everyone's help here.

source/schemas/shopping/types/available_payment_instrument.json
```diff
@@ -2,27 +2,28 @@
...
               "uniqueItems": true,
-              "description": "Credential specific constraints accepted for the payment instrument branch selected by `type`. Concrete instrument schemas can narrow known credential entries while preserving handler-extended credential entries."
+              "description": "Accepted credential families — each a Typed Constraint keyed by `type` with per-branch `constraints`. Applied by data lookup on the submitted credential's `type`; unknown types are handler/extension branches."
             }
           },
-          "description": "Constraints on this instrument type. Base payment instrument constraints can require and describe `billing_address` and `credentials`; concrete instrument schemas SHOULD add instrument-specific constraint keys while remaining open to extensions."
+          "additionalProperties": {
+            "type": "array",
+            "items": { "type": "string" },
+            "uniqueItems": true
```
raginpirate, line 24: This seems problematic to close the supported additional properties set, right?
I'm imagining other data types are acceptable keys to continue extending onto available instrument.

source/schemas/shopping/types/available_payment_instrument.json
```diff
@@ -2,27 +2,28 @@
...
       "type": "object",
       "properties": {
-        "constraints": {
+        "options": {
           "type": "object",
           "properties": {
-            "billing_address": {
-              "$ref": "object_constraint.json",
-              "description": "Local constraint on the instrument's `billing_address` field."
-            },
```
raginpirate, line 17: Without this, we'll fail to reject a constraint over field "billing_addr" when it should be "billing_address", right?

raginpirate (commented):
I appreciate the direction you’ve steered this in, Ilya. I’ll be non-blocking on this moving forward.

I do want us to name the primary trade-off: a more fragmented API shape in exchange for simpler constraint-to-overlay compilation. In some cases, the requiredness and accepted shape of the same target are split across mirrored paths (`constraints.required: ["credential"]` alongside `options.credentials`) which weakens locality.

That may be a fair trade-off. However, resolving an instrument’s type already requires selecting its concrete branch from the negotiated handler schema, so I question how much end-to-end complexity this saves. Put another way: this branch removes per-key role classification from the overlay compiler; it does not remove handler-schema resolution, branch selection, target resolution, or end-to-end validation complexity.

source/schemas/shopping/types/available_payment_instrument.json
jamesandersen: +1. The docs describe `options` as a uniform map — *"a scalar list (e.g. `brands`) or a typed family (e.g. `credentials`)"* — but the schema only makes the scalar case extensible: `additionalProperties` is `{array of strings}`, so `credentials` validates only because it's hardcoded as a named property. Any other typed family under a new key would fail.

I'd suggest widening `additionalProperties` to admit either shape (distinguished by item type, so still a pure data-shape check):

```json
"additionalProperties": {
  "anyOf": [
    { "type": "array", "items": { "type": "string" }, "uniqueItems": true },
    { "type": "array", "items": { "$ref": "type_constraint.json" }, "uniqueItems": true }
  ]
}
```

With this, `credentials` no longer needs to be a named property either — it just becomes the canonical typed-family value, and `options` is genuinely uniform (shape decides dispatch vs. membership). The tradeoff is losing `credentials` as signposting for a known family, so worth keeping it named if we'd rather advertise it explicitly.

Otherwise, if `options` is meant to be scalar-only with `credentials` a deliberate one-off, let's tighten the docs to say so. Either way, schema and prose should agree — which did you intend, @igrigorik?

source/schemas/shopping/types/available_payment_instrument.json
jamesandersen: If I follow correctly, I think this is safe to remove — the field-name check belongs against the target schema (`payment_instrument`, where `billing_address` actually lives), not against `available_payment_instrument`. That's the lint @igrigorik defers in Trade-offs ("does the named field exist on the target?").

PR #627, opened 2026-07-23
fix: authority binding accepts exact name-host match

igrigorik:
Authority Binding currently requires (specced in #530) an entity's `name` to begin with the reversed `schema` host **plus at least one more label**. That rejects a legitimate, common shape: an entity whose name *is* a controlled domain — e.g. a payment handler named `com.example.pay` served from `pay.example.com`. 

This PR allows the **exact-match** case (`name == authority_prefix`) alongside the existing prefixed case, and clarifies that binding proves **provenance only** — the `{reverse-domain}.{service}.{capability}` structure is a separate Naming Convention, not a binding requirement.

- **Before:** valid iff `name` starts with `authority_prefix + "."` (non-empty remainder required).
- **After:** valid iff `name == authority_prefix` (**exact match**, new) **or** `name` starts with `authority_prefix + "."`

---

## Checklist

### DIFF
--- a/docs/specification/overview.md
+++ b/docs/specification/overview.md
@@ -297,8 +297,9 @@ from any host (e.g. a docs subdomain or third-party docs host). Only the
 
 The authority is derived **from the `schema` URL host** — which names the
 owning domain directly, with no ambiguity about where the domain ends — and
-validated as a label prefix of the entity's name. For the `schema` URL of an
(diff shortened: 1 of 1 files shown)


drewolson-google (approved):

raginpirate (approved):
🚀

PR #628, opened 2026-07-23
feat: typed error responses and checkout conditional writes

sakinaroufid:
# Description

**TL;DR:** UCP is built on agents consuming machine-readable contracts, and the discovery profile points them straight at `rest.openapi.json`. But that file only modeled success responses, so every SDK, server stub and conformance test generated from it was blind to the error contract the docs promise, and error handling is exactly where autonomous agents need structure most. This PR declares the full error contract in the OpenAPI/OpenRPC definitions (`4XX`/`5XX` on all 13 operations, `409` on the state-changing ones, and `search_catalog`/`lookup_catalog` finally able to return the standard error envelope like every other op), so generated clients get typed errors and conformance can test failure paths. It also adds `ETag`/`If-Match` conditional writes to checkout so two writers on one session, say an agent on the API and a buyer in a trusted UI, can't silently overwrite each other's changes. No schema changes anywhere.

--------------------------------------------------------------

The REST service definition only declares the success response for each operation. The docs list a full status code contract (400 through 503, including 409 on idempotency-key reuse) with a JSON error body, but none of that exists in `rest.openapi.json`, so anything generated from it (clients, server stubs, contract tests) has no idea errors exist. I hit this while generating a client: the documented 409 comes back as an unmodeled exception instead of something you can parse the error envelope out of.

The first part fixes that. It adds reusable error response components carrying the standard envelope and wires them into every operation: `4XX`/`5XX` on all 13 operations, an explicit `409` on the seven state-changing ones, and failed-delivery responses on the order event webhook. While in there I noticed `search_catalog` and `lookup_catalog` are the only two operations whose result has no `error_response` branch, in both the REST and MCP definitions, so they can't return the standard envelope at all. Every other operation can, including `get_product` in the same capability. Both results now use the same `oneOf` pattern as everything else, and the behavior is documented on the catalog pages.

The second part adds optimistic concurrency to checkout. Update Checkout is a full replacement, and a session can have more than one writer: an agent on the API plus a buyer parked in a trusted UI via `continue_url`, or the business repricing things server side. Two writers that read the same state and write different replacements lose one of the updates silently. Idempotency-Key doesn't help, it only dedupes identical retries of one request.

The fix is plain RFC 9110 conditional requests, no new schema fields. Businesses return a strong `ETag` on checkout responses. Platforms echo it in `If-Match` on Update, Complete and Cancel. On a mismatch the business applies nothing and returns `412 Precondition Failed` with the standard envelope (`precondition_failed`, severity `recoverable`); the platform re-reads, reconciles and retries with the fresh tag. `428 Precondition Required` is there for businesses that want to force conditional writes. Support is discovered from responses: if a platform never saw a tag it just writes unconditionally, so existing implementations keep working untouched. Only REST gets a binding for now, since MCP/A2A/Embedded have no natural place to carry the tag.

The conditional writes add normative behavior, so if that part needs to go through an enhancement proposal first I'm fine splitting it out into its own PR. The error declarations are just making the machine-readable contract match what the docs already say.

## Category (Required)

Core Protocol covers the conditional writes; the error declarations are Capability plus Documentation.

## Related Issues

None.

## Checklist

The unchecked ones: the new conditional update examples (applied and rejected) are exercised by the docs example corpus rather than separate unit tests, and nothing under `source/schemas/` changes. That's the point of putting the tag in HTTP fields, so there's also nothing to regenerate on the SDK side.

```
$ ucp-schema lint source/
✓ 96 files checked, all passed

$ python3 scripts/validate_examples.py --schema-base source/schemas/
279 passed, 0 failed, 0 errors, 47 skipped

$ python3 scripts/test_validate_examples.py
48 passed, 0 failed
```

`mkdocs build --strict` passes in both site modes and the link checker comes back clean. The doc generation macros only read success responses, so rendered pages don't change apart from the added notes and the two new catalog entity sections.

### DIFF
--- a/docs/specification/catalog/lookup.md
+++ b/docs/specification/catalog/lookup.md
@@ -107,6 +107,11 @@ from the response.
 
 {{ extension_schema_fields('catalog_lookup.json#/$defs/lookup_response', 'catalog') }}
 
+Like every other operation, `lookup_catalog` MAY return the standard error
+envelope (`ucp.status: "error"` with `messages`) in place of a lookup
+response payload when the request cannot be served. See
+[Error Handling](../overview.md#error-handling).
+
 ---
 
 ## Get Product (`get_product`)
--- a/docs/specification/catalog/mcp.md
+++ b/docs/specification/catalog/mcp.md
@@ -117,6 +117,11 @@ Maps to the [Catalog Search](search.md) capability.
   'catalog_search.json#/$defs/search_response', 'catalog/mcp'
 ) }}
 
+When the request cannot be served, the tool returns a successful JSON-RPC
+result carrying the standard error envelope (`ucp.status: "error"` with
+`messages`) instead of a search response payload, mirroring
+[Product Not Found](#product-not-found) for `get_product`.
+
 #### Search Example
 
 === "Request"
@@ -276,6 +281,11 @@ The `catalog.ids` parameter accepts an array of identifiers and optional context
 {{ extension_schema_fields(
   'catalog_lookup.json#/$defs/lookup_response', 'catalog/mcp'
 ) }}
+
+When the request cannot be served, the tool returns a successful JSON-RPC
+result carrying the standard error envelope (`ucp.status: "error"` with
+`messages`) instead of a lookup response payload, mirroring
+[Product Not Found](#product-not-found) for `get_product`.
 
 #### Lookup Example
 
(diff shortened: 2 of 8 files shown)


sakinaroufid:
@carolinerg1 Should I split this PR to facilitate review? Happy to submit the ETag/If-Match conditional checkout writes as a separate PR, or push the error response components, 4XX/5XX wire-ins, and the search_catalog/lookup_catalog envelope fixes (these touch no core procol behavior). Thanks!

PR #341, opened 2026-04-08
style!: Clean up transition & deprecated schema in preparation of a version cut

jingyli:
# Description

Prior to the new UCP version cut, we need to clean up schema fields that are transitioning to a different state & drop any deprecated fields.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] **Breaking change** (fix or feature that would cause existing
      functionality to not work as expected, **including removal of schema files
      or fields**)
- [ ] Documentation update

---

### Is this a Breaking Change or Removal?

If you checked "Breaking change" above, or if you are removing **any** schema
files or fields:

- [x] **I have added `!` to my PR title** (e.g., `feat!: remove field`).
- [x] **I have added justification below.**

```text
## Breaking Changes / Removal Justification

The breaking schema changes are already forecasted via other PRs. This PR is simply to clean up annotations prior to the formal version cut.
```

---

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #332, opened 2026-04-02
docs: omit deprecated checkout id from playground payload

gaureshpai:
# Description

While reviewing the interactive playground script in `docs/specification/playground.md`, I noticed that the `id` property was still being passed inside the `patch` object for update requests. Since this field was recently marked as `deprecated_required_to_omit` (as it is provided via the URL path), the playground was inadvertently generating non-compliant payloads. This PR fixes the playground payload creation by properly initializing an empty patch object.

## Type of change

Please delete options that are not relevant.

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] Documentation update

---

## Checklist

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules

---

### Pull Request Title

docs: omit deprecated checkout id from playground payload

PR #633, opened 2026-07-24
fix: Transition cart.id to be omitted in update requests instead of a required field

jingyli:
# Description

Transition `id` field from `required` to `omit` in cart capability for `update` requests.

The current state presents a redundancy given the cart being updated is already identified at the transport level (e.g., via the resource path in REST `/carts/{cart_id}` or via tool arguments in MCP) - this is also the pattern used in checkout capability.

Also augmented logic in `main.py` to properly process schema with `transition` annotations so the rendering of the field requirement in `reference.md` clarifies the transition (and all `ucp_request`/`ucp_response` annotations properly). See attached screenshot for an example of how it will look like post this change.

## Category (Required)

## Related Issues

N/A

## Checklist

cart-fix-render

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -173,6 +173,9 @@ yarn-error.log
 *.code-workspace
 .vscode/
 
+# Local agent config
+/.agents/
+
 # Site output from scripts/build_local.sh
 /local_preview/
 
--- a/docs/specification/cart-rest.md
+++ b/docs/specification/cart-rest.md
@@ -313,7 +313,7 @@ All REST endpoints **MUST** be served over HTTPS with minimum TLS version 1.3.
     Content-Type: application/json
 
     {
-      "id": "cart_abc123",
+      "id": "cart_abc123",  // deprecated: id is provided in URL path
       "line_items": [
         {
           "item": {
--- a/docs/specification/cart-rest.md
+++ b/docs/specification/cart-rest.md
@@ -313,7 +313,6 @@ All REST endpoints **MUST** be served over HTTPS with minimum TLS version 1.3.
     Content-Type: application/json
 
     {
-      "id": "cart_abc123",
       "line_items": [
         {
           "item": {
--- a/docs/specification/cart-rest.md
+++ b/docs/specification/cart-rest.md
@@ -313,6 +313,7 @@ All REST endpoints **MUST** be served over HTTPS with minimum TLS version 1.3.
     Content-Type: application/json
 
     {
+      "id": "cart_abc123",
       "line_items": [
         {
           "item": {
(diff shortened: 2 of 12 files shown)


raginpirate (changes requested):
Good spot @jingyli, I am aligned with fixing this id advertisement and the `main.py` bug. 

However, the immediate removal of id from all scaffolds looks wrong, as `from` is supposed to be the authoritative definition in the schema until a future cutoff is hit. The established sequence is #145 announce (keep the field, comment it) → #341 flip at version cut → #332 drop from docs. Unless we come up with a safe way to condense versioning, I think this blocks the PR from shipping as-is.

igrigorik (changes requested):
Agree with @raginpirate, the end state is right but the sequence is wrong. The examples must resolves against the *production* schema shape, not future state. Let's break this apart and follow same sequence we landed for checkout: land the transition first, then stage transition removal and concrete 'omit' with updated examples.

jingyli:
@raginpirate @igrigorik Good prompt and callout.

Originally I was thinking whether the established sequence had areas for improvement (i.e. changing fields in docs #332 probably should have came before the version cut finalization #341 ) so [draft ucp.dev](https://ucp.dev/draft/specification/overview/) always have the right documentation accompanying on the schema. But it should be done in another PR - though I do think we should do the documentation/scaffolding update in the same PR where we finalize the transition so both takes effect into the version cut. Will take a stab at it when closer to the version cut.

For now, reverted back the changes on scaffolding/documentation updates and attached a screenshot on how the `main.py` rendering will look like on `reference.md`.

raginpirate (commented):
@jingyli did you want to keep deprecation markers in the docs for this interim period?

P.S. While writing this I realized we never removed the deprecated IDs with https://github.com/Universal-Commerce-Protocol/ucp/pull/341! Opened a doc fix: https://github.com/Universal-Commerce-Protocol/ucp/pull/646

I wonder if theres some simple tooling that can help us catch deprecated fields in examples and ensure they match the existing transition states.

igrigorik:
> I wonder if theres some simple tooling that can help us catch deprecated fields in examples and ensure they match the existing transition states.

our schema validator with --strict flag would/should have caught this.

jingyli:
@raginpirate PTAL at https://github.com/Universal-Commerce-Protocol/ucp/pull/633/changes/7869d546509e7e4b120abaecccce2eb76de55bca!

_Although just a side note from me personally_: I feel the need to add deprecation markers to be a little bit unnatural given json doesn't allow for inline comments structurally. I guess it's fine since these are just doc examples and it's better for readability (but still feel a little bit odd given these are technically json code blocks we have in our markdown files).

raginpirate (approved):
🚀 Agreed, these are example payloads rather than schemas, and there's no clean way in base JSON to show this inline, so I'm personally happy with the // convention here.

igrigorik (approved):

amithanda (approved):

PR #634, opened 2026-07-24
feat: add return policy extension (dev.ucp.shopping.policy.return)

amithanda:
Defines the body of the `dev.ucp.shopping.policy.return` policy type on the core
`policies[]` primitive (#572), adding pre-purchase, machine-readable return terms
across catalog, cart, checkout, and order. The primitive's container, `applies_to`
targeting, same-type precedence, and `messages[]` disclosure are unchanged; this
extension only structures the type-specific body.

## Context

- The `policies[]` primitive landed in #572, which **already lists
  `dev.ucp.shopping.policy.return` as a well-known type in `overview.md`** with an
  empty body, explicitly left for this PR to define. This is not a new namespace
  claim; it realizes an already-documented type.
- This is the sequencing agreed in #572: land the primitive first, then define the
  return type on top of it. It supersedes the standalone structure in #257 (a
  `return_policies` map keyed by id plus `return_policy_id` on line items), which
  the primitive's `applies_to` targeting replaced.
- The design record is the discussion across #257 and #572.

## What this adds

- `source/schemas/shopping/policy_return.json` - the extension. Composes onto
  `catalog.search`, `catalog.lookup`, `cart`, `checkout`, and `order`.
- `docs/specification/return.md` - the spec.
- `mkdocs.yml` - nav entry and llms.txt section.

Five type-specific fields, all optional (the base policy already guarantees `type`
and `description`):

- `final_sale` - non-returnable flag.
- `window` `{ days, anchor }` - the return window as a policy statement.
- `supported_resolutions[]` - what the buyer gets back (money, store credit, exchange).
- `methods[]` - return channels, each with its logistics `fee`.
- `restocking_fee` - a refund deduction independent of channel.

## Design decisions

1. **Extension, not core; type body gated by `if type == ... then <body>`.** Because
   the base policy sets `additionalProperties: true`, a platform that has not
   negotiated the type still renders the policy from `type` + `description`, and
   unknown policy types are never rejected. The container is untouched.
2. **Targeting via the primitive's `applies_to`, not item -> policy ids** (the #572
   decision). A response-wide default is one entry with no `applies_to`; overrides
   target specific items; precedence is inherited from the primitive.
3. **Five fields, by a deliberate inclusion bar.** A structured field ships only if
   it is needed to *compute or compare on* (not merely display) and reduces to a
   shape that is closed-and-final (boolean, integer) or open-by-design (string
   vocab), so future additions stay non-breaking. Everything else stays in
   `description`/`url` until its shape settles.
4. **Two costs modeled separately:** per-channel logistics `methods[].fee` vs. a
   policy-level `restocking_fee`. This resolves the limitation raised in #257
   (§1.4): "free returns by mail with a restocking fee" is inexpressible when a
   single per-method fee must be `free` xor `fixed_fee`.
5. **Explicit `window.anchor`, no default.** Anchors genuinely vary, so `window` 
   requires both `days` and `anchor`.
7. **Open-string vocabularies throughout** (`anchor`, `supported_resolutions`,
   `method.type`, `fee.type`), never `enum`, so new channels and outcomes are
   non-breaking additions.

## Potential future additions

Fields we may add to this type later, once their shape settles. Each would be an
optional, non-breaking addition; until then the information is carried in
`description`/`url`:

- **Resolved `window.deadline`** - the absolute cutoff, computed once the anchor
  event (e.g., delivery) fires; added with the Order-side resolution work.
- **Item-condition eligibility** - which conditions are returnable (new / used /
  damaged / refurbished). The strongest near-term candidate.
- **Reason-differentiated fees** - distinct fee/label for buyer remorse vs. item
  defect.
- **Seasonal / holiday window overrides** - dated changes to the return window.
- **Richer fee breakdown** - itemizing a logistics fee into components
  (e.g., shipping + handling).
- **Explicit unlimited-window signal** - to distinguish "explicitly unlimited"
  from "not stated structurally."

## Non-goals

Out of scope for this type; they belong to other extensions:

- **Post-purchase resolution** - RMA issuance, label generation, refund/exchange
  execution, and reason/condition capture at return time. A separate future
  extension; this type must not grow those fields.
- **Tiered, anchor-relative cancellation** - stepped partial-refund schedules for
  bookings/events. A sibling `dev.ucp.shopping.policy.cancellation` type over the
  same primitive.

## Category 

## Related Issues

Builds on #572 (policies primitive). Supersedes the standalone approach in #257.

## Checklist

- [x] I have followed the [Contributing Guide](https://github.com/Universal-Commerce-Protocol/.github/blob/main/CONTRIBUTING.md) (Conventional Commits; additive, no `!`).
- [x] I have updated the documentation (`return.md`, mkdocs nav, llms.txt).
- [x] I have added tests that prove the feature works (schema-validated `ucp:example` blocks in `return.md`, checked by the corpus example validator).

### DIFF
--- a/docs/specification/return.md
+++ b/docs/specification/return.md
@@ -1,0 +1,261 @@
+<!--
+   Copyright 2026 UCP Authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+
+# Return Policy Extension
+
+## Overview
+
+The Return Policy Extension defines the `dev.ucp.shopping.policy.return` policy
+type on the core [`policies[]`](overview.md#policies) primitive. It adds
+pre-purchase, machine-readable return terms to policies that carry this type, so
+platforms and agents can answer questions like "How long do I have to return
+this?", "Do I get my money back or store credit?", "Where and at what cost do I
+send it back?", and "Can I return this at all?" without leaving to read a policy
+page.
+
+**Key features:**
+
+- A return window as a policy statement (`window`), with an explicit anchor
+- What the buyer receives on an accepted return (`supported_resolutions`)
+- Permitted return channels and their logistics cost (`methods`)
+- A restocking fee that deducts from the refund independent of channel
+  (`restocking_fee`)
+- A hard `final_sale` flag for non-returnable items
+
+**Dependencies:**
+
+- The core `policies[]` primitive (see [Policies](overview.md#policies)).
+- One or more of the parent capabilities this type extends: Catalog Search,
+  Catalog Lookup, Cart, Checkout, or Order.
+
+This extension only structures the type-specific body. The `policies[]`
+container, `applies_to` targeting, same-type precedence, and buyer-facing
+disclosure through `messages[]` are defined once by the primitive and are
+**unchanged** here.
+
+## Discovery
+
+Businesses advertise return policy support in their profile. The type extends any
+surface that carries `policies[]`:
+
+<!-- ucp:example schema=profile def=business_schema extract=$.ucp.capabilities target=$.ucp.capabilities -->
+```json
+{
+  "ucp": {
+    "version": "{{ ucp_version }}",
+    "capabilities": {
+      "dev.ucp.shopping.policy.return": [
+        {
+          "version": "{{ ucp_version }}",
+          "extends": [
+            "dev.ucp.shopping.catalog.search",
+            "dev.ucp.shopping.catalog.lookup",
+            "dev.ucp.shopping.cart",
+            "dev.ucp.shopping.checkout",
+            "dev.ucp.shopping.order"
+          ],
+          "spec": "https://ucp.dev/{{ ucp_version }}/specification/return",
+          "schema": "https://ucp.dev/{{ ucp_version }}/schemas/shopping/policy_return.json"
+        }
+      ]
+    }
+  }
+}
+```
+
+A business MAY advertise support on a subset of these surfaces. A platform that
+has negotiated this type validates the return body and MAY reason over it (for
+example, computing a return deadline). A platform that has not negotiated it
+still renders the policy from the base `type` and `description`, because the
+return body is purely additive.
+
+## Schema
+
+When this type is active, a `policies[]` entry whose `type` is
+`dev.ucp.shopping.policy.return` MAY carry the following fields in addition to the
+base `type`, `description`, `applies_to`, and `url`. All fields are optional; a
+return policy with no structured fields is still presentable from its
+`description`.
+
+### Return Policy
+
+{{ extension_schema_fields('policy_return.json#/$defs/return_body', 'return') }}
+
+### Return Method
+
+{{ extension_schema_fields('policy_return.json#/$defs/return_method', 'return') }}
+
+## Return terms
+
+### Window and anchor
+
+`window.days` is the length of the return window as a **policy statement**, not a
+live countdown. A platform computes a concrete deadline by adding `days` to the
+event named by `window.anchor`.
+
+Businesses differ on whether the clock starts at delivery, purchase, shipment, or
+fulfillment, so when `window` is present a business MUST provide both `days` and
+`anchor`. There is no default anchor, and a platform MUST NOT assume one. Omit
+`window` entirely for an unlimited window, or when the window is not stated
+structurally, and convey it in `description`.
+
+The resolved, absolute cutoff is an Order-time fact: once the anchor event has
+occurred, a business republishes the policy on the Order as the source of truth
+(see [Order-time resolution](#order-time-resolution)). Pre-purchase, the window
+stays a duration plus an anchor.
+
+### Non-returnable items
+
+To signal that the covered items cannot be returned, a business MUST set
+`final_sale` to `true` and SHOULD omit the other return-body fields, since return
+terms do not apply. A business MUST NOT use an empty `supported_resolutions` array
+to signal non-returnability; `final_sale` is the single, unambiguous signal a
+platform acts on.
+
+When a business requires the buyer to be shown that an item is final sale, it
+emits a `messages[]` warning with `presentation: "disclosure"` and `code` equal
+to `dev.ucp.shopping.policy.return`, targeting the item. The disclosure pairs
+with the governing return policy at that node, as defined in
+[Presenting policies](overview.md#presenting-policies).
+
+### Two kinds of cost
+
+Returns carry two independent costs, modeled separately:
+
+- **Logistics cost** is the price of using a channel (return shipping or
+  handling). It lives on each `methods[].fee` because it varies by channel: an
+  in-store drop-off may be `free` while a mailed return is a `fixed_fee`.
+- **Restocking fee** is a deduction from the refund charged regardless of how the
+  item comes back. It lives once at the policy level as `restocking_fee`.
+
+Keeping them separate lets a business express, for example, "free returns by mail
+with a 15% restocking fee" - which a single per-method fee cannot represent. A
+percentage restocking fee uses `restocking_fee.percentage_bps`; a flat or
+precomputed amount uses `restocking_fee.amount`.
+
+When a return method's `fee.type` is `fixed_fee`, `amount` MUST be present, so the
+charge is never left uninterpretable.
+
+## Targeting and precedence
+
+Targeting and precedence are provided by the `policies[]` primitive and are not
+redefined here. In short: a policy with no `applies_to` is the response-wide
+default; a policy that targets specific items overrides it where they overlap,
+with the narrowest same-type target winning. See
+[Targeting](overview.md#targeting) and [Precedence](overview.md#precedence).
+
+For return policies this means a business states a single default return policy
+once, then adds targeted overrides only for the exceptions (a final-sale item, a
+category with a different window), rather than repeating a policy on every line.
+
+## Responsibilities
+
+Return policies are business-stated facts. They are response-only data
+(`ucp_request: omit` on cart, checkout, and order) that a platform never submits;
+there is no buyer selection and no request-side machinery. They carry no
+buyer-asserted claims and no PII, so they cross no new trust boundary beyond the
+base response.
+
+A business SHOULD populate the structured fields it can author accurately and MAY
+state anything it cannot yet structure (condition requirements, category
+nuances) in the policy `description` or link to it through `url`. A platform
+SHOULD surface `url` alongside the structured terms so the buyer can review the
+full policy.
+
+A policy's `description` and its structured fields are two representations of the
+same policy. The `description` MAY add terms not modeled here, but a business MUST
+NOT let it contradict a structured field. A platform that models this type MUST
+treat the structured fields as authoritative (for computing the return deadline,
+net refund, and returnability); the `description` is the human-readable summary it
+renders, and the fallback for a platform that does not model the type.
+
+## Order-time resolution
+
+Pre-purchase, the return window is a duration plus an anchor. The concrete
+deadline a buyer actually has can only be computed once the anchor event (for
+example, delivery) has occurred. That resolution is an Order-time concern: on the
+Order, a business republishes the same return policy carrying the resolved cutoff,
+so the platform reads it rather than tracking fulfillment events and recomputing.
+Modeling the resolved deadline field is deferred to that Order-side work.
+
+## Examples
+
+### Checkout with a default policy and a final-sale override
+
+A response-wide return policy states the standard terms once. An item-scoped
+policy marks one engraved line as final sale; because it targets that line
+directly, it governs it under same-type precedence, while every other line keeps
+the default. A paired disclosure compels the final-sale notice to be shown.
+
+<!-- ucp:example schema=shopping/checkout target=$.policies -->
+```json
+[
+  {
+    "type": "dev.ucp.shopping.policy.return",
+    "description": {
+      "plain": "30-day returns from delivery. Refund to original payment or exchange. Free in-store; $5 return shipping by mail. 15% restocking fee on opened items."
+    },
+    "window": { "days": 30, "anchor": "delivered" },
+    "supported_resolutions": ["original_payment_method", "exchange"],
+    "methods": [
+      { "type": "in_store", "fee": { "type": "free", "display_text": "Free in-store return" } },
+      { "type": "by_mail", "fee": { "type": "fixed_fee", "amount": 500, "display_text": "Return shipping" } }
+    ],
+    "restocking_fee": { "percentage_bps": 1500, "display_text": "15% restocking fee on opened items" },
+    "url": "https://example.com/returns"
+  },
+  {
+    "type": "dev.ucp.shopping.policy.return",
+    "description": { "plain": "This engraved item is final sale and cannot be returned." },
+    "applies_to": ["$.line_items[?@.id=='engraved_watch']"],
+    "final_sale": true,
+    "url": "https://example.com/returns#final-sale"
+  }
+]
+```
+
+The paired disclosure that compels display of the final-sale term:
+
+<!-- ucp:example schema=shopping/checkout target=$.messages -->
+```json
+[
+  {
+    "type": "warning",
+    "code": "dev.ucp.shopping.policy.return",
+    "path": "$.line_items[?@.id=='engraved_watch']",
+    "presentation": "disclosure",
+    "content": "This engraved item is final sale and cannot be returned."
+  }
+]
+```
+
+### Catalog with a store-credit-only return policy
+
+On a catalog surface, a return policy targets one product and signals that
+returns are store credit only, a material purchase-decision factor:
+
+<!-- ucp:example schema=shopping/catalog_search op=search direction=response target=$.policies -->
+```json
+[
+  {
+    "type": "dev.ucp.shopping.policy.return",
+    "description": { "plain": "Returns accepted within 30 days for store credit only." },
+    "applies_to": ["$.products[0]"],
+    "window": { "days": 30, "anchor": "purchased" },
+    "supported_resolutions": ["store_credit"]
+  }
+]
+```
--- a/docs/specification/return.md
+++ b/docs/specification/return.md
@@ -98,6 +98,14 @@ return policy with no structured fields is still presentable from its
 
 {{ extension_schema_fields('policy_return.json#/$defs/return_method', 'return') }}
 
+### Return Method Fee
+
+{{ extension_schema_fields('policy_return.json#/$defs/fee', 'return') }}
+
+### Fixed Fee
+
+{{ extension_schema_fields('policy_return.json#/$defs/fixed_fee', 'return') }}
+
 ## Return terms
 
 ### Window and anchor
--- a/source/schemas/shopping/policy_return.json
+++ b/source/schemas/shopping/policy_return.json
@@ -1,0 +1,260 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/policy_return.json",
+  "name": "dev.ucp.shopping.policy.return",
+  "title": "Return Policy Extension",
+  "description": "Defines the `dev.ucp.shopping.policy.return` policy type on the core `policies[]` primitive, adding return terms to catalog, cart, checkout, and order.",
+  "$defs": {
+    "return_body": {
+      "title": "Return Policy",
+      "description": "Return terms for a `dev.ucp.shopping.policy.return` policy. All fields are optional.",
+      "type": "object",
+      "properties": {
+        "final_sale": {
+          "type": "boolean",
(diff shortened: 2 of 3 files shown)


source/schemas/shopping/policy_return.json
```diff
@@ -0,0 +1,260 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/shopping/policy_return.json",
+  "name": "dev.ucp.shopping.policy.return",
+  "title": "Return Policy Extension",
+  "description": "Defines the `dev.ucp.shopping.policy.return` policy type on the core `policies[]` primitive, adding return terms to catalog, cart, checkout, and order.",
```
gsmith85, line 6: This description could be more clear that the policy MAY advertise across one or more these capabilities, but it doesn't necessarily advertise across all of them.

source/schemas/shopping/policy_return.json
```diff
@@ -0,0 +1,260 @@
...
+  "$defs": {
+    "return_body": {
+      "title": "Return Policy",
+      "description": "Return terms for a `dev.ucp.shopping.policy.return` policy. All fields are optional.",
```
gsmith85, line 10: Nit: Drop "All fields are optional" as it's redundant.  The schema itself communicates required/non-required.

source/schemas/shopping/policy_return.json
```diff
@@ -0,0 +1,260 @@
...
+      "type": "object",
+      "properties": {
+        "final_sale": {
+          "type": "boolean",
+          "description": "When `true`, the covered items cannot be returned. When `false` or omitted, the items are returnable subject to the other fields."
+        },
```
gsmith85, line 16: If the item cannot be returned, and `final_sale` obviates the others fields.

Should we lean more into structure:

```base_return_body
{ "final_sale": { type: "boolean" } }
```

```final_sale_return_body
{ "final_sale": { type: "boolean", const: true } }
```

```non_final_sale_return_body
{
  "final_sale": { type: "boolean", const: false },
  "window": ...
}
```

This makes the schema explicitly self-documenting and avoids contradictory states.

source/schemas/shopping/policy_return.json
```diff
@@ -0,0 +1,260 @@
...
+        "window": {
+          "title": "Return Window",
+          "description": "Return window as a policy statement. Omit for an unlimited or unstated window.",
+          "type": "object",
+          "required": ["days", "anchor"],
+          "properties": {
+            "days": {
+              "type": "integer",
+              "minimum": 0,
+              "description": "Length of the return window in whole days from `anchor`. `0` means non-returnable."
```
gsmith85, line 26: How does `days == 0` relate to `final_sale` being true?

source/schemas/shopping/policy_return.json
```diff
@@ -0,0 +1,260 @@
...
+          "required": ["type"],
+          "properties": {
+            "type": {
+              "type": "string",
+              "description": "Cost structure. Well-known values: `free`, `fixed_fee` (see `amount`), `customer_responsibility`."
+            },
+            "amount": {
+              "$ref": "../common/types/amount.json",
+              "description": "Flat channel fee in ISO 4217 minor units, when `type` is `fixed_fee`."
+            },
```
gsmith85, line 94: Consider splitting out a `fixed_fee` type and have `amount` live on it:

```json
"fixed_fee": {
  "title": "Fixed Fee Return Method Fee",
  "description": "Fixed logistics fee charged for a return channel.",
  "type": "object",
  "allOf": [{ "$ref": "#/$defs/fee" }],
  "properties": {
    "type": {
      "type": "string",
      "const": "fixed_fee"
    },
    "amount": {
      "$ref": "../common/types/amount.json",
      "description": "Flat channel fee in ISO 4217 minor units. Required when type is fixed_fee."
    }
  }
}
```

This does not change the contract of the base `fee` but relatedness between `type=fixed_fee` and `amount` is now explicitly captured by the schema.

source/schemas/shopping/policy_return.json
```diff
@@ -0,0 +1,260 @@
...
+      }
+    },
+    "policies_item": {
+      "description": "A `policies[]` entry; the return body applies when `type` is `dev.ucp.shopping.policy.return`.",
+      "allOf": [
+        {
+          "$ref": "types/policy.json"
+        },
+        {
+          "if": {
```
gsmith85, line 110: I flagged similar usage of `if/then` blocks in https://github.com/Universal-Commerce-Protocol/ucp/pull/688#discussion_r3731782164 and there suggested we avoid it entirely.  This is primarily motivated by (i) readability of the schema and (ii) interoperability with code generation.

Is there a reason to not have a `return_policies_item` that inherits from `"#/types/policy"`?

In `_response` definitions you would change:

```
"policies": {
  "type": "array",
   "items": {
      "anyOf": [
        { "$ref": "#/types/policy" }
      ]
    }
}
```

I believe the primary driver of the  `allOf`/`if/then` pattern is decentralized extensibility, but I believe that can be achieved using this approach as well.

source/schemas/shopping/policy_return.json
amithanda: Updated the description to clarify that businesses advertise return terms across one or more supported capabilities (catalog, cart, checkout, and order), matching the discovery specification in `return.md`.

source/schemas/shopping/policy_return.json
amithanda: Good catch, dropped "All fields are optional" from the description.

source/schemas/shopping/policy_return.json
amithanda: Thanks for flagging! Having `0 means non-returnable` created competing signals with `final_sale: true`. Updated `days` to require `"minimum": 1` and removed the non-returnable clause from its description so `final_sale: true` is the sole, unambiguous signal.

source/schemas/shopping/policy_return.json
amithanda: Good suggestion. Split out base `$defs/fee` (open `type`, `display_text`) and added `$defs/fixed_fee` inheriting via `allOf` with `"type": { "const": "fixed_fee" }` and `"required": ["type", "amount"]`. Both definitions are now documented in `return.md`.

source/schemas/shopping/policy_return.json
amithanda: Thanks for raising this, @gsmith85! I reviewed the resolution in #688 (`fulfillment_method.json` and `fulfillment_destination.json`) to align the patterns.

In JSON Schema draft 2020-12, using `anyOf: [return_policy, types/policy.json]` in `policies.items` introduces a validation masking issue: because `types/policy.json` sets `additionalProperties: true`, any malformed return policy (e.g. bad fee type, invalid `window.days`) would fail the `return_policy` branch but match the generic `policy` branch, causing `anyOf` to evaluate to true and silently bypass validation.

Following the pattern that landed in #688, we get the best of both worlds:
1. Added an explicit `$defs/return_policy` definition inheriting from `types/policy.json` via `allOf` with `const: "dev.ucp.shopping.policy.return"`. This gives SDK generators and reviewers the clean, self-contained `ReturnPolicy` model.
2. In `policies_item`, retain the `allOf` + `if/then` dispatch pointing to `$defs/return_policy`. This ensures that when `type == dev.ucp.shopping.policy.return`, the return policy schema is strictly validated without masking errors, while still allowing unknown/other policy types to pass through cleanly.

Does that work?

source/schemas/shopping/policy_return.json
amithanda: Thanks for raising this, @gsmith85. The core concern about contradictory states is closely tied to your comment on line 26 (regarding `days == 0`), which is now resolved in the latest commit.
I am leaning towards keeping `return_body` as a single model, and I think the latest commit addresses the contradictory state concern:
1. **Eliminating the competing signals:** The main source of contradictory states was that `window.days: 0` previously claimed to mean "non-returnable" alongside `final_sale: true`. In the latest commit, we updated `window.days` to require `"minimum": 1` and removed that clause. `final_sale: true` is now the single, canonical signal for non-returnability across the protocol.
2. **UCP's Precedence & Override Pattern:** In UCP, `final_sale` is typically authored as a targeted policy override rather than mixed into a single object. A merchant defines a response-wide default policy with the standard return terms (`window`, `methods`), and then publishes an item-scoped policy entry with `final_sale: true` targeting the specific non-returnable items (e.g. `applies_to: ["$.line_items[2]"]`). Under same-type precedence, that narrow entry governs line 2 and overrides the default. Because it's a separate entry, the business simply emits `final_sale: true` and omits the window/methods on that entry.
3. **Omission semantics & Code Generation DX:** For normal returnable goods, `final_sale` is omitted rather than explicitly declared as `false`. Splitting this into polymorphic sub-schemas (`final_sale_return_body` vs. `non_final_sale_return_body`) would create a discriminated union (`Union[FinalSaleReturnBody, NonFinalSaleReturnBody]`) in SDK generators. This would force SDK consumers to perform type narrowing on every policy read to access `window` or `methods`, conflicting with our goal of keeping the generated models clean and ergonomic (as discussed in #688).
4. **Authoritative fallback:** As specified in `return.md`, platforms treat `final_sale: true` as authoritative for eligibility decisions regardless of whether benign template defaults were carried over by upstream catalog exporters.

Let me know if that makes sense.

mangelmartinezfer-hue:
Some field data on what the return policy body has to be able to express, if it is useful while this is in review.

On 28 August 2026 I captured return policies from 14 US retailer pages using a real browser, forcing en-US, and recorded the source URL and capture date for each. Two are excluded here: one page returned 404 and one redirected to a non-US locale, so I can quote the text but cannot attest the source. That leaves 12.

Of those 12, 8 have a return window that is not a property of the merchant alone:

| Merchant | Window | Varies with |
|---|---|---|
| REI | 365 days for Co-op Members, 90 days otherwise | membership |
| IKEA | 365 days unopened, 180 days opened | item condition |
| Kohl's | 90 days, 120 days for Kohl's Card transactions | payment instrument |
| Bombas | 30 days, 60 days if purchased Nov-Dec | purchase date |
| Target | 90 days, 30 days when sold by Target Plus | seller |
| Walmart | 90 days, 30 days when sold by a Marketplace seller | seller |
| eBay | none set by the marketplace | seller, entirely |
| Nike | 60 days, from delivery online or purchase in store | channel |

Three observations that bear on the body's shape rather than on returns policy generally.

**1. A day count is ambiguous without the clock it runs on.** Nike's window starts at delivery for online orders and at purchase for store purchases. Walmart's reads "90 days after purchase or upon receipt." The same integer resolves to two different deadlines depending on the order, so the basis needs to travel with the number.

**2. Eligibility and resolution are different axes.** Bombas: past 30 days (60 if bought Nov-Dec) the item is still returnable, but for store credit rather than a refund. A boolean "returnable" is true and false at once. Brooklinen: 365 days, but "All returns are subject to a $9.95 return fee deducted from your refund amount." A window alone is an incomplete answer to "can I return this."

**3. Some windows are not the merchant's to author at all.** eBay sets none; the seller does. Target and Walmart halve theirs when a marketplace seller ships. Seller identity looks like an input to the policy rather than something the merchant can state once.

Kohl's is the one I did not anticipate - a payment instrument changing a return window is not something I had seen in any proposed schema, including my own.

Caveats: 12 stores is a sounding, not a sample. US only, one capture per page, one day. Happy to share the full set with source URLs and capture dates if that helps.

docs/specification/return.md
```diff
@@ -0,0 +1,269 @@
...
+# Return Policy Extension
+
+## Overview
+
+The Return Policy Extension defines the `dev.ucp.shopping.policy.return` policy
+type on the core [`policies[]`](overview.md#policies) primitive. It adds
+pre-purchase, machine-readable return terms to policies that carry this type, so
+platforms and agents can answer questions like "How long do I have to return
+this?", "Do I get my money back or store credit?", "Where and at what cost do I
+send it back?", and "Can I return this at all?" without leaving to read a policy
```
gsmith85: The issue isn't just leaving to read a policy page.  Policy pages often are incomplete or don't provide per product instructions.

docs/specification/return.md
```diff
@@ -0,0 +1,269 @@
...
+   limitations under the License.
+-->
+
+# Return Policy Extension
+
+## Overview
+
+The Return Policy Extension defines the `dev.ucp.shopping.policy.return` policy
+type on the core [`policies[]`](overview.md#policies) primitive. It adds
+pre-purchase, machine-readable return terms to policies that carry this type, so
```
gsmith85, line 23: Nit: "pre-purchase" seems to imply that the extension does not apply throughout the entire funnel, which it does.

docs/specification/return.md
```diff
@@ -0,0 +1,269 @@
...
+platforms and agents can answer questions like "How long do I have to return
+this?", "Do I get my money back or store credit?", "Where and at what cost do I
+send it back?", and "Can I return this at all?" without leaving to read a policy
+page.
```
gsmith85, line 27: What you're solving is really more significant than this.  Many policy pages lack product level granularity that can lead to unresolvable ambiguity for humans and agents 🙌 .

docs/specification/return.md
```diff
@@ -0,0 +1,269 @@
...
+    "description": {
+      "plain": "30-day returns from delivery. Refund to original payment or exchange. Free in-store; $5 return shipping by mail. 15% restocking fee on opened items."
+    },
+    "window": { "days": 30, "anchor": "delivered" },
+    "supported_resolutions": ["original_payment_method", "exchange"],
+    "methods": [
+      { "type": "in_store", "fee": { "type": "free", "display_text": "Free in-store return" } },
+      { "type": "by_mail", "fee": { "type": "fixed_fee", "amount": 500, "display_text": "Return shipping" } }
+    ],
+    "restocking_fee": { "percentage_bps": 1500, "display_text": "15% restocking fee on opened items" },
```
gsmith85, line 225: `return.md` mandates that platforms MUST treat structured fields as authoritative for computing the net refund and that `description` MUST NOT contradict structured fields.

If the structured fields are treated as authoritative here, then the "on opened items" clause in `description` / `display_text` is lost when computing the net refund from `percentage_bps: 1500`.

Either `condition` needs to be communicated structurally, or the normative language needs to state that the structured data for `restocking_fee` represents a maximum deduction with `display_text` / `description` carrying any condition qualifiers.

docs/specification/return.md
```diff
@@ -0,0 +1,269 @@
...
+
+A response-wide return policy states the standard terms once. An item-scoped
+policy marks one engraved line as final sale; because it targets that line
+directly, it governs it under same-type precedence, while every other line keeps
+the default. A paired disclosure compels the final-sale notice to be shown.
+
+<!-- ucp:example schema=shopping/checkout target=$.policies -->
+```json
+[
+  {
```
gsmith85, line 214: Because of the precedence scheme of policies (see: docs/specification/overview.md (lines 2308–2347, specifically line 2339)) it's not possible to add two policies at the same depth and precision.

What that practically means is that you're not going to be able to represent: 30 days for original_payment_method, 60 days for store_credit.

I don't think we need to block on this but we should consider documenting their primary return window using the policy with others subsumed into the description.

source/schemas/shopping/policy_return.json
gsmith85: It's been hard-fought but we have much more clarity now around how we want to handle polymorphic types 😅 .

The good news here is that we can enforce the invariant without a discriminated union:

```json
"if": {
  "properties": { "final_sale": { "const": true } },
  "required": ["final_sale"]
},
"then": {
  "properties": {
    "window": false,
    "supported_resolutions": false,
    "methods": false,
    "restocking_fee": false
  }
}
```

Philosophically if we can represent the constraint in the schema, I'd prefer that to relying on carve outs in the normative language.  I worry that an "authoritative fallback" creates an intra-object property precedence that is non-obvious.

source/schemas/shopping/policy_return.json
```diff
@@ -0,0 +1,310 @@
...
+              "minimum": 1,
+              "description": "Length of the return window in whole days from `anchor`."
+            },
+            "anchor": {
+              "type": "string",
+              "description": "Event `days` is measured from. Well-known values: `delivered`, `purchased`, `shipped`, `fulfilled`."
+            }
+          }
+        },
+        "supported_resolutions": {
```
gsmith85, line 34: return.md states:

>  business "MUST NOT use an empty supported_resolutions array"

Should this be marked `"minItems": 1`?

source/schemas/shopping/policy_return.json
```diff
@@ -0,0 +1,310 @@
...
+          "type": "array",
+          "description": "Accepted return outcomes, as an unordered set. Well-known values: `original_payment_method`, `store_credit`, `gift_card`, `exchange`, `replacement`.",
```
gsmith85, line 36: Per "unordered set", should we add `"uniqueItems": true` too?

source/schemas/shopping/policy_return.json
```diff
@@ -0,0 +1,310 @@
...
+            "type": "string"
+          }
+        },
+        "methods": {
+          "type": "array",
+          "description": "Permitted return channels and the logistics cost of each.",
+          "items": {
+            "$ref": "#/$defs/return_method"
+          }
+        },
```
gsmith85, line 47: Per the spec I believe methods should be omitted or be non-empty?

docs/specification/return.md
```diff
@@ -0,0 +1,269 @@
...
+so the platform reads it rather than tracking fulfillment events and recomputing.
+Modeling the resolved deadline field is deferred to that Order-side work.
+
+## Examples
+
+### Checkout with a default policy and a final-sale override
+
+A response-wide return policy states the standard terms once. An item-scoped
+policy marks one engraved line as final sale; because it targets that line
+directly, it governs it under same-type precedence, while every other line keeps
```
gsmith85, line 208: The matching pattern used in the example (`$.line_items[?@.id=='engraved_watch']`) does not target the line as a singular match: `[?...]` is a set match (`precision = 0`). A singular query (`precision = 1`) only permits name and integer index selectors.

If a checkout also carries a category filter (`$.line_items[?@.item.category=='watches']`), both evaluate to (`depth = 2, precision = 0`) and tie with undefined behavior.

Change the example selector in both `applies_to` and `messages[].path` to `$.line_items[2]`.

source/schemas/shopping/policy_return.json
```diff
@@ -0,0 +1,310 @@
...
+          "$ref": "order.json"
+        },
+        {
+          "type": "object",
+          "properties": {
+            "policies": {
+              "type": "array",
+              "items": {
+                "$ref": "#/$defs/policies_item"
+              }
```
gsmith85, line 284: Missing `"ucp_request": "omit"`.

source/schemas/shopping/policy_return.json
```diff
@@ -0,0 +1,310 @@
...
+        "type",
+        "amount"
+      ],
+      "properties": {
+        "type": {
+          "type": "string",
+          "const": "fixed_fee"
+        },
+        "amount": {
+          "$ref": "../common/types/amount.json",
```
gsmith85, line 106: Consider: We are not specifying currency here, which is likely fine, I think the context implies it, but you could note in the documentation that:

> Amounts are in the currency of the enclosing cart, checkout, or order, or of the covered product's price on catalog surfaces

source/schemas/shopping/policy_return.json
```diff
@@ -0,0 +1,310 @@
...
+          "type": "string",
+          "description": "Cost structure. Well-known values: `free`, `fixed_fee` (see `fixed_fee`), `customer_responsibility`."
+        },
+        "display_text": {
+          "type": "string",
+          "description": "Human-readable fee label (e.g., 'Return shipping')."
+        }
+      }
+    },
+    "fixed_fee": {
```
gsmith85, line 87: `fixed_fee` is not referenced. It should be wired into `return_method.properties.fee`:

```json
"fee": {
  "description": "Logistics cost of using this channel.",
  "allOf": [
    { "$ref": "#/$defs/fee" },
    {
      "if": {
        "properties": { "type": { "const": "fixed_fee" } },
        "required": ["type"]
      },
      "then": { "$ref": "#/$defs/fixed_fee" }
    }
  ]
}
```

Note: `#/$defs/fee` will also need `"additionalProperties": true` (like `types/policy.json`) so `ucp-schema` strict mode doesn't inject `"additionalProperties": false` on the base `fee` and reject `fixed_fee.amount`.

mangelmartinezfer-hue:
Thanks @gsmith85 — I really appreciate you taking the time on this, and your two notes on product-level granularity landed exactly where the field data does 🙌

Picking up from that data: the granularity point, the conditional restocking fee, and the 30/60-day precedence case all expose the same thing — a scalar value being serialized without the predicate that makes it true.

In the 12-policy sounding above, 8 return windows varied with membership, item condition, payment instrument, purchase date, seller, or channel. So it does not look rare, at least in that small set.

That is why I would be a little cautious about documenting one branch as the "primary" structured window and subsuming the rest into `description`. If structured fields are authoritative, one branch will look complete. Choosing the shortest window avoids an optimistic answer, but it still creates a false negative — denying a 60-day store-credit return after day 30, for example. That is still a wrong machine answer.

A bounded v1 rule could be:

- populate `window` only when one window applies across the policy's `applies_to` scope and supported resolutions;
- if the branches differ and cannot yet be modeled, omit `window` (or explicitly mark structured terms incomplete, if such a signal is added) and carry the branches in `description`;
- treat `restocking_fee` as a maximum only if the schema or normative semantics explicitly define it as a cap rather than an exact unconditional deduction.

That keeps "structured fields are authoritative" true: missing information stays visible instead of a conditional policy turning into an apparently complete scalar. A typed condition/scope model can stay future work.

Caveat as before: 12 retailers is a sounding, not a sample 🙂

Happy to keep feeding field data into this as the extension evolves, and genuinely glad to hear where you and @amithanda want to take it. This is the part of the problem I spend my time on, so any direction you point me in is useful.

Issue #636, opened 2026-07-24
[Bug]: Request-Signature marked required in 2026-01-23 shopping spec contradicts signatures.md


sbeashwar:
**Describe the bug**

In the `2026-01-23` release, the shopping REST OpenAPI defines the `Request-Signature` header parameter as `required: true`, and `$ref`s it into every shopping operation:

```jsonc
// source/services/shopping/openapi.json  (and the generated spec/services/shopping/rest.openapi.json)
"request_signature": {
  "name": "Request-Signature",
  "in": "header",
  "required": true,
  ...
}
```

This contradicts the same release's Message Signatures spec. `signatures.md` ("When Signatures Apply") states that platforms **SHOULD** sign requests and **MAY** use alternative authentication (API keys, OAuth, mTLS) instead - i.e. request signing is optional. Because the OpenAPI marks the header unconditionally required, clients and validators generated from the spec reject otherwise-valid requests that authenticate by another mechanism.

**To Reproduce**

1. Open `source/services/shopping/openapi.json` on `release/2026-01-23`.
2. Find `components/parameters/request_signature` -> `"required": true`.
3. Compare with `signatures.md` "When Signatures Apply" (requests SHOULD be signed; alternative auth MAY be used).
4. Generate a client/validator from the OpenAPI: the `Request-Signature` header is treated as mandatory on every operation.

**Expected behavior**

`Request-Signature` should be optional (`required: false`), consistent with `signatures.md`. `main` already treats request signing as optional, so the released spec is the only place still marking it required.

**Additional context**

The fix is a minimal, backward-compatible flag change (`required: true` -> `false`) in `source/services/shopping/openapi.json` and the generated `spec/services/shopping/rest.openapi.json` - loosening `required` never invalidates a previously-valid request. It does not require backporting main's larger RFC 9421 header restructure.

Since `main` already resolves this, would you accept a minimal backport to `release/2026-01-23` Based on past precedent, I have the change ready to open as a [PR](https://github.com/Universal-Commerce-Protocol/ucp/pull/637).

- [x] I agree to follow this project's Code of Conduct

vishkaty:
Confirmed for `v2026-01-23` — `source/services/shopping/openapi.json` does mark the
`Request-Signature` header `required: true` and `$ref`s it into every operation, which
contradicts the conditional signing model in `signatures.md` (§"When Signatures Apply":
[...]

PR #637, opened 2026-07-24
fix: make Request-Signature header optional in shopping spec

sbeashwar:
### What

Makes the `Request-Signature` header parameter optional (`required: false`) in the shopping REST spec, in both the `source/` and generated `spec/` OpenAPI, and clarifies its description.

### Why

The shopping OpenAPI marks `Request-Signature` as `required: true` on every operation, which contradicts this release's own Message Signatures spec (`signatures.md`, "When Signatures Apply"): platforms **SHOULD** sign requests, and **MAY** use alternative authentication (API keys, OAuth, mTLS) instead. As written, the OpenAPI forces a signature header even when a platform authenticates by another mechanism, and clients/validators generated from this spec reject otherwise-valid unsigned requests.

`main` already treats request signing as optional (the header is not unconditionally required there). This backports that intent to the `2026-01-23` release with a minimal, backward-compatible change - it does not pull in main's larger RFC 9421 header restructure. Existing signed requests remain valid; the only effect is that unsigned requests using alternative auth are no longer rejected by the spec.

### Scope

- `source/services/shopping/openapi.json`
- `spec/services/shopping/rest.openapi.json`

No behavioral change for signed requests; no schema fields added or removed. Consistent with prior backports to this release branch (e.g. #212, #228).

Addresses #636.

### DIFF
--- a/source/services/shopping/openapi.json
+++ b/source/services/shopping/openapi.json
@@ -430,11 +430,11 @@
       "request_signature": {
         "name": "Request-Signature",
         "in": "header",
-        "required": true,
+        "required": false,
         "schema": {
           "type": "string"
         },
-        "description": "Ensure the authenticity and integrity of an HTTP message."
+        "description": "Ensure the authenticity and integrity of an HTTP message. Optional; required only when the platform authenticates with HTTP Message Signatures. Platforms MAY use alternative authentication mechanisms (API keys, OAuth, mTLS) instead, per the Message Signatures specification."
       },
       "idempotency_key": {
         "name": "Idempotency-Key",
--- a/spec/services/shopping/rest.openapi.json
+++ b/spec/services/shopping/rest.openapi.json
@@ -455,11 +455,11 @@
       "request_signature": {
         "name": "Request-Signature",
         "in": "header",
-        "required": true,
+        "required": false,
         "schema": {
           "type": "string"
         },
-        "description": "Ensure the authenticity and integrity of an HTTP message."
+        "description": "Ensure the authenticity and integrity of an HTTP message. Optional; required only when the platform authenticates with HTTP Message Signatures. Platforms MAY use alternative authentication mechanisms (API keys, OAuth, mTLS) instead, per the Message Signatures specification."
       },
       "idempotency_key": {
         "name": "Idempotency-Key",


westonale:
Right call to align the OpenAPI with signatures.md's SHOULD/MAY. Two things worth making explicit so "optional" does not quietly become a downgrade:

1. Optional to send is not optional to validate. When `Request-Signature` IS present, the server MUST verify it and reject on failure, not treat a malformed signature as merely absent. Otherwise `required: false` widens into "signature errors are ignored."

2. The auth mechanism is the server's policy, not the caller's choice. A server that requires signatures for an operation MUST NOT accept an unsigned request just because the header is schema-optional; otherwise stripping the header is a downgrade path to whatever the weakest accepted method is.

The OpenAPI change is correct. A one-line note in signatures.md that omission is honored only where the server's policy actually permits an alternative would close the gap.

sbeashwar:
Thanks @westonale - agree on both points; the intent is exactly that "optional to send" should not become "optional to validate" or a silent downgrade path.

One snag on where to put the note: this PR targets `release/2026-01-23`, and that branch has no `signatures.md`. The message-signatures doc (with "When Signatures Apply") was added on `main` after this release was cut, so there's no signing doc in this tree to add the clarification to: the only signing surface on the release is the OpenAPI parameter this PR touches. Did you have a specific `signatures.md` on this branch in mind that I'm missing? I couldn't find one in the `release/2026-01-23` tree.

Happy to address your points either way:
1. Capture them directly in the `request_signature` parameter description in this PR - server MUST verify when the header is present; on omission, acceptance is governed by the server's auth policy, not the caller's choice - so the clarification lives in the release artifact itself.
2. Land the OpenAPI flip here and add the wording to `main`'s `signatures.md` "When Signatures Apply" as a small follow-up (main doesn't state the verify/policy rule today either).

westonale:
No, you are not missing anything. I checked the tree: `release/2026-01-23` has no signing doc at all, and `signatures.md` exists only on `main`, at `docs/specification/signatures.md`. The OpenAPI parameter really is the only signing surface on this release, so your option 1 is the right call for this PR.

Worth splitting the two points, though, because they are not in the same state upstream.

Point 1, MUST verify when present, is already normative on `main`. It is not in "When Signatures Apply", it is the closing rule of the Identity Resolution Algorithm: "The request MUST be rejected (`key_not_found`, `algorithm_unsupported`, or related error) only when every signature has been skipped or fails verification." Once a present signature fails, the request has no verifying signature, so that rule rejects it. That does not need restating on `main`. It is missing only here, on a branch with no IRA text.

Point 2, that omission is acceptable only where the server's policy permits an alternative, is absent on both. `main` says platforms SHOULD sign and that "alternative authentication mechanisms (API keys, OAuth, mTLS) may be used instead", which is a caller-side permission with no matching server-side statement. So the follow-up you offered is worth doing, and it only needs to carry the policy sentence.

One detail that makes that sentence load bearing rather than cosmetic. In `main`'s verification steps, `signature_invalid` and `digest_mismatch` both return `skip_signature(...)`, the same return as no signature at all, and step 5 says a component the signature does not cover "is treated as unsigned". The IRA's closing rule is the only thing that converts an all-skipped request into a rejection. That holds while the operation requires a signature. Where a server's policy accepts unsigned requests, a skipped signature and an absent one become the same input, and stripping or corrupting the header is a downgrade to whatever the weakest accepted method is. Stating that acceptance is the server's call is what keeps those two cases apart.

sbeashwar:
Gentle nudge on this one, @vixdug @sdedeo2025. It's a small, self-contained change (a single OpenAPI flag alignment). Whenever you have a moment, it should be a quick review. Thanks!

PR #638, opened 2026-07-26
refactor: compose Order payment instruments from the shared base

raginpirate:
Stacked ontop of https://github.com/Universal-Commerce-Protocol/ucp/pull/584 as an alternative direction to extend payment instruments beyond checkout.

The main differences are:
 - order.payment.instruments reuses the shared payment_instrument base instead of introducing a parallel order_payment_instrument.
- handler_id moves into selected_payment_instrument, since handler routing is specific to Checkout processing. Handler defined instrument schemas still extend the same base once.
- credential remains on the base because payment handlers extend it with their concrete credential shape. Sensitive and reusable credential fields must use ucp_response: omit, allowing responses to retain safe metadata such as credential type without returning tokens. It is also nil-able so it is not a requirement for request either.
- Per-instrument amount remains owned by the Split Payments extension. Split Payments now composes onto both Checkout and Order. For a completed split, Order preserves the Checkout instrument IDs and gross contribution amounts; refunds and credits remain separate Order adjustments.
- Order requires display, while leaving its handler-defined shape open.


igrigorik (approved):
👍

raginpirate:
Closed as this was merged into the base 🚀

PR #642, opened 2026-07-27
feat: Support location context hint in platform-initiated requests

jingyli:
# Description

Addressing a gap that would extend existing UCP shopping service schema/capabilities to support hyper-local goods (i.e. grocery) use cases:

- **Store/Branch Context**: Enable platform to provide location hints on top of user addresses, especially if the user has a known "favourite store".

### Standardized `location` Field in `context.json`
* **Problem**: Cart pricing & availability for these purchase journeys are often tied to specific physical stores, which requires passing a store identifier as hint to businesses (if the location is retrieved through upper-funnel journeys like catalog search).
* **Solution**: Add an optional `location` string property directly to `context.json`. This also allows the hint to be overwritten by more authoritative data constructs like `filters.fulfills_to` in `catalog_search` and `fulfillment` extension's pickup location ID.

## Category (Required)

## Checklist

### DIFF
--- a/source/schemas/shopping/types/context.json
+++ b/source/schemas/shopping/types/context.json
@@ -13,6 +13,10 @@
       "type": "object",
       "additionalProperties": true,
       "properties": {
+        "location": {
+          "type": "string",
+          "description": "Identifier or name hint for the specific location (e.g., physical store, property, or restaurant) associated with the user. A common example in retail shopping is the default home store selected and saved by the user when purchasing grocery."
+        },
         "intent": {
           "type": "string",
           "description": "Background context describing buyer's intent (e.g., 'looking for a gift under $50', 'need something durable for outdoor use'). Informs relevance, recommendations, and personalization."
--- a/source/schemas/shopping/types/context.json
+++ b/source/schemas/shopping/types/context.json
@@ -15,7 +15,7 @@
       "properties": {
         "location": {
           "type": "string",
-          "description": "Stable, opaque identifier for a Location in the Business's namespace. This provisional, non-binding hint is distinct from the Buyer's locality. The operation specification or an active capability/extension defines its effects. A common example in retail shopping is the default home store ID selected and saved by the user when purchasing grocery."
+          "description": "Stable, opaque identifier for a Location in the Business's namespace. This provisional, non-binding hint is distinct from the Buyer's locality. The operation specification or an active capability/extension defines its effects. A common example in retail shopping is the default home store ID selected and saved by the user when purchasing groceries."
         },
         "intent": {
           "type": "string",
--- a/source/schemas/shopping/types/context.json
+++ b/source/schemas/shopping/types/context.json
@@ -15,7 +15,7 @@
       "properties": {
         "location": {
           "type": "string",
-          "description": "Identifier or name hint for the specific location (e.g., physical store, property, or restaurant) associated with the user. A common example in retail shopping is the default home store selected and saved by the user when purchasing grocery."
+          "description": "Reference to the specific location (e.g., physical store, property, or restaurant) associated with the user. A common example in retail shopping is the default home store selected and saved by the user on their profile when purchasing grocery."
         },
         "intent": {
           "type": "string",
--- a/source/schemas/shopping/types/context.json
+++ b/source/schemas/shopping/types/context.json
@@ -15,8 +15,7 @@
       "properties": {
         "location": {
           "type": "string",
-          "description": "Stable, opaque identifier for a Location in the Business's namespace. This provisional, non-binding hint is distinct from the Buyer's locality. The operation specification or an active capability or extension defines its effects.",
-          "ucp_request": "optional"
+          "description": "Stable, opaque identifier for a Location in the Business's namespace. This provisional, non-binding hint is distinct from the Buyer's locality. The operation specification or an active capability or extension defines its effects."
         },
         "intent": {
           "type": "string",
--- a/source/schemas/shopping/types/context.json
+++ b/source/schemas/shopping/types/context.json
@@ -15,7 +15,8 @@
       "properties": {
         "location": {
           "type": "string",
-          "description": "Reference to the specific location (e.g., physical store, property, or restaurant) associated with the user. A common example in retail shopping is the default home store selected and saved by the user on their profile when purchasing grocery."
+          "description": "Stable, opaque identifier for a Location in the Business's namespace. This provisional, non-binding hint is distinct from the Buyer's locality. The operation specification or an active capability or extension defines its effects.",
+          "ucp_request": "optional"
         },
         "intent": {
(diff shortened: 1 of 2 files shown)


source/schemas/shopping/types/context.json
```diff
@@ -13,6 +13,10 @@
       "type": "object",
       "additionalProperties": true,
       "properties": {
+        "location": {
+          "type": "string",
+          "description": "Identifier or name hint for the specific location (e.g., physical store, property, or restaurant) associated with the user. A common example in retail shopping is the default home store selected and saved by the user when purchasing grocery."
```
jamesandersen: Acknowledging that #589 intends to add a location search capability should this PR be dependent on that one and then specify that values should be from `location.json#id`?  That feels like it would provide both platform and business a clearer expectation about this field (relative to it being an identifier *or* name hint)

jingyli: Good point! I updated the description here to be more explicit that it should be a reference to the location (so it's consistent with the field we have today in catalog filters - https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/shopping/types/fulfillment_destination_filter.json#L18

I don't want to specifically reference back to the json field (i.e. like `location.json#id`) in the description here since that will imply a required dependency between UCP capabilities. In most cases, that would be the source of which we pull the data from, but if there is another mechanism of which platforms & businesses can communicate this reference back-and-forth then it's also fine to use its value here as-is.

source/schemas/shopping/types/context.json
```diff
@@ -13,6 +13,10 @@
       "type": "object",
       "additionalProperties": true,
       "properties": {
+        "location": {
+          "type": "string",
+          "description": "Reference to the specific location (e.g., physical store, property, or restaurant) associated with the user. A common example in retail shopping is the default home store selected and saved by the user on their profile when purchasing grocery."
+        },
         "intent": {
           "type": "string",
           "description": "Background context describing buyer's intent (e.g., 'looking for a gift under $50', 'need something durable for outdoor use'). Informs relevance, recommendations, and personalization."
```
gsmith85, line 22: I agree with @jingyli that cross-referencing capabilities within schema descriptions introduces potential coupling between capabilities, which we want to avoid.

Still, there is an opportunity to create conceptual connectivity across specifications, which seems to be the crux of @jamesandersen's comment. Instead of cross-referencing capabilities in `context.location.description`.

I suggest we establish **Location** as a first-class commerce concept in UCP via the [Glossary](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/docs/specification/glossary.md):

  ```
     Term     | Acronym | Definition
    ----------|---------|---------------------------------------------------------------------------------------------------
     Location | -       | A physical entity (e.g., retail store, restaurant, pickup locker, warehouse) identified by a
              |         | stable, business-scoped string (id). Referenced across UCP for buyer localization
              |         | (context.location), store pickup fulfillment, and location discovery.
    ──────
  ```

This keeps the JSON Schema decoupled while giving implementers a clear, unified definition of what a location identifier represents across all UCP flows.

jingyli: A fair callout, I actually think establishing this as a UCP concept should belong to #589 instead of this PR here (as that is where the overarching capability will be introduced). PTAL at [5e3ff05](https://github.com/Universal-Commerce-Protocol/ucp/pull/589/commits/5e3ff05ab6cd83e4a645612fc6e3a4310cfca2d3)!

igrigorik (approved):
@jingyli @gsmith85 @jamesandersen pushed a small update (https://github.com/Universal-Commerce-Protocol/ucp/pull/642/changes/0b5035ecddb34bb3cd721514b8931e182f7fd5b5) to hopefully bring this one home. Taking Greg's suggestion on Glossary, ptal.

jingyli:
Thanks @igrigorik for https://github.com/Universal-Commerce-Protocol/ucp/commit/0b5035ecddb34bb3cd721514b8931e182f7fd5b5 - update LG though one nit I'd like to update on top of is to remove the explicit `ucp_request: optional` annotation on this new field so it remains consistent with other provisional hints where the optionality is determined by the enclosing capability.

source/schemas/shopping/types/context.json
jingyli: Resolved by https://github.com/Universal-Commerce-Protocol/ucp/commit/0b5035ecddb34bb3cd721514b8931e182f7fd5b5.

source/schemas/shopping/types/context.json
```diff
@@ -13,6 +13,10 @@
       "type": "object",
       "additionalProperties": true,
       "properties": {
+        "location": {
+          "type": "string",
+          "description": "Stable, opaque identifier for a Location in the Business's namespace. This provisional, non-binding hint is distinct from the Buyer's locality. The operation specification or an active capability or extension defines its effects."
```
gsmith85: > The operation specification or an active capability or extension defines its effects.

Minor nit: Replace the first "or" with a comma

gsmith85 (commented):
LGTM

source/schemas/shopping/types/context.json
igrigorik: No I think 'or' is correct. It's saying operation (e.g. create_cart) or extension (e.g. `org.foo.fancy-fulfillment` that extends dev.ucp.cart) define the semantics of this field.

igrigorik:
@jingyli @gsmith85 lgtm on updated shape. @amithanda final pass?

source/schemas/shopping/types/context.json
amithanda: @jingyli - can we add the examples back in the schema file to make it clearer?

amithanda (approved):
Added a minor suggestion, PTAL.

source/schemas/shopping/types/context.json
jingyli: @gsmith85 To @igrigorik's callout above, I think the first `or` should still be preserved. However, the second one (`an active capability or extension`) can be simplified to just a `/` so the sentence runs a bit more smoothly. Made the corresponding update, PTAL!

@amithanda Added back the common example for retail shopping, PTAL!

source/schemas/shopping/types/context.json
```diff
@@ -13,6 +13,10 @@
       "type": "object",
       "additionalProperties": true,
       "properties": {
+        "location": {
+          "type": "string",
+          "description": "Stable, opaque identifier for a Location in the Business's namespace. This provisional, non-binding hint is distinct from the Buyer's locality. The operation specification or an active capability/extension defines its effects. A common example in retail shopping is the default home store ID selected and saved by the user when purchasing grocery."
```
gsmith85: Nit: groceries.

gsmith85 (approved):

PR #651, opened 2026-07-30
fix(openapi): reference request-shape schemas for checkout and cart request bodies

### LINKED ISSUE #649: [Bug]: OpenAPI schema still exposes checkout.id in update_checkout request body
### Describe the bug

The OpenAPI specification `source/services/shopping/rest.openapi.json` (and its generated versions) still references the base `checkout` schema for the `PUT /checkout-sessions/{id}` (`update_checkout`) request body.

In PR #341, `checkout.id` was marked as deprecated in the `update_checkout` request (and subsequently removed from examples in PR #646).
The JSON schema `source/schemas/shopping/checkout.json` correctly marks `id` with `"ucp_request": "omit"`.

However, because the OpenAPI spec references `checkout.json` directly and does not use `readOnly: true` or a separate update request schema, standard OpenAPI tools will still expose `id` in the request body for `update_checkout`.

### To Reproduce

1. Open `source/services/shopping/rest.openapi.json`.
2. Locate the `/checkout-sessions/{id}` path and the `put` operation.
3. Observe that `requestBody` references `#/components/schemas/checkout`, which in turn references `../../schemas/shopping/checkout.json`.
4. Open `source/schemas/shopping/checkout.json` and see that `id` is defined as a property.
5. Standard OpenAPI parsers (like Swagger UI or Redoc) will show `id` as a valid field in the request body.

### Expected behavior

The OpenAPI specification should correctly reflect that `id` is omitted in the `update_checkout` request body. This can be achieved by:
*   Adding `readOnly: true` to the `id` property in `checkout.json` (if supported by the UCP toolchain).
*   Or defining a separate `checkout_update_request` schema in `rest.openapi.json` that excludes `id` and referencing it in the `PUT` request body.

### Additional context

*   Deprecation PR: https://github.com/Universal-Commerce-Protocol/ucp/pull/341
*   Example cleanup PR: https://github.com/Universal-Commerce-Protocol/ucp/pull/646

Mubeen-Fatima:
Fixes #649.

Scope per maintainer confirmation in #649: checkout + cart in one PR, same root cause.

## Problem

All five REST request bodies (`create_checkout`, `update_checkout`, `complete_checkout`, `create_cart`, `update_cart`) referenced the full base schema, so standard OpenAPI tooling (Swagger UI, Redoc, codegen) exposed every server-assigned field as writable — 12 `ucp_request: "omit"` fields on checkout and 9 on cart (`id`, `status`, `totals`, `expires_at`, `continue_url`, …), not just `checkout.id` as originally filed.

`readOnly: true` was considered and rejected: 28 `ucp_request` annotations are per-operation maps (e.g. `cart.id` is `{"create": "omit", "update": "required"}`, `checkout.payment` is required only on complete), which a binary, operation-global `readOnly` cannot express.

## Change

Follows the existing `catalog_search.json` → `$defs/search_request` convention (also used by `ask_request` in #538):

- **`checkout.json`**: add `$defs/create_request`, `$defs/update_request`, `$defs/complete_request`. Properties `$ref` the top-level definitions (`#/properties/...`) so descriptions stay single-sourced; the `ucp_request` annotations remain the source of truth.
- **`cart.json`**: add `$defs/create_request`, `$defs/update_request` alongside the existing `$defs/checkout`.
- **`rest.openapi.json`**: point the five request bodies at new `*_request` components, mirroring the catalog wiring.

Each `$defs` shape was verified byte-equivalent (properties + required) to the canonical derivation:

```
ucp-schema resolve source/schemas/shopping/checkout.json --request --op create|update|complete
ucp-schema resolve source/schemas/shopping/cart.json --request --op create|update
```

## Verification

- `ucp-schema lint source/` — 99 files, all passed
- `scripts/validate_examples.py` full corpus — 282 passed, 0 failed
- `scripts/test_validate_examples.py` — 48 passed
- `pre-commit run --all-files` — passed (stylelint fails locally on a missing node module, unrelated; no CSS touched)
- `mkdocs build` — no new warnings vs `main` (reference pages only render `$defs` via explicit macros, so schema pages are unchanged)

## Notes

- The now-unreferenced `checkout`/`cart` components are left in place (the `payment` component was already name-only) in case downstream codegen references them by name; happy to remove them if preferred.
- #633 transitions `cart.id` to omitted in update requests; if it lands first, `cart.json#/$defs/update_request` needs a two-line follow-up (drop `id` from `properties`/`required`). Current shapes reflect the annotations on `main` today.
- `mcp.openrpc.json` has the same defect (`create_checkout`/`update_checkout`/`complete_checkout` tool params reference the full `checkout.json`). Left out of scope here as filed; can follow up if wanted.

### DIFF
--- a/source/schemas/shopping/cart.json
+++ b/source/schemas/shopping/cart.json
@@ -26,6 +26,62 @@
           }
         }
       ]
+    },
+    "create_request": {
+      "title": "Cart Create Request",
+      "description": "Request body for create_cart. Includes only buyer-writable fields; server-assigned fields are omitted per the ucp_request annotations on the base schema, which remain the source of truth (equivalent to `ucp-schema resolve --request --op create`).",
+      "type": "object",
+      "required": [
+        "line_items"
+      ],
+      "additionalProperties": true,
+      "properties": {
+        "line_items": {
+          "$ref": "#/properties/line_items"
+        },
+        "context": {
+          "$ref": "#/properties/context"
+        },
+        "signals": {
+          "$ref": "#/properties/signals"
+        },
+        "attribution": {
+          "$ref": "#/properties/attribution"
+        },
+        "buyer": {
+          "$ref": "#/properties/buyer"
+        }
+      }
+    },
+    "update_request": {
+      "title": "Cart Update Request",
+      "description": "Request body for update_cart. Server-assigned fields are omitted per the ucp_request annotations on the base schema, which remain the source of truth (equivalent to `ucp-schema resolve --request --op update`).",
+      "type": "object",
+      "required": [
+        "id",
+        "line_items"
+      ],
+      "additionalProperties": true,
+      "properties": {
+        "id": {
+          "$ref": "#/properties/id"
+        },
+        "line_items": {
+          "$ref": "#/properties/line_items"
+        },
+        "context": {
+          "$ref": "#/properties/context"
+        },
+        "signals": {
+          "$ref": "#/properties/signals"
+        },
+        "attribution": {
+          "$ref": "#/properties/attribution"
+        },
+        "buyer": {
+          "$ref": "#/properties/buyer"
+        }
+      }
     }
   },
   "type": "object",
--- a/source/schemas/shopping/checkout.json
+++ b/source/schemas/shopping/checkout.json
@@ -4,6 +4,86 @@
   "name": "dev.ucp.shopping.checkout",
   "title": "Checkout",
   "description": "Base checkout schema. Extensions compose onto this using allOf.",
+  "$defs": {
+    "create_request": {
+      "title": "Checkout Create Request",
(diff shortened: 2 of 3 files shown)


Mubeen-Fatima:
Rebased onto `main` now that #633 has landed, and re-verified.

#633 changed `cart.id`'s update annotation to the `transition` form (`from: required` → `to: omit`). `ucp-schema resolve --request --op update` still reports `id` as required for cart update today, so `cart.json#/$defs/update_request` is unchanged and remains byte-equivalent to the canonical derivation. All five shapes re-checked against the resolver after the rebase — still matching.

Re-ran on the rebased branch: `ucp-schema lint` 99 files passed, example corpus 282 passed / 0 failed, validator unit tests 48 passed.

Worth flagging for whoever picks up the transition follow-up: when `cart.id` completes its move to `omit` on update, the only change needed here is dropping `id` from `properties` and `required` in `cart.json#/$defs/update_request` — the OpenAPI wiring stays as-is.

Issue #525, opened 2026-06-17
key ordering for UCP map registries


igrigorik:
UCP represents extensible registries as JSON objects keyed by reverse-DNS identifiers. One limitation of this contract is that JSON object member order is not guaranteed, and we've run into cases where we want to declare the preferred traversal order of their keys — e.g. payment handlers, identity providers, etc. 

```json
{
  "payment_handlers": {
    "com.google.pay": [{"id": "gpay", "version": "2026-01-11"}],
    "dev.shopify.shop_pay": [{"id": "shop_pay", "version": "2026-01-11"}]
  },
  "providers": {
    "app.example.login": [{"type": "oauth2", ... }],
    "com.google": [{"type": "oauth2", ...}]
  }
}
```

Note: this shape is intentional. Reverse-DNS keys provide decentralized namespace governance, JSON object keys provide uniqueness by construction, and schemas can enforce key shape with `propertyNames` / `patternProperties`. However, JSON object member order is not semantically reliable. Canonicalization, signing, intermediaries, and ordinary JSON tooling may reorder object keys. 

---

## Proposal
Introduce a reserved local companion object named `$order`.

```json
{
  "providers": {
    "app.example.login": [{"type": "oauth2"}],
    "com.google": [{"type": "oauth2"}]
  },
  "$order": {
    "providers": ["app.example.login", "com.google"]
  }
}
```

> `$order` is an object scoped to its containing object. Each property of `$order` names a sibling map-valued field. The value is an array of keys from that sibling map and defines ordered traversal of those keys.

`$order` defines **key order only**. It does not, by itself, define display priority, selection priority, trust level, negotiation preference, endorsement, or fallback behavior. The referenced field’s specification must state what ordered traversal means in that context.

### Normative rules

For a sibling map field `<field>` and companion array `$order.<field>`:

1. Producers MUST NOT rely on JSON object member order for UCP map-valued registries.
2. `$order.<field>` contains keys from `<field>` in preferred traversal order.
3. The order array MAY be partial. Listed keys are traversed first, in array order.
4. Unlisted map keys remain valid and available. They are appended after listed keys using the field-defined fallback, or lexicographic order if no fallback is defined.
5. The order array is not a whitelist. Omission from `$order.<field>` MUST NOT mean removal, ineligibility, or reduced support.
6. Producers SHOULD only list keys present in the sibling map. Robust consumers SHOULD ignore keys that are not present.
[...]

PR #657, opened 2026-07-31
feat: reserve the `ucp` protocol namespace, add `map_order` for registry ordering

igrigorik:
UCP represents extensible registries as JSON objects keyed by reverse-DNS names — `payment_handlers`, `services`, `capabilities`. That shape is deliberate (decentralized governance, uniqueness by construction, `propertyNames` enforcement), but JSON object member order is not a protocol contract: [RFC 8785](https://www.rfc-editor.org/rfc/rfc8785.html) canonicalization, which UCP signing relies on, sorts object member names. There is currently no reliable way for a Business to declare preferred traversal order for registry keys — the concrete driver is merchant-preferred payment-handler presentation order (#170; design discussion in #525).

### Proposed model

- **The member name `ucp` is reserved at every object scope as the protocol namespace.** The top-level envelope that profiles and responses already carry is not a special wrapper — it is the root manifestation of this reservation: a reserved member of the root object, grammatically identical to any nested scope carrying protocol annotations. Placement carries no semantics.
- **The namespace is ambient and its vocabulary is schema-registered.** Any object scope MAY carry a `ucp` member; domain schemas never declare it — the member is part of the UCP document grammar, like the name reservation itself. Its contents are defined exclusively by the vocabulary registered in `ucp.json#/$defs/members`. Consumers process the members they recognize and MUST ignore the rest.
- **Admission rule:** a member enters the vocabulary only if it is safe to ignore — a non-processing consumer loses only that member's benefit, never correctness. No structural member can ever be load-bearing. One exception is an object closed with `additionalProperties: false` does not get an exemption from the grammar — it MUST declare the optional `ucp` property explicitly.

Because registration is central, future members work at every scope the moment they land: #655's request constraints rebase onto this vocabulary as one property registration, immediately usable on any leaf object (e.g. per-line-item quantity constraints).

### First member: `map_order`

`map_order` declares preferred key-traversal order for sibling map-valued fields, carried in an array value because JCS preserves array element order while sorting object keys. Partial lists are valid; unlisted keys remain valid and follow via field-defined fallback or lexicographic order; the list is not an allowlist; absent means no order is declared, and consumers MUST NOT fall back to member order.

At the root envelope, it orders the registries directly beside it:

```json
{
  "ucp": {
    "version": "2026-01-11",
    "payment_handlers": {
      "com.google.pay": [{ "id": "gpay", "version": "2026-01-11" }],
      "dev.shopify.shop_pay": [{ "id": "shop_pay", "version": "2026-01-11" }]
    },
    "map_order": {
      "payment_handlers": ["dev.shopify.shop_pay", "com.google.pay"]
    }
  }
}
```

At any deeper scope, the same member rides the ambient `ucp` member — no schema change to the host object. Ordering an identity-provider registry inside a capability config:

```json
{
  "config": {
    "providers": {
      "app.example.login": [{ "type": "oauth2" }],
      "com.google": [{ "type": "oauth2" }]
    },
    "ucp": {
      "map_order": { "providers": ["app.example.login", "com.google"] }
    }
  }
}
```

For `payment_handlers`, the declared order is the Business's presentation preference (#170): suggestive — the Platform SHOULD consider it and MAY apply its own ordering. 

### Why this shape / alternatives considered in #525:

- **`display_order` embedded rank (#176):** cannot attach to a map bucket whose value is an array; unnecessary where arrays already self-order.
- **Reserved key inside the map:** breaks `propertyNames` — a breaking change for every shipped validator; typed maps become unions.
- **Per-field siblings (`payment_handlers_order`):** open-ended reserved-name sprawl; `_order` suffix collides with the commerce Order noun.
- **`$`-prefixed member (original #525 proposal):** empirically taxed in codegen — `$` is an illegal identifier in Python/Go/Java/C#/Rust, and Pydantic's asymmetric alias defaults silently emit non-conforming payloads with stock `datamodel-code-generator`. `ucp` is a clean identifier in every ecosystem, and the name is already de facto reserved (zero domain usages in the corpus).
- **Array-of-keyed-entries registries:** cleanest greenfield shape, breaking for every shipped UCP map.
- **Per-scope opt-in declaration:** wastes the universal reservation and pollutes every host schema; rejected in favor of the ambient grammar plus central registration.

### Compatibility

Additive. The envelope is open today; existing consumers ignore `map_order`. No shipped registry is reshaped; registry maps stay pure and homogeneous (`propertyNames` already rejects `ucp` and `map_order` as registry keys — verified). Consumers that ignore the member lose nothing: order was never reliable before.

---

## Checklist


docs/specification/overview.md
```diff
@@ -1778,6 +1924,17 @@ governing body.
 the context of the cart (e.g., removing "Buy Now Pay Later" for subscription
 items, or filtering regional methods based on shipping address).
 
+**Presentation Order:** Businesses **MAY** declare a preferred presentation
```
jamesandersen, line 1973: This is a tricky as per the comments on #170 and #176 citing that available *instruments* within handlers are really the presentation grain that a business may be most interested in ordering and `map_order` doesn't cover that (in current form).  However, in practice this may still address a large number of use cases (where payment handlers are advertising more than one available instrument or where businesses are content to order at the handler level) riding on a broadly usable ordering mechanism.

igrigorik: Pushed an update that clarifies this: [6391c451](https://github.com/Universal-Commerce-Protocol/ucp/pull/657/changes/6391c451c26a2ef3c8736425bfc2c68235e1c9c1) — within a handler, the business's `available_instruments` array order carries the instrument presentation preference, so combined with `map_order.payment_handlers` the full presentation order can be derived (covers the #176 example: order the handlers, tokenizer advertises `[card, bank]`).

To be fair, this doesn't cover every case: a business can rank handlers, and rank instruments within each handler, but it can't mix them freely across handlers — e.g. card via gpay first, then bank via tokenizer, then back to paypal via gpay. I'd suggest we leave that as is though.

jamesandersen: Yeah - it was the mixing case that I had in mind ... but - agreed - not a critical / blocking use case.

jamesandersen:
Two comments coming to mind:
1. An ambient ucp that domain schemas never declare is invisible to schema-driven codegen (untyped escape hatch to read it, and non-uniform given the closed-object exception), and its contents fall outside ordinary instance validation. More importantly, the tolerant-reader mandate isn't free at every scope — e.g. Java's Jackson fails on unknown by default; Go/Pydantic drop it (round-trip/re-sign loss). Given $ was rejected partly on codegen grounds, this trades one codegen tax for another.

2. Does the reserved `ucp` member have defined meaning on the request side, or only responses? The container looks request/response-agnostic, and gating is per-member via ucp_request: omit — which #655 uses for request_constraints, but map_order isn't marked, so it's schema-valid on a request with no defined semantics. Since this is the foundational PR, worth stating the general rule here: either per-member applicability (with map_order marked/ignored on requests) or response/profile-only for the whole container?

docs/specification/overview.md
```diff
@@ -958,6 +958,152 @@ example:
...
 
+### The `ucp` Protocol Namespace
+
+The member name `ucp` is reserved in every UCP object scope as the **protocol
+namespace**. The top-level `ucp` member that profiles and responses carry —
+described in [Profile Structure](#profile-structure) above — is not a special
+wrapper; it is the root manifestation of this reservation: a reserved member
+of the root object. The same reservation applies at every nested object
+scope. Schema authors **MUST NOT** define a member named `ucp` in domain
+schemas or extensions.
```
amithanda: The ambient reach seems deliberate, and #655 is the reason: it needs the
member on arbitrary leaf objects, and "immediately usable on any leaf object"
is the stated payoff. My question is narrower, whether "every object scope"
is meant to include dictionaries, where neither registered member can do
anything.

Both registered members need schema-declared field names. #655 requires that
"every field name in `required` or under `properties` exists at the
corresponding level of the actual composed, operation-specific request schema,"
and `map_order` targets map-valued fields of the annotated scope. A dictionary
has neither. Every use in both PRs is on a named-property object: a Line Item,
a payment instrument, a capability `config`. And for the one case where you
might want to constrain a dictionary's keys, `request_constraints` reaches it
from the parent through `properties`, so nothing needs `ucp` inside the
dictionary itself.

That matters because "reserved in every UCP object scope" plus "any object
scope in a UCP payload **MAY** carry a `ucp` member" currently reaches scopes
where the reservation is unusable or actively conflicts. The authoring guide
explains why registry maps are safe (`propertyNames` rejects `ucp`), but
several shipped scopes are dictionaries with no `propertyNames`:

- `shopping/types/attribution.json` is `additionalProperties: {"type": "string"}`.
  A `ucp` key is legal there today and its value **must** be a string, so a
  protocol-namespace object at that scope is schema-invalid. The ambient "MAY
  carry a `ucp` member" is therefore not true everywhere. It is also plausible
  in practice: the field is documented as "the same parameters platforms
  communicate via URL query parameters," and `ucp` is exactly what a platform
  would name a UCP-attribution param. Reachable from checkout, cart, order,
  catalog_search, and catalog_lookup.

Would scoping the reservation to object scopes whose members are named by a
schema, with dictionary scopes called out as exempt the way registry maps are,
cost anything? It would keep every use #655 needs and drop the scopes where no
member can apply. If you agree, here is a quick suggestion:

```suggestion
of the root object. The same reservation applies at every nested object
scope whose members are named by a schema. Schema authors **MUST NOT**
define a member named `ucp` in domain schemas or extensions.

The reservation does not extend to dictionary scopes: objects whose keys are
open data rather than schema-declared members, such as `attribution`. 
Those scopes carry no protocol namespace, and a `ucp` key appearing in one is
ordinary data.
```

docs/specification/overview.md
```diff
@@ -958,6 +958,152 @@ example:
...
+member names while preserving array element order. An array
+value is therefore the only order carrier that survives canonicalization and
+signing — and `map_order` uses one.
+
+`map_order` declares a preferred key-traversal order for sibling map-valued
+fields of its containing scope. Each key of `map_order` names a sibling map
+field; its value is an array of that map's keys in preferred traversal order.
+At the root envelope, `map_order` appears directly beside the registries it
+orders — both are protocol-namespace members. In a nested scope, it lives
+inside that scope's `ucp` member.
```
amithanda: "Each key of `map_order` names a **sibling** map field" does not hold nested:
in the `providers` example the target is a sibling of `ucp`, not of `map_order`.
At the root the targets really are siblings of `map_order`, because the
registries live inside `ucp`. The two sentences that follow state both cases, so
this is close to a wording fix, except that the two readings differ on a case
that now exists in `main`.

#458 merged `actions`, a reverse-DNS registry map with array values, at
`checkout.json:88`, as a sibling of `ucp` on the response root. Under the
uniform reading ("`ucp` annotates the object that carries it"),
`$.ucp.map_order.actions` orders it. Under the root reading ("`map_order` orders
the registries directly beside it"), `actions` is outside the resolution scope
and rule 6 says to ignore the entry. The schema accepts the spelling either way,
since `map_order` keys are unconstrained, so two implementers can diverge on the
same document.

#655 states the uniform reading directly: "The `ucp` object containing
`request_constraints` annotates its parent logical object." That is the rule I
would expect here too, and it is what the suggestion below codifies, with the
root called out as the one place it does not hold.

I am not arguing `actions` should be orderable. overview.md:148 says the map
defines no cross-type processing order and :151 hands sequencing to the
declaring extension, which seems right. I am asking that the spec say which
reading holds, so `map_order` on a root-level domain field is either defined or
explicitly out of scope. Does the following suggestion help clarify it better?

```suggestion
`map_order` declares a preferred key-traversal order for the map-valued
fields of the scope its `ucp` member annotates. Each key of `map_order` names
one of those map fields; its value is an array of that map's keys in
preferred traversal order.

Targets resolve one level out from `map_order`, in the object that carries
the `ucp` member. The document root is the exception: there the protocol
registries and their annotations share a single object, so `map_order` names
its own siblings inside `ucp` and does not reach domain fields of the
document root, such as a checkout response's `actions`.
```

source/schemas/ucp.json
```diff
@@ -46,6 +46,25 @@
...
     },
 
+    "map_order": {
+      "type": "object",
+      "description": "Preferred key-traversal order for sibling map-valued fields of the containing scope. Each property names a sibling map field and its array lists that map's keys in preferred order. Lists may be partial: unlisted keys remain valid and are appended using the field-defined fallback or lexicographic order. This member is not an allowlist; consumers ignore unknown or absent target fields.",
+      "additionalProperties": {
+        "type": "array",
+        "items": { "type": "string" },
+        "uniqueItems": true
+      }
```
amithanda, line 55: Should a malformed `map_order` be able to reject a whole profile?
`uniqueItems: true` means a duplicated key fails validation for the entire
document. Verified on this branch:

```
{"valid":false,"errors":[{"path":"/ucp/map_order/payment_handlers",
 "message":"[\"com.example.pay\",\"com.example.pay\"] has non-unique elements"}]}
```

`profile.json` already takes the opposite position for the other optional,
skippable profile element: an unsupported key "affects only the signature that
references it ... and **MUST NOT** cause whole-profile rejection." An advisory
presentation hint seems like the last thing that should be able to break
discovery for a business, and rule 7 already defines the recovery. With the
suggestion applied, the payload above returns `{"valid":true}`.

```suggestion
      "additionalProperties": {
        "type": "array",
        "items": { "type": "string" }
      }
```

docs/specification/overview.md
```diff
@@ -958,6 +958,152 @@ example:
...
+4. Unlisted map keys remain valid and available; consumers traverse them
+    after the listed keys, using the field-defined fallback order or, if the
+    field defines none, lexicographic order.
+5. The order array is not an allowlist: consumers **MUST NOT** interpret
+    omission of a key as removal, ineligibility, or reduced support.
+6. Producers **SHOULD** list only keys present in the sibling map. Consumers
+    **SHOULD** ignore entries naming keys that are not present, and entries
+    whose target field is absent or is not a map.
+7. Duplicate keys are invalid; consumers **SHOULD** honor the first
+    occurrence of a key and ignore later duplicates.
```
amithanda: Companion to my other comment, making the recovery normative rather than advisory.

```suggestion
7. Duplicate keys are invalid, but a duplicate **MUST NOT** invalidate the
    containing document: consumers **SHOULD** honor the first occurrence of a
    key and ignore later duplicates.
```

docs/specification/overview.md
```diff
@@ -958,6 +958,152 @@ example:
...
+grammar, like the name reservation itself. Guidance for schema authors on
+working within this reservation lives in the Schema Authoring Guide's
+[The Reserved `ucp` Member](/documentation/schema-authoring/#the-reserved-ucp-member)
+section. A consumer encountering a `ucp` member at any scope processes the
+members it recognizes, each per its own definition, and **MUST** ignore
+unrecognized members (see *Openness* above).
+A member is admitted to the vocabulary only if it is safe to ignore: a
+consumer that does not process it loses only that member's benefit, never
+correctness. A consumer that ignores `map_order`, for example, simply
+traverses the map unordered — the status quo before ordering existed.
```
amithanda: Building on @jamesandersen's round-trip point rather than repeating it: ambient
`ucp` members can now appear at any depth inside a checkout payload, and
`merchant_authorization` is a JWS over the JCS-canonicalized payload, so the
question is what the signature is meant to cover.

`ap2` is excluded from that computation on the stated rationale that one
exclusion rule handles all future `ap2` fields. Ambient `ucp` members are
non-load-bearing by the admission rule, which is a similar argument for
excluding them. But the root `ucp` carries load-bearing handshake members, so
the exclusion cannot be uniform, which is another place the "identical at every
scope" model has to bend. If they stay inside the signature, is it worth saying
that ignore means "do not process" rather than "may drop"? Grepping
`docs/specification/` for a preservation rule turns up nothing today.

Should we clarify it in the documentation:

```suggestion
A member is admitted to the vocabulary only if it is safe to ignore: a
consumer that does not process it loses only that member's benefit, never
correctness. A consumer that ignores `map_order`, for example, simply
traverses the map unordered, which is the status quo before ordering existed.
Ignoring a member means not processing it, not discarding it. A party that
re-serializes a payload covered by a signature, including an AP2 mandate
computed over the JCS-canonicalized body, **MUST** preserve unrecognized
`ucp` members verbatim.
```

docs/specification/overview.md
```diff
@@ -958,6 +958,152 @@ example:
...
+inside `ucp` that they do not recognize (tolerant reader). Openness exists so
+documents produced under a newer UCP version remain readable by older
+consumers — it is *not* extension space. Only UCP core defines members inside
+`ucp`, and extension authors **MUST NOT** place extension data there. An
+unrecognized member inside `ucp` means "defined by a newer UCP version,"
+never "extension data."
+
+**Idempotence.** Within `ucp`, protocol members appear directly: `ucp.ucp`
+never exists. Producers **MUST NOT** emit a `ucp` member inside `ucp`, and no
+UCP version will define one. Note that because the container is open, an
```
amithanda: "Producers **MUST NOT** emit a `ucp` member inside `ucp`" is literally violated
by the nested example: in a real profile that `config` sits at
`$.ucp.capabilities["dev.ucp.common.identity_linking"][0].config`, so
`config.ucp` is a `ucp` member inside `ucp`. I read the intent as "direct
child." Worth saying so, and perhaps showing the nested example in its real
profile position?

```suggestion
**Idempotence.** Within `ucp`, protocol members appear directly: `ucp.ucp`
never exists. Producers **MUST NOT** emit a `ucp` member as a direct child of
`ucp`, and no UCP version will define one. Deeper nesting is unaffected: a
`ucp` member on an object that happens to sit inside `ucp`, such as a
capability's `config`, is that object's own protocol namespace rather than a
nested envelope. Note that because the container is open, an
```

docs/specification/overview.md
igrigorik: Yep, great call. Addressed.

igrigorik:
@jamesandersen put together [ucp-schema#48](https://github.com/Universal-Commerce-Protocol/ucp-schema/pull/48) as ref implementation: it boils down to a preprocessor pass that materializes `ucp` properties at eligible objects. The resulting output is ordinary JSON Schema, so downstream validation and codegen work unchanged. Overall, I agree that this adds another UCP-specific requirement, but we already have the preprocessor requirement for resolving ucp request/response annotations, so it's not a net new ask.

@amithanda great feedback, ty! Addressed in followup commits -- ptal.

jamesandersen:
Thanks @igrigorik for https://github.com/Universal-Commerce-Protocol/ucp-schema/pull/48 to incorporate the optional `ucp` members into the overall schema (so every consumer won't have to).

Some good callouts from @amithanda to tighten up the docs as well.

I'm feeling pretty comfortable with this one and inclined to approve but do want to vet that #655 which builds on this also is trending to approval.

jamesandersen (approved):

amithanda (approved):

docs/documentation/schema-authoring.md
```diff
@@ -198,6 +198,59 @@ The same registry structure appears in three contexts with different field requi
...
+  This is why structural metadata about a registry map — such as `map_order`
+  — sits in the parent structured scope's protocol namespace rather than
+  inside the map itself.
+- **Closed nested structured objects must declare optional `ucp`
+  explicitly.** Closing a nested structured object does not exempt it from the
+  protocol grammar. A schema author defining one with
+  `additionalProperties: false` **MUST** declare an optional `ucp` property
+  referencing `ucp.json#/$defs/members`. This exception applies only to nested
+  structured objects; closing a dictionary does not create an ambient namespace
+  there.
```
raginpirate, line 246: jsonrpc.json is another example of why you might close an obj and not include the ucp property. 

With that in mind, is it better to just say that closed objects don't get ucp, rather than forcing it onto authors before closing? Will we ever make a mistake of closing an object purposefully without ucp in it?

docs/specification/overview.md
```diff
@@ -969,6 +969,184 @@ example:
...
+**Schema processing.** UCP source schemas are open by default, so ordinary
+validation against them may accept ambient `ucp` without applying the protocol
+vocabulary. For a selected message direction, a UCP-aware resolver **MUST**
+produce a resolved schema that recognizes and validates ambient `ucp` at every
+eligible structured scope against the central vocabulary in
+`ucp.json#/$defs/members`, subject to that vocabulary's applicability in the
+selected direction. The `ucp` namespace remains open to unrecognized members
+for forward compatibility, even if the resolved schema rejects other unknown
+domain fields. The result is ordinary JSON Schema that standard validators and
+code generators can consume.
```
raginpirate, line 1047: Just a note that I like the open-ness of this base; although you can't mix into the base ucp as a domain, you can borrow the base as its written today and mix it into other edges with the property 👍

raginpirate (approved):
Great fundamental base for us to build ontop of!

docs/specification/ap2-mandates.md
```diff
@@ -255,6 +255,10 @@ immediately over the same HTTP connection. Mandates are different:
 JCS ensures that semantically identical JSON produces byte-identical output,
 making signatures reproducible across implementations and time.
 
+Verification and mandate construction operate on the complete checkout JSON,
+not a projection of schema-recognized fields; removing any member covered by
+`merchant_authorization` changes the JCS payload and invalidates the signature.
```
gsmith85, line 260: In context of this PR, this comment entirely makes sense.  However, if you were looking at this directive in isolation it may not be obvious what you're alluding to.  I recognize you're trying to capture something more general, but you may want to specifically cite the protocol namespace as a specific example of members that should not be dropped between mandate construction and verification.

docs/documentation/schema-authoring.md
```diff
@@ -198,6 +198,59 @@ The same registry structure appears in three contexts with different field requi
...
+  **MUST NOT** declare a domain field named `ucp` in a structured object; the
+  name denotes the protocol namespace there.
+- **Do not declare ambient `ucp` in open nested objects.** The root `ucp`
+  envelope is not ambient placement; UCP root schemas already declare its
+  required protocol metadata. Below the root, the member is UCP document grammar
+  defined centrally, so schema authors **MUST NOT** declare it in an open nested
+  object. A Business or Platform **MAY** include it in an eligible open
+  structured scope without the domain schema saying so. Ordinary instance
+  validation against an open UCP source domain schema treats it as an ignored
+  unknown object; validating its contents is a conformance-tooling concern.
```
gsmith85, line 220: Not for this PR, but for the sake of implementors and the community, we should aim to bend `schema-authoring.md` to be (i) as obvious as possible and (ii) example driven.

So in this case something like:

```
- DO NOT declare ucp in open schemas; DO declare it if you close the schema.
  Because JSON Schema is open by default, ambient ucp metadata blocks are
  allowed on any structured object automatically. You only declare ucp in your
  .json file if you explicitly set `additionalProperties: false`.

❌ DON'T (Open Schema — default behavior):
{
  "type": "object",
  "properties": {
    "carrier_name": { "type": "string" }
    // DO NOT put "ucp" here! JSON Schema allows it automatically.
  }
}

✅ DO (Closed Schema — explicit restriction):
{
  "type": "object",
  "additionalProperties": false,
  "properties": {
    "carrier_name": { "type": "string" },
    "ucp": { "$ref": "https://.../ucp.json#/$defs/members" } // Required ONLY when closed
  }
}
```

More academic reasoning around why protocol level choices were made is rolled up into the specification itself.

gsmith85 (commented):
LGTM - This is an elegant solution, I'm really happy with where this landed.  The concepts are well explained in `overview.md` and overall I find the scheme to be intuitive.

docs/documentation/schema-authoring.md
igrigorik: jsonrpc is a different beast altogether, it's a service spec that doesn't participate in any of the extension mechanics. That aside, let's keep the flexibility to say "yes" to `ucp` in closed objects -- it's an explicit opt-in.

docs/specification/ap2-mandates.md
igrigorik: Fair, clarified 👍

igrigorik:
Thanks all for thorough reviews! 🙇🏻

PR #658, opened 2026-07-31
docs(split-payments): validate the allowed_combinations profile example

vishkaty:
### Problem

The only business-profile example for the split payments extension is annotated `<!-- ucp:example skip -->`, so CI never validates it, and it does not match the profile schema in three ways:

1. `capabilities` is an array, but `ucp.json` defines it as an object map keyed by reverse-domain name.
2. The capability entry omits the required `schema` URL (`capability.json` `business_schema` requires `schema`).
3. It pins `"version": "2026-01-23"`, a release that does not contain this extension.

Validating the fragment against the profile schema fails on the `capabilities` type and then on the missing `schema` property; it passes once rewritten as a map with a `schema` URL.

### Fix

Rewrite the example in the `ucp.capabilities` map form with `spec`, `schema`, `extends`, and `{{ ucp_version }}`, mirroring the buyer-consent profile example, and replace the `skip` annotation with a validated `schema=profile def=business_schema` annotation so CI covers it going forward.

`extends: ["dev.ucp.shopping.checkout"]` reflects `split_payments.json`'s `requires` on checkout.

### Verification

`scripts/validate_examples.py` full corpus goes from 343 to 344 passing (50 to 49 skipped): the example flips from skipped to validated. `ucp-schema lint source/` unaffected.

### DIFF
--- a/docs/specification/payment/extensions/split-payments.md
+++ b/docs/specification/payment/extensions/split-payments.md
@@ -107,13 +107,18 @@ if any valid assignment exists, the combination matches.
 A business that supports (a) a card with up to 2 redeemables, (b) up to 5
 gift cards alone, and (c) two credit cards:
 
-<!-- ucp:example skip reason="illustrative business profile fragment showing allowed_combinations shape" -->
+<!-- ucp:example schema=profile def=business_schema extract=$.capabilities target=$.ucp.capabilities -->
 ```json
 {
-  "capabilities": [{
+  "capabilities": {
     "dev.ucp.common.payment.split_payments": [
       {
-        "version": "2026-01-23",
+        "version": "{{ ucp_version }}",
+        "spec": "https://ucp.dev/{{ ucp_version }}/specification/split-payments",
+        "schema": "https://ucp.dev/{{ ucp_version }}/schemas/common/payment_split_payments.json",
+        "extends": [
+          "dev.ucp.shopping.checkout"
+        ],
         "config": {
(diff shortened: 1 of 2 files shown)


vishkaty:
A gentle note on this and its sibling docs and schema PRs (#659, #660, #661, #662, #669, #670, plus #625): all eight are rebased onto current main, conflict free, and validate_examples passes on each. They are independent of one another, so they can be reviewed in any order or batched, whatever is easiest. Happy to adjust any of them.

vishkaty:
Following up on this docs and schema batch (#658, #659, #660, #661, #662, #669, #670, #625): all eight are still green on current main, conflict free, and independent, so any subset can be merged in any order. If the batch size is the awkward part, I am glad to fold the docs-only ones into a single PR to cut the review overhead, whatever is easiest on your side. No urgency at all, just flagging they are ready whenever the queue allows. Thanks @damaz91.

vishkaty:
@jingyli you are a requested reviewer here, so flagging this quiet one. Small fix so the split payments profile example actually validates against the profile schema. Glad to adjust anything.

dkoch74 (approved):

dkoch74:
@knightlin-shopify I am supportive of fixing this example - could you please review?

vishkaty:
Updated the spec URL in the example to the post-reorganization path, specification/payment/extensions/split-payments, matching what the other extension profiles carry on current main. The rest of the example is unchanged.

PR #659, opened 2026-07-31
docs(order): sign Webhook-Id, Webhook-Timestamp, and UCP-Agent in the webhook signature

vishkaty:
### Problem

The order event webhook lists `Webhook-Id` and `Webhook-Timestamp` as required headers and states that headers follow [Standard Webhooks](https://www.standardwebhooks.com/), but the signing and verification steps never cover them, and the example `Signature-Input` signs only `@method`, `@authority`, `@path`, `content-digest`, and `content-type`.

Webhooks carry no `Idempotency-Key`, so `Webhook-Id` and `Webhook-Timestamp` are the only per-delivery replay signal. As written, a captured, validly signed delivery can be replayed with an altered `Webhook-Id` or `Webhook-Timestamp`, defeating any dedup or freshness check keyed on them.

### Fix

The overview verification gate is a closed set (`ucp-agent`, `signature-agent`, `idempotency-key`) with the clause "a header added to UCP later is gate-required only if its defining section says so." This PR has the order webhook section say so:

- `Webhook-Id` and `Webhook-Timestamp` MUST be covered components; platforms MUST verify their coverage and MUST deduplicate on `Webhook-Id`.
- Update the signing and verification steps and the example `Signature-Input` accordingly.

The example also now covers `ucp-agent`. The overview gate already requires it whenever the `UCP-Agent` header is present, and the example carried the header without covering it; adding it binds the signer identity on the webhook. This subsumes the webhook-example change proposed in #568, which is now closed unmerged. That PR additionally covered `ucp-agent` in the checkout REST binding request example (now `shopping/checkout/rest.md`), which this PR does not change.

It also drops the stale `event_id`/`created_time` from the example body and the playground payload builder (moved to headers in #342), and shows the required headers in the example.

### Verification

`validate_examples.py` and `ucp-schema lint` unaffected (the change touches an `http` example and prose, not validated JSON). Playground `eventId`/`eventTime` remain used elsewhere (no dead vars).

### DIFF
--- a/docs/specification/order.md
+++ b/docs/specification/order.md
@@ -482,6 +482,16 @@ to ensure authenticity and integrity. Signatures follow the
 | `Signature`      | Contains the signature value               |
 | `Content-Digest` | Body digest (RFC 9530)                     |
 
+`Webhook-Id` and `Webhook-Timestamp` (the required headers listed for the
+[Order Event Webhook](#order-event-webhook)) **MUST** be included as covered
+components in the signature, and platforms **MUST** verify their coverage.
+Webhooks carry no `Idempotency-Key`, so these headers are the only per-delivery
+replay signal, and a signature that omits them leaves an otherwise valid, signed
+delivery open to replay with an altered identity or freshness value. Platforms
+**MUST** reject deliveries whose signature does not cover both components, and
+**MUST** deduplicate on `Webhook-Id` so that a re-sent (or replayed) delivery is
+processed at most once.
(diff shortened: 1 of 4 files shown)


vishkaty:
@gsmith85 flagging as a requested reviewer. This signs `Webhook-Id`, `Webhook-Timestamp`, and `UCP-Agent` so the replay signal is covered by the signature. Happy to iterate.

PR #660, opened 2026-07-31
docs(consent): exempt buyer.consent from the cart and checkout full-replacement rules

vishkaty:
### Problem

The buyer-consent extension says omitting `consent` on a request "retains the prior position," but core Update Cart and Update Checkout are full replacements: the platform sends the entire resource and it replaces the stored session state. A business implementing the literal full-replacement contract drops captured consent on the next update that omits `buyer.consent` (for example a fulfillment change), re-exposing advertised defaults. A buyer who explicitly opted out can be reset toward an opted-in default, which is the failure mode a consent record most needs to prevent. Two conformant implementations diverge on the outcome.

### Fix

Consent rides both `cart.buyer.consent` and `checkout.buyer.consent`, so the exception is added to every place the full-replacement rule is stated:

- `buyer-consent.md`: state that consent persistence is an explicit exception to the cart and checkout full-replacement rules, cross-referencing both.
- `shopping/checkout/index.md` and `shopping/cart/index.md`: add the exception note to each full-replacement rule (the checkout document already carries the `complete_in_progress` exception).
- `shopping/checkout/rest.md`: the rule is restated **twice** here, and both now carry the carve-out — the general Update Checkout rule, and the buyer-specific **Update Buyer Info** paragraph, which draws the drop-if-omitted consequence as a `MUST` over the `buyer` object (and so is the most on-point site, since consent is `buyer.consent`).

The MCP and cart-rest bindings link back to the capability docs and do not restate the rule; the `embedded-*` docs mention it only in example code comments.

### Alternative

The other option is to require platforms to re-send the full consent map on every update once submitted. The exemption approach is proposed here as the smaller change; happy to switch if maintainers prefer the re-send contract.

### DIFF
--- a/docs/specification/buyer-consent.md
+++ b/docs/specification/buyer-consent.md
@@ -371,7 +371,16 @@ Handling](shopping/checkout/index.md#error-handling) flow.
 
 2. **Confirm semantics.** The `consent` field is optional on requests;
    omitting it provides no consent update and the business retains its prior
-   position. When submitting `consent`, platforms MUST include every advertised
+   position. This persistence is an explicit exception to the cart and checkout
+   full-replacement rules (see
+   [Update Cart](shopping/cart/index.md#update-cart) and
+   [Update Checkout](shopping/checkout/index.md#update-checkout)): a business
+   MUST retain the last submitted `buyer.consent` state across subsequent cart
+   or checkout updates that omit `consent`, and MUST NOT reset it to
+   advertised defaults, so that an update sent for an unrelated field (for
+   example a fulfillment change) does not silently discard captured consent.
+   When submitting `consent`,
+   platforms MUST include every advertised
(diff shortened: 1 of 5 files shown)


vishkaty:
@jingyli companion to #658. This keeps captured buyer consent from being dropped by the full replacement update rules. Glad to rework if you see a better shape.

vishkaty:
Re-anchored on released v2026-08-25: the branch already targets the new docs/specification/shopping paths and applies cleanly. The release sharpens the case rather than absorbing it, since buyer-consent.md L372-373 now asserts that the business retains its prior position while checkout/rest.md L242-245 states a MUST over the same buyer object that replaces everything not resent. The discount extension already carries per-field replacement semantics in its own doc (discount.md L134), which is the shape proposed here for consent.

PR #661, opened 2026-07-31
fix(policies): require non-empty applies_to and fix the set-match example and coverage wording

vishkaty:
### Problem

Three small `policies[]` corrections found while reading the merged policies capability:

1. **`applies_to: []` is undefined.** An empty array is schema-valid (no `minItems` on `policy.json`) but matches none of the three defined targeting forms (singular, set, omitted). Response-wide scope is expressed by omitting `applies_to`, so an empty array has no defined meaning.
2. **The Set-match example matches nothing.** `overview/index.md` filters on `$.products[?@.category=='electronics']`, but the product schema defines `categories` (array of category objects) and `tags` (array of strings), not `category`, so the canonical example selects the empty set against a schema-conformant product.
3. **The coverage definition is circular.** "the matched node's Normalized Path is a prefix of the target node's" is self-referential (the target matches a set, not a single node); it is the covered node whose path is prefixed.

### Fix

- `policy.json`: add `minItems: 1` to `applies_to` (omit the field for response-wide scope).
- `overview/index.md`: change the example filter to `$.products[?@.tags[?@=='electronics']]`, a field products actually carry.
- `overview/index.md`: reword the coverage sentence to "a prefix of that node's Normalized Path".

### Verification

No example in `docs/` or `source/` uses `"applies_to": []`, so `minItems: 1` breaks nothing (confirmed by grep and a full `validate_examples.py` run: 343 passing, unchanged, re-measured 2026-09-01 on a test merge into main `1d39948`). `ucp-schema lint source/` passes. The new JSONPath is valid RFC 9535.

### DIFF
--- a/docs/specification/overview/index.md
+++ b/docs/specification/overview/index.md
@@ -3018,7 +3018,7 @@ policy targets nodes in one of three forms:
   §2.3.5.1](https://www.rfc-editor.org/rfc/rfc9535#section-2.3.5.1)), e.g.
   `$.line_items[2]`.
 - **Set match** — a filter, wildcard, or slice matching a set of nodes, e.g.
-  `$.products[?@.category=='electronics']`.
+  `$.products[?@.tags[?@=='electronics']]`.
 - **Response-wide** — an omitted `applies_to`; the policy applies to the entire
   response. This is the common case: a single site-wide policy is one entry with
   no targeting, never repeated per item.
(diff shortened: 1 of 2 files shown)


vishkaty:
@jamesandersen flagging as a requested reviewer. Three small policies corrections, the only schema change is `minItems: 1` on `applies_to`. Happy to split or adjust.

Issue #665, opened 2026-07-31
Extension-defined Actions: empty actions map passes schema despite the MUST-omit rule; A2A binding missing the complete_in_progress freeze


vishkaty:
Two small gaps around the new extension-defined Actions primitive (#582).

### 1. An empty `actions` map passes schema despite the MUST-omit rule

The prose requires omitting the map when nothing is outstanding:

> the Business **MUST** include every outstanding Action and **MUST** omit `actions` when none are outstanding.
> (`docs/specification/overview.md`, Actions)

But `source/schemas/common/types/actions.json` places no `minProperties` on the map (only `minItems: 1` on the array values), so `"actions": {}` validates while violating the MUST. Adding `minProperties: 1` would let a schema-driven validator catch it. Happy to open a PR.

### 2. The A2A checkout binding is missing the `complete_in_progress` update-freeze

The freeze MUST is present in the core, REST, MCP, OpenAPI, and OpenRPC contracts (`checkout.md`, `checkout-rest.md`, `checkout-mcp.md`, and the service definitions), but `docs/specification/checkout-a2a.md` does not restate it; it has only a `messageId`-based Request Idempotency section and its own `complete_checkout` action verbs. Restating the freeze in the A2A binding would close the transport-parity gap so an implementer reading only that binding sees it.

---

Note: I initially also raised pre-runtime discovery, key-to-extension dispatch, and a core execution-integrity floor. On closer reading, the schema-composition and negotiation model ("Each declaring extension contributes its Action-type keys to the containing capability's schema through `allOf` composition ... The composed JSON Schema can validate ... each declared type's key and `config` shape"), Namespace Governance (an extension may declare only types within a namespace controlled by its schema authority), and the explicit "Trust and Execution Boundaries" scope-out already address those, so I have withdrawn them (see comment below).

vishkaty:
On re-reading, I have edited this down to the two observations that hold: the empty-`actions` map versus the MUST-omit rule, and the missing `complete_in_progress` freeze in the A2A binding.

[...]

PR #669, opened 2026-08-01
fix(schema): require the actions map to be non-empty

vishkaty:
### Problem

`docs/specification/overview/index.md` requires the Business to omit the `actions` map when nothing is outstanding:

> the Business **MUST** include every outstanding Action and **MUST** omit `actions` when none are outstanding.

But `source/schemas/common/types/actions.json` placed no `minProperties` on the map (only `minItems: 1` on the array values), so an empty `"actions": {}` validated while violating that MUST.

### Fix

Add `minProperties: 1` to the map. The map should be absent, not empty, when there is no outstanding work, and a schema-driven validator now catches the empty case.

### Verification

`ucp-schema lint source/` clean (124 files); `validate_examples.py` 343 passed (unchanged). Re-measured 2026-09-01 on a test merge into main `1d39948`. No example emits an empty actions map. Verified with a JSON Schema validator that `{}` is now rejected and a non-empty map still passes.

Raised from #665.

### DIFF
--- a/source/schemas/common/types/actions.json
+++ b/source/schemas/common/types/actions.json
@@ -26,6 +26,7 @@
     }
   },
   "type": "object",
+  "minProperties": 1,
   "propertyNames": {
     "$ref": "reverse_domain_name.json"
   },


vishkaty:
@sdedeo2025 flagging as a requested reviewer. One line: `minProperties: 1` so an empty `actions` map cannot validate against the MUST in overview.md. Happy to iterate.

PR #671, opened 2026-08-01
feat: define Fulfillment behavior for location context

igrigorik:
Builds on #642, which added the vertical-neutral `context.location` carrier to base Context and the **Location** glossary term. #642 deferred the field's effects to "the operation specification or an active capability/extension." This PR takes up that deferral for Fulfillment.

## What this defines

When the Fulfillment extension is active, `context.location` names the provisional Business Location that a Business evaluates fulfillment availability against and generates initial fulfillment choices for. It never selects a destination.

Four fields carry distinct roles rather than precision levels of one value:

| Field | Role |
| --- | --- |
| `context.location` | Provisional Business Location to evaluate fulfillment against |
| Catalog `filters.fulfills_to` | Explicit destination the order is fulfilled to, and a filter on results |
| Catalog `methods[].location` | The Business Location resolved for a place-based method |
| Checkout `methods[].selected_destination_id` | The destination selected from that method's `destinations[]` |

Precedence is scoped to what each field governs: `filters.fulfills_to` supersedes `context.address_country`, `context.address_region`, `context.postal_code`, and `context.location` **only** for fulfillment destination and `availability` resolution. Once a Platform sets `selected_destination_id` on a method, that selection governs that method's fulfillment scope.

Base Cart Context already carries `location`, and cart-to-checkout conversion initializes Checkout from the Cart's `context`, so a Catalog -> Cart -> Checkout journey preserves the identifier and an explicit Checkout selection then governs it.


jingyli:
Thanks @igrigorik! Adding a general thought for further discussion here:

While I agree that we should have a unified way of representing the explicit Location identifier when it comes to binding-contracts with respect to fulfillment (examples 2, 3, 4 in the table above), I'm wondering if grouping `context.location` (a provisional, non-binding hint shared on behalf of the user) here as well is too constraining, especially if we look at use cases outside of Shopping vertical that may materialize in the future?

An example:
- A user is looking to browse room availability or make a booking at a specific hotel branch.
- Business may require a location hint provided to establish the session (this ID may or may not come from the common Location capability we are looking to introduce in #589).
- However, lodging is an industry where there is **no concept of physical fulfillment**.
- We are then forced into picking one of the 2 options below:
    1. Adopt a consistent pattern where `context.location` is exposed and negotiated via a dummy "fulfillment" capability. (_con_: semantic irrelevance)
    2. Model its own bespoke `location_id` in context. (_con_: fragmentation)

jamesandersen:
Thanks @igrigorik — this answers the concern I raised on #642, where I'd pushed on `context.location` being an "identifier *or* name hint." The new `common/types/location_id.json` + "Location Identifier" glossary term make it a stable, well-defined ID (also capturing @gsmith85's first-class-concept suggestion). On the binding fields — method `location`, `fulfills_to`, `selected_destination_id` — this is a clear win.

On @jingyli's point about `context.location` being provisional and cross-vertical: a provisional location signal on base Context isn't actually novel. `shopping/types/context.json` already `allOf`s `common/types/locality.json`, so Context carries coarse `address_country`/`address_region`/`postal_code` today — framed as *"provisional buyer signals… not authoritative data."* The "base Context advertises nothing about who honors it" concern applies equally to `locality`, yet it works fine under the tolerant-reader contract — which suggests the objection is really about *binding* fields, not provisional ones.

So the narrow question: **does an opaque location ID add provisional value beyond the coarse `locality` Context already carries?** If yes (branch-level granularity, e.g. the hotel case), it may belong *alongside* `locality` on Context rather than scoped under Fulfillment — which is what makes it feel stranded for non-fulfillment verticals. If no, coarse `locality` already establishes the session and the branch-specific choice looks more like a binding selection. Worth noting the "non-binding" description sits in tension with the lodging example's Business that *requires* the hint — resolving that probably resolves the placement.

igrigorik:
@jamesandersen @jingyli rebased this PR to build on top of what we landed in #642, ptal. Should be a simple one 🤞🏻

docs/specification/fulfillment.md
```diff
@@ -53,6 +53,49 @@ On Checkout, the `fulfillment` field contains:
...
+to that capability names the provisional Business Location that the Business
+evaluates fulfillment availability against, and generates initial fulfillment
+choices for. It does not select a fulfillment destination.
+
+These fields carry distinct roles, not precision levels of one value:
+
+| Field | Role |
+| --- | --- |
+| `context.location` | Provisional Business Location to evaluate fulfillment against. |
+| Catalog `filters.fulfills_to` | Explicit destination the order is fulfilled to, and a filter on results. |
```
jingyli: nit: Should this be `Explicit destination the product or variant is fulfilled to` instead of `order`? There is no concept of `order` during catalog operations.

source/schemas/shopping/fulfillment.json
```diff
@@ -141,7 +141,7 @@
           "properties": {
             "fulfills_to": {
               "$ref": "types/fulfillment_destination_filter.json",
-              "description": "Where the order is fulfilled to — may differ from the buyer's `context` location (e.g. a gift). A `location` id, or a coarse address (`address_country`/`address_region`/`postal_code`). Restricts results to what can be fulfilled there and seeds method `availability`. Supersedes `context`."
+              "description": "Explicit destination where the order is fulfilled. It may differ from the locality or Business Location supplied in `context` (e.g. for a gift) and can be specified by a `location` identifier or a coarse locality (`address_country`/`address_region`/`postal_code`). The filter restricts results to what can be fulfilled there and seeds method `availability`. It supersedes `context.address_country`, `context.address_region`, `context.postal_code`, and `context.location` only for fulfillment destination and availability resolution."
```
jingyli: > Explicit destination where the order is fulfilled.

Similar to the comment above - referencing order here feels a bit odd given there shouldn't any concept of it available during catalog search time..

>(e.g. for a gift)

nit: `(e.g. shopping for a gift that should be delivered directly to the recipient)`?

> It supersedes `context.address_country`, `context.address_region`, `context.postal_code`, and `context.location`

Instead of listing out all the fields explicitly here, do we feel keeping it as generic as "Supersedes respective fields in `context` only for fulfillment destination..." is better in case new fields are added?

jingyli (approved):
Thanks @igrigorik for updating the prose here ! Update LGTM (with 2 nit on some wording choices).

igrigorik:
@jingyli ty, good suggestions, addressed in https://github.com/Universal-Commerce-Protocol/ucp/pull/671/changes/94609eef9b7a0d10ee8ea5d065dbe206ec918af1.

jamesandersen (approved):

amithanda (approved):

PR #687, opened 2026-08-05
fix!: define deterministic operating hours for Location service

igrigorik:
The current Location PR (https://github.com/Universal-Commerce-Protocol/ucp/pull/687) introduces weekly and exceptional operating hours,
   but leaves several wire and evaluation semantics ambiguous. In particular,
   closures rely on an artificial midnight interval, `open_now` depends on an
   implicit server clock, exception date bounds are unclear, and the specification
   does not define timezone, overnight, DST, overlap, or precedence behavior.

   Close those gaps with a UCP-native schedule model informed by Schema.org's
   OpeningHoursSpecification:
   https://schema.org/OpeningHoursSpecification

   Schema.org is design input only. UCP owns the field names, values, and
   evaluation rules defined here.

   Make weekly intervals explicit and reusable:

       "hours": [
         {
           "day": "tuesday",
           "opens": "09:00",
           "closes": "12:00"
         },
         {
           "day": "tuesday",
           "opens": "13:00",
           "closes": "21:00"
         }
       ]

   Rename `open` and `close` to `opens` and `closes`, and define `day` as a
   stable UCP weekday identifier rather than localized display text. Multiple
   entries for one day represent split shifts, and an interval whose closing time
   is earlier than its opening time continues into the next local date.

   Refactor the shared time interval schema so `opens` and `closes` are an
   optional but inseparable pair. Weekly hours require both fields, while
   exception hours may omit both to represent a full closure. Reject the ambiguous
   `00:00` to `00:00` pair and reserve `00:00` to `23:59` as the full-local-day
   sentinel.

   Replace the previous exception shape:

       {
         "from": "2026-11-26",
         "through": "2026-11-27",
         "label": "Thanksgiving",
         "open": "00:00",
         "close": "00:00"
       }

   with inclusive local-date bounds and an actual closure representation:

       {
         "title": "Thanksgiving",
         "valid_from": "2026-11-26",
         "valid_through": "2026-11-26"
       }

   Rename `from`, `through`, and `label` to `valid_from`, `valid_through`, and
   `title`. Treat `title` as optional presentation metadata that does not affect
   schedule evaluation. Allow timed exceptions with paired `opens` and `closes`,
   including multiple entries with identical bounds for split shifts.

   Define every returned schedule in the Location's Business-owned IANA timezone.
   Require `timezone` whenever regular or exception hours are present, and keep
   the canonical schedule independent of the requesting Platform or Buyer's
   timezone.

   Specify deterministic evaluation:

   - convert an exact instant into each Location's local date, weekday, and time
   - use half-open timed intervals, except for the reserved full-day sentinel
   - let overnight intervals carry into the following local date
   - replace regular hours with exception hours at local midnight
   - treat omitted weekdays as having no interval starting that day
   - treat absent schedules as unknown rather than closed
   - evaluate DST gaps and folds pointwise without shifting nonexistent times
   - reject equal time pairs and intersecting non-identical exception ranges as Business conformance errors where JSON Schema cannot express the constraint

   Remove the redundant `open_now` filter. It makes results depend on an implicit
   processing clock and creates undefined precedence when combined with
   `open_at`. Require one caller-supplied RFC 3339 instant instead:

       "filters": {
         "hours": {
           "open_at": "2026-05-18T17:00:00Z"
         }
       }

   Require `open_at` to include `Z` or a numeric offset. The offset identifies the
   instant only; the Business still evaluates that instant using each candidate
   Location's authoritative IANA timezone. Keep the nested hours filter open so
   extensions can add qualifiers without changing the standard predicate.

   Move complete Search and Lookup examples into the transport-neutral capability
   documents. Cover hours with serviceability and amenities, inventory with
   distance, split shifts, full closures, and partial Lookup success there.

   Reduce REST and MCP examples to equivalent binding envelopes that link to the
   same canonical payload examples. This keeps both transports on equal footing,
   avoids duplicating domain semantics, and prevents one binding's examples from
   becoming more complete or authoritative than the other. Preserve MCP's
   required `meta["ucp-agent"].profile` contract while separating protocol
   metadata from the Location request.

   This is a breaking correction to the Location PR's draft wire shape:

   - `open` becomes `opens`
   - `close` becomes `closes`
   - `from` becomes `valid_from`
   - `through` becomes `valid_through`
   - `label` becomes `title`
   - `open_now` is removed
   - full closures omit both time fields instead of using `00:00` to `00:00`

### DIFF
--- a/docs/specification/location/index.md
+++ b/docs/specification/location/index.md
@@ -95,9 +95,10 @@ interpreted using its IANA timezone.
 
 ### Evaluation
 
-Timed intervals ordinarily include the opening time and exclude the closing
-time. If `closes` is earlier than `opens`, the interval continues into the next
-local date. As a reserved exception to those half-open semantics, the exact
+Timed intervals are open at `opens` and closed at `closes`. For example,
+`10:00`–`17:00` is open immediately before `17:00` and closed at `17:00`. If
+`closes` is earlier than `opens`, the interval continues into the next local
+date. As a reserved exception to this closing-boundary rule, the exact
 `00:00`–`23:59` pair represents the entire local civil date, including daylight
 saving time transitions. Every pair with equal `opens` and `closes` is invalid,
 including `00:00`–`00:00`.
@@ -169,8 +170,10 @@ to the requesting Platform's or Buyer's timezone. Because JSON Schema does not
 enforce every semantic constraint, a Business **MUST** emit only schedules that
 follow the rules above, including unequal `opens` and `closes`, a `valid_from`
 value no later than `valid_through`, and valid exception-range intersections.
-When a request provides a language preference, a Business **SHOULD** localize
-`title` accordingly, unless that language is unavailable.
+A Business **SHOULD** omit an `exception_hours` entry once it can no longer
+affect the schedule at any current or future instant. It **SHOULD** publish
+known future exceptions through the planning horizon for which its schedule is
+authoritative.
 
 #### Platform
 
--- a/docs/specification/location/index.md
+++ b/docs/specification/location/index.md
@@ -77,9 +77,8 @@ other capabilities (like Catalog, Cart, and Checkout in Shopping):
 ### Representation
 
 The hours filter's [`open_at`](search.md#hours-based-filter) value is an exact
-instant. By contrast, `day`, `opens`, `closes`, `valid_from`, and
-`valid_through` are canonical local civil values interpreted in the containing
-Location's `timezone`; they are not timestamps.
+instant. Operating hours use the Location's local date and clock time,
+interpreted using its IANA timezone.
 
 * `hours` is a list of regular weekly intervals. Each item contains `day`,
     `opens`, and `closes`. `day` is a stable UCP day-of-week identifier for the
@@ -114,7 +113,9 @@ current schedule shape cannot distinguish the two fold occurrences.
 Multiple `hours` items for the same `day` combine as split shifts. An omitted
 day means no regular interval begins that day, but an interval from the
 preceding day can carry into it. Absent `hours` means the regular schedule is
-unknown, not closed.
+unknown, not closed. To represent a temporary full closure, a Business retains
+its regular `hours` and adds an `exception_hours` entry for the affected dates
+without `opens` or `closes`.
 
 An exception schedule replaces the regular schedule on every covered local
 date. If an interval carries into a local date governed by an exception, that
--- a/docs/specification/location/index.md
+++ b/docs/specification/location/index.md
@@ -47,8 +47,9 @@ This is vertical-agnostic and enables key commerce flows such as:
     This is used to determine if a location can serve a specific user (e.g., delivery area check).
     Clients can perform proximity searches (`distance` filter) or coverage checks (`geofence_point` filter)
     using the filter.
-* **Operating Hours**: Weekly schedules (`hours`) and date-specific overrides
-    (`exception_hours` - e.g., holidays, temporary closures) associated with a timezone.
+* **Operating Hours**: Regular weekly schedules (`hours`) and date-specific
+    exceptions (`exception_hours`), interpreted in the Location's `timezone`.
+    See [Operating Hours](#operating-hours).
 
 ### Relationship to Other Capabilities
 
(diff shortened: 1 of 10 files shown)


docs/specification/location/lookup.md
```diff
@@ -70,6 +72,121 @@ Filters apply **after** identifier resolution. For example, if a client requests
...
+            {"day": "tuesday", "opens": "13:00", "closes": "21:00"},
+            {"day": "wednesday", "opens": "09:00", "closes": "21:00"},
+            {"day": "thursday", "opens": "09:00", "closes": "21:00"},
+            {"day": "friday", "opens": "09:00", "closes": "22:00"},
+            {"day": "saturday", "opens": "10:00", "closes": "20:00"}
+          ],
+          "exception_hours": [
+            {
+              "title": "Thanksgiving",
+              "valid_from": "2026-11-26",
```
karangoel16, line 132: should we add some kind of description as well , with the title for the exception hours ? 

"business might close early due to Thanksgiving or something like that " , 

at times description could be useful

docs/specification/location/lookup.md
```diff
@@ -70,6 +72,121 @@ Filters apply **after** identifier resolution. For example, if a client requests
...
+    ```
+
+=== "Response"
+
+    <!-- ucp:example schema=common/location_lookup op=lookup direction=response -->
+    ```json
+    {
+      "ucp": {
+        "version": "{{ ucp_version }}",
+        "capabilities": {
```
karangoel16, line 97: should we also add in expiry_at after this platform or agent should pull new info, this could help agent understand that the location info is stale and they need to pull new information ?

docs/specification/location/search.md
```diff
@@ -70,12 +70,23 @@ custom filters via `additionalProperties`.
...
-
-* `open_now`: A quick boolean filter to find locations currently open.
-* `open_at`: An RFC 3339 date-time string to find locations open at a
-    specific future time (e.g., planning a visit or ordering ahead).
-    The business resolves this against the location's local time and timezone.
+The standard `filters.hours` object requires `open_at`, an exact RFC 3339
+instant expressed with `Z` or a numeric offset, and can include
+extension-defined fields. Before sending a Buyer-local wall time, a Platform
+**MUST** resolve it to an exact instant for `open_at`. When asking which
+Locations are open now, a Platform **MUST** supply the current instant. No
```
jingyli: Just a naive thought: but I feel like replacing `open_now` boolean filter with this approach may not be ideal in the sense that `current instant` is extremely difficult to enforce in practice.

Platforms can issue the request at `Z` instant, but because there is always some intrinsic delay before business receives the request at `Z'` instant, it always look like platform has issued a search request with an instant in the past. I'd imagine there would be other legitimate reasons for businesses wanting to implement some logic to invalidate platform requests searching for stores that are "opened in the past", if this case is mixed in there, it introduces a rather grey area.

Instead, if we separate out the use case with `open_now`, then the boolean can clearly indicate that business should use `Z'` when processing the search request and can also easily invalidate all requests where `open_at` < `Z'`.

docs/specification/location/index.md
```diff
@@ -71,6 +72,115 @@ other capabilities (like Catalog, Cart, and Checkout in Shopping):
...
+
+### Representation
+
+The hours filter's [`open_at`](search.md#hours-based-filter) value is an exact
+instant. By contrast, `day`, `opens`, `closes`, `valid_from`, and
+`valid_through` are canonical local civil values interpreted in the containing
+Location's `timezone`; they are not timestamps.
+
+* `hours` is a list of regular weekly intervals. Each item contains `day`,
+    `opens`, and `closes`. `day` is a stable UCP weekday identifier for the day
```
jingyli: nit: Would recommend removing this `weekday` word from the sentence to factor in weekends as well.

docs/specification/location/index.md
```diff
@@ -71,6 +72,115 @@ other capabilities (like Catalog, Cart, and Checkout in Shopping):
...
+    `valid_through` in `YYYY-MM-DD` form; equal bounds select one date. Both
+    `opens` and `closes` define a timed interval, while omitting both defines a
+    full closure. The optional `title` is a short, human-readable heading. It is
+    presentation metadata and does not affect schedule evaluation.
+* `timezone` identifies the Business-owned canonical local civil-time frame for
+    both schedules.
+
+### Evaluation
+
+Timed intervals ordinarily include the opening time and exclude the closing
```
jingyli: Given this is documentation for location, I'm wondering if we should call out explicitly that the intervals used here are actually inclusive of both ends instead (also consistent of how we plan to represent the exception valid date ranges) of this standard half open representation?

The current way will result in a very unnatural way of representation opening hours. Example: a store that opens from 10am to 5pm. Under this rule, businesses would need to return `[10:00, 17:01)` vs. the more humanly understandable format of `[10:00, 17:00]`.

docs/specification/location/index.md
```diff
@@ -71,6 +72,115 @@ other capabilities (like Catalog, Cart, and Checkout in Shopping):
...
+
+Schedule evaluation is deterministic for each instant: convert the instant to
+the Location's local date, weekday, and time using `timezone`, then apply the
+effective schedule to those local values. During a forward daylight saving time
+(DST) transition, local clock labels in the gap correspond to no instants and
+are not shifted. During a backward DST transition, both instants in the fold
+that map to the same repeated local time receive the same schedule result. The
+current schedule shape cannot distinguish the two fold occurrences.
+
+Multiple `hours` items for the same `day` combine as split shifts. An omitted
```
jingyli, line 114: Just want to clarify if I understood these 2 sentences correctly:
- An omitted entry in `hours` (i.e. missing an object for `day: friday`) means the location is closed on the day.
- If the entire `hours` object is missing from the response, then the regular schedule is unknown.
- For a location under temporary closure, business would need to return the location result with no `hours` and an `exception_hours` rule set.

If the above is not correct, then how do we plan to represent communicating closed days to platforms?

docs/specification/location/index.md
```diff
@@ -71,6 +72,115 @@ other capabilities (like Catalog, Cart, and Checkout in Shopping):
...
+}
+```
+
+The two entries share date bounds, so they define split shifts that apply
+independently on each date in the inclusive range. A full closure instead uses
+one item that omits both `opens` and `closes`.
+
+### Guidelines
+
+#### Business
```
jingyli, line 161: Given we are providing some definition here, should we also add a guideline mentioning business **MUST** emit only currently "active" exception schedules in `exception_hours`?

source/schemas/common/types/location.json
```diff
@@ -25,27 +25,26 @@
         },
         "hours": {
           "type": "array",
-          "description": "Regular weekly operating hours. Support split hours through multiple entries on the same day and overnight hours (open > close). Omission of days implies full day closures.",
```
jingyli, line 28: Should we keep this sentence (`Omission of days implies full day closures.`) to make it mirror the documentation and make it a bit easy for implementors who are reading the schema fields only?

source/schemas/common/types/time_interval.json
```diff
@@ -2,19 +2,31 @@
...
-      "description": "End time (e.g., '18:00')."
+      "description": "Local civil closing time in 24-hour HH:MM format, interpreted in the containing Location's `timezone`.",
+      "ucp_request": "omit"
     }
+  },
+  "dependentRequired": {
+    "opens": ["closes"],
+    "closes": ["opens"]
+  },
+  "not": {
```
jingyli: I'm wondering if this block should be left out of the json schema given its current representation is incomplete (we forbid all location time interval ranges ranges where `opens` = `closes`, not just the `00:00` scenario).

source/schemas/common/types/time_interval.json
```diff
@@ -2,19 +2,31 @@
...
+  "description": "Reusable local civil-time fields for a containing schedule schema. Containing schemas determine whether the `opens` and `closes` pair is required; this fragment's standalone `{}` is not an interval.",
   "type": "object",
-  "required": ["open", "close"],
   "properties": {
-    "open": {
+    "opens": {
       "type": "string",
       "pattern": "^([01][0-9]|2[0-3]):[0-5][0-9]$",
-      "description": "Start time (e.g., '09:00')."
+      "description": "Local civil opening time in 24-hour HH:MM format, interpreted in the containing Location's `timezone`.",
```
jingyli: I'm wondering if these descriptions should be more generic and not tied specifically to `Location` if the goal is to define it as a generically reusable type (as hinted in the overall schema description)? i.e. dropping the `interpreted in the containing...` sub-sentence would be good?

jingyli (commented):
Thanks @igrigorik for the enhancement on hours representation! Overall the proposed shape LGTM but added a couple of inline comments (would especially love to get some reaction on the request filter front).

docs/specification/location/lookup.md
igrigorik: I’d keep description out for now: Schema.org’s OpeningHoursSpecification expresses exceptions through opens, closes, validFrom, and validThrough, while omission of opens means closed. UCP’s optional title already names the exception, and the time fields communicate its actual effect.

docs/specification/location/lookup.md
igrigorik: valid_from to valid_through effectively give you that; platform can refresh past those intervals.

docs/specification/location/index.md
igrigorik: I don't think half-open is right contract here. 10:00–17:00 is open immediately before 17:00 and closed at 17:00, matching the ordinary meaning of "closes at 5" without requiring 17:01. A value of 17:02 means the Location closes exactly at 17:02; it is not rounded to 17:00.

source/schemas/common/types/time_interval.json
igrigorik: removed this, ptal at updated commit.

docs/specification/location/index.md
igrigorik: I think "active" would add more confusion: active meaning right now, or active as known/planned. I think current infrastructure already gives us a clean and clear contract: the date bounds already define when each entry applies, while the Business can publish known exception ranges within its relevant planning horizon.

docs/specification/location/search.md
igrigorik: Clarified in https://github.com/Universal-Commerce-Protocol/ucp/pull/687/changes/c23dc7ee28063ae3af46cc441701782400c88c1c, ptal.  

`open_at` identifies the instant relevant to the buyer—such as expected arrival or pickup time—rather than making a freshness claim tied to when the Business receives it. We don't need to chase false precision and request/response timing here.

docs/specification/location/index.md
igrigorik: Clarified in [c23dc7e](https://github.com/Universal-Commerce-Protocol/ucp/commit/c23dc7ee28063ae3af46cc441701782400c88c1c). 

An omitted day means no regular interval begins that day, although an overnight interval from the preceding day may carry into it; omitting hours entirely means the regular schedule is unknown. A temporary closure retains the regular hours and adds an exception_hours entry without opens or closes for the affected dates.

source/schemas/common/types/location.json
igrigorik: clarified in [c23dc7e](https://github.com/Universal-Commerce-Protocol/ucp/pull/687/commits/c23dc7ee28063ae3af46cc441701782400c88c1c), ptal.

docs/specification/location/index.md
jingyli: I actually think we are agreeing on the same thing conceptually 😃 but the wording of the sentence here remains the issue...

`exclude the closing time` seems to suggest that the closing time (value in `closes`) should be discarded and naively reads more like a [10:00, 17:01) and there is no further clarification sentence in this paragraph talking about why for location specifically we are leveraging a closed interval - the only exception builds on top of the half-open semantics for entire date.

Can I suggest rewording this sentence instead to call out that time intervals in Location capability deviates from the ordinary half-open timed intervals and instead enforces a closed interval where opening and closing times are both included?

docs/specification/location/index.md
jingyli: My only worry (from the platform interpretation angle) is that a business _may_ choose to include all of its exception hours, including ones where it is very clear that it has expired (i.e. Thanksgiving in 2025), and it may end up making this object extremely bloated.. Having some guideline to limit `exception_hours` to only ones that have not expired will help in that regard. 

But given it's an edge case to a specific business interpretation, we can tackle it through other ways.

docs/specification/location/search.md
jingyli: Thanks! The clarification works well for the `scheduled`/future-looking intent but would love to still dig a bit deeper into this thread as I personally think the correctness/precision remains important when it comes to hyper local goods that need an `immediate` representation that doesn't depend on the platform's clock. Related, this data contract will require 2 additional translations on both platform & business sides to express `immediate` - so it complicates implementation and sacrifices precision..

The things I feel are still unresolved:
- If the business receives a platform request where `open_at` is in the past relative to the business's clock, it must decide whether to allow it or reject it now through leveraging a "tolerance window".
- If platform's clock is misconfigured or agent hallucinates when it attempts to translate a user's "immediate" intention to an absolute local instant, then the response will be incorrect.

jingyli (approved):

jingyli:
Thanks @igrigorik for the latest update! All my outstanding concerns are now addressed through [8bb2b74](https://github.com/Universal-Commerce-Protocol/ucp/pull/687/commits/8bb2b74ec83e49143b0c373f0000b02465456ebc). Proceeding with the merge back into the feature branch.

PR #688, opened 2026-08-05
fix!: make destination types explicit

igrigorik:
This PR makes Fulfillment destination responses self-describing while preserving existing Platform shipping-address requests.

Destination `type` is required and open in Business responses. In Platform requests, the enclosing method contract defines destination authorship and any default shape:

- `shipping` accepts Platform-written Shipping Destinations; omitted destination `type` defaults to `shipping_address`;
- `pickup` destinations are Business-authored and response-only; the Platform selects one through `selected_destination_id`;
- other method types, including `curbside`, use the base default: request `destinations[]` is omitted, the Business enumerates typed destinations in its response, and the Platform selects one through `selected_destination_id`. A defining contract **MAY** narrow the response shape or opt into a Platform-writable request shape.

The PR also introduces a bounded `Location Summary` that Checkout can render independently and #589 can extend with richer Location fields.

### Context

Fulfillment originally used an untagged `oneOf` between Shipping Destination and Retail Location. The enclosing method suggested the intended shape, but every destination was validated against both branches.

The branches overlap: Shipping Destination is open and requires only `id` in responses, so every Retail Location also matches it and strict `oneOf` rejects intended pickup destinations. The old request shape also mixed authority: `selected_destination_id` could select one Location while Platform-written `destinations[]` asserted another name/address, with no precedence rule.

#507 opened the method vocabulary but left this debt unchanged. #589 exposes it again and adds a capability-boundary issue: Checkout must not inherit the full Location Search/Lookup entity and its independently evolving geo, hours, amenities, and service-area fields.

The directional contract is:

| Direction/method | Destination behavior |
| --- | --- |
| Platform request, `shipping` | Platform writes Shipping Destination facts; omitted type defaults to `shipping_address`. |
| Platform request, `pickup` | Platform submits `selected_destination_id`; destinations are response-only. |
| Business response | Every destination includes required, open `type`. |
| Other method type | Request destinations are omitted by default. The Business enumerates typed response destinations and the Platform selects one through `selected_destination_id`; a defining contract may opt into a Platform-writable request shape. |

#### Shipping request — unchanged

```json
{
  "type": "shipping",
  "line_item_ids": ["line_1"],
  "destinations": [
    {
      "street_address": "123 Main St",
      "address_country": "US"
    }
  ]
}
```

A Shipping Destination can contain the address itself or an `id` that references a saved address. The Business can resolve that `id` from its own records or through a trusted Credential Provider, such as a digital wallet or identity provider.

#### Shipping response — typed

```json
{
  "id": "method_shipping",
  "type": "shipping",
  "line_item_ids": ["line_1"],
  "destinations": [
    {
      "type": "shipping_address",
      "id": "address_1",
      "street_address": "123 Main St",
      "address_country": "US"
    }
  ]
}
```

#### Pickup request — scalar selection

```json
{
  "type": "pickup",
  "line_item_ids": ["line_1"],
  "selected_destination_id": "loc_downtown"
}
```

#### Pickup response — typed Location summary

```json
{
  "id": "method_pickup",
  "type": "pickup",
  "line_item_ids": ["line_1"],
  "selected_destination_id": "loc_downtown",
  "destinations": [
    {
      "type": "business_location",
      "id": "loc_downtown",
      "name": "Downtown Store",
      "address": {
        "street_address": "500 Market St",
        "address_country": "US"
      }
    }
  ]
}
```

### What changes with this PR

- **Moves** request destination authorship to the Fulfillment Method contract.
- **Keeps** existing untyped Platform shipping-address requests valid.
- **Requires** open destination `type` in Business responses.
- **Makes** pickup destinations response-only; selection uses `selected_destination_id`.
- **Requires** method `type` when a request writes `destinations[]`; selection-only updates may target the method by `id`.
- **Adds** bounded `Location Summary` and Business Location Destination contracts.
- **Generalizes** Retail Location to Business Location Destination; retail stores remain supported alongside lockers and partner pickup points.

Destination selection is source-agnostic. The Platform may obtain an ID from Catalog, Location Search or Lookup, an earlier Checkout, or any other source the Business recognizes for that method. When the Business accepts the selection, it returns the same `selected_destination_id` and exactly one matching typed destination. The Business revalidates current availability and terms and does not silently substitute another destination; a rejected Update leaves the Checkout unchanged and returns a recoverable Message.

This is the Fulfillment prerequisite for #589. After it lands, #589 can compose its rich Location entity from `location_summary.json`, remove its direct Fulfillment rewrite, and keep discovery fields owned by Location Search/Lookup.

## Breaking changes

- Business response producers must add destination `type`.
- Retail Location responses migrate to `type: "business_location"`.
- Platforms that wrote pickup destination objects must select through `selected_destination_id`.
- Platforms that wrote `destinations[]` for method types other than `shipping` must omit them unless the method's defining contract explicitly opts into a Platform-writable request shape.
- `retail_location.json` is retired; retail stores remain supported through Business Location Destination.

Existing untyped Platform Shipping Destination requests remain valid.

---

## Checklist

- [x] **Capability**: breaking Fulfillment schema update
- [x] **Documentation**: directional request/response contract and examples
- [x] Documentation, schema lint, examples, and validator tests pass locally.


richmolj (approved):

source/schemas/shopping/types/shipping_destination.json
```diff
@@ -16,10 +16,17 @@
           "type": "string",
           "description": "ID specific to this shipping destination.",
           "ucp_request": "optional"
+        },
+        "type": {
+          "type": "string",
+          "const": "shipping_address",
+          "description": "Destination type discriminator. Required in responses; optional in requests.",
```
gsmith85: Nit: Let `"ucp_request": "optional"` speak for itself and drop the redundant comment.

source/schemas/shopping/types/fulfillment_method.json
```diff
@@ -4,38 +4,115 @@
...
+      }
     },
     "type": {
       "type": "string",
       "description": "Fulfillment method type. Well-known values: `shipping`, `pickup`. Businesses MAY use additional values.",
-      "ucp_request": {"create": "required", "update": "optional"}
+      "ucp_request": {
+        "create": "required",
+        "update": "optional"
+      }
```
gsmith85, line 27: I have some concerns regarding using nested `if`/`then` conditional blocks inside `allOf` in `fulfillment_method.json` (and `fulfillment_destination.json`) from both schema maintainability and DevRel/SDK tooling perspective:

1. **Schema Readability & Maintainability:** Nested conditional logic inside `allOf` makes the JSON Schema significantly harder for humans to read, trace, and reason about when reviewing or extending protocol contracts.
2. **SDK Code Generation & DX:** Most standard SDK generators (`datamodel-code-generator` for Python/Pydantic, `quicktype` for TypeScript/Go/C#, `openapi-generator`, etc.) do not natively parse or resolve dynamic JSON Schema `if`/`then` conditionals. As a result, code generation tools either drop these branches, emit untyped `Any` / `unknown` arrays, or fail to produce clean discriminated unions.

source/schemas/shopping/types/fulfillment_method.json
```diff
@@ -4,38 +4,115 @@
...
+      "items": {
+        "$ref": "fulfillment_group.json"
+      }
     }
+  },
+  "allOf": [
+    {
+      "if": {
+        "properties": {
+          "type": {
```
gsmith85, line 67: I believe the intent here is to (i) avoid limiting the value of `type` (ii) while still adding the **correctly typed** `destinations` array either `shipping_destination[]` or `location_destination[]`)  in recognized cases?

In the spirit of avoiding conditionals in our JSON Schema and keeping it simple(r) for code generation, I think that we can achieve the same result using an object-oriented approach.

  **fulfillment_method.json (Open Base Object)**

  Defines common fields shared across all fulfillment methods without imposing destination rules:

```json
{
  "$id": "https://ucp.dev/schemas/shopping/types/fulfillment_method.json",
  "title": "Fulfillment Method",
  "type": "object",
  "ucp_shared_request": true,
  "required": ["id", "type", "line_item_ids"],
  "properties": {
    "id": { "type": "string", "ucp_request": {"create": "omit", "update": "optional"} },
    "type": {
      "type": "string",
      "description": "Fulfillment method type (e.g. shipping, pickup). Open vocabulary."
    },
    "line_item_ids": {
      "type": "array",
      "items": { "type": "string" },
      "ucp_request": {"create": "optional", "update": "required"}
    },
    "selected_destination_id": { "type": ["string", "null"] }
  }
}
```

 **shipping_fulfillment_method.json (Specialized Subtype)**

  Extends fulfillment_method.json via allOf, constrains "type": "shipping", and adds Platform-writable shipping destinations:
```json
{
  "$id": "https://ucp.dev/schemas/shopping/types/shipping_fulfillment_method.json",
  "title": "Shipping Fulfillment Method",
  "type": "object",
  "ucp_shared_request": true,
  "allOf": [
    { "$ref": "fulfillment_method.json" },
    {
      "type": "object",
      "required": ["type"],
      "properties": {
        "type": { "type": "string", "const": "shipping" },
        "destinations": {
          "type": "array",
          "description": "Platform-authored shipping addresses.",
          "ucp_request": "optional",
          "items": { "$ref": "shipping_destination.json" }
        }
      }
    }
  ]
}
```

  **pickup_fulfillment_method.json (Specialized Subtype)**

```json
{
  "$id": "https://ucp.dev/schemas/shopping/types/pickup_fulfillment_method.json",
  "title": "Pickup Fulfillment Method",
  "type": "object",
  "ucp_shared_request": true,
  "allOf": [
    { "$ref": "fulfillment_method.json" },
    {
      "type": "object",
      "required": ["type"],
      "properties": {
        "type": { "type": "string", "const": "pickup" },
        "destinations": {
          "type": "array",
          "description": "Business-authored business locations. Response-only.",
          "ucp_request": "omit",
          "items": { "$ref": "location_destination.json" }
        }
      }
    }
  ]
}
```

A bit of an aside about polymorphism and this pattern: most of our SDK generators are going to be fine serializing from `subtype -> JSON` but will need custom code (cued off of `type`) to handle deserialization to `JSON -> subtype`.  We are already doing this for `fulfillment_method`'s current approach, so we'll need to make minor modifications to support this change.

igrigorik:
@gsmith85 great catch on the SDKs. Spent some time working through options...

Open strings are a load-bearing UCP design principle, not an accident of this schema. We use open `type` vocabularies for fulfillment methods and destinations, totals and adjustments, media and links, identity providers, payment instruments and credentials, policies, and other concepts. This lets future UCP versions and negotiated extensions add values without making older generic consumers reject otherwise usable payloads.

Fulfillment is one of the cases where recognized values also select additional constraints. JSON Schema's `if`/`then` branches express that open-world contract correctly: recognized values get their complete shape, while unknown values remain valid against the open base. The gotcha is that common SDK generators take a shortcut and emit only the structural base. That output is useful, but its validation is incomplete because it does not evaluate the known type branches.

The named subtype pattern you proposed helps readability and class generation, but it does not solve dynamic dispatch. Something must still select those classes when decoding `methods[]`. Closing the active union would sacrifice the open vocabulary, while a generic `{type: string}` fallback overlaps the known branches. Either schema organization therefore requires the SDKs to implement known-first dispatch: recognized values decode strictly as their known subtype and fail there, while only unrecognized values may reach Unknown.

Asked my agent to prototype known-first decoding against both official SDK stacks and recap its findings. It works against current schema proposed in this PR, but yes it does require UCP-specialization. Implementation notes and test matrix are here: https://gist.github.com/igrigorik/29eb5dbc09e7c43a2fc25bda281c8af0

My proposal is:

1. Keep the conditional branches in this PR as the normative open contract.
2. Document the bounded shape-selecting open-discriminator pattern and known-first rule.
3. Teach both official SDK generators to lower it before they consume this UCP release.

source/schemas/shopping/types/fulfillment_method.json
```diff
@@ -4,38 +4,115 @@
...
+            }
+          }
+        }
+      }
+    },
+    {
+      "if": {
+        "properties": {
+          "type": {
+            "const": "pickup"
```
gsmith85, line 92: One edge-case dynamic worth flagging that affects both the `if`/`then` and OO subtype approaches is: what happens if an `extension` tries to introduce a secondary shape that reuses the same type value?  In this case a second `{ "type": { "const": "pickup" } }`.

Because any deserialization relies on the `type` as a 1:1 hint to select a target model, tag collisions break string-hinted dispatch. To keep dynamic dispatch deterministic you'd need to solve this through either:

1. **Mandating Reverse-Domain Namespacing (RDN) for extension type(s):** Prescribing that extensions introducing distinct shapes use namespaced `type` strings (`"type": "com.acme.curbside_pickup"`) rather than redefining core specification literals like `"pickup"`.
2. **Static graph closure checks in `ucp-schema`:** Enhancing `ucp-schema` composition passes to inspect the active closure of core schemas + extensions and emit a build/CI failure if multiple schemas claim overlapping `const` string discriminators.

An advantage of the OO approach is that it eliminates the ambiguity of "am I intentionally extending a specific type?" or "am I (perhaps, unintentionally) extending a type because I had a name collision in my extension set?"

gsmith85:
> Open strings are a load-bearing UCP design principle, not an accident of this schema. We use open type vocabularies for fulfillment methods and destinations, totals and adjustments, media and links, identity providers, payment instruments and credentials, policies, and other concepts. This lets future UCP versions and negotiated extensions add values without making older generic consumers reject otherwise usable payloads.

We're fully in agreement here, the proposed OO shape does not limit type vocabularies.

> Fulfillment is one of the cases where recognized values also select additional constraints. JSON Schema's if/then branches express that open-world contract correctly: recognized values get their complete shape, while unknown values remain valid against the open base. The gotcha is that common SDK generators take a shortcut and emit only the structural base. That output is useful, but its validation is incomplete because it does not evaluate the known type branches.

Also agree here.  The ability to associate structured data with specific typing is a baseline affordance UCP should provide.

> The named subtype pattern you proposed helps readability and class generation, but it does not solve dynamic dispatch.  Something must still select those classes when decoding methods[].

The common pattern for this would be using the type name as a back mapping for polymorphic deserialization.  What's nice about this approach is that it would be easy to extend across instances where use open type vocabularies with type-specific shape mappings.

> Closing the active union would sacrifice the open vocabulary, while a generic {type: string} fallback overlaps the known branches.

I believe this is a false dilemma that supposes you must either (a) create a Closed Union (`type FulfillmentMethod = ShippingMethod | PickupMethod`) or (b) an unreliable Open Union (`ShippingMethod | PickupMethod | BaseFulfillmentMethod`).  In the case of (b) the portrayed risk is that if deserialization occurs on the basis of property deduction `BaseFulfillmentMethod` would match everything.

The way polymorphic deserialization typically would work is as follows:

```
1. Inspect json["type"].
2. If json["type"] == "shipping", decode as ShippingFulfillmentMethod.
3. If json["type"] == "pickup", decode as PickupFulfillmentMethod.
4. Else (any unknown string like "acme.drone"), decode as BaseFulfillmentMethod.
```

A few examples of this pattern:
- [Swagger](https://swagger.io/docs/specification/v3_0/data-models/inheritance-and-polymorphism/#discriminator)
- [Jackson](https://medium.com/@sandesh.pande/polymorphic-deserialization-in-java-using-jackson-edf9f5296479)

> Either schema organization therefore requires the SDKs to implement known-first dispatch: recognized values decode strictly as their known subtype and fail there, while only unrecognized values may reach Unknown.

For decoding tolerance: when an application developer imports SDK DTOs at a specific schema version, standard tolerant decoding simply ignores unmodeled extra fields (unless marked `required`). If a payload carries an unrecognized `type` string (e.g., a new extension), it naturally decodes into the open base `FulfillmentMethod` class. We don't need synthetic `Unknown` wrapper classes or  complex negative-set string refinements to keep the type hierarchy safe and clean

> Asked my agent to prototype known-first decoding against both official SDK stacks and recap its findings. It works against current schema proposed in this PR, but yes it does require UCP-specialization. Implementation notes and test matrix are here: https://gist.github.com/igrigorik/29eb5dbc09e7c43a2fc25bda281c8af0

I agree that deterministic dynamic dispatch is required either way, but looking at your prototype gist highlights the core technical divide here: **property inference versus explicit type mapping.**

Under the `if`/`then` schema, standard code generators discard conditional branches entirely. To recover them, your SDK preprocessor script is forced to perform **property inference**—crawling the JSON AST to find `if` blocks, inferring that whatever dynamic attribute assertions sit inside the corresponding `then` block (`...branch fields`) form a distinct subclass, and artificially synthesizing model classes (plus a synthetic `Unknown` negative-set refinement class) in script code.

Under explicit OO subtyping (`shipping_fulfillment_method.json` extending `fulfillment_method.json`), standard tools (`quicktype`, `datamodel-code-generator`) natively emit first-class `ShippingFulfillmentMethod` and `PickupFulfillmentMethod` models out of the box. All the SDK build pipeline has to do is wire a 1-line **type mapping** lookup table (`raw_json["type"]` $\rightarrow$ Model  Class).

> My proposal is:
> - Keep the conditional branches in this PR as the normative open contract.
> - Document the bounded shape-selecting open-discriminator pattern and known-first rule.
> - Teach both official SDK generators to lower it before they consume this UCP release.

I realize what I'm proposing isn't established in the protocol up to this point.  Still, it provides an opportunity to:
+ Maintain UCPs extensibility, it does not constrict which `types` can be specified or which shapes can be used.  It avoids the pitfalls of the `oneOf` pattern we have intentionally stepped away from. 
+ Improves schema readability, both by using a more common inheritance model and by naturally grouping related properties in a dedicated (sub)type definition.  By avoiding `if/else` we keep the UCP schema declarative, rather than procedural.
+ Code generation works out of the box without additional preprocessing.
+ SDKs will require custom type-matching logic for polymorphic deserialization. While we must define these mappings in our SDK generators, the implementation is straightforward because the type field inherently acts as the discriminator.
+ It avoids the need for documenting the if/then pattern and relying on the developer community to learn it.  `if`/`then` has existed since Open API 3.1, which came out in 2021, but using `anyOf` with declarative `const` properties is still more typical.

igrigorik:
@gsmith85 thanks, this discussion is helping sharpen the ownership boundary.

I agree on the runtime semantics: SDKs should use explicit known-first type mapping—`shipping → ShippingFulfillmentMethod`, `pickup → PickupFulfillmentMethod`, otherwise the open base; a recognized-type decode failure must not fall through. Where I differ is how we get there. Pushing codegen representation into the authored protocol schemas to me is a net regression, and I think we can achieve the outcome in a more ergonomic way.

UCP source schemas should optimize first for being easy to author and understand, while retaining standard JSON Schema semantics. We augment JSON Schema only where necessary with UCP annotations such as `ucp_request` and `ucp_response`; `ucp-schema` then composes and resolves those sources into standard JSON Schema that vanilla validators can consume -- this is an important property that I think we **must** retain. We should not make every schema author maintain a base, named subtype files, an active dispatcher, and potentially a mapping solely to accommodate generator limitations.

This also sent me reviewing our SDKs and made realize that none of them leverage `ucp-schema` and, worse, both Python and JavaScript SDKs independently reimplement substantial parts of its operation/direction projection, reference rewriting, and normalization. The better model is one source, and a preprocesor with two compiler targets:

```text
                         ┌─ validation target: resolved, complete standard JSON Schema; optionally bundled
schemas ─ ucp-schema ───┤  
                         └─ codegen target: resolved, factored named types + discriminator manifest
```

For the pattern in this PR: `if type == literal, then apply schema` can be mechanically materialized as a generated base schema, one named schema per known literal, plus a `type → schema` manifest SDKs if needed. The SDK generators then receive first-class models and explicit mappings without inferring semantics or requiring duplicated authored files. The same pass can reject discriminator collisions in the active composed graph. Language SDKs retain only the small runtime dispatch adapter.

---

If we step back, the underlying principles behind this direction are..

1. **Author once for protocol correctness**. UCP source schemas should be readable and express the complete contract using standard JSON Schema plus minimal UCP annotations—not duplicate structures or mappings to accommodate individual generators.
2. **Compile for each consumer.** Validators and SDK generators need different representations of the same schema graph: validators need a complete, optionally bundled schema; generators need factored named types and discriminator mappings. ucp-schema should derive both.
3. **Centralize complexity in shared tooling.** A small number of UCP tooling/SDK maintainers should solve projection and codegen once. Requiring every schema author—or every SDK—to recreate that logic multiplies complexity, drift, and bugs.

docs/specification/fulfillment.md
```diff
@@ -171,6 +176,143 @@ method.
...
+`type` is omitted.
+
+Destination authorship — whether the Platform may write entries into a
+method's `destinations[]` — is keyed by the method's `type`:
+
+| Method `type` | Request `destinations[]` |
+| --- | --- |
+| `shipping` | Platform-writable. The Platform writes the Buyer's shipping-address facts. |
+| `pickup` | Not Platform-writable (schema-enforced). The Business enumerates locations in its response; the Platform selects one by ID (see [Selection and Location Identity](#selection-and-location-identity)). |
+| other method type | The method's defining contract specifies the destination shape and whether it is Platform-writable. |
```
maximenajim: The “other method type” row leaves `curbside` without a defining request contract even though this same specification still lists it as well-known, and #507  deliberately promoted it as a first-class method for Drive Up. With the new conditional model, only `pickup` selects the Business Location branch. One client can therefore treat `curbside` as scalar Business-location selection while another treats it as an unknown method with no writable destination contract.

Add a `curbside` row and a schema branch with the same Business-authored Location Destination and `selected_destination_id` rules as `pickup` (or define a named location-pickup method family that both values use). The alternative is to remove `curbside` from the well-known vocabulary until it has a contract. For Target, this is a hot path, so leaving the first-class Drive Up value in the generic fall-through is not interoperable.

docs/specification/fulfillment.md
```diff
@@ -171,6 +176,143 @@ method.
...
+```
+
+### Selection and Location Identity
+
+`selected_destination_id` identifies the selected destination in a method by
+its `id`, and is the sole channel for selecting a Business Location. It accepts
+any stable, Business-scoped Location ID the Business recognizes for that
+method, including IDs the Business has not (yet) enumerated in that method's
+`destinations[]` — for example an ID discovered through Catalog or a separately
+negotiated Location capability. When Catalog represents a location as
```
maximenajim: The paragraph says an ID discovered through a separately negotiated Location capability is accepted, but the following **MUST** return the typed destination only when Catalog supplied the ID. A Platform that selects a Location Search result can therefore receive `selected_destination_id` without a corresponding self-describing destination, and the contract also does not make the selection relation single-valued. Different clients can render different places, especially when cached discovery meets a store closure or changed eligibility.

Make acceptance source-agnostic and testable: “When the Business accepts a non-null `selected_destination_id`, its response MUST contain exactly one destination with the same `id` in that method. If revalidation cannot accept the ID, the Business MUST NOT substitute another destination; for an update it MUST retain the prior selection and return the current Checkout with a recoverable error Message whose `path` selects `selected_destination_id`.” This should apply equally to Catalog, Location, and IDs learned from an earlier Checkout response.

igrigorik:
@maximenajim good eye, addressed both points in https://github.com/Universal-Commerce-Protocol/ucp/pull/688/changes/e7e74c34bd3a360622006aca99331869971101d7. 

@gsmith85 closing the loop on the SDK, prototyped https://github.com/igrigorik/ucp-sdk. The flow is:

```text
Platform profile
  → ucp-schema resolve
  → shared nominal SDK IR
  ├── portable projection
  ├── Python/Pydantic SDK
  └── TypeScript/Zod SDK
```

Some takeaways from the poc..

We can keep the authored UCP contract as standard JSON Schema, including the open `type` vocabulary and `if`/`then` refinements in this PR. `ucp-schema` resolves the UCP-specific operation/direction and composition semantics; the projection then materializes the named Base/Shipping/Pickup models and deterministic known-first mapping SDKs need.

- Generated known values decode strictly and cannot fall through to the open base when malformed; unmapped values retain the tolerant base behavior.
- Rather than port every JSON Schema constraint into Pydantic and Zod, the generated public parsers validate against the retained operation-resolved schema first, then decode the nominal model. A shared differential corpus agrees across canonical `ucp-schema validate`, Python, and TypeScript for the exercised Checkout/Fulfillment slice.

Neither official SDK currently uses `ucp-schema`; both independently implement substantial operation projection, ref rewriting, normalization, and constraint repair. Across the two repositories that is roughly 5,087 lines of handwritten generation/repair machinery. The POC is roughly 1,080 production lines. This is not an apples-to-apples but I think meta point stands that we can remove a lot of duplication and bug-prone replication.

The POC’s distinctive capability is profile-specific generation. Its sample Platform profile selects Checkout, Fulfillment, and `org.foo.custom`; both generated SDKs contain that exact supported universe, including the third-party extension, without emitter-specific extension code. The existing SDKs generate against an official core snapshot/fixed extension set rather than a supplied Platform profile, which is incomplete subset of what you'd want as a platform.

With all this in backpocket, I am confident that the SDK concern is solvable downstream without moving generator-oriented scaffolding into every authored schema. The current conditional shape preserves the complete vanilla JSON Schema contract; `ucp-schema → projection/IR → language emitter` gives SDKs the explicit representation they need while centralizing the UCP-specific complexity once.

docs/specification/fulfillment.md
```diff
@@ -171,6 +175,173 @@ method.
...
 
+## Destinations
+
+A fulfillment method's `type` describes how items are fulfilled, while each
+destination's `type` describes where fulfillment occurs. The two
+discriminators play different roles by direction: in Business responses,
+every destination carries a required `type`, so responses are
+self-describing; in Platform requests, the method's contract determines who
+authors `destinations[]` and MAY define a default destination shape when
+`type` is omitted.
```
gsmith85, line 229: Writing style: Using examples here may help with comprehension.

> A fulfillment method's `type` specifies **how** items are fulfilled (`shipping`, `pickup`), while each destination's `type` specifies **where** fulfillment occurs (`shipping_address`, `business_location`).

docs/specification/fulfillment.md
```diff
@@ -171,6 +175,173 @@ method.
...
+
+Destination authorship — whether the Platform may write entries into a
+method's `destinations[]` — is Business-authored by default; a contract keyed
```
gsmith85: Nit: Double "authorship".

> Destination ... is business-authored by default.

docs/specification/fulfillment.md
```diff
@@ -171,6 +175,173 @@ method.
...
+on the method's `type` **MAY** override that default:
+
+| Method `type` | Request `destinations[]` |
+| --- | --- |
+| `shipping` | Platform-writable. The Platform writes the Buyer's shipping-address facts. |
+| `pickup` | Not Platform-writable (schema-enforced). The Business enumerates locations in its response; the Platform selects a location by its Business-scoped ID (see [Selection and Location Identity](#selection-and-location-identity)). |
+| any other method `type` | Not Platform-writable (the base default). The Business enumerates destinations in its response; the Platform selects one via `selected_destination_id`. |
+
+Destination `type` is **required in responses and optional in requests**. In
+responses, every destination carries a required, open `type`, so destinations
```
gsmith85: Nit: Strike "a required, open" it's not relevant to the primary point that it's required for self-description.

docs/specification/fulfillment.md
```diff
@@ -171,6 +175,173 @@ method.
...
+remain self-describing wherever they appear. In requests, `destinations[]` is
+absent unless a method-specific contract opts into a Platform-writable shape:
```
gsmith85: Nit: Prefer "omitted" versus "absent" to align with UCP's grammar.

docs/specification/checkout-rest.md
```diff
@@ -371,9 +371,10 @@ include all previously set fields it intends to retain.
 
 #### Update Fulfillment
 
-Fulfillment is an extension to the checkout capability. Most fields are provided
-by the business based on buyer inputs, which includes desired fulfillment
-type & addresses.
+Fulfillment is an extension to the checkout capability. The Business provides
+most fields based on Buyer inputs, including desired fulfillment method types
+and typed destinations such as shipping addresses and Business Location
+references.
```
gsmith85: Nit: Avoid "most" for its non-specificity.

docs/specification/fulfillment.md
```diff
@@ -171,6 +175,173 @@ method.
...
+specific to such a value are validated by the negotiated extension's schema.
+
+### Shipping Destination
+
+For a Shipping Destination, the Platform supplies shipping-address facts as
+flat Postal Address fields directly on the destination. The Platform **MAY**
+include `id` and **MAY** include `type: "shipping_address"` in its request.
+The Business **MUST** include `type: "shipping_address"` and assign `id` in
+its response. An `id`-only destination that references a saved or
+provider-held address is still a Shipping Destination; conveying provider
```
gsmith85: Make it clear that we're talking about a Credential Provider?

> An id-only destination referencing a saved address (such as from a digital wallet or user profile)

source/schemas/shopping/types/fulfillment_method.json
```diff
@@ -4,38 +4,115 @@
...
+            }
+          }
+        }
+      }
+    }
+  ],
+  "dependentRequired": {
+    "destinations": [
+      "type"
+    ]
```
gsmith85, line 116: Created #732 to reenable JSON schema formatting.

gsmith85 (commented):
LGTM.  I added some minor comment around documentation that you can have an agent accept (or not) but otherwise I'm happy with this.

gsmith85 (approved):

docs/specification/fulfillment.md
igrigorik: Simplified this whole paragraph.

igrigorik:
@gsmith85 ty for great suggestions. Landed & refreshed PR description to match final shape.

vishkaty:
Two notes ahead of the rebase over #671. Main now carries the location bullet as a MUST accept (fulfillment.md:358) while this PR specifies MAY submit with revalidated acceptance and a rejection path, so the rebase will have to pick one contract; deciding it explicitly before the release keeps that from resolving mechanically. Also, the pickup row at fulfillment.md:196 says Not Platform-writable (schema-enforced), but in default validation mode ucp_request omit only drops destinations from the request projection and leaves the object open, so a pickup update carrying destinations, even destinations: "garbage", validates with ucp-schema 1.4.1; adding destinations: false to the non-shipping request branches would make the claim true in every mode.

igrigorik:
@vishkaty good catch, addressed in https://github.com/Universal-Commerce-Protocol/ucp/commit/30144b580e50a7573e6c0ed4f3b2fc665f40e310.
@amithanda ready for final pass.

amithanda (approved):

PR #690, opened 2026-08-06
feat(media): model image, video, and 3D variants

igrigorik:
The existing `Media` shape is image-oriented: one URL and one pair of dimensions cannot faithfully describe alternate video resolutions and codecs, adaptive-streaming manifests, device-specific 3D formats, poster images, or an externally hosted player. This update extends `Media` as an open type discriminator with well-known `image`, `video`, and `model_3d` variants. It follows the schema.org [`MediaObject`](https://schema.org/MediaObject) family while preserving UCP's existing wire contract and generated-source APIs.

- Adds per-type schemas for image, video, and 3D media using open `if`/`then` composition.
- Adds `sources` for the complete set of available video renditions or 3D encodings.
- Adds `preview` for a poster or thumbnail still.
- Adds optional `name`, distinct from accessibility-oriented `alt_text`.
- Adds video `duration` in seconds.
- Models externally hosted playback as `type: "video"` plus `embed_url`, rather than introducing a redundant `external_video` type.
- Adds source metadata for MIME type, format, dimensions, and byte size.
- Keeps `Media.type` open: a Platform must ignore media types it does not recognize.

When `sources` is present and `url` identifies a rendition, the Business should include a source whose `url` matches the media item's `url`. The exception is an externally hosted video, where `url` identifies a watch page rather than a playable rendition.

### Self-hosted video

The primary rendition appears in both `url` and the complete `sources` set:

```json
{
  "type": "video",
  "url": "https://cdn.example.com/video-720.mp4",
  "name": "Product demonstration",
  "duration": 24,
  "preview": {
    "url": "https://cdn.example.com/video-poster.jpg",
    "width": 1280,
    "height": 720
  },
  "sources": [
    {
      "url": "https://cdn.example.com/video-720.mp4",
      "mime_type": "video/mp4",
      "width": 1280,
      "height": 720
    },
    {
      "url": "https://cdn.example.com/video.m3u8",
      "mime_type": "application/vnd.apple.mpegurl"
    }
  ]
}
```

### 3D model

A single logical model can advertise the formats supported by different devices:

```json
{
  "type": "model_3d",
  "url": "https://cdn.example.com/product.glb",
  "preview": {
    "url": "https://cdn.example.com/product-poster.jpg"
  },
  "sources": [
    {
      "url": "https://cdn.example.com/product.glb",
      "mime_type": "model/gltf-binary"
    },
    {
      "url": "https://cdn.example.com/product.usdz",
      "mime_type": "model/vnd.usdz+zip"
    }
  ]
}
```

### Externally hosted video

The watch page remains the primary URL while `embed_url` identifies third-party active player content:

```json
{
  "type": "video",
  "url": "https://videos.example.com/watch/123",
  "embed_url": "https://videos.example.com/embed/123",
  "preview": {
    "url": "https://cdn.example.com/video-poster.jpg",
    "width": 1280,
    "height": 720
  }
}
```

A Platform may decline to load third-party active content under its security, privacy, or presentation policy and use the watch-page URL instead.

## Backward compatibility

This is backward compatible for conforming uses of the released `Media` contract:
- the required set remains `type` and `url`;
- no existing field is removed, renamed, or retyped;
- `width` and `height` remain on base `Media` for generated-source compatibility;
- all new fields are optional;
- existing payloads continue to validate;
- unknown media types remain valid.

---

## Checklist

### DIFF
--- a/docs/specification/catalog/index.md
+++ b/docs/specification/catalog/index.md
@@ -132,7 +132,56 @@ as the first element. Platforms SHOULD treat the first element as featured.
 
 ### Media
 
+A single logical media item on a Product or Variant: an image, a video, or a
+3D model. `url` identifies the item's primary resource, and its interpretation
+depends on `type`. `preview` is a still image the Platform can render before —
+or instead of — loading that resource.
+
+`type` is an open string with three well-known values: `image`, `video`, and
+`model_3d`. A Platform **MUST** ignore a media item whose `type` it does not
+recognize.
+
+When present, `sources` contains the complete set of available renditions or
+encodings. A Business that provides `sources` **SHOULD** include a source whose
+`url` equals the media item's `url`, unless `url` is an externally hosted
+video's watch page rather than a rendition.
+
+For an image, a self-hosted video, or a 3D model, a Platform loads the
+renderable or loadable primary resource at `url`. For an externally hosted
+video with `embed_url`, `url` identifies the watch page and `embed_url`
+identifies third-party active player content. A Platform **MAY** decline to
+embed that content under its security, privacy, or presentation policy and use
+the watch-page `url` instead. For a thumbnail, a Platform can use `preview.url`
+and fall back to `url` for an image, which is directly displayable.
+
 {{ schema_fields('types/media', 'catalog') }}
+
+#### Image
+
+{{ schema_fields('types/media_image', 'catalog') }}
+
+#### Video
+
+A self-hosted video's `url` identifies its directly playable default rendition.
+An externally hosted video is distinguished by `embed_url`, which identifies
+the third-party active player content; its `url` identifies the watch page, not
+a playable rendition.
+
+A Platform that selects a rendition from `sources` **SHOULD** use the `width`
+and `height` of that entry rather than those on the media item, unless the
+entry omits them.
+
+For an externally hosted video, a Business **SHOULD** provide both
+`preview.width` and `preview.height` with a ratio that matches the intended
+player, unless it has no suitable preview. A Platform **SHOULD** reserve the
+initial player layout using that ratio, unless its presentation policy supplies
+another ratio.
+
+{{ schema_fields('types/media_video', 'catalog') }}
+
+#### 3D Model
+
+{{ schema_fields('types/media_model_3d', 'catalog') }}
 
 ### Product Option
 
--- a/source/schemas/common/types/media_video.json
+++ b/source/schemas/common/types/media_video.json
@@ -1,0 +1,35 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/common/types/media_video.json",
+  "title": "Media Video",
+  "description": "Type-specific properties for video media.",
+  "type": "object",
+  "properties": {
+    "url": {
+      "description": "The directly playable default rendition for a self-hosted video, or the watch-page URL for an externally hosted video with `embed_url`. When `sources` is present and `url` identifies a rendition, the Business SHOULD include a source whose `url` equals the media `url`; this does not apply when `url` is an external watch-page URL."
+    },
+    "embed_url": {
+      "type": "string",
+      "format": "uri",
+      "description": "Player embed (iframe) URL for an externally hosted video, through which the Platform may play the video; absent for self-hosted video. A Platform MAY decline third-party active content under its security, privacy, or presentation policy and use the watch-page `url` instead."
+    },
+    "duration": {
+      "type": "integer",
+      "minimum": 1,
+      "description": "Duration in seconds."
+    },
+    "width": {
+      "description": "The video's display width in pixels, for layout and aspect ratio. A selected rendition's exact width is in that `sources[]` entry."
+    },
+    "height": {
+      "description": "The video's display height in pixels, for layout and aspect ratio. A selected rendition's exact height is in that `sources[]` entry."
+    },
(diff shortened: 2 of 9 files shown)


docs/specification/catalog/index.md
amithanda, line 106: In this file, the **Product** (line 106) and **Variant** (line 120) sections right above `### Media` still refer specifically to returning the featured *"image"* as the first element of `media` and state that *"Platforms SHOULD treat the first element as featured."*

With `Media` now supporting `video`, `model_3d`, and open `type` values where *"A Platform **MUST** ignore a media item whose `type` it does not recognize"* (line 141):
- Should we update lines 106 and 120 from *"image"* to *"media item"* (to match [`product.json`](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/shopping/types/product.json#L56): *"First item is the featured media for listings."*), or recommend that if the first element is not `type: "image"`, the Business SHOULD provide `preview` so listing cards always have a still thumbnail?
- Should we also clarify that Platforms SHOULD treat the **first recognized** media item as featured (so that if `media[0]` has an unrecognized `type` that the Platform ignores, it falls back cleanly to the next supported item in `media[]`)?

source/schemas/common/types/media_video.json
```diff
@@ -0,0 +1,35 @@
...
+      "minimum": 1,
+      "description": "Duration in seconds."
+    },
+    "width": {
+      "description": "The video's display width in pixels, for layout and aspect ratio. A selected rendition's exact width is in that `sources[]` entry."
+    },
+    "height": {
+      "description": "The video's display height in pixels, for layout and aspect ratio. A selected rendition's exact height is in that `sources[]` entry."
+    },
+    "sources": {
```
amithanda, line 27: On scope: should `sources` also be available for `image`?
The motivation in the description — "one URL and one pair of dimensions cannot
faithfully describe alternate … resolutions and codecs" — reads to me as applying
just as strongly to images. AVIF/WebP/JPEG fallbacks and DPR/width variants are
probably the highest-volume multi-rendition case in a product catalog, but
`media_source.json` scopes itself to "one resolution/format of a video, or one
file format of a 3D model."
Two things that make this feel like a small delta rather than new design work:
1. `media_source.json` is already generic enough as-is — `url`, `mime_type`,
   `format`, `width`, `height`, `filesize` describe an image rendition perfectly
   (`image/avif`, `image/webp`, `image/jpeg`).
2. `media_image.json` currently carries three description-only overrides and no
   constraints, so it's the one variant file that doesn't yet earn its place.
   Adding `sources` would give it a reason to exist alongside the other two.

source/schemas/common/types/media_video.json
```diff
@@ -0,0 +1,35 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ucp.dev/schemas/common/types/media_video.json",
+  "title": "Media Video",
+  "description": "Type-specific properties for video media.",
+  "type": "object",
+  "properties": {
+    "url": {
+      "description": "The directly playable default rendition for a self-hosted video, or the watch-page URL for an externally hosted video with `embed_url`. When `sources` is present and `url` identifies a rendition, the Business SHOULD include a source whose `url` equals the media `url`; this does not apply when `url` is an external watch-page URL."
+    },
+    "embed_url": {
```
amithanda, line 11: Security question on the two new vectors this PR introduces for externally hosted video: `<iframe src={embed_url}>` and the watch-page fallback `<a href={url}>`.

Before this PR, `media.url` was consumed by passive media elements (``, `<video>`), where browsers ignore executable schemes like `javascript:` or `data:text/html`. With externally hosted video:

1. **Scheme safety on both `embed_url` and watch-page `url`:** Because `"format": "uri"` allows any RFC 3986 scheme, `javascript:...`, `data:text/html;base64,...`, and plain `http://` all validate (including under `--strict true`) for both `embed_url` and `url`. Unlike ``, both `<iframe src={embed_url}>` (on load) and a watch-page link `<a href={url}>` (on click) will execute `javascript:` / `data:text/html` URIs. Could we enforce `"pattern": "^https://..."` on media URLs (mirroring [`permalink.json`](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/schemas/shopping/permalink.json#L11)), or require Platforms to reject non-`https:` schemes before rendering or linking?

2. **Iframe isolation floor for `embed_url` (even over `https://`):** Even with a valid `https://` URL, `embed_url` is the first catalog field that loads an **active third-party JS runtime** (`<iframe>`) rather than a passive media stream. In [`embedded-protocol.md`](https://github.com/Universal-Commerce-Protocol/ucp/blob/main/docs/specification/embedded-protocol.md), UCP requires business iframes to use `sandbox`, `credentialless`, and strict `postMessage` origin checks, whereas `embed_url` currently only says a Platform *"MAY decline third-party active content."* Should we add a normative baseline that if a Platform renders `embed_url`, it **MUST** sandbox the iframe (without granting `allow-top-navigation` or same-origin access to the host)?

source/schemas/common/types/media_video.json
igrigorik: Good call, addressed.

source/schemas/common/types/media_video.json
igrigorik: 👍 updated

igrigorik:
@amithanda thanks, great feedback! Addressed, and with benefit of fresh eyes, caught a few additional issues. PTAL.

PR #700, opened 2026-08-08
ci: serialize gh-pages deploys and add job timeouts

sakinaroufid:
# Description

`deploy_main` and `deploy_release` both push to `gh-pages`, and neither has a concurrency group. Two merges landing close together means two runs racing on the push, and one gets rejected.

It's hit twice in the last 100 main pushes (2026-04-20 to 2026-08-07), from two different push sites, the raw `git push` in the root deploy and mike's own:

```
run 26217319631  2026-05-21
  ! [remote rejected] gh-pages -> gh-pages (cannot lock ref
    'refs/heads/gh-pages': is at 37c3742 but expected dcf2a70)

run 27422226472  2026-06-12
  error: failed to push branch gh-pages to origin:
  ! [rejected]  gh-pages -> gh-pages (fetch first)
```

Whether it matters depends on which run loses. In May the loser was the older commit, and since the winner was a descendant on main its build already had that content, so all we got was a red X on someone's merge for something they didn't break. June was the bad case: the loser was `feat!: buyer consent with per-segment extensibility`, and those docs didn't publish until the next merge to main.

Call it one stale publish and one false failure a quarter at the current rate. Not an outage. But the window is "two merges inside one deploy" and the median gap between deploys is already 6 minutes, so it won't get better on its own.

### Fix

Both deploy jobs go in one group:

```yaml
concurrency:
  group: gh-pages-deploy
  cancel-in-progress: false
```

`cancel-in-progress: false` matters here. A cancelled deploy is just one that didn't happen, and for `deploy_release` that means a version never publishing.

One gap this doesn't close: if two runs queue behind a third, GitHub keeps only the newest pending one. That's fine for `deploy_main` since a newer build supersedes an older one, but two release versions queueing at once could still drop one. I had retry-and-rebase loops around mike and the raw push to cover it, then took them back out. The root deploy builds a single commit that replaces the whole root tree, and rebasing that onto a moved tip can conflict. I'd rather not ship a retry path I can't exercise in CI, wrapped around a deploy. At least a cancelled run shows up in the Actions tab. Happy to put them back if you'd rather have them.

### Also in here

Two things I noticed while reading the workflows:

* No job anywhere sets `timeout-minutes`, so the default is 6 hours. Nothing here takes more than a few minutes. Set 30 on all six jobs in `docs.yml`.
* The `linter.yaml` paths filter lists `.github/workflows/schema-validation.yml`. That file doesn't exist.

## Category (Required)

## Related Issues

N/A

## Checklist

Workflow-only, so nothing to update for docs, schemas or models. No tests because there's no way to assert on Actions concurrency from this repo; the evidence is the two failed runs above.

`yamllint -c .github/linters/.yamllint.yml .github/workflows/` shows nothing new. The two line-length warnings left in `docs.yml` are on existing `if: contains(fromJson(...))` lines that just shifted down, 127 chars before and after.

Job logs quoted above. No rendering change.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -40,6 +40,7 @@ on:
 jobs:
   lint:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     # Don't run this workflow if this is a release branch creation event
     # (event.before SHA is all zeros)
     if: >-
@@ -80,6 +81,7 @@ jobs:
   build_and_verify_main:
     needs: lint
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     if: github.ref == 'refs/heads/main' || github.event.pull_request.base.ref == 'main'
     permissions:
       contents: read
@@ -107,6 +109,7 @@ jobs:
   build_and_verify_release:
     needs: lint
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     if: >-
       startsWith(github.ref, 'refs/heads/release/') ||
       startsWith(github.event.pull_request.base.ref, 'release/')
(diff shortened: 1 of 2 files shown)


damaz91 (approved):
Really appreciate the help here, thanks @sakinaroufid !

ptiper (approved):

PR #704, opened 2026-08-10
fix: run pre-commit validators through uv

horaceewang-TT:
# Description

Update the local pre-commit validator hooks to run through `uv run python`
instead of the system `python3`.

This keeps local hook execution aligned with the project's declared Python
requirement (`>=3.10`) and avoids failures on machines where `python3` points to
an older interpreter, such as Python 3.9.

Directly running scripts/test_validate_examples.py fails at runtime because it uses Python 3.10+ type syntax such as dict | None, for example at validate_examples.py中 _collect_properties 的     props.update(branch.get("properties", {}))
```text
 def _collect_properties(schema: dict) -> dict:
  """Collect properties, merging allOf branches."""
  props = dict(schema.get("properties", {}))
  for branch in schema.get("allOf", []):
    props.update(branch.get("properties", {}))
  return props
```

## Category (Required)

## Checklist

- [ ] My changes pass all local linting and formatting checks.

Validation performed:

```text
uv run python --version
Python 3.12.x

pre-commit run validate-examples-tests --all-files
Validator unit tests.....................................................Passed

pre-commit run validate-examples-full --all-files
Validate JSON examples (full corpus).....................................Passed

pre-commit run validate-examples-changed --files docs/specification/order.md
Validate JSON examples (changed docs)....................................Passed

git diff --check HEAD~1..HEAD
# no output
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,6 +13,8 @@ repos:
       # install with `cargo install ucp-schema`. CI runs the full-corpus
       # check on every PR as the mandatory backstop
       # (.github/workflows/docs.yml).
+      # Run through uv so local hooks use the project's Python version
+      # requirement from pyproject.toml instead of the system python3.
       #
       # Two-hook split for fast local feedback:
       #  - validate-examples-changed: doc edits validate only changed files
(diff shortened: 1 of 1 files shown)


damaz91 (approved):
Thanks!

ptiper (approved):

igrigorik:
@damaz91 @ptiper @horaceewang-TT 

Ran into an unintended side effect: plain `uv run` is allowed to re-resolve and update `uv.lock`.

This makes the pre-commit/pre-push hooks dependent on developer-global uv settings. For example, when uv applies a global `exclude-newer` policy, a docs-only commit or push causes each validator hook to rewrite `uv.lock`. Pre-commit then correctly fails because the hook modified a tracked file. Reverting the lockfile does not help because the next hook run rewrites it again.

I reproduced this on the current `main` configuration:
- plain pre-push validation rewrote hundreds of lines in `uv.lock`;
- the same validation with `UV_FROZEN=1` passed and left the lockfile byte-identical.

I recommend making the hooks explicitly frozen:

```yaml
entry: uv run --frozen python scripts/validate_examples.py --schema-base source/schemas/ --file
```

and applying the same change to the full-corpus and validator-test hooks.

`--frozen` is preferable to `--locked` here: these hooks validate source and documentation; they should consume the committed lockfile rather than fail or mutate it because of developer-local resolution policy. Dependency updates should remain an explicit `uv lock` operation.

Immediate workaround:

```bash
UV_FROZEN=1 git commit ...
UV_FROZEN=1 git push ...
```

damaz91:
@igrigorik sorry we haven't catch this, and thanks for promptly raising the issue. Addressing in https://github.com/Universal-Commerce-Protocol/ucp/pull/717

horaceewang-TT:
@igrigorik @damaz91  Thanks for the quick turnaround! The --frozen change in #717 looks like exactly the right fix — it matches the behavior I verified locally with UV_FROZEN=1 (hooks passed and uv.lock stayed byte-identical)

PR #708, opened 2026-08-11
fix(scripts): correctly resolve internal absolute root site URLs to root index.html in check_links.py

yappermoar-boop:
Updated check_links.py to correctly recognize internal absolute root site URLs (e.g. https://ucp.dev or https://ucp.dev/#anchor) and resolve them to ROOT_DIR/index.html instead of falling back to file_path. Also handles unslashed site URL matches.

### DIFF
--- a/scripts/check_links.py
+++ b/scripts/check_links.py
@@ -160,11 +160,17 @@ def check_links():
         continue
 
       parsed = urlparse(link)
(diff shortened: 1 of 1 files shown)


damaz91:
@yappermoar-boop can you please sign the CLA so we can accept your contribution? Thank you for this PR!

PR #709, opened 2026-08-11
fix(scripts): correctly resolve internal absolute root site URLs to root index.html in check_links.py

yappermoar-boop:
Resend because google cla  ,Updated check_links.py to correctly recognize internal absolute root site URLs (e.g. https://ucp.dev or https://ucp.dev/#anchor) and resolve them to ROOT_DIR/index.html instead of falling back to file_path. Also handles unslashed site URL matches.

### DIFF
--- a/scripts/check_links.py
+++ b/scripts/check_links.py
@@ -160,11 +160,17 @@ def check_links():
         continue
 
       parsed = urlparse(link)
+      is_absolute_internal = False
       if parsed.scheme and parsed.scheme in ("http", "https"):
-        if not link.startswith(SITE_URL):
+        site_url_no_slash = SITE_URL.rstrip("/")
+        if not link.startswith(SITE_URL) and not link.startswith(site_url_no_slash):
           continue  # External link
-        # Internal absolute URL (e.g. https://ucp.dev/foo) -> /foo
-        link = link[len(SITE_URL) - 1 :]  # Keep the leading slash
+        is_absolute_internal = True
+        if link.startswith(SITE_URL):
+          link = link[len(SITE_URL) - 1 :]
+        else:
+          link = "/" + link[len(site_url_no_slash) :]
+        parsed = urlparse(link)
 
       path_part = parsed.path
       anchor_part = parsed.fragment
@@ -179,7 +185,7 @@ def check_links():
 
       # Resolve Target File
       if not path_part:
-        target_file = file_path
+        target_file = ROOT_DIR / "index.html" if is_absolute_internal else file_path
       elif path_part.startswith("/"):
         # Absolute path from root
         rel_path = path_part[1:]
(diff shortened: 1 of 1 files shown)


damaz91:
Hi @yappermoar-boop,

Thank you for this contribution! Fixing the absolute root URL resolution in the link checker is a great improvement.

I have a couple of suggestions to make the internal link detection and resolution logic more robust:

### 1. Robust Internal Link Detection
The current check `link.startswith(site_url_no_slash)` can be fragile. For example, it would incorrectly classify an external link containing userinfo like `https://ucp.dev@evil.example/x.json` as internal because it starts with `https://ucp.dev`.

We can make this more robust by parsing the URL and comparing the hostnames directly:

```python
      parsed = urlparse(link)
      is_absolute_internal = False
      if parsed.scheme and parsed.scheme in ("http", "https"):
        parsed_site = urlparse(SITE_URL)
        if parsed.hostname == parsed_site.hostname:
          path = parsed.path if parsed.path else "/"
          site_path = parsed_site.path
          site_path_no_slash = site_path.rstrip("/")
          if (path == site_path_no_slash) or path.startswith(site_path):
            is_absolute_internal = True
            
            # Normalize link path to be relative to ROOT_DIR
            if path.startswith(site_path):
              rel_link_path = "/" + path[len(site_path):]
            else: # path == site_path_no_slash
              rel_link_path = "/"
            
            query_part = f"?{parsed.query}" if parsed.query else ""
            fragment_part = f"#{parsed.fragment}" if parsed.fragment else ""
            link = rel_link_path + query_part + fragment_part
            parsed = urlparse(link)
        
        if not is_absolute_internal:
          continue  # External link
```

### 2. Redundant Code Cleanup
With the normalization logic, internal absolute links will always have a path of at least `/`. This means `path_part` is never empty for these links, and the `not path_part` block is only entered for relative anchor links (where `is_absolute_internal` is always `False`).

We can revert this line back to its original state to keep it clean:
```python
      # Resolve Target File
      if not path_part:
        target_file = file_path
```
The root URL resolution (when path is `/`) will still be correctly handled by the `elif path_part.startswith("/")` branch.

What do you think?

Thanks again for the PR!

yappermoar-boop:
Hi @damaz91 ,

Thank you for the thoughtful review and detailed feedback! 

I have updated the PR with your suggestions:
1. Implemented hostname-based comparison using `urlparse` to prevent edge-case external links (like `userinfo` format) from being misclassified.
2. Normalized the path resolution so internal absolute links start at `/`, allowing us to keep the `if not path_part` target assignment clean and concise.

All tests pass cleanly. Thanks again!

damaz91:
Thanks! LGTM - this PR actually uncovered some broken links - fixing those in https://github.com/Universal-Commerce-Protocol/ucp/pull/710

damaz91 (approved):

carolinerg1 (approved):

damaz91:
Let's wait for https://github.com/Universal-Commerce-Protocol/ucp/pull/710 to merge first, then we can go ahead with this

carolinerg1 (approved):

PR #712, opened 2026-08-11
refactor: simplify totals, required terms, split payments, and subscriptions

raginpirate:
Fork of https://github.com/Universal-Commerce-Protocol/ucp/pull/602

Update: payment terms after a thorough review.

Four changes, all narrowing. The fundamental simplifications were all great additions, thanks @igrigorik! This pushes some of the remaining edges to simplify a bit further.

1. Amounts over Totals
The proposal actually started with amounts, and we grew it into totals to try and address the complexities of discounts or additional fees from using certain payment terms. However, as this proposal's prose became more verbose, we find ourselves with an array of one-element — minContains: 1 + maxContains: 1, all pinned to a single total. I think the right analogue in UCP is split_payments.instruments[].amount (a scalar), not fulfillment_option.totals (a real breakdown). Drops ~16 lines of schema machinery and introduced the right prose to encourage the correct discovery of discount and fees context in the right place; the checkout total.

2. Payment terms are optional.
 The biggest behavioural change. terms and selected_term_id were required, so negotiating the extension forced every checkout to carry a synthetic "Pay now" term, an id, a mandatory prose sentence, and an accepted_term on the order for a buyer who owes nothing. Absence is now defined: payment is due in full at completion, exactly as core Checkout already means. The extension is silent on the majority of checkouts, and adopting it costs a Platform no changes to what it sends right now.

3. Drop subscription examples
A subscription is a series of recurring orders, and the recurrence is carried by policy — structured, applied to the line items it governs — not by a payment schedule. That's why it's out of scope here: you don't need a payment term to express that a subscription recurs or that the associated method will be vaulted to pay for it. And now that terms are optional, you don't need one to take the first charge either. So I'd drop the subscription framing from this extension rather than leave the two adjacent. Leaving them side by side invites a Business or Platform to reach for a deferred term to express unbounded fulfillment and payment recurrence — the mistake that misstates checkout.total and writes a moving obligation into a record that must not move. 

4. Split Payments does not compose with a multi-schedule term.
Split Payments allocates by amount, in instrument order, against one total. A term with several schedules has several dates, so any allocation across it decides when each instrument is charged — and an amount bounds what an instrument pays, never where it lands, so small adjustments ($0.01) in capture amounts can accidentally change when the buyer would be charged. This problem feels ill-fitted to solve with prose; the safest path is to defer split payments support outright for now and fast-follow it.

### DIFF
--- a/docs/specification/payment-terms.md
+++ b/docs/specification/payment-terms.md
@@ -34,11 +34,24 @@ This extension adds two properties to `checkout.payment`:
 
 * `terms[]` — the payment terms the Business offers for this checkout.
   Response-only.
-* `selected_term_id` — the selected term. Always present in a response;
-  a Platform writes it to change the selection.
+* `selected_term_id` — the selected term. Present in a response whenever `terms`
+  is; a Platform writes it to change the selection.
 
 A Buyer picks one of the options in `terms[]`, the same way they pick
 one [fulfillment option](fulfillment.md#platform-responsibilities).
+
+## When a Checkout carries terms
+
+Payment terms are optional. A Business offers them when a Checkout has something
+to say about payment timing — a choice between terms, or a single term whose
+schedules must be disclosed. Where payment is due in full at completion, a
+Business may omit `terms` and `selected_term_id` entirely, and the Checkout
+`total` is the amount due, as core Checkout already states.
+
+Where `terms` is present it **MUST** contain at least one term, and
+`selected_term_id` **MUST** name one of them. The Checkout `total` is the
+selected term's total, so a list of terms without a selection would show an
+amount matching no stated term.
 
 ## Presenting payment terms
 
@@ -47,7 +60,7 @@ installments, or trade credit to present a term meaningfully.
 
 Given the Checkout's `currency`, a Platform that recognizes no `type` value and
 reads no payment-term field other than `title`, `description`,
-`schedules[].description`, and `schedules[].totals` **MUST** be able to present
+`schedules[].description`, and `schedules[].amount` **MUST** be able to present
 what is owed and when, for every term. A Business **MUST** author terms so this
 holds.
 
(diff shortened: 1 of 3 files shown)


archrao:
These are great improvements Daniel, thank you for the thorough diligence. Agree with these changes as they simplify the specification. 
While the one instrument-one schedule, restricts the buyers from being able to use multi-cards(split payments) for a multi-schedule booking, it seems like the right step for v1.

vishkaty:
Read this from the conformance side, with the four changes checked against the schemas and examples on the branch:

**Validation:** ran `scripts/validate_examples.py` against the branch — 299 passed, 0 failed, so every example (including the new `$.totals` discount fragment and the reworked schedule `amount` examples) validates against the simplified schemas. The move from the one-element `totals` array to a scalar `amount` also quietly keeps the non-negativity constraint: `common/types/amount.json` carries `"minimum": 0` itself, so the old `minimum: 0` machinery survives the simplification even though the prose no longer states it. And the invariants from the #692 round — the accepted-term freeze with `adjustments[]` carrying post-purchase drift, the `payment_term_changed` warning, sum-to-checkout-total on the selected term — all carry through intact.

**One suggestion on change 2 (terms optional).** The new co-occurrence rule — `selected_term_id` "present in a response whenever `terms` is" — became prose-only when `required` was dropped, but its response direction is exactly what `dependentRequired` expresses. As the schema stands, a response of

```json
{ "terms": [{ "id": "pt_pay_now", "...": "..." }] }
```

with no selection validates clean, which is the "amount matching no stated term" case the prose exists to prevent. Adding one line to the `payment` def:

```json
"dependentRequired": { "terms": ["selected_term_id"] }
```

rejects that shape while accepting everything legal. I checked the inverse direction too ("absent when it is not") and it deliberately does not belong in the schema: the def also validates update requests, where a Platform legally sends `selected_term_id` alone because `terms` is response-only — adding `"selected_term_id": ["terms"]` fails your own update-request example (I ran it: 298/1 vs 299/0 with the forward direction only). So the forward direction is machine-enforceable for free, and the absence direction stays prose, which seems like the right split.

The reasoning on 3 and 4 reads right to me: recurrence-as-policy avoids exactly the moving-obligation trap the Order freeze rule exists to prevent, and deferring split-payments composition rather than specifying allocation-order timing in prose is the honest call — the one-instrument-per-multi-schedule-term rule is at least cleanly wire-checkable when it fast-follows.

PR #731, opened 2026-08-14
docs(identity-linking): repair the garbled JWT grant single-use bullet

vishkaty:
The JWT grant single-use bullet in Security Considerations reads as a broken sentence: an explanatory clause was inserted mid-sentence, leaving the requirement itself split across a full stop.

Currently:

> Businesses **MUST** enforce single-use JWT;
> a short exp narrows the replay window, but only jti tracking closes it.
> authorization grants by tracking the `jti` claim within the grant's
> validity window.

This PR restores the requirement as one sentence and keeps the explanation after it, with `exp` and `jti` in code formatting to match their use in the neighbouring bullets. No normative change: the MUST, the mechanism, and the window are the same.

### DIFF
--- a/docs/specification/common/identity-linking/index.md
+++ b/docs/specification/common/identity-linking/index.md
@@ -1018,10 +1018,10 @@ field conveys the business's value prompt to the platform (e.g.,
 * **JWT grant lifetime.** JWT authorization grants **MUST** be short-lived;
   the `exp` claim **SHOULD** be no more than 60 seconds after `iat`. Short
   lifetimes limit the window for grant theft and replay.
-* **JWT grant single-use.** Businesses **MUST** enforce single-use JWT;
-  a short exp narrows the replay window, but only jti tracking closes it.
+* **JWT grant single-use.** Businesses **MUST** enforce single-use JWT
   authorization grants by tracking the `jti` claim within the grant's
-  validity window.
+  validity window. A short `exp` narrows the replay window, but only
+  `jti` tracking closes it.
 * **Grant relay.** Businesses **MUST NOT** store or forward JWT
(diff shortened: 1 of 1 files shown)


westonale:
Confirming this reads correctly now. The original had the clause boundary in the wrong place: "MUST enforce single-use JWT" ran into the `exp` aside, and then "authorization grants by tracking the `jti` claim" resumed a sentence that had already been closed, so the normative requirement was split across a parenthetical and lost its object.

The rewrite restores it to one requirement ("MUST enforce single-use JWT authorization grants by tracking the `jti` claim within the grant's validity window") and demotes the `exp` versus `jti` point to its own sentence, which is where it belongs: it is rationale, not a second MUST. Backticking `exp` and `jti` matches the surrounding bullets.

Worth saying that the rationale sentence is the part doing the work for implementers. A short `exp` feels like replay protection and is not, and that is a genuinely common mistake to make when reading quickly. Keeping it as a standalone sentence after the requirement is the right ordering.

niranjanmanjunath (approved):
This is a clean, non-normative documentation fix that repairs a spliced sentence in the
  Security Considerations section.

westonale:
Worth flagging that this is no longer only a docs nit: the garbled bullet is now inside the `v2026-08-25` tag at `docs/specification/common/identity-linking/index.md` lines 1021 to 1023, so implementers pinning to the release read a normative replay-protection MUST with a rationale clause spliced through the middle of it and an orphaned fragment after the semicolon.

Two independent contributors have now filed the same +3/-3 fix in nine days (this PR and #764, which was self-closed as a duplicate), which is a reasonable signal that the sentence is actively confusing readers rather than merely untidy. Checks here are green and it only needs a maintainer.

PR #644, opened 2026-07-28
docs: refactor AGENTS.md into a dual-persona playbook

damaz91:
# Description

Refactored `AGENTS.md` to cleanly separate instructions for repository contributors (Persona B) and external implementers (Persona A).
- Grouped reference sections for Checkout and Order capabilities to combine their REST and MCP transport specifications.
- Added ucp-schema `resolve` and `validate` helper commands under local command reference.
- Restored instructions advising AI agents on when to fetch external resource indices (like `llms.txt`).
- Updated Python and JavaScript/TypeScript SDK sections to list PyPI and npm package URLs alongside GitHub repositories and installation commands.
- Removed legacy SDK model generation references from contributor commands.

## Category (Required)

- [ ] **Core Protocol**: Changes to the base communication layer, global context, or breaking refactors. (Requires Technical Council approval)
- [ ] **Governance/Contributing**: Updates to GOVERNANCE.md, CONTRIBUTING.md, or CODEOWNERS. (Requires Governance Council approval)
- [ ] **Capability**: New schemas (Discovery, Cart, etc.) or extensions. (Requires Maintainer approval)
- [x] **Documentation**: Updates to README, or documentations regarding schema or capabilities. (Requires Maintainer approval)
- [ ] **Infrastructure**: CI/CD, Linters, or build scripts. (Requires DevOps Maintainer approval)
- [ ] **Maintenance**: Version bumps, lockfile updates, or minor bug fixes. (Requires DevOps Maintainer approval)
- [ ] **SDK**: Language-specific SDK updates and releases. (Requires DevOps Maintainer approval)
- [ ] **Samples / Conformance**: Maintaining samples and the conformance suite. (Requires Maintainer approval)
- [ ] **UCP Schema**: Changes to the `ucp-schema` tool (resolver, linter, validator). (Requires Maintainer approval)
- [ ] **Community Health (.github)**: Updates to templates, workflows, or org-level configs. (Requires DevOps Maintainer approval)

## Related Issues

None.

## Checklist

- [x] I have followed the [Contributing Guide](https://github.com/Universal-Commerce-Protocol/.github/main/CONTRIBUTING.md) (including Conventional Commits title requirements and `!` for breaking changes).
- [x] I have updated the documentation (if applicable).
- [x] My changes pass all local linting and formatting checks.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] New and existing unit tests pass locally with my changes.
[...]

PR #734, opened 2026-08-17
docs: fix stale references found in fact-accuracy audit

PulseCheckAI:
## Summary
- `AGENTS.md`: drops the "Regenerate SDK Models" step referencing `sdk/python/generate_models.sh` — that path has never existed in this repo (confirmed via full git history search); SDKs are published from separate org repos per README.md's own "Use our SDKs" section.
- `README.md`: corrects the `build_local.sh` description — it builds the site only and does not start a local server (confirmed by reading the script, which ends by printing the serve command rather than running one). `AGENTS.md` already documented this correctly as two separate steps.
- `pyproject.toml`: version pin left unchanged — an earlier commit on this branch bumped it to match the current release, then reverted that change since package versioning is the maintainers' release-process concern, not a doc correction.

## Test plan
- [ ] Confirm `sdk/python/generate_models.sh` genuinely doesn't exist anywhere in this repo's history
- [ ] Confirm `build_local.sh`'s actual behavior matches the corrected description

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -49,7 +49,6 @@ Validate your changes locally using these commands before considering a task com
 Then validate your changes with:
 
 * **Lint Schemas:** `ucp-schema lint source/` (after schema changes)
-* **Regenerate SDK Models:** `bash sdk/python/generate_models.sh` (after schema changes)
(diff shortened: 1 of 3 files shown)


damaz91:
Already covered by https://github.com/Universal-Commerce-Protocol/ucp/pull/644

PR #723, opened 2026-08-13
refactor: organize specification hierarchy by domain verticals (Shopping, Payment, Common) 

niranjanmanjunath:
# Description
This PR builds on @jingyli 's RFC https://github.com/Universal-Commerce-Protocol/ucp/issues/520 specifically addressing  Phase 1. Please note there's changes in this implementation from the original RFC, however the intent remains the same.

 ## Summary                                                                                                                                                  
This PR refactors the specification documentation structure by grouping specification documents into dedicated, domain-aligned namespaces and vertical directories (`shopping/`, `payment/`, `common/`, and `overview/`). This allows us to add more verticals in the future along the same hierarchy.                                                                                  

It aligns the documentation information architecture with the UCP protocol taxonomy (`{reverse-domain}.{service}.{capability}`), simplifies navigation, updates all internal links and macros, and implements comprehensive backwards-compatible redirects for legacy URLs.                                           
   
## Key Changes & Restructuring                                                                                                                                
                                                                                                                                                                
  ### 1. Shopping Vertical (docs/specification/shopping/)                                                                                                       
                                                                                                                                                                
  • Checkout Capability: Moved to shopping/checkout/ (index.md, rest.md, mcp.md, a2a.md, embedded.md).                                                          
  • Cart Capability: Moved to shopping/cart/ (index.md, rest.md, mcp.md, embedded.md).                                                                          
  • Catalog Capability: Moved to shopping/catalog/ (index.md, search.md, lookup.md, rest.md, mcp.md).                                                           
  • Order Capability: Moved to shopping/order/ (index.md, rest.md, mcp.md).                                                                                     
  • Playground: Relocated from specification/playground.md to specification/shopping/playground.md.                                                             
                                                                                                                                                                
  ### 2. Payment Vertical (docs/specification/payment/)                                                                                                         
                                                                                                                                                                
  • Core Guides & Protocols: Moved to payment/ (guide.md, template.md, tokenization.md, authentication.md, split-payments.md).                                  
  • Payment Actions: Grouped into payment/actions/ (device-data-collection.md, three-ds-challenge.md).                                                          
  • Payment Examples: Grouped into payment/examples/ (processor-tokenizer-payment-handler.md, platform-tokenizer-payment-handler.md, encrypted-credential-      
  payment-handler.md).                                                                                                                                          
                                                                                                                                                                
  ### 3. Common Vertical (docs/specification/common/)                                                                                                           
                                                                                                                                                                
[...]

Issue #520, opened 2026-06-16
[RFC] Proposal: Repository Refactoring for Multi-Vertical Support


jingyli:
>**Goal:** Improve UCP repo structure & developer tooling to facilitate new vertical support.
>**Status:** Proposal (Inviting feedback from Community & Tech Council discussion)
>**Related:** #499, #484 

## Summary

This proposal outlines the restructuring of the Universal Commerce Protocol (UCP) repositories and schemas to support multiple verticals (such as Food) alongside the existing Shopping specification. It details the phases required to transition from a retail-centric codebase to a modular, vertical-agnostic architecture, enabling UCP to scale as a multi-vertical standard.

## Motivation
UCP was started with "Shopping" as the first domain. However, as UCP expands to verticals like Food (menu exploration, cart building, checkout fulfillment), the current repository structure is inefficient.

Currently, retail-specific schemas and documentation are treated as the default "core" of UCP (e.g., checkout and cart specs living at the root of `docs/specification/`). Furthermore, foundational schemas like `signals.json` live under shopping specific paths (e.g., `source/schemas/shopping/types`), and shared transactional concepts like payment instruments are not consolidated as a horizontal layer.

To support multiple verticals without duplication and allow them to evolve independently, we must:
- Delineate core/common types from vertical-specific types
- Restructure the directory layout of the main `ucp` repo (specs and schemas)
- Align developer tooling (`samples`, `conformance`) and SDKs to support vertical-scoped constructs
- *[Not in scope for this RFC]* Design a standard process to deal with PR triage and review

This refactor will ensure that UCP remains robust, easy to navigate for readers, and simple to maintain for implementers.

### Phase 1: Restructure specification documentation
**Target:** [`ucp`](https://github.com/Universal-Commerce-Protocol/ucp)

**Problem statements**
- **Problem 1 (File Structure):** The `docs/specification/` directory contains retail-specific files at the root (e.g., `checkout.md`, `cart.md`).
- **Problem 2 (Rendering Logic):** The documentation rendering script (`main.py`) is tightly coupled to the Shopping vertical layout and assumes a single-vertical structure when it comes to macro definitions. We also need to make `reference.md` and site navigation bar clearly categorize capabilities under vertical headers (Shopping, Common).

**What lands**
[...]

PR #736, opened 2026-08-18
refactor!: Refactor common primitive types from shopping/types to common/types

jingyli:
# Description

A continuation of #436, addressing Phase 3 + 4 in that PR. Also related to #723 when it comes to schema types clean-ups that go hand-in-hand with specification refactoring.

Design choices made in this PR:
- Refactor 7 common primitives from `source/schemas/shopping/types/` to `source/schemas/common/types/`:
    - `context.json`
    - `signals.json`
    - `totals.json`
    - `total.json`
    - `policy.json`
    - `price_filter.json`
    - `price_range.json`
- Decided to leave `attribution.json` out-of-scope for now as we haven't fully vetted the design for x-vertical usages.
- No changes to documentation/specification rendering (taken care of in #723) and payment constructs (separate PR).

## Category (Required)

## Related Issues

Phase 2 in #520 

## Checklist

Manually validated via local ucp.dev to ensure that none of the reference links in the existing `reference.md` will break.

Also a screenshot example of how the inline rendering now looks like in capability pages:
inline-rendering


igrigorik (commented):
Overall, looks good. Two documentation flags before we land this...

### Can we preserve inline schema rendering?

Moving these schemas into `common/types/` activates the existing shared-type behavior in `schema_fields`: inline field tables on the Catalog, Cart, Checkout, Fulfillment, and Order pages are replaced with links to the Schema Reference. That's a lot of external refs that were previously directly visible in the document, which I think is a meaningful regression.

This should require only small surgery in `main.py`. I would make inline rendering the default and make link-only rendering an explicit opt-out:

```python
def schema_fields(entity_name, spec_file_name, inline=True):
    ...

    if (
        not inline
        and spec_file_name != "reference"
        and _resolves_to_shared_type(base_name) is not None
    ):
        return _render_shared_type_link(base_name, spec_file_name)
```

Call sites that intentionally want only the canonical-reference link can then opt out explicitly:

```jinja
{{ schema_fields('types/postal_address', 'checkout', inline=False) }}
```

### Nit: stale `totals.json` reference

The complete Checkout example in `docs/documentation/schema-authoring.md` still contains:

```json
"totals": {"$ref": "types/totals.json"}
```

After this move, the correct relative reference from `shopping/checkout.json` is `../common/types/totals.json`.

jingyli:
Thanks @igrigorik!

## Can we preserve inline schema rendering?

I'd like to take a step back and make sure we are aligned on the rendering principles here. Based on #536, we aligned that:
- Inline rendering on capabilities page: any vertical specific types (e.g., `shopping/types/buyer.json`, `shopping/types/line_item.json`)
- Redirect to `reference.md`: reserved for `common/types` only

Personally I think the increase in external refs/redirects from this PR is WAI and consistent with the expectation we want to set with developers. The net new effects:
| Capability Page | Pre-PR #736 (PR #536 Baseline) | Post-PR #736 | Newly Redirected Types (+Δ) |
| :--- | :---: | :---: | :--- |
| **`cart`** | 5 | **8** | +3 (`Context`, `Signals`, `Total`) |
| **`checkout`** | 8 | **11** | +3 (`Context`, `Signals`, `Total`) |
| **`catalog/index`** | 5 | **8** | +3 (`Context`, `Signals`, `Price Range`) |
| **`catalog/search`** | 2 | **3** | +1 (`Price Filter`) |
| **`fulfillment`** | 1 | **2** | +1 (`Total`) |
| **`order`** | 1 | **2** | +1 (`Total`) |
| **`embedded-checkout`** | 1 | **1** | 0 *(no refactored types referenced)* |
| **Total** | **23** | **35** | **+12 newly redirected instances** |

I'm also fine to always inline schema rendering by default on capabilities documentation, but I want to make sure it then means we are doing it consistently to all `common/types` (and not have an split between the two styles selectively), which means technically it will remove all the effects of the redirect predicates and force them to render inline on capability pages.

## Nit: stale totals.json reference

Good catch! I realized that there was one more stale reference to (`link.json`) so fixed both in [2de5f4a](https://github.com/Universal-Commerce-Protocol/ucp/pull/736/commits/2de5f4a31bf8087a8191ec56ed6ba30ffd900ad8).

igrigorik:
> which means technically it will remove all the effects of the redirect predicates and force them to render inline on capability pages.

@jingyli my vote is for inline render. The main reason to avoid it is due to risk of them getting out of sync in different places, but since these are auto-resolved + inlined refs and we rebuild the whole tree, that risk is a no-op. By inlining we get the benefit of much easier and friendlier reading and comprehension -- no need to jump across tabs, etc.

jingyli:
@igrigorik Sounds good, PTAL at [623b2c8](https://github.com/Universal-Commerce-Protocol/ucp/pull/736/commits/623b2c87f9b1111ccc2daa53f72e0fb1b8bb8848) - this updates both `schema_fields` and `extension_schema_fields` **consistently to by default force inline rendering on all `common/types` schemas** (also attached a sample example screenshot of `checkout.md` in the PR description).

igrigorik (approved):
@jingyli pushed a small fix: https://github.com/Universal-Commerce-Protocol/ucp/pull/736/changes/de7f17ea6737a2d264e4eaed584e0cf27fc7114f — ptal.

LGTM.

main.py
```diff
@@ -1091,7 +1082,7 @@ def _render_shared_type_link(canonical_name, spec_file_name):
 
   # --- MACRO 1: For Standalone JSON Schemas ---
   @env.macro
-  def schema_fields(entity_name, spec_file_name):
+  def schema_fields(entity_name, spec_file_name, render_inline=True):
```
gsmith85: Nit / Type Annotations: Consider adding PEP 484 type annotations for consistency with the top-level helpers in `main.py`:

```python
def schema_fields(
    entity_name: str,
    spec_file_name: str,
    render_inline: bool = True,
) -> str:
```

main.py
```diff
@@ -1136,14 +1129,18 @@ def schema_fields(entity_name, spec_file_name):
...
+    # Dynamically choose rendering mechanisms (redirect to reference.md
+    # for `common/types` vs. inline rendering) based on specifications
+    # from capability-page callers via `render_inline`.
+    # By default, `render_inline` is always true to promote better readability,
+    # but callers can override it to false to explicitly trigger the redirect
+    # rendering for non vertical-namespaced schemas. reference.md
+    # itself always renders full tables via auto_generate_schema_reference.
+    # base_name is passed (suffix stripped) so the link anchor targets
+    # the canonical heading.
     if (
```
gsmith85, line 1141: Nit: Note that `_resolves_to_shared_type(base_name)` strips suffixes but assumes `base_name` does not already end with `.json`. If a caller passes a `.json` path or pointer with `#`, `_resolves_to_shared_type` can construct `.json.json`. Normalizing `clean = str(ref_path).split('#', 1)[0]` and `name = Path(clean).stem` in `_resolves_to_shared_type` avoids this edge case.

main.py
```diff
@@ -1176,35 +1173,37 @@ def schema_fields(entity_name, spec_file_name):
...
     """Parse a standalone JSON Schema file and render a table.
 
     Usage: {{ extension_schema_fields('fulfillment_option') }}
 
-    When the embedded schema lives under `common/types/`, the macro
-    emits a link to the canonical entry on `reference.md` instead of
-    rendering the table inline. Vertical-namespaced extensions render
-    inline as before. See `_resolves_to_shared_type` for scope and
-    #412 for motivation.
+    When the embedded schema lives under `common/types/`, the macro looks
```
gsmith85: Nit: Phrasing in the docstring (*"the macro looks at the additional `render_inline` args to understand whether..."*) is a bit informal. Consider:

```python
"""When the embedded schema lives under `common/types/` and `render_inline`
is False, the macro emits a link to the canonical entry on `reference.md`
instead of rendering the table inline. Vertical-namespaced extensions always
render inline."""
```

main.py
```diff
@@ -1176,35 +1173,37 @@ def schema_fields(entity_name, spec_file_name):
...
 
     """
     # entity_name has the form `<file>.json#/$defs/<def>`. Redirect to
     # reference.md only when `<file>.json` resolves to a
-    # `common/types/` schema; vertical-namespaced extensions keep
-    # their inline render so same-page `$defs` anchors continue to
-    # resolve. file_part preserves any `types/` prefix from the
-    # caller so the lookup is unambiguous against the basename
-    # namespace.
+    # `common/types/` schema and `render_inline = False`; vertical-namespaced
```
gsmith85: Nit: Use `render_inline is False` or `not render_inline` rather than assignment formatting `render_inline = False`.

gsmith85 (approved):

main.py
jingyli: I agree with the suggestion to strengthen our parsing logic in `_resolves_to_shared_type`, though just calling out the gap more explicitly here with our current logic:

We may unintentionally introduce false negatives: i.e. `schema_fields('context.json', ..., render_inline=False)` will cause `_resolves_to_shared_type` to attempt resolve whether `/common/types/context.json.json` exists - which will result in `None` being returned from the function and the redirect rendering will not happen.

Another downside here is that it also expects the author to read through `main.py` to understand what `ref_path` we truly expect (which may or may not be a huge problem - if I play the devil's advocate here, we are forcing consistency since all other forms will result in broken documentation rendering and and the author will need to correct/fix the incorrect `ref_path).

amithanda:
Overall, this refactor is in great shape and is a solid step forward for the multi-vertical roadmap (#520). The switch to `render_inline=True` by default in `main.py` is a great documentation UX win.

Before we land this breaking refactor, here are a few non-blocking suggestions to address before merging:

### 1. SDK Model Regeneration
- In the PR checklist, `- [ ] I have regenerated Python Pydantic models by running generate_models.sh under python_sdk` is currently unchecked.
- Because moving these 7 schemas to `common/types/` changes class namespaces and import paths (e.g., `ucp.models.shopping.types.context` $\rightarrow$ `ucp.models.common.types.context`), please ensure SDK models are regenerated and verified prior to release (and this box is checked).


### 2. Schema Authoring Guide Update (`render_inline`)
- The new `render_inline` parameter in `schema_fields` and `extension_schema_fields` is great. Could we add a brief example in `docs/documentation/schema-authoring.md` demonstrating how capability authors can pass `render_inline=False` (e.g., `{{ schema_fields('types/context', 'checkout', render_inline=False) }}`) when they explicitly want to emit a reference link instead of an inlined table?

amithanda (approved):

jingyli:
Thanks @amithanda!

## 1. SDK Model Regeneration
Ack on this - I think a known gap is that our current generation script cannot be executed on draft PRs (even if there is a way to hack around it, it's not very well documented) and we will need to think about how to streamline that so that checkbox in the PR template makes sense for developers (@gsmith85 FYI).

FWIW, when I tried regenerating the Python SDK based on the current `main` branch, there are already some warnings popping up to which we should sift through as part of the upcoming release.

## 2. Schema Authoring Guide Update (render_inline)
I gave this some thoughts when updating the script (e.g., documenting down our principles like always force inline type rendering on specification pages). But I think right now `Schema Authoring` is very targeted to `conventions for authoring UCP JSON schemas` and this kind of content/examples don't really fit well.

I think as a follow-up, maybe we should create a sibling page called `Specification Authoring` that would capture this, along with how our specification markdown files should be structured (as a consequence of #723), etc. Let me track this as a TODO to follow-up!

PR #744, opened 2026-08-18
add path targeting to Request Constraints

igrigorik:
#655 defines a reusable closed Constraint Expression grammar, but the response location alone does not always identify the request data to validate. For example, a constraint attached to a response Line Item could mean the item at the same array position, the item with the same identity, or some other request object. The ambiguity becomes unavoidable when arrays reorder or when response and request shapes differ, such as an available payment instrument constraining a submitted instrument.

---

This PR adds an optional, outer-only `path` to Request Constraints. An explicit `path` is a complete RFC 9535 JSONPath query evaluated against the next logical UCP request to the same resource. When omitted, the effective path is derived from the RFC 9535 Normalized Path of the structured response object whose `ucp` member contains the constraint.

The omission behavior preserves concise ambient authoring. A constraint on the response root derives `$`; a constraint on a nested object derives that object's positional path. That positional correspondence is deterministic but does not imply stable identity. Authors use an explicit query when a rule must survive reordering, select multiple objects, or bridge different response and request shapes.

A path may select zero, one, or many objects. Zero matches have no effect, while every selected object must satisfy the complete Constraint Expression. Overlapping constraints compose conjunctively without precedence or override, and the Business must ensure potentially overlapping expressions are jointly satisfiable. These checks narrow the resolved request schema; they cannot make an otherwise invalid request valid.

## Binding shape

The chosen wire shape keeps `path` beside the root members of the expression:

```json
{
  "path": "$['line_items'][?@['id'] == 'line_123']",
  "properties": {
    "quantity": {"const": 100}
  }
}
```

One considered alternative was separating path from constraint expression: 

```json
{
  "path": "$['line_items'][?@['id'] == 'line_123']",
  "expression": {
    "properties": {
      "quantity": {"const": 100}
    }
  }
}
```

The benefit of the latter is that keeps expression strict subset of JSONSchema. The cost is permanent wire noise in the common case. Most constraints use ambient placement and omit `path`, so even a simple local requirement would need an otherwise meaningless wrapper:

```json
{
  "expression": {
    "required": ["billing_address"]
  }
}
```

The selected shape optimizes for the authoring and wire surface rather than schema-factoring purity. `constraint_expression.json` remains direction-neutral and contains no `path`; `request_constraints.json` is the closed binding that admits `path` only at its root and re-exposes the root Object Constraint members. Recursive constraints continue through the pure expression schema, so nested `path` members remain invalid. The trade-off is that future changes to the root Object Constraint must also be reflected in `request_constraints.json`.

This preserves the core invariant that one binding carries one schema. It avoids introducing a list of independently targeted rules, repeated navigation, target precedence, or another orchestration model alongside JSON Schema.

### DIFF
--- a/docs/specification/overview.md
+++ b/docs/specification/overview.md
@@ -177,58 +177,51 @@ uses the shared representation.
 After capabilities and extensions are negotiated, the resolved UCP request
 schema defines the fields and structure allowed for an operation. In an
 authoritative response, a Business can use `ucp.request_constraints` to signal
-additional rules it will apply when evaluating request data in a subsequent
+additional rules it will apply when evaluating request data in the next
 Platform request. For example, it can constrain a Line Item quantity to exactly
 `100` sale-basis steps or require a submitted payment instrument to include
 `billing_address`. A Platform can evaluate these constraints before submission,
 avoiding a round trip for request data the Business has already indicated it
 will reject.
 
-`ucp.request_constraints` applies only to authoritative operation responses. A
-Business or Platform **MUST** ignore the member if it receives it in a discovery
-profile or operation request.
+`ucp.request_constraints` is used only in authoritative operation responses and
+has no effect in discovery profiles or operation requests.
 
 ### Validation model
 
-Request Constraints combine with the resolved request schema to form the
-effective rules for subsequently submitted request data:
+A submitted request is valid under Request Constraints only if it satisfies the
+resolved request schema and every object selected by Request Constraints
+satisfies the corresponding Constraint Expression. Request Constraints only
+narrow the resolved request schema; they cannot make a request valid when that
+schema rejects it. A Business evaluates the request as follows:
 
 ```text
-effective_request_schema = allOf(
-  resolved_request_schema,
-  request_constraints
-)
+valid = validate(resolved_request_schema, request)
 
-validate(effective_request_schema, submitted_request_data)
-  = validate(resolved_request_schema, submitted_request_data)
-    AND validate(request_constraints, submitted_request_data)
+for each constraint:
+  objects = select(request, effective_path(constraint))
+  valid = valid AND validate_all(
+    constraint_expression(constraint),
+    objects
+  )
+
+return valid
 ```
 
-An implementation can construct the effective schema with JSON Schema `allOf`
-and validate the submitted request data once, or run the two validations
-separately. Both validations must pass for authoritative Business evaluation.
-Request Constraints can narrow the inputs allowed by the resolved request
-schema, but cannot make an input valid when that schema rejects it.
-
-A Platform **MUST** validate the `ucp.request_constraints` value against the
-shared [Request Constraints](site:schemas/common/types/request_constraints.json)
-schema before using it for preflight. This is ordinary JSON Schema Draft 2020-12
-validation.
+A Platform **MAY** perform the same checks as preflight. For each chosen value,
+the Platform **MUST** use its effective path and complete Constraint Expression.
+The Platform **MAY** submit the request regardless of the preflight result;
+Business evaluation is authoritative.
 
 ### Constraint Expression
 
-The Request Constraints schema binds the shared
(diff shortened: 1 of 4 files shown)


source/schemas/common/types/request_constraints.json
```diff
@@ -2,6 +2,26 @@
...
+      "$ref": "constraint_expression.json#/properties/required"
+    },
+    "properties": {
+      "type": "object",
+      "description": "Constraints keyed by property name.",
+      "additionalProperties": {
+        "oneOf": [
+          {"$ref": "constraint_expression.json"},
+          {"$ref": "constraint_expression.json#/$defs/value_constraint"}
+        ]
```
raginpirate:
```suggestion
      "$ref": "constraint_expression.json#/properties/properties"
```

raginpirate (approved):
Thanks for all the deep thinking here Ilya. Approving, with a note from our discussion in the deep dive earlier.

Another solution is to put checkout.ucp.request_constraints on the response root as a single property, which carries an array of constraints and their paths in one place. As you mentioned, an implementer could even choose to do this.

In a vacuum, I think thats the simple solution, as it avoids having to walk the tree to assemble all the fragments you'll evaluate to restrict your json schema. However, I'll concede that it is very nice to keep the constraints authored in a semantically-relevant place in the schema; something I care deeply about for our payment requests 😄

Lets land this thing!

gsmith85:
Thanks for driving this through, @igrigorik! This looks great and aligns with what we discussed in the deep dive. 

Adding RFC 9535 path targeting gives us the precision needed for array reordering and payment shape bridging while keeping the common ambient case clean and flat. LGTM! 🚀

(Add `feat:` to the title to make CI happy and ship it).

gsmith85 (approved):
Thanks for driving this through, @igrigorik! This looks great and aligns with what we discussed in the deep dive.

Adding RFC 9535 path targeting gives us the precision needed for array reordering and payment shape bridging while keeping the common ambient case clean and flat. LGTM! 🚀

docs/specification/overview.md
```diff
@@ -177,167 +177,266 @@ uses the shared representation.
...
+No other member is admitted at either grammar position.
+
+### Path
+
+`path` is optional. When omitted, the effective path is the
+[RFC 9535](https://www.rfc-editor.org/rfc/rfc9535.html){ target="_blank" }
+Normalized Path from the authoritative response root to the structured response
+object whose `ucp` member contains `request_constraints`. When the Business
+provides `path`, the Business **MUST** use a complete RFC 9535 JSONPath query
+evaluated against the next logical UCP request to that resource.
```
jamesandersen, line 241: Two little nits here:
* I think it's useful to take a little extra care to contrast how these `path` values are used relative to others that can be returned in the response but which target the *response* (rather than next `request`)
* Theres a potentially strange behavior where a business could embed `request_constraints` *anywhere* in a response that and specify a `path` to the intended subsequent request node without any logical change of behavior.   Perhaps some acknowledgement of this behavior can help avoid confusion for implementers?


```suggestion
evaluated against the next logical UCP request to that resource.  Note this differs
from other RFC 9535 paths such as `messages` which are evaluated against the
response body itself. While the response node in which `ucp.request_constraints`
is provided is no longer strictly relevant when `path` is provided, businesses
SHOULD select a semantically related node.
```

source/schemas/common/types/request_constraints.json
igrigorik: Nice.

jamesandersen (commented):
I was a bit late to the party here ... but agree this is a good catch and addition @igrigorik  - thanks!

docs/specification/overview.md
igrigorik: Ty James, addressed in https://github.com/Universal-Commerce-Protocol/ucp/pull/655/commits/f471521bb17b247c8d7316e72ca2cbd855eef832.

PR #753, opened 2026-08-20
fix!: Location spatial relations and Lookup correlation

igrigorik:
Addresses some of the outstanding feedback on #589:

- moves `distance` and `serves` out of `filters.geo` and makes them independent root-level relations in Search and Lookup;
- defines deterministic distance and serviceability semantics without contextual fallback;
- preserves `filters` for predicates over inherent or current Location facts;
- mirrors Catalog’s batch Lookup response shape with required `inputs[]` correlation on every returned Location;
- keeps empty, hint-only, pagination-only, and filters-only Search requests as bounded browse operations.

## Why spatial relations are not filters

The previous `filters.geo` shape mixed two different kinds of constraints:

| Input | Meaning |
| :--- | :--- |
| `distance` | A relation between a candidate Location and an explicit Platform-supplied center and radius |
| `serves` | A relation between a candidate Location and an explicit Platform-supplied service target |
| `filters` | Predicates over inherent or current Location facts such as hours, amenities, and inventory |
| `context` / `signals` | Provisional ranking and localization hints, never spatial proof |

`distance` and `serves` now sit at the request root because their operands are external to the Location. Every supplied spatial relation and filter predicate combines with AND. Keeping the relations independent also permits requests such as “within five miles of this point and able to serve that address.” If both relations use the same point, the request repeats it explicitly rather than relying on hidden cross-field inheritance.

```json
{
  "distance": {
    "center": {
      "latitude": 40.707,
      "longitude": -74.011
    },
    "max": 10000
  },
  "serves": {
    "address": {
      "address_country": "US",
      "postal_code": "11201"
    }
  },
  "filters": {
    "inventory": [
      {
        "id": "item_id_phone_15_pro",
        "availability_status": "in_stock"
      }
    ]
  }
}
```

A returned Location must satisfy all three conditions:

```text
within 10,000 meters of distance.center
AND can provisionally serve serves.address
AND satisfies the inventory predicate
```

### `distance`

- requires an explicit `center` and inclusive `max` in meters;
- compares the unrounded shortest distance;

### `serves`

- contains exactly one target: `point`, coarse `address`, or a negotiated reverse-domain extension target;
- treats the supplied target as authoritative and never derives one from contextual hints;
- means that at least one currently available method at the Location can provisionally serve the target;
- does not expose coverage geometry, identify the qualifying method, reserve capacity, or guarantee checkout success;
- rejects unsupported or unevaluable targets rather than ignoring them or broadening results.

## Bounded browse

Search remains valid without `query`, `distance`, or `serves`. Empty, context-only, signals-only, pagination-only, and filters-only requests are bounded browse operations over the Business’s policy-controlled selection. They never mean an unbounded export, and an omitted `pagination.limit` retains the existing default of 10.

## Lookup correlation

Location Lookup now mirrors Catalog’s response shape: every returned Location carries a non-empty `inputs[]` array identifying the request identifiers that resolved to it.

```json
{
  "ids": ["loc_downtown", "downtown-store"]
}
```

```json
{
  "locations": [
    {
      "id": "loc_downtown",
      "inputs": [
        {"id": "loc_downtown"},
        {"id": "downtown-store"}
      ],
      "name": "Downtown Store"
    }
  ]
}
```

This supports unordered batch responses, secondary identifiers, multiple identifiers resolving to one Location, and one identifier resolving to multiple Locations.

Requiring `inputs[]` for direct canonical-ID matches adds minor payload overhead, but preserves the same uniform, schema-enforceable correlation rule as Catalog. Location does not copy Catalog’s `match` classification because it has no equivalent of product-to-featured-variant resolution.

## Batch limits

Batch limits are applied after duplicate identifiers are removed. When a request exceeds the Business maximum, the Business processes the first maximum number of distinct identifiers in request order and returns a successful partial response.

The response includes an informational message:

```json
{
  "type": "info",
  "code": "batch_limit_applied",
  "content": "Processed the first 2 of 3 distinct identifiers; 1 identifier was not processed."
}
```

The omitted suffix was not evaluated and must not be interpreted as unresolved. The Platform may submit it in a subsequent request. REST returns HTTP 200 and MCP returns a successful JSON-RPC result for this business outcome.

### DIFF
--- a/docs/specification/catalog/mcp.md
+++ b/docs/specification/catalog/mcp.md
@@ -710,5 +710,6 @@ A conforming MCP transport implementation **MUST**:
 4. Return successful result for lookup requests; unknown identifiers result in fewer products returned (MAY include informational `not_found` messages).
 5. Validate tool inputs against UCP schemas.
 6. Return products with valid `Price` objects (amount + currency).
-7. Support cursor-based pagination with default limit of 10.
+7. Support cursor-based pagination for Search according to the shared
+    pagination contract (see [Pagination](search.md#pagination)).
 8. Return `-32602` (Invalid params) for requests exceeding batch size limits.
--- a/docs/specification/catalog/rest.md
+++ b/docs/specification/catalog/rest.md
@@ -602,6 +602,7 @@ A conforming REST transport implementation **MUST**:
 
 1. Implement endpoints for each catalog capability advertised in the business's UCP profile, per their respective capability requirements ([Search](search.md), [Lookup](lookup.md)). Each capability may be adopted independently. When the Lookup capability is advertised, both `/catalog/lookup` and `/catalog/product` MUST be available.
 2. Return products with valid `Price` objects (amount + currency).
-3. Support cursor-based pagination with default limit of 10.
+3. Support cursor-based pagination for Search according to the shared
+    pagination contract (see [Pagination](search.md#pagination)).
 4. Return HTTP 200 for lookup requests; unknown identifiers result in fewer products returned (MAY include informational `not_found` messages).
 5. Return HTTP 400 with `request_too_large` error for requests exceeding batch size limits.
--- a/docs/specification/catalog/search.md
+++ b/docs/specification/catalog/search.md
@@ -63,16 +63,19 @@ merchants MAY support additional custom filters via `additionalProperties`.
 
 ## Pagination
 
-Cursor-based pagination for list operations. Cursors are opaque strings
-that implementations MAY encode as stateless keyset tokens.
+Cursor-based pagination for list operations. Cursors are opaque strings. A
+Business **MAY** encode them as stateless keyset tokens.
 
 ### Page Size
 
-The `limit` parameter is a requested page size, not a guaranteed count.
-Implementations SHOULD accept a page size of at least 10. When the
-requested limit exceeds the implementation's maximum, implementations
-MAY clamp to their maximum silently — returning fewer results without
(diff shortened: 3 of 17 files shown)


jingyli:
Thanks @igrigorik! Per our discussion, pushed up an update [3e34a6b](https://github.com/Universal-Commerce-Protocol/ucp/pull/753/commits/3e34a6bdfef124b26f2c99944f45d9e8f9caa7d4), PTAL! I will merge it back to #589 so we can take the final review there 😃 

2 things to call out:
- **Loosening up the default `pagination.limit`**: I'm unable to find the mirroring guideline in catalog search/lookup so unsure why we want to enforce this default limit. From a platform's perspective, if the request is sent in the form of an open browse (omitting `pagination`), the motivation is to get the most data back in the response as permitted by the business. Instead of enforcing this arbitrary 10 value, I want to strongly suggest to let Businesses apply their arbitrary threshold that they are comfortable with.
- **Removing unnecessary `ucp_request: required` annotation**: These fields are already required in the json type schema so I'm unsure why we need to define these duplicated required annotation again.

igrigorik:
@jingyli we have a `default` specced in schema, which is inherited by location via pagination.json. That said, I think your nudge is correct, the intent of the contract is to require that some pagination limit should exist but defer to the business on what that ought to be. Updated to use aligned language across catalog and location search, ptal.

Also made a few other small adjustments. If all that looks good, let's merge and continue in #589.

PR #759, opened 2026-08-21
fix: preserve fragments in root spec links

igrigorik:
Follow-up to #723, which moved the specification overview from `overview.md` to `overview/index.md` as part of the vertical documentation hierarchy. Root documentation links were rewritten to the versioned specification but for links containing fragments, the rewriter checked for `index.md` before separating the fragment, so links such as:

```text
/latest/specification/overview/index.md#protocol-version
```

were left pointing at the generated Markdown source instead of the rendered page:

```text
/latest/specification/overview/#protocol-version
```

The existing link checker did not catch this because the generated `.md` file exists and `latest/specification` paths are otherwise ignored during partial builds.

## Changes

- Separate fragments before normalizing `index.md` and restore them on the canonical rendered URL.
- Reject internal links to raw `.md` files before applying `.linkignore` patterns.
- Continue allowing external Markdown links, such as GitHub source files.

## Checklist


scripts/check_links.py
```diff
@@ -178,6 +170,27 @@ def get_file_ids(path):
       anchor_part = parsed.fragment
       path_part = unquote(path_part)
 
+      # Built documentation should link to rendered pages, never copied source
+      # Markdown. A raw .md target can exist and fool the ordinary existence
```
jingyli: nit: Lower case `m` in `Markdown` looks more natural when used in the sentence context?

Also maybe we can add an example in this comment blocker here to clearly callout what we are trying to catch:
```
#  Example of a broken link: https://ucp.dev/specification/cart.md/#cart-vs-checkout
```

jingyli (approved):

amithanda (approved):

PR #757, opened 2026-08-21
feat: Introduce anyOf constraint with payments example

raginpirate:
# Description

Resolving our open follow-up to #655, introducing the anyOf constraint to the base UCP constraints grammar.
Added a simple example using it with payments, and #424 is stacked on top of this to extend the usage of constraints across payments normatively.

## Category (Required)

_Please select one or more categories that apply to this change._

- [x] **Core Protocol**: Changes to the base communication layer, global context, or breaking refactors. (Requires Technical Council approval)
- [ ] **Governance/Contributing**: Updates to GOVERNANCE.md, CONTRIBUTING.md, or CODEOWNERS. (Requires Governance Council approval)
- [ ] **Capability**: New schemas (Discovery, Cart, etc.) or extensions. (Requires Maintainer approval)
- [ ] **Documentation**: Updates to README, or documentations regarding schema or capabilities. (Requires Maintainer approval)
- [ ] **Infrastructure**: CI/CD, Linters, or build scripts. (Requires DevOps Maintainer approval)
- [ ] **Maintenance**: Version bumps, lockfile updates, or minor bug fixes. (Requires DevOps Maintainer approval)
- [ ] **SDK**: Language-specific SDK updates and releases. (Requires DevOps Maintainer approval)
- [ ] **Samples / Conformance**: Maintaining samples and the conformance suite. (Requires Maintainer approval)
- [ ] **UCP Schema**: Changes to the `ucp-schema` tool (resolver, linter, validator). (Requires Maintainer approval)
- [ ] **Community Health (.github)**: Updates to templates, workflows, or org-level configs. (Requires DevOps Maintainer approval)

## Checklist

- [x] I have followed the [Contributing Guide](https://github.com/Universal-Commerce-Protocol/.github/blob/main/CONTRIBUTING.md) (including Conventional Commits title requirements and `!` for breaking changes).
- [x] I have updated the documentation (if applicable).
- [x] My changes pass all local linting and formatting checks.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] New and existing unit tests pass locally with my changes.
- [x] (For Core/Capability) I have included/updated the relevant JSON schemas.
- [ ] I have regenerated Python Pydantic models by running generate_models.sh under python_sdk.

PR #760, opened 2026-08-21
Ig/anyof review suggestions

igrigorik:
Nits/suggestions for #757.

**`docs: complete abridged payment instrument examples`** — Two `complete` examples elided `id`, `handler_id`, and `type` behind a `// ... more instrument required field` comment, and one also omitted `credential.type`. They only validate because shipped `ucp-schema` never applies the self-root `allOf` requirements in `selected_payment_instrument`, so `payment.instruments[]` is never checked against `payment_instrument`'s required members. [ucp-schema#44](https://github.com/Universal-Commerce-Protocol/ucp-schema/pull/44) fixes that; built against its head, this branch's base reports `316 passed, 2 failed` at exactly these two lines. Completing them costs one line each. 

**`feat: require non-empty anyOf branches in the constraint grammar`** — An empty Object Constraint is a valid no-op everywhere else, but an empty `anyOf` branch is satisfied by every object, so it neutralizes the alternation rather than doing nothing. Adds `minProperties: 1` as a sibling to the recursive `$ref` in both grammar schemas (legal 2020-12 sibling composition, applies at every nesting depth) and mirrors the constraint into both grammar tables.

**`docs: discriminate the anyOf credential example`** — The example's undiscriminated form does not enforce what the surrounding prose claims: `{"card_number_type": "anything", "cvc": "1"}` satisfies the `cvc` branch. Pins `card_number_type` with `const` in each branch and names it in that branch's `required`, matching the idiom #424 already uses in `payment/guide.md`. Uses the existing closed enum in `card_credential.json`, so it introduces no dependency on #424's new credential types.

Also documents the consequence that pinning creates — because every branch pins the discriminator, the branch set closes the accepted values, so a `dpan` credential valid under `card_credential.json` satisfies neither branch — and replaces the "would restate the path" justification. Request Constraints conjoin, so two separately targeted constraints would require *both* `cvc` and `cryptogram`; the real alternative is discriminating in the path filter, which puts conditional logic in the selector.

### DIFF
--- a/docs/documentation/schema-authoring.md
+++ b/docs/documentation/schema-authoring.md
@@ -541,7 +541,7 @@ registered member.
 
 | Position | Admitted members | Shape |
 | :-- | :-- | :-- |
-| Object Constraint | `required`, `properties`, `anyOf` | `required` contains unique field names; `properties` maps field names to Object or Value Constraints; `anyOf` is a non-empty array of Object Constraints; an empty object is a no-op. |
+| Object Constraint | `required`, `properties`, `anyOf` | `required` contains unique field names; `properties` maps field names to Object or Value Constraints; `anyOf` is a non-empty array of non-empty Object Constraints; an empty object is a no-op except as an `anyOf` branch. |
 | Value Constraint | `enum`, `const` | `enum` is non-empty and unique; at least one member is present; both apply when present together. |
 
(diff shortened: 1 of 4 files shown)


raginpirate (approved):
Good improvements!

PR #765, opened 2026-08-24
fix!: make Location amenities self-describing

igrigorik:
Amends #589: clarify fulfillment-handoff, fix refs, remove unnecessary 'offerings' concept, plus...

We previously discussed amenities as a map with the premise of potential amenity-specific metadata, but deferred. However, re-reading the PR with fresh eyes, there is a bigger gap and violation that we missed: amenities are not self-describing. We bought collision free and extensible contract, but we do not provide a mechanism for a platform to meaningfully present a `com.foo.custom` amenity - this is a gap and footgun, which we already address in other places by requiring a business-provided description and a presentation contract for the platform. 

## Self-describing amenities

Location responses now represent amenities as a map:

```json
{
  "amenities": {
    "dev.ucp.amenity.shopping.curbside_pickup": {
      "description": "Curbside pickup"
    },
    "com.example.amenity.in_person_installation": {
      "description": "In-person installation by appointment"
    }
  }
}
```

The map key remains the amenity identity and the exact value used for filtering. Each value **MUST** contain a nonempty, plain-text, buyer-facing `description`; the Business **SHOULD** make it a short, self-contained label or phrase suitable for direct use in a compact list. An entry without `description` is schema-invalid, and the Platform **MUST** reject the response rather than infer text from the key.

**Request filtering remains identifier-based:** `filters.amenities` is still an array of exact keys with AND semantics. When that filter is present, every returned Location must disclose every requested key in its `amenities` map.

## Fulfillment handoff

This PR also makes the Location-to-Fulfillment boundary explicit:

- `Location.id` identifies the Location.
- A Platform **MAY** submit a returned Location ID as `selected_destination_id` on an applicable Fulfillment method.
- In that use, the ID selects only the destination; `fulfillment.methods[].type` identifies the mode.
- The Business **MUST** revalidate the selected Location's current availability and terms; recognizing the ID neither reserves inventory nor guarantees eligibility.


docs/specification/common/location/index.md
```diff
@@ -56,21 +57,26 @@ This is vertical-agnostic and enables key commerce flows such as:
...
-    business-scoped `location.id` values. These IDs are referenced further in other requests & responses
-    (e.g., associating product variants to specific locations in Catalog filters, passed directly
-    in `selected_destination_id` to indicate pickup fulfillment mode).
+The Location capability provides the foundation for localized commerce by
+integrating tightly with other capabilities (like Catalog, Cart, and Checkout
+in Shopping):
+
+1. **Stable Identifiers**: Location Search and Lookup return stable,
+    Business-scoped `Location.id` values. A Platform **MAY** submit a returned
+    ID as `selected_destination_id` on an applicable Fulfillment method. The
```
jingyli: > A Platform **MAY** submit a returned ID as `selected_destination_id` on an applicable Fulfillment method. The method's `type`, not the Location ID, identifies the fulfillment mode; the ID selects only the destination.

Just a nit: I don't think we should reference into Fulfillment's schema shape in the Locations documentation here as part of the core sentence (instead of having it as part of the `e.g.,` parentheses) to avoid leaky abstractions. Suggest to reword this bullet here to:

```
1. **Stable Identifiers**: Location Search and Lookup return stable, Business-scoped `Location.id` values 
representing physical entities. A Location ID selects only the physical entity; it does not determine the 
interaction mode, guarantee availability, or bind operational terms. Downstream capabilities authoritatively 
negotiate and revalidate all terms during transaction processing (for example, see [Selection and Location 
Identity](../../fulfillment.md#selection-and-location-identity)).
```

docs/specification/common/location/index.md
```diff
@@ -190,6 +196,28 @@ Platform **MAY** present `title` according to its presentation policy and
...
+The Business **MUST** provide a nonempty, plain-text, buyer-facing
+`description` for every entry. The Business **SHOULD** make it a short,
+self-contained label or phrase suitable for direct use in a compact list and
+**SHOULD** localize it for the request when possible. It does not participate
+in amenity identity or filtering. An entry without that description is
+schema-invalid; a Platform **MUST** reject the invalid response rather than
+infer presentation text from the key.
+
+Amenity presentation is identifier-agnostic: the Business-provided
+`description` alone is sufficient to present any amenity. A Platform
```
jingyli: > A Platform **SHOULD** present every returned amenity using its `description`, including entries whose identifier it does not recognize.

I think an angle that is missing here is that Platforms ultimately has control on whether or not amenities renders based on the list - if it does choose to render, then following the contract here makes sense. A business that wants to force rendering should leverage warning messages.

Recommend to enhance this sentence with:
```
A Platform governs its
own presentation policy and **MAY** determine whether or where
amenities will be displayed (e.g., omitting them on compact summary cards).
When the Platform elects to present amenities, it **SHOULD** use the
provided `description` on every returned amenity and **SHOULD NOT**
suppress returned entries solely because their identifier is unrecognized.
```

docs/specification/common/location/search.md
```diff
@@ -246,6 +250,21 @@ The following table contains a non-exhaustive list of well-known values:
...
+presentable without assigning them standardized semantics.
+
+A Platform supplies amenity filters as an array of identifiers. Matching uses
+exact whole-key equality; namespace prefixes carry no wildcard or inheritance
+semantics. A Business **MUST NOT** ignore a supplied identifier: a Location
+whose authoritative amenity set lacks any supplied key does not match. This
+includes identifiers the Business does not use; no diagnostic is required for
+an ordinary non-match. When `filters.amenities` is present, the Business
+**MUST** include every supplied identifier in each returned Location's
+`amenities` map. A Platform **MUST** reject as non-conforming a returned
```
jingyli: nit on wording: Instead of `A Platform **MUST** reject as non-conforming a returned Location that omits`, we probably should have `A Platform **MUST** reject a returned Location as non-conformation if it omits...` to have the sentence flow better?

docs/specification/common/location/search.md
```diff
@@ -361,8 +393,10 @@ The following requests and responses are transport-neutral UCP payloads.
 The explicit `serves.point` is the authoritative service target; the coarse
 `context` hints only shape ranking and localization. At the supplied instant,
 it is Monday at `10:00` in `America/Los_Angeles`, within the returned
-interval. See [Operating Hours](index.md#operating-hours) for complete
-schedule evaluation rules.
+interval. The custom amenity remains presentable through its description
```
jingyli: > The custom amenity remains presentable through its description
without assigning the identifier any standardized UCP meaning.

Do we need this sentence? If yes, we should move it to behind the `See [Operating Hours](...)...` sentence so everything concerning hours filtering are grouped together.

source/schemas/common/types/location.json
```diff
@@ -2,7 +2,24 @@
...
+    "amenity": {
+      "title": "Amenity",
+      "description": "Buyer-facing presentation metadata for one amenity identifier. The containing map key, not this metadata, defines amenity identity and filter matching.",
+      "type": "object",
+      "required": ["description"],
+      "additionalProperties": true,
+      "properties": {
+        "description": {
+          "type": "string",
+          "minLength": 1,
```
jingyli, line 16: Should we also add a `uniqueItems: true` annotation here?

source/services/common/mcp.openrpc.json
```diff
@@ -33,14 +33,9 @@
               "profile": {
                 "type": "string",
                 "format": "uri",
-                "description": "URL to the platform's UCP profile document."
+                "description": "URL to the Platform's UCP profile document."
               }
             }
-          },
-          "idempotency-key": {
```
jingyli, line 40: If we are reconciling this to standard HTTP header - then we should probably also remove it from https://github.com/Universal-Commerce-Protocol/ucp/blob/main/source/services/shopping/mcp.openrpc.json#L64? Otherwise, any harm to keep it optionally for read operations?

source/schemas/common/types/location.json
igrigorik: Unnecessary, since it's a map now.

docs/specification/common/location/index.md
igrigorik: reworked, ptal

source/services/common/mcp.openrpc.json
igrigorik: No because shopping has mutations where this is required whereas location is purely reads.

PR #772, opened 2026-08-26
fix: treat off-site links as external in the docs link checker

Ectsang:
**Nothing on the current docs is broken by this. It is hardening plus the first tests for
this script.** I grepped for every shape below and found none of them in the docs today.

`check_links.py` decides whether a link is external by matching its scheme against
`("http", "https")`, plus a skip list for `mailto:`/`tel:`/`javascript:`/`data:`. Anything
else falls through and its path gets resolved against the local build, so a link pointing
somewhere else entirely is reported as a broken internal one:

| Link in a page | Reported as | Should be |
|---|---|---|
| `//example.com/x` | `Target: local_preview/x (Not Found)` | skipped, another host |
| `ftp://example.com/x` | `Target: local_preview/x (Not Found)` | skipped, another host |
| `vscode:extension/ms-python.python` | `Target: local_preview/extension/… (Not Found)` | skipped, not a web page |

This matters because the docs job treats any finding as fatal. A false positive blocks a
deploy on a link that is not ours to fix, and the reverse case ships a broken link to
readers.

Classifying on what the URL carries fixes all three: a host, or a scheme we do not
resolve, means off-site. A relative path that happens to contain a colon still parses with
no scheme and no host, so it keeps being resolved as before.

Two adjacent gaps in the same block, both one-liners:

- The raw-Markdown guard from #759 compares a case-sensitive suffix, so `page.MD` is
  accepted while `page.md` is rejected.
- `path` is percent-decoded but `fragment` is not, so `target/#re%61l` reports
  `Anchor not found` against an id of `real`.

## Tests, and where they run

`check_links.py` had no tests. This adds `scripts/test_check_links.py` on the same harness
as `test_validate_examples.py`, covering off-site classification, resolution, the
raw-Markdown guard, exit codes, `.linkignore` precedence and `DOCS_MODE=spec`.

These fail on unpatched `main`:

```
off_site_skipped[//example.com/x]
off_site_skipped[//example.com/x.md]
off_site_skipped[ftp://example.com/x]
off_site_skipped[ws://example.com/socket]
off_site_skipped[vscode:extension/ms-python.python]
on_site_resolves[target/#re%61l]
on_site_failure_reported[target/page.MD]
```

`.pre-commit-config.yaml` gains a `check-links-tests` hook mirroring
`validate-examples-tests`, so the tests run rather than sit unused.

I did not add a CI step. Editing `.github/workflows/docs.yml` makes zizmor scan it, and it
reports seven `unpinned-uses` findings on `actions/checkout@v5` and
`tj-actions/changed-files@v47` that have nothing to do with this change. They only surface
when someone touches the file. Happy to send the CI step as a follow-up if you would like
it, but that decision looked like yours rather than mine to bundle in here.

The script reads `sys.argv` and the environment at module level, so the tests drive it as a
subprocess against a synthetic site rather than importing it. That exercises the same entry
point CI uses. Say the word if you would rather have an importable seam instead.

## Verification

Built the site locally and diffed findings between this branch and `main`: 0 findings
either way, in both `DOCS_MODE=root` and `DOCS_MODE=spec`, running both versions as real
subprocesses over the same build.

`pre-commit run --all-files` passes. `test_validate_examples.py` still passes.

## One removal to flag

The diff drops `link = link[len(SITE_URL) - 1 :]` after the site-prefix strip. That value
is never read again: the last read is the `startswith(SITE_URL)` test directly above it,
and everything downstream uses `parsed.path` or `original_link`. Behavior-neutral, but not
required by the fix, so I would rather name it than leave it for review to find.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -44,6 +44,13 @@ repos:
         pass_filenames: false
         files: ^scripts/(validate_examples|test_validate_examples)\.py$
         stages: [pre-commit, pre-push]
+      - id: check-links-tests
+        name: Link checker unit tests
+        entry: uv run --frozen python scripts/test_check_links.py
+        language: system
+        pass_filenames: false
+        files: ^scripts/(check_links|test_check_links)\.py$
+        stages: [pre-commit, pre-push]
   - repo: https://github.com/streetsidesoftware/cspell-cli
     rev: v9.3.3
     hooks:
--- a/scripts/check_links.py
+++ b/scripts/check_links.py
@@ -160,20 +160,23 @@ def check_links():
         continue
 
       parsed = urlparse(link)
-      if parsed.scheme and parsed.scheme in ("http", "https"):
+      if parsed.scheme in ("http", "https"):
         if not link.startswith(SITE_URL):
           continue  # External link
-        # Internal absolute URL (e.g. https://ucp.dev/foo) -> /foo
-        link = link[len(SITE_URL) - 1 :]  # Keep the leading slash
+      elif parsed.netloc or parsed.scheme:
+        # Anything else carrying a host (protocol-relative //host/path) or a
+        # scheme we do not resolve (ftp:, ws:, vscode:, ...) points off-site.
+        # Falling through would read its path as a site-absolute path and
+        # report a spurious "Not Found" against the local build.
+        continue
 
-      path_part = parsed.path
-      anchor_part = parsed.fragment
-      path_part = unquote(path_part)
+      path_part = unquote(parsed.path)
+      anchor_part = unquote(parsed.fragment)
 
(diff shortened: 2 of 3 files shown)


damaz91 (approved):

Ectsang:
Still merges clean against `main` after the nine commits that have landed since this opened. The one red check is `Governance / Approvals` reporting `Pending required approvals`, so it is waiting on a reviewer rather than failing anything. I can rebase onto `ee07e9f0` if that makes it easier to look at.

carolinerg1 (approved):

carolinerg1:
Hi @Ectsang, thank you for your patience. PR approved and merged.

Ectsang:
Thanks both. @damaz91 for the review back in August, @carolinerg1 for closing it out.

Issue #738, opened 2026-08-15
[Proposal]: State-bound Offer extension for merchant-proposed Cart changes


arjun2075:
### Summary

Design discussion: Universal-Commerce-Protocol/ucp#702
Proposal discussion: Universal-Commerce-Protocol/ucp#733

This proposes a small `Offer` extension for Cart to represent a merchant-proposed change to the buyer's current cart.

The main idea is that an Offer is not just merchandising text or a discount. It is a concrete proposal against a specific Cart state:

> given this Cart state, the Business is proposing this resulting Cart state.

The design discussion is in #702.

The current direction from that discussion is:

* reuse existing Cart `line_items` semantics for the proposed state;
* identify an Offer with `id + revision`;
* bind it to the Cart state it was generated against using an opaque `state_ref`;
* consider an Offer valid only while the source state still matches, it has not expired, and the Business can still honor it;
* apply it through a narrow `apply_offer(cart_id, offer_id, revision)` operation;
* keep the returned Cart authoritative.

### Motivation

Today UCP can represent a Cart and allow the Platform to update it, but there is no explicit protocol distinction between:

1. the buyer asking for an item,
2. the agent deciding to add an item, and
3. the Business proposing a specific cart change that the agent then accepts.

That distinction matters more when an agent is making commerce decisions on behalf of someone else.

For example, a merchant may want to propose:

> Add this charger to the current order for an additional $20.

The agent should be able to inspect the actual proposed cart mutation and its impact without having to infer the meaning from natural-language copy.

It should also be impossible to take that proposal and silently apply it to a different Cart state later.

### Goals

* Represent a Business-proposed Cart transition explicitly.
* Bind that proposal to the Cart state it was generated against.
* Reuse existing Cart mutation semantics rather than inventing a second mutation model.
* Make price/recurrence impact available as structured data.
* Make stale Offers fail closed.
* Preserve enough identity/provenance to know that a resulting Cart change came from a specific Offer.
* Leave room for evidence/signing systems to bind to the exact decision-relevant proposal.

### Non-Goals

This proposal is intentionally not trying to solve:

* general Cart/Checkout versioning;
* agent authorization policy;
* a new subscription model;
* a new discount model;
* a generic JSON Patch mechanism;
* a new signing/attestation system.

In particular, I don't think Offer should revive or duplicate the general Cart revision mechanism discussed in #113.

[...]

Issue #113, opened 2026-01-23
[Spec] Concurrent checkout updates can cause silent data loss - need ETag/If-Match support


alidogankaya:
## The Problem

I noticed there's no mechanism to handle concurrent updates to checkout sessions. This can cause silent data loss:

**Example scenario:**
- User opens checkout in Tab 1, starts editing address
- Same user (or an AI agent) opens checkout in Tab 2, changes quantity
- Tab 1 saves → success
- Tab 2 saves → success, but it just overwrote Tab 1's address change

No error, no warning. The address change is just gone.

This happens because Update Checkout does a "full replacement" (as spec says), but there's no way to detect if someone else modified the resource in between.

## When does this actually happen?

- Two browser tabs open (pretty common)
- Phone + laptop at the same time
- AI agent making changes while user is also editing
- Network hiccup causes a retry

## Suggestion

Standard HTTP already solves this with ETag/If-Match headers (RFC 7232). Something like:

**Response includes a revision:**
```json
{
  "id": "chk_123",
  "revision": "r_a8f2e1",
  ...
}
```

**Update request includes If-Match:**
```http
PUT /checkout-sessions/chk_123
If-Match: "r_a8f2e1"
```

**If revision changed, return 412:**
```json
{
  "status": "requires_escalation",
  "messages": [{
    "type": "error",
    "code": "revision_mismatch", 
    "message": "Checkout was modified by another request.",
    "severity": "recoverable"
  }]
}
```

This is what Google Cloud APIs do, Stripe does something similar with versioning, etc.

## Happy to help

If this makes sense, I can put together a PR with the spec changes. Just wanted to check if this is something you'd consider adding first.

PR #773, opened 2026-08-26
feat: add state-bound Offer extension for Cart proposals

arjun2075:
## Summary

Implements the smallest structural/transport portion of #738:

* adds a state-bound `Offer` extension for Cart (`dev.ucp.shopping.offer`);
* represents the Business-computed proposed target `line_items`;
* binds the Offer to `applies_to.cart_id` + `applies_to.state_ref`;
* identifies the Offer operationally using `id` + `revision`;
* adds a reference-based `apply_offer` operation;
* keeps REST and MCP semantics aligned;
* reuses normal UCP idempotency;
* keeps the returned Cart authoritative.

## Why

UCP can represent a Cart and let the Platform update it, but has no explicit
representation for:

> given this exact Cart state, the Business proposes this resulting Cart state.

This turns the design settled in #702 into an executable schema/transport slice.

The extension follows the `discount.json` composition pattern (`$defs` + `allOf`
over `cart.json`) rather than modifying the Cart schema. `proposed_update.line_items`
reuses existing Cart full-replacement semantics, so there is no second mutation
vocabulary.

`apply_offer` deliberately diverges from the discount precedent, which routes
application through the normal cart update request. A discount code is just a
string, so there is nothing precomputed to echo. An Offer's value is that the
Business already computed the resulting state, so requiring the Platform to
resubmit those `line_items` would duplicate authoritative state and create a
drift case between the stored proposal and the resubmitted one. The Platform
references `offer_id` + `revision`; the Business applies the exact stored revision.

## Key invariant

```text
applicable =
    source state still matches
    AND not expired
    AND Business can still honor the proposal
```

These are conjunctive, so a stale-but-unexpired Offer fails closed. Applying one
Offer moves the Cart to a new state and makes sibling Offers bound to the previous
state stale.

Runtime conditions are expressed as normative operation semantics rather than
forced into JSON Schema, since they depend on live Cart state.

## Scope deliberately deferred

This PR does **not** implement:

* commercial-intervention provenance from #724;
* generic Cart or Checkout versioning (see #113);
* recurrence/subscription modeling;
* cryptographic digest/signature infrastructure — `id` + `revision` is operational
  identity here; evidence identity is left to a shared convention;
* generic mutation / JSON Patch semantics;
* SDK, sample, or conformance implementations.

`state_ref` is intentionally opaque and implementation-assigned. If another
extension needs the same state identity, that seems like the right point to
introduce a shared Cart primitive rather than inventing one here.

## Testing

Commands actually run, with results:

| Command | Result |
| --- | --- |
| `ucp-schema lint source/` | 125 files checked, all passed |
| `uv run python scripts/validate_examples.py --schema-base source/schemas/` | 343 passed, 0 failed, 0 errors |
| `uv run python scripts/test_validate_examples.py` | 52 passed, 0 failed |
| `uv run yamllint -c .github/linters/.yamllint.yml .` | pre-existing line-length warnings only |
| `cspell lint` (3 changed files) | 0 issues |
| `DOCS_MODE=spec uv run mkdocs build --strict` | 47 warnings, byte-identical to `main` baseline; no new warnings |

Additional local verification (ad-hoc, not committed): resolved every `$ref` in
both service documents; confirmed the extension composes against `cart.json` and
validates a full Offer example; confirmed each decision-relevant field is required,
that an empty `state_ref` is rejected, that `presentation` is optional, and that
`impact.total_delta` accepts negative values; confirmed REST and MCP take the same
`{id, offer_id, revision}` inputs, both require idempotency, and neither carries a
request body.

## Relationship

Implements the structural/transport portion of #738. Design history in #702.


arjun2075:
Hi maintainers — following up on this Offer extension PR. The CLA check is passing, and the PR is currently awaiting the required governance approvals.

The scope is limited to the state-bound Offer schema and REST/MCP apply_offer operation from #738. Commercial-intervention provenance (#724) and the broader lifecycle contract (#788) remain separate.

Could someone from the TC take an initial pass, or point me to the appropriate reviewer? Feedback on the reference-based application operation and source-state validity semantics would be particularly helpful. Happy to address changes or narrow the scope further.

arjun2075:
Hi @Universal-Commerce-Protocol/maintainers — could someone help route this PR through its two remaining gates?

The five fork-PR workflows are still marked action_required pending approval, and the PR carries gov:needs-tc-review, but no TC reviewer has been assigned. My September 5 request for an initial review has not received a response.

Could a maintainer please approve the pending workflow runs and route the PR to the appropriate TC reviewer for the Offer extension? If there is a prerequisite or scope change needed before review, please let me know and I’ll address it.

Issue #790, opened 2026-09-01
[Proposal]: Carry token issuance context in token credentials


prasad-stripe:
## Summary

Add optional `binding` and `identity` members to `TokenCredential` so a Platform
can carry token issuance context into checkout. Existing `{type, token}`
credentials remain valid, and concrete handlers can require either member.

## Motivation

UCP requires every token to be bound to a capability resource and issued to one
participant. `/tokenize` receives both values, but `TokenCredential` carries
neither. In a tokenize-to-process flow there is no `/detokenize` request, so UCP
provides no second binding value for the Processor to compare with its issuance
record.

The same gap affects delegated flows: the Business or PSP otherwise has to
reconstruct issuance context from surrounding configuration. That is unreliable
for binding because `binding.type` is an open namespace and need not identify a
checkout directly.

## Goals

- Carry the original binding and participant identity with a token credential.
- Preserve compatibility with existing token credentials.
- Let concrete handlers require the context their flow needs.
- Support exact binding comparison in tokenize-to-process flows.
- Keep participant identity separate from caller authentication.

## Non-Goals

- Defining a payment-processing endpoint.
- Making either new member required in the shared schema.
- Treating carried context as proof or authentication.
- Changing the existing `PaymentIdentity` contract.

## Detailed Design

Add two optional properties to
`source/schemas/common/types/token_credential.json`:

```json
{
  "binding": {
    "$ref": "binding.json",
    "ucp_response": "omit"
  },
  "identity": {
    "$ref": "payment_identity.json",
    "ucp_response": "omit"
  }
}
```

Neither property is added to `required`. Both are omitted from UCP response
shapes, like the token itself, because they are checkout request credentials.

When either value is carried into checkout, the Platform must copy it unchanged
from the tokenization request. The Business validates binding according to the
handler's binding policy. A tokenize-to-process handler should require binding
in its concrete credential schema and must define how the Business passes it to
the Processor. The Processor compares it by exact equality with the value
recorded at issuance.

`identity` retains its existing UCP meaning: it identifies the participant the
token was issued to. The processing caller is authenticated independently using
the handler-defined mechanism. Carried identity must not be accepted as
authentication.

## Risks and Mitigations

**Carried context is not proof.** The Platform supplies these values. Binding
assurance still comes from comparison with the Tokenizer or Processor's stored
issuance record, and caller authority is verified independently.

**Compatibility.** Both members are optional in the shared schema. Existing
[...]

PR #792, opened 2026-09-01
feat(payment): carry token issuance context through checkout

prasad-stripe:
# Description

`TokenCredential` currently carries only `type` and `token`. That leaves no
standard way to carry the resource binding and participant identity established
during tokenization into checkout.

This change adds optional `binding` and `identity` members to
`TokenCredential`. Existing credentials remain valid. A concrete handler can
require either member when its flow needs it.

The guidance requires Platforms to preserve carried values unchanged,
Businesses to validate binding according to the handler's policy, and
Processors to compare a presented binding with their issuance record. Caller
authentication remains separate and handler-defined. `identity` continues to be
a participant identifier and must not be accepted as authentication.

## Category (Required)

## Related Issues

- Enhancement Proposal: #790
- Follow-up to #746
- Participant-identity documentation cleanup: #791

## Checklist

- [ ] I have followed the [Contributing Guide](https://github.com/Universal-Commerce-Protocol/.github/blob/main/CONTRIBUTING.md). Enhancement Proposal #790 is awaiting approval.
- [x] My changes pass all available local linting and formatting checks.
- [x] I have added schema-annotated examples that exercise the new credential shape in CI.
- [ ] New and existing unit tests pass locally. The validator contract suite passes 42 tests; two integration cases require the unavailable `ucp-schema` binary.
- [x] I have updated the relevant JSON schema.
- [ ] Pydantic model regeneration is not applicable; this repository has no `python_sdk` directory or `generate_models.sh` script.

## Screenshots / Logs

No visual change requiring a screenshot.

Local checks:

- `git diff --check`
- JSON parsing
- markdownlint
- cspell
- validator contract tests: 42 passed; 2 `ucp-schema` integration cases skipped

### DIFF
--- a/docs/specification/payment/examples/platform-tokenizer-payment-handler.md
+++ b/docs/specification/payment/examples/platform-tokenizer-payment-handler.md
@@ -261,10 +261,11 @@ The response config includes runtime token lifecycle information.
 Upon receiving a checkout with a token credential:
 
 1. **Validate Handler:** Confirm `instrument.handler_id` matches the expected handler ID.
-2. **Detokenize or Delegate:**
+2. **Validate Binding:** Validate `credential.binding` according to the handler's binding policy.
+3. **Detokenize or Delegate:**
    * **Option A (Direct):** Call the platform's **credential provider** `/detokenize` endpoint directly, then process payments.
    * **Option B (Delegated):** Forward the token to a PSP for detokenization and payment processing.
-3. **Return Response:** Respond with the finalized checkout state.
+4. **Return Response:** Respond with the finalized checkout state.
 
 For option B, see section [PSP Integration](#psp-integration).
 
@@ -397,7 +398,14 @@ Content-Type: application/json
         },
         "credential": {
           "type": "token",
-          "token": "ptok_x9y8z7w6v5u4"
+          "token": "ptok_x9y8z7w6v5u4",
+          "binding": {
+            "type": "dev.ucp.shopping.checkout",
+            "id": "checkout_789"
+          },
+          "identity": {
+            "access_token": "business_abc123"
+          }
         }
       }
     ]
@@ -408,6 +416,10 @@ Content-Type: application/json
   }
 }
 ```
+
+The Platform **MUST** copy `binding` and any `identity` unchanged from the token
+issuance context. Before detokenization or delegation, the Business **MUST**
+validate the binding according to the handler's binding policy.
 
 ---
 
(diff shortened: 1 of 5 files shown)


karangoel16:
I think we should separate caller authentication from token binding identity here.

For authentication, access tokens should be sent via `Authorization: Bearer <token>` at the transport layer. This matches the existing REST auth model and identity-linking guidance.

The `identity.access_token` field in the tokenization payload appears to mean “the participant/business this payment token is issued to,” not “the caller credential used to authenticate this request.” If so, calling it `access_token` is confusing and potentially unsafe, because consumers may treat it as an auth credential or copy it into checkout/payment payloads.

Recommendation:
- Keep caller auth in the HTTP `Authorization` header.
- Keep token/resource binding in the body.
- Rename `identity.access_token` to a non-secret subject identifier, e.g. `business_id`, `merchant_id`, `participant_id`, or `subject`.
- If this value is actually secret/auth-bearing, it should not be part of the credential/body model.

prasad-stripe:
Thanks, agreed. I removed `identity` from `TokenCredential` and from the new examples and guidance. This PR now only carries `binding` through the checkout credential; the existing participant-identity model and transport-layer authentication remain unchanged.

prasad-stripe:
@googlebot I signed the CLA; please recheck the updated commit.

prasad-stripe:
Thanks for raising this. Caller authentication and participant identity are
already separate in this proposal. #792 does not treat
`identity.access_token` as an authentication credential. It reuses the existing
`PaymentIdentity` schema, which defines that field as the identifier of the
participant the token was issued to. The caller is authenticated independently
using the handler-defined transport mechanism, such as `Authorization`.

Carrying `identity` in the checkout credential preserves the token's issuance
context for downstream verification; it does not turn the value into a bearer
credential or replace caller authentication.

archrao:
Thanks for the proposal and PR! A couple of questions: 
1.  In 'POST /checkout-sessions/{checkout_id}/complete', the checkout_id is already present in the URL. Is it fair to say that explicitly carrying 'binding' inside TokenCredential is primarily to enable the Business to forward the credential object to their PSP without manually reconstructing the binding object?

2. Today, PSPs are agnostic to UCP, and Businesses use their existing proprietary PSP APIs  to process tokens. With 'TokenCredential' carrying 'binding' and 'identity', the Business will now need to make code changes to map and pass those fields into their proprietary PSP calls.  
   - Does this introduce extra integration burden on the Business, and is the primary gain for the Business or the PSP?  
   - Since this protects the Business against fraud and chargebacks, would there be benefit in explicitly calling out this fraud/chargeback value proposition in the specification docs so Businesses understand the incentive for passing these fields to their PSP?

Issue #486, opened 2026-05-27
[Proposal]: Providing the ability to return after buyer navigates to continue_url


sioked:
### Summary

`continue_url` is a great solution for ensuring buyers can always complete their transaction, but it's currently a one-way door. Once a buyer navigates away, the originating platform has no visibility into what
 happened — did they complete? cancel? abandon? — even though UCP has a full order model that tracks exactly that. There's also an assumption baked in that continue_url always handles the entire remaining checkout
 session, but that's not always true. PayPal, for example, typically handles just the payment step — and 3DS is similar, where a bank redirect resolves one piece of the flow before the buyer should return to the
 partner surface to finish.

Adding return_url as an optional, platform-provided field makes the handoff bidirectional. Businesses and payment handlers redirect the buyer back when their hosted experience ends, appending a status hint so the
 platform knows how to route. This keeps complete callable on the original surface, gives platforms visibility into post-handoff outcomes, and enables redirect-style flows that are scoped to a single step rather
 than the whole session.

### Motivation

Currently, the `continue_url` is exposed in the spec for a variety of different purposes:
1. After cart creation, the merchant can return a continue_url where the buyer can complete the entire checkout session
2. During checkout, continue_url is returned - especially if the merchant `requires_escalation`; for example, if an error occurs or if the platform does not support the capabilities of the merchant

This is a great solution that ensures there's always a pattern that allows a buyer to complete their transaction. However, there's a couple challenges with the current flow:

1. This is currently a one-way door; if a buyer navigates to the continue_url they are pushed off the original surface and may not come back. The original platform has no visibility into what happened on the external site and the buyer has no incentive to share that data back. Considering that UCP also has specs for confirmed transactions (orders), there would be no way for the partner to know whether this became a transaction was successful
2. This assumes that the continue_url will always handle the full checkout_session completion, but at PayPal we typically handle just the payment in the middle of a checkout_session. We'd like the ability to continue to PayPal, then redirect back to the platform to finish checkout there. 3DS might be another similar example - the 3DS experience might include a redirect to the bank for resolution, but ultimately could go back to the partner surface to complete

Adding a return_url would ensure that partner surfaces can get buyers back after the hosted experience ends, give partners better visibility into what happened after the user navigated away (completion or
 cancellation), and give merchants and payment providers (e.g. PayPal as a payment method) a way to enable redirect flows scoped to a subset of the checkout session - keeping the complete call on the original
[...]

PR #801, opened 2026-09-02
docs(cart): define the continue_url destination and format

YanisMtcr:
## Motivation

The cart page says businesses SHOULD provide `continue_url`, and the schema describes it as "URL for cart handoff and session recovery", but nothing says where the URL should lead or what it has to preserve. The requirements list still carries `TODO: discuss continue_url destination - cart vs checkout` from #73, where two readings met: for some implementers the URL is the checkout URL, for others the business decides the intended use (convert to checkout, or keep building the cart).

The checkout page answers the same questions for checkout in its Continue URL section (availability, format, server-side state or permalink); the cart page has no equivalent. And the permalink specification already lets a business merge items into an existing cart, create a new cart, or stage a checkout, so the destination question has an answer elsewhere in the spec: the business chooses.

This came up while implementing the cart capability: a cart session is served by the business's UCP endpoint, which does not always share state with the storefront, and a `continue_url` pointing at the storefront's own cart page lands the buyer on an empty basket.

## What this PR changes

- **`shopping/cart/index.md`** - adds a `## Continue URL` section after Cart-to-Checkout Conversion. **Destination**: the business chooses (cart page, checkout entry, embedded flow); whatever it picks MUST present the cart's current `line_items`, quantities and selected variants included, without the buyer re-adding them; items no longer available follow the permalink redirect resolution rules. **Format**: absolute HTTPS URL that MUST carry or reference the cart contents, using either approach the checkout page already describes **by reference** (server-side cart state or a stateless permalink), rather than restating them.
- Removes the TODO bullet and appends a link to the new section to the existing SHOULD bullet; the requirement level is unchanged.

Normative addition, scoped to the field: two MUSTs on `continue_url` (absolute HTTPS, contents preserved). Both mirror what the checkout page already requires for its own `continue_url`. No schema change.

Overlaps with nothing open: #740 (cart terminal status) edits the same requirements list but leaves this bullet as is, and #486 (return after handoff) is checkout-side.

## Category (Required)

## Related Issues

None.

## Checklist

### DIFF
--- a/docs/specification/shopping/cart/index.md
+++ b/docs/specification/shopping/cart/index.md
@@ -89,6 +89,37 @@ SHOULD be linked for the duration of the checkout.
     on a cleared cart ID return `not_found`; the platform can start a new
     session with `create_cart`.
 
+## Continue URL
+
+The `continue_url` field hands a cart from the platform to the business UI, so
+a buyer can share, recover, or keep building the cart in a browser, or move on
+to purchase.
+
+### Destination
+
+The destination is the business's choice: its cart page, its checkout entry,
+an embedded flow, or any surface that fits its purchase path. Whatever the
+destination, it **MUST** present the cart's current `line_items`, with their
+quantities and selected variants, without requiring the buyer to add them
+again. Items that can no longer be offered at resolution time follow the
+[permalink redirect resolution rules](../../permalink.md#redirect-resolution).
+
+A cart session is served by the business's UCP endpoint, and that endpoint does
+not always share state with the storefront. A `continue_url` that points at the
+storefront's own cart page then lands the buyer on an empty basket.
+
+### Format
+
+The `continue_url` **MUST** be an absolute HTTPS URL and **MUST** carry or
+reference the cart contents. Businesses **MAY** use either approach described
+in the [checkout Continue URL format](../checkout/index.md#format):
+
+* an opaque URL backed by server-side cart state, for example
+    `https://business.example.com/carts/{cart_id}`;
+* a stateless URL that encodes the cart's items, such as a
+    [permalink](../../permalink.md), which the business resolves into a cart
+    or a checkout.
+
 ## Quantity and sale basis
 
 Cart line items apply the shared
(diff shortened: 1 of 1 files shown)


proshoumma:
Good addition to the cart capability. I'm wondering about the Platform behavior.

At which exact moment Platform is supposed to display this `continue_url` to the user. For checkout its clear - when an escalation is required, Platform would surface the url. However, since cart do not have a state-machine, it's not clear for Platform when to surface this.

Would be great to hear your thoughts. Also, please let me know if I missing something. 🤔

YanisMtcr:
Not missing anything: the page does not say when the platform surfaces `continue_url`, before or after this PR. The platform guideline is "MAY display `continue_url` for handoff to business UI", and this PR only defines the business side, destination and format.

The cart has no status lifecycle (Cart vs Checkout table, Actions section), so the trigger is buyer intent or a platform-side dead end, not a status:

1. The buyer asks for it: share the cart, come back to it later, keep building on the business's site. That is the "sharing and recovery" use the page names in its introduction.
2. The buyer expresses purchase intent and the platform cannot convert the cart to a checkout, because the business does not expose the checkout capability or the platform does not support it. Same moment as the "SHOULD convert cart to checkout" guideline; when conversion is not available, `continue_url` is the buyer's path to complete the purchase on the business's site.
3. The cart returns a message the platform cannot act on. Cart reuses the checkout message schema, and the Create Cart example already pairs an `unrecoverable` error with a `continue_url`; `requires_buyer_input` carries the same handoff meaning as in checkout.

proshoumma:
Hmm... so the only time this `continue_url` will come to play is either Buyer explicitly asks for it, or an error occurred during cart operations. Would be great to add a sentence or two mentioning these - Platforms will have a clear instruction to follow for when to surface the url.

YanisMtcr:
Added a "When to surface the continue URL" list with the three situations. I kept the no-checkout case separate from errors: nothing fails there, the business exposes cart without checkout, which nothing in the capability model prevents, and the link is then the buyer's only path to purchase. It also covers the case where the buyer builds the cart on the platform and completes the purchase on the business's site.

Issue #817, opened 2026-09-10
[RFC] Decoupling SDK Data Transfer Objects from JSON Schema Semantic Validation


gsmith85:
# RFC: Decoupling SDK Data Transfer Objects from JSON Schema Semantic Validation

**Related Specs**: [UCP Issue #800](https://github.com/Universal-Commerce-Protocol/ucp/issues/800) (*Standardized Intermediate Representation Strategy for UCP*), [UCP Issue #484](https://github.com/Universal-Commerce-Protocol/ucp/issues/484) (*UCP Open Source Unified Toolchain and SDK Generation*)

---

## 1. Summary

This RFC establishes an architectural boundary for Universal Commerce Protocol (UCP) client SDKs: **Language-native SDK models are strictly structural Data Transfer Objects (DTOs), decoupled from JSON Schema Draft 2020-12 predicate validation engines.**

Official UCP client SDKs (`python-sdk`, `js-sdk`, `java-sdk`, etc.) will generate directly from the canonical OpenAPI 3.1 Intermediate Representation established in [Issue #800](https://github.com/Universal-Commerce-Protocol/ucp/issues/800) (`ucp-schema export-openapi`) using standard community code generators. We formally deprecate and remove all post-generation Abstract Syntax Tree (AST) monkey-patching scripts (~4,000 lines of code across `python-sdk` and `js-sdk`).

For applications that require explicit schema validation—whether **pre-dispatch (outbound)** to catch errors prior to network transit, or **on-receipt (inbound)** to validate incoming webhooks, callbacks, or merchant responses—SDKs will provide a decoupled validation helper powered by standard JSON Schema engines (`jsonschema` in Python, `@hyperjump/json-schema` in TypeScript) executing against the bundled specification.

---

## 2. Background: DTOs vs. Domain Invariants

In standard software architecture (Domain-Driven Design, Fowler's *Data Transfer Object* pattern), DTOs serve as simple data carriers:
- Their responsibility is serialization, deserialization, static typing, and transferring data across system boundaries.
- Domain rules, complex invariants, and stateful constraints belong in domain entities or dedicated validation services, not in boundary transfer types.

[...]

Issue #800, opened 2026-09-02
[RFC] Proposal: Standardized Intermediate Representation Strategy for UCP


gsmith85:
# RFC: Standardized Intermediate Representation Strategy for UCP

## 1. Summary

The objective of this RFC is to produce a standardized intermediate representation for UCP that enables downstream code generation.

This is in part to support UCP's managed SDKs (`js-sdk`, `python-sdk`) which are costly to maintain, particularly in lockstep with schema changes. Moreover, this is an approach to enable merchants and platform builders to self-generate types for their own systems, *tailored to their specific UCP profile* in a format (OpenAPI 3.1) that has broad industry adoption, potentially eliminating the need for UCP to maintain a large number of first-party SDKs.

This is not new ground. Rather, it is the outcome of previous Technical Council discussions—most notably [Issue #484](https://github.com/Universal-Commerce-Protocol/ucp/issues/484) (*Proposal: UCP Open Source Unified Toolchain and SDK Generation*), alongside key PRs from the `2026-08-25` release cycle (e.g. [PR #688](https://github.com/Universal-Commerce-Protocol/ucp/pull/688) on tagged union scaffolding and [PR #684](https://github.com/Universal-Commerce-Protocol/ucp/pull/684) on generator class breakage) where the TC navigated the tension between maintaining a fully-featured JSON Schema authoring surface and comprehension.

### Proposed Solution: Extension of `ucp-schema` Toolchain

We propose extending the shared Rust compiler toolchain (`ucp-schema`) with an `export-openapi` command.

OpenAPI 3.1 natively adopts the JSON Schema 2020-12 specification, eliminating dialect translation loss and keyword stripping (`$defs`, validation rules, etc.).  As such, it is well suited to handle the dynamicism and compositional extensibility of UCP.

The `export-openapi` command produces a fully-resolved IDL from UCP source schema and specified extensions producing:

1. Canonical Types & DTOs (`components.schemas`) for Type Generation:
   - Directional Models: Slices schemas into explicit request/resource models (e.g. `CheckoutCreateRequest`, `CheckoutSession`) based on `(direction, op)` annotations.
   - Tagged Unions: Maps union variants (such as PR #688 fulfillment destinations) to standard OpenAPI `discriminator`s so code generators emit real typed unions instead of `Any`/`unknown`.
[...]

PR #833, opened 2026-09-17
docs(schema): clarify that service schema references a transport definition, not a JSON Schema

gsmith85:
## Summary

The `schema` field on a service binding is inherited from `ucp.json#/$defs/entity`, whose description reads:

> *"URL to JSON Schema defining this entity's structure and payloads."*

That is accurate for **capabilities**. It is not accurate for **services** — every service binding in the specification points `schema` at an OpenAPI or OpenRPC document, not a JSON Schema.

This PR documents the actual contract. **Documentation only — no structural or validation change.**

## The inconsistency

`service.json#/$defs/base` composes the shared entity base and adds only `transport` and `endpoint`:

```json
"allOf": [
  { "$ref": "ucp.json#/$defs/entity" },
  { "required": ["transport"], "properties": { "transport": {…}, "endpoint": {…} } }
]
```

So `schema` arrives with a description written for the generic entity, and the same field name ends up carrying two different media types depending on which registry it appears in:

| Location | `schema` value | Actual format |
| --- | --- | --- |
| `services["dev.ucp.shopping"][transport=rest]` | `…/services/shopping/rest.openapi.json` | OpenAPI 3.1 |
| `services["dev.ucp.shopping"][transport=mcp]` | `…/services/shopping/mcp.openrpc.json` | OpenRPC 1.3.2 |
| `services["dev.ucp.shopping"][transport=embedded]` | `…/services/shopping/embedded.openrpc.json` | OpenRPC 1.3.2 |
| `services["dev.ucp.shopping"][transport=a2a]` | *(absent — not required)* | Agent Card, via `endpoint` |
| `capabilities["dev.ucp.shopping.checkout"][0]` | `…/schemas/shopping/checkout.json` | JSON Schema 2020-12 |

The distinction matters because the two fields do different jobs. `capability.schema` supplies **types**; `service.schema` supplies **routes and method names**, and per *Endpoint Resolution* those are what get appended to `endpoint` to form a call:

> *"The `endpoint` field provides the base URL for API calls. OpenAPI paths are appended to this endpoint to form the complete URL."*

A tool author reading only the field description would reasonably attempt to parse `rest.openapi.json` as a JSON Schema.

## Changes

**`source/schemas/service.json`** — add a service-specific `schema` description covering the transport-to-format mapping and its relationship to `endpoint`.

**`source/schemas/ucp.json`** — generalise the inherited `entity.schema` description so it is accurate for every entity type that composes it, rather than describing only the capability case.

## Verification

- `ucp-schema lint source/` → `121 files checked, all passed`; `schemas/service.json` and `schemas/ucp.json` both clean. The 8 warnings are pre-existing and untouched by this change.
- Recompiled `shopping.openapi.json` before/after: paths identical, schema count unchanged at 177, and the only deltas are the two intended `description` strings on `ServiceBase` and `UcpEntity` — confirming the clarification propagates into generated artifacts automatically.

## Out of scope — raising for TC discussion

Two follow-ups this touches but deliberately does not attempt:

1. **Should the field be split?** A distinct name for the service case (`interface`, `definition`) would remove the overload structurally rather than documenting around it. That is a breaking change and needs an Enhancement Proposal, so it is not attempted here.
2. **Tooling does not read this field yet.** `ucp-schema`'s profile parser reads `services[].endpoint` but neither `services[].transport` nor `services[].schema` — so profile-driven generation currently emits OpenAPI regardless of the transport a business advertises. Tracked separately against the #800 work; noting it here because this field is the input that fix will depend on.

Related: #800, #817


proshoumma:
Just out of curiosity - since there is already machine-readable JSON schema available for Platforms, what are the benefits of also having an OpenAPI / OpenRPC schema? Would this act as an additional supporting data points for Platforms?

Not related to this PR, just a general concern - more schema a Business needs to maintain, it adds complexity during UCP onboarding, as well as long term maintenance. Since multiple schema's (capabilities and service) now needs to stay in sync.

Issue #797, opened 2026-09-02
[Proposal]: Expand Signals for Fraud Risk Assessment


sw-tang:
### Summary

Extend `signals.json` with 6 well-known signals that provide device environment and browser context to support fraud risk assessment.  [#630](https://github.com/Universal-Commerce-Protocol/ucp/issues/630) proposes trust-level signals such as authentication, verification, and delegation. This proposal complements it by standardizing the device and browser environment signals that businesses need for fraud scoring — IP country, device timezone, screen dimensions, browser language, and Java support. Browser signals apply when the buyer interacts via a browser-based agent surface (e.g. Google AI Mode in Chrome); they will be absent in headless agent flows. Two signals in this proposal (`dev.ucp.session_start_time`, `dev.ucp.device_type`) overlap with the existing RFC and are included here for completeness.

### Motivation

Businesses receiving checkout requests need device and browser environment data to run fraud scoring models. Today, `signals.json` provides `dev.ucp.buyer_ip` and `dev.ucp.user_agent`, which are insufficient for effective risk assessment.

In agent-originated commerce, the platform controls the buyer's session environment. The platform observes the buyer's IP country, device timezone, screen dimensions, and browser configuration — but has no standard keys to relay this context to businesses. Without these signals, businesses must either accept degraded fraud detection quality or fall back to proprietary signal exchanges with each platform individually.

The target beneficiaries are businesses performing fraud risk assessment at checkout time. 

### Goals

 - Define well-known 6 signal keys for device and browser environment context
- Enable businesses to receive IP geolocation, device timezone, screen dimensions, browser language, and Java support status from platforms via a standardized vocabulary
- Maintain full backward compatibility — all new signals are optional
- Complement the existing agent-trust signals RFC with environment-level fraud context

#### API Changes

No new API endpoints. This is an additive change to the `signals` object already present on catalog, cart, and checkout requests.

#### Data Structures

Add 6 properties to `/schemas/common/types/signals.json`:

| Signal | Type | Description |
|--------|------|-------------|
| `dev.ucp.buyer_ip_country` | string | IP Country |
| `dev.ucp.device_time_zone` | string | Current timezone of the device being used in session |
| `dev.ucp.browser_screen_height` | integer | Screen height |
| `dev.ucp.browser_screen_width` | integer | Screen width |
| `dev.ucp.browser_language` | string | Language being used in browser on the device |
| `dev.ucp.browser_java_enabled` | boolean | Java enabled |

[...]

Issue #630, opened 2026-07-23
RFC: Expanding Well-Known Signals for Agent-Originated Commerce


prasad-stripe:
# RFC: Expanding Well-Known Signals for Agent-Originated Commerce

> **Goal:** Extend the `signals` schema with well-known signals that support risk assessment for agent-originated transactions.
> **Status:** Proposal (Inviting feedback from TC)
> **Related:** #153 (original signals EP), #203 (signals landing PR)

## 1. Problem Statement

The signals schema ([PR #203](https://github.com/Universal-Commerce-Protocol/ucp/pull/203)) established a framework with two well-known signals (`dev.ucp.buyer_ip`, `dev.ucp.user_agent`) and an extensibility model via `additionalProperties` + reverse-domain naming. This was sufficient as a baseline for catalog, cart, and checkout flows where browser-originated requests still provide useful environment data.

Agent-originated commerce changes the signal landscape structurally. `buyer_ip` resolves to a cloud provider range, `user_agent` identifies an agent runtime, and browser-based signals (fingerprinting, behavioral patterns) are entirely absent. Handlers performing authorization risk assessment on agent-originated transactions have two options today, both bad:

1. **Rely on the two existing well-known signals** and accept that they carry minimal signal value in agent contexts.
2. **Use proprietary signals via `additionalProperties`** and accept N-to-N integration burden (each platform invents its own keys, each handler must understand all of them).

The extensibility model works for vendor-specific signals, but the absence of standardized well-known signals for agent commerce means:

1. Platforms have no guidance on WHAT to send. The spec says signals support "authorization and abuse prevention" but doesn't define what's useful beyond IP and UA.
2. Handlers cannot build common risk infrastructure for agent transactions because there's no shared signal vocabulary.
3. Platforms already possess relevant context (how the buyer authenticated, what instrument verification was performed, session timing) but have no standard keys to express it.

## 2. Proposal

Add 6 well-known signals to `signals.json`. All are optional (the `signals` object itself remains optional, and no signal within it is required). All satisfy the existing spec constraint: "Values MUST NOT be buyer-asserted claims -- platforms provide signals based on direct observation or independently verifiable third-party attestations."

### Signal 1: `dev.ucp.instrument_verification`

**Type:** object keyed by payment instrument ID

[...]

PR #835, opened 2026-09-18
feat: add well-known signals describing agent behavior during a transaction

dpaul0501:
Opening this against the discussion in #630 and #797 rather than in place of it — happy for it to
be folded into either, or restated as an Enhancement Proposal if that is the right route.

`signals.json` carries two properties today: `dev.ucp.buyer_ip` and `dev.ucp.user_agent`. #630
proposes six more covering how the buyer was authenticated before the agent acted, and #797
proposes six covering the device and browser. None of the fourteen describes what the agent did
during the transaction itself.

That gap has a structural cause. A payment layer commonly receives only a hash of the request
body, so the agent's reasoning cannot reach the component making the authorization decision even
in principle. Signals of this class have to be carried explicitly by the protocol, because
nothing downstream can reconstruct them.

This adds five optional, platform-observed signals:

| Signal | Type | Description |
|---|---|---|
| `dev.ucp.instruction_provenance` | String | What caused this purchase: the buyer's own input, the agent's plan, or content the agent read |
| `dev.ucp.agent_spend_context` | Object | Amount relative to the agent's own baseline, rather than an absolute figure |
| `dev.ucp.payee_continuity` | Object | Whether the destination matches prior settlements to this counterparty |
| `dev.ucp.request_retry_context` | Object | Idempotency key and attempt number |
| `dev.ucp.agent_instance_continuity` | String | Opaque, platform-scoped agent instance identifier, stable across sessions |

Two notes on specific fields.

`instruction_provenance` is distinct from delegated authority. Delegation establishes that the
buyer authorized the agent to spend at all; provenance establishes what caused *this* spend. Both
seem necessary and #630's `delegation_context` covers only the first.

`payee_continuity` is documented as a scoring input and explicitly **not** as grounds for a
decline. That is from measurement rather than caution: on a live agent payment rail, a
payee-change control fired 14 times across 1,068 settlements and all 14 were legitimate.
Specifying the field without that note would invite implementers to repeat the mistake.

The docs change also states that the absence of a browser-environment signal must not itself be
scored as risk, since the natural implementation — missing signal raises risk — penalizes
agent-originated traffic by construction.

All five are optional, platform-observed rather than buyer-asserted, and purely additive. **This
is not a breaking change.**

Field data behind the proposal comes from one deployment at modest volume — 1,068 settled
payments, 43 agents, 295 services — and should be read as field observations rather than
established distributions. Median payment was $0.007 (p95 $0.15), and 217 of 1,068 settlements
failed, which is what motivates the retry field.


proshoumma:
Hi @dpaul0501 thanks for the PR. Similar to #630, could you please also add rationale behind each of the proposed signals? From the short description, it's not super clear why they might be necessary.

pratham-bhatnagar:
Rationale from the agent-builder side, since that is the perspective these signals serve:

We build shopping agents that browse and buy on real storefronts. The failure modes below are ones any platform operating agents at volume will recognize; none of them require access to our internal data to state.

**instruction_provenance**: Agents read untrusted content - product pages, reviews, seller descriptions. Content injection steering an agent into a purchase is the documented failure mode of web-browsing agents, and at the payment layer an injected purchase is byte-identical to a buyer-instructed one. Delegation establishes that the buyer authorized the agent to spend; it says nothing about whether this specific spend traces back to the buyer's intent or to something the agent read. Without provenance carried explicitly, a handler has no way to distinguish the two cases.

**agent_spend_context**: Absolute thresholds catch almost nothing for agents. Agent baskets are small and skewed, so a $400 order from an agent whose baseline is $25 is a far stronger anomaly than the same amount from a human account, and only a relative signal captures that.

**payee_continuity**: First-time-payee is the standard pattern in account-takeover and injection fraud. Continuity against prior settlements is a cheap check that adds no buyer friction, which is exactly why it belongs as a scoring input rather than a hard control - the note in the docs about it not being decline grounds is the right call.

**request_retry_context**: Agents retry. Retry loops from agent runtimes present to a handler as duplicate-purchase attempts, and without attempt number plus idempotency context there is no way to tell a buggy retry storm from a buyer genuinely ordering twice. Merchants see the downstream version of this as double-charge complaints.

**agent_instance_continuity**: Without a stable platform-scoped identifier, every agent transaction presents as first-seen. Reputation can never accrue to well-behaved agents and history can never attach to misbehaving ones. This is the signal that makes the other four useful over time rather than only per-transaction.

One supporting note on the docs point about absence of signal: treating a missing browser-environment signal as risk penalizes agent traffic by construction, and it creates an incentive for agents to emulate browser environments, which degrades the signal ecosystem for everyone. Worth keeping that statement in the spec.

dpaul0501:
@proshoumma Added. The description now gives the reasoning for each signal, the measurements
behind it, and the limits.

Two of the five I no longer think belong here, and I would rather write that than delete them
quietly. `request_retry_context` repeats the `Idempotency-Key` header, which already has the
24-hour store, the 409 on body mismatch, and signature coverage. The gap is that the header is
optional, and requiring it on `complete` is a better change than my field.
`agent_instance_continuity` is a value the caller picks, so any caller can present it, while the
request signature is already verifiable. Both objections are @westonale's from #630. Both hold.

`instruction_provenance` is worth more discussion. An injected purchase and an instructed one are
byte-identical at the payment layer, so the problem is real. But the classification comes from the
same runtime the injected content influenced, which is what the signals rule excludes. My view is
that it should be recorded as evidence and checked later, rather than consumed at decision time.
#836 and #837 already express the case. I will drop the key from this PR on that basis.

That leaves `agent_spend_context` and `payee_continuity`. If a smaller PR with only those two is
easier to review, say so and I will cut it down.

@pratham-bhatnagar Thank you. Your point about agents emulating browser environments states the
problem better than I did, and I have credited it in the description.

source/schemas/common/types/signals.json
```diff
@@ -16,6 +16,75 @@
     "dev.ucp.user_agent": {
       "type": "string",
       "description": "Client's HTTP User-Agent header or equivalent."
+    },
+    "dev.ucp.instruction_provenance": {
+      "type": "string",
+      "enum": ["user_turn", "agent_plan", "tool_output", "retrieved_content", "unknown"],
+      "description": "Origin of the instruction that produced this purchase, as observed by the platform running the agent. `user_turn` means the purchase traces to the buyer's own input; `agent_plan` to the agent's own planning; `tool_output` and `retrieved_content` to data the agent read during the task. Distinct from delegated authority: delegation establishes that the buyer authorized the agent to spend, provenance establishes what caused this particular spend."
+    },
+    "dev.ucp.agent_spend_context": {
```
karangoel16, line 25: Keeping `agent_spend_context` in agent memory is fragile: the agent can hallucinate or supply incorrect values, and the receiving side has no way to verify agent-asserted spend context. Spend-related signals that feed into payment decisions should be platform-asserted (or signed), not agent memory. At minimum, the trust assumption here should be explicit in the spec.

